PR #2, opened 2026-06-18
Fix frontend loading deadlock in Vite/Rolldown by deferring extension registration

FreesoSaiFared:
Vite loads extension scripts alphabetically early during startup. The comfyui-mcp-panel JS was imported before window.comfyAPI.app was populated, throwing a synchronous TypeError in /scripts/app.js shim, which deadlocked the Vite module loader chain.

This change replaces the static ES module imports with an asynchronous check, deferring extension registration until window.comfyAPI is fully ready.

### DIFF
--- a/web/js/comfyui-mcp-panel.js
+++ b/web/js/comfyui-mcp-panel.js
@@ -36,22 +36,23 @@
 // `window.app` is no longer assigned before extension eval, so the module
 // import is the canonical access path. The absolute specifier works from any
 // nesting depth under /extensions/<pack>/.
-import { app } from "/scripts/app.js";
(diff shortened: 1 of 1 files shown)


artokun:
Hi @FreesoSaiFared — thank you for this, and for pinpointing the Vite/Rolldown load-order deadlock! 🙏

We were mid a big round of updates on this branch (a major streaming + subgraph + UX pass that heavily rewrote this file), so we couldn't merge your PR in cleanly — but we didn't want to lose the fix. We adopted your deferred-registration approach and credited you as co-author; it shipped in #10 (v0.1.2).

Really appreciate the contribution — hope to see more from you! 🙌

PR #26, opened 2026-06-26
feat(panel): backend-driven Defaults settings (seed-only, loop-proof)

artokun:
## What

Re-implements the backend-driven **Defaults** settings page that was reverted (squash `cf950df`) for causing a reconnect/effort-oscillation storm — this time **seed-only / one-way**, so it is loop-proof by construction.

Feature parity with `cf950df`: a render-fn fetched-model `<select>`, a per-backend reasoning-effort `<select>`, a per-backend Bridge-URL input, with the **Agent backend as the parent control**. Order: **Agent backend → Default model → Default reasoning effort → Bridge URL**.

## Single-source-of-truth design (the whole point)

At runtime there is **exactly one** source of truth for model / effort / backend: the panel's existing `prefs` + composer-picker state (`prefs.model`, `prefs.effort`, `prefs.modelAuto`, `selectedBackend`, `urlInput`).

Settings are **SEED-ONLY / one-way**:
- **settings → prefs** happens only at **mount** (`seedFromSettings`) and on a **genuine user edit** in the Settings dialog (the render-fn change handlers → `panelHooks.applyModel/applyEffort/applyBackend`).
- Settings are **never re-read on a reconnect**. The reconnect/handshake path reads `prefs` only.
- **prefs → settings** reflection (composer pick → settings value) is one-directional: `setSetting(...)` runs under the existing `suppressSettingOnChange` guard (no `onChange`), and the open-dialog repaint (`rebuildDependentSettings`) is pure DOM (`.value` / `replaceChildren`, dispatches no `change` event). Neither can call `set_options` or `connect`.

`settingsBackendState` (per-backend fetched catalog + discovered ports) is a **display mirror for the Settings dialog only** — never read on a reconnect, never drives `set_options`.

## Anti-oscillation trace (a / b / c)

**(a) On Connect/reconnect, is `set_options(effort/model)` ever sent automatically?**
The only automatic push is the **pre-existing** `applyModelCatalog` post-handshake push (`if (prefs.userSet)`), unchanged in trigger — it lets a freshly-spawned agent adopt the saved pick. It now reads `prefs.modelAuto` (sends `model:null` for Auto) but still sends **stable `prefs` values**. No *new* reconnect-time push was added, and it never issues a *changed* effort/model on a reconnect (the cf950df failure mode). Settings are not consulted here.

**(b) Can the settings effort and prefs effort disagree so successive reconnects send different values?**
No. A reconnect runs `onModels → applyModelCatalog`, which reads `prefs` (not settings) and snaps `prefs.effort` via `nearestInList(prefs.effort, effortsForModel(prefs.model))`. On a same-backend reconnect `connectedBackend` and `prefs.model` are stable, so the snap is idempotent and `prefs.effort` is unchanged → the push sends the **same** value every reconnect. `prefs` is mutated only by mount-seed and user edits, never by a reconnect. There is no second writer, so the two can't drive divergent reconnect pushes (the medium↔default flip that looped before).

**(c) Does a backend switch send at most one `set_options`, with no follow-on flip?**
Yes. The Settings backend `onChange` runs `applyBackendDefaults(v)` (seeds the dependent **settings** under `suppressSettingOnChange`; sends nothing) then `panelHooks.applyBackend(v)`. `applyBackend` resets the **live prefs once** (`modelAuto=true`, effort mapped to the new scale), saves, refreshes the chip, and calls `connectBackend` — it **sends no `set_options`**. The single push comes from the post-handshake `applyModelCatalog`. Because prefs were pre-reset to agree (Auto → `model:null`, mapped effort), there is exactly one push and no follow-on correction.

## Reused (no new mechanisms)

`modelCatalog` / `presentableModels` / `normalizeModels` / `effortsForModel` / `pickDefaultModel`, `BACKEND_EFFORTS` / `ALL_EFFORTS` / `nearestInList` / `effortMeta`, `prefs` / `savePrefs`, `suppressSettingOnChange` + `settingsArmed` gates, `setSetting`, `connectBackend`, `panelHooks.*`, `GET /comfyui_mcp_panel/backends` (already returns `port`).

## P2 polish

- Blank/**Auto** model and **"Model default"** effort now CLEAR prefs (live via `applyModel`/`applyEffort`; on reopen via `seedFromSettings`) instead of being ignored.
- Token button shows a recovery `alert` after retry exhaustion instead of silently no-op'ing.
- Logo gains an `onerror` fallback to the served path (recovers from a 404, not just a `URL()` construction failure).

## Untouched

The bridge-client connect/reconnect/flicker state machine (`markConnected`, `connectPatienceAttempts`, `gaveUp`, soft-reload/auto-respawn) is **not modified** — this PR is settings-only.

`node --check web/js/comfyui-mcp-panel.js` → exit 0.

### DIFF
--- a/web/js/comfyui-mcp-panel.js
+++ b/web/js/comfyui-mcp-panel.js
@@ -264,6 +264,34 @@ let suppressSettingOnChange = false;
 // decision; only genuine post-load user edits in the Settings dialog take effect.
 let settingsArmed = false;
 
+// Backend display names available at module scope (the panel keeps its own copy in
+// buildPanel's closure for the chips; this one is for the Settings dialog's helpers).
+const BACKEND_TEXT = { claude: "Claude", codex: "ChatGPT" };
+
+// Live panel state PUBLISHED for the Settings dialog's backend-driven Default-model
+// / Default-effort / Bridge-URL controls. Those render at MODULE scope (the settings
+// `type:()=>HTMLElement` render-fns run outside buildPanel's closure), so the open
+// panel keeps this current; it also persists (last value) after the panel unmounts so
+// the dialog can still render sensible options. `modelsByBackend` is the per-backend
+// FETCHED catalog (cached as each backend connects at least once); `ports` is the
+// per-backend bridge port discovered from GET /backends (claude 9180, codex 9181).
+//
+// IMPORTANT (anti-loop): this is a DISPLAY mirror for the Settings dialog only. It is
+// NEVER read on a reconnect/handshake and NEVER drives set_options. The single source
+// of truth for the live agent stays the panel's prefs/composer-picker state.
+const settingsBackendState = {
+  modelsByBackend: {}, // backend id -> presentable model rows (last fetched)
+  ports: { claude: 9180, codex: 9181 }, // per-backend bridge port (from /backends)
+};
+// References to the dialog's live dependent controls, (re)set by their render-fns each
+// time the dialog opens, so a backend change can RESET + repopulate them IN PLACE (a
+// render-fn setting has no static options to re-key). Updating these never fires the
+// elements' change listeners (we only assign .value / replaceChildren), so reflecting
+// a composer pick into the open dialog can't trigger set_options or a reconnect.
+let settingsModelSelectEl = null;
+let settingsEffortSelectEl = null;
+let settingsBridgeInputEl = null;
+
 function getSetting(id) {
   try {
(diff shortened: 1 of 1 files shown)


artokun:
Superseded by the split-into-per-backend-groups design (Claude group + Codex group) — that structure is inherently loop-proof (no shared dropdown that resets/flips on backend switch). Re-doing as grouped settings.

PR #48, opened 2026-07-01
fix(panel): connect dropdown shows 'connect <this-comfyui-url>' (not --panel-orchestrator)

artokun:
The connect dropdown copied `npx -y comfyui-mcp --panel-orchestrator` — the local-spawn form that targets the LOCAL ComfyUI. So opening a **remote pod's** ComfyUI and copying it gave an agent that drove/reported the **laptop** (the '4090 instead of the pod's 5090' confusion). Now it shows `npx -y comfyui-mcp connect <window.location.origin>` so the orchestrator targets the ComfyUI you're actually viewing. OS-aware (cmd /c on Windows). node --check passes; no pyproject change.

### DIFF
--- a/web/js/comfyui-mcp-panel.js
+++ b/web/js/comfyui-mcp-panel.js
@@ -5630,10 +5630,22 @@ function buildPanel() {
   const helpDiv = document.createElement("div");
   helpDiv.className = "cmcp-help";
   helpDiv.textContent =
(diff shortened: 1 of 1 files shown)


artokun:
Superseded by #50, which ships the corrected `connect` command (OS-aware) alongside the orchestrator-readiness fix, rebased on current main (post #49 browser-host targeting).

PR #52, opened 2026-07-01
feat(panel): prefill this pod's URL into the connect command (secure bridge in every browser)

artokun:
## Why

A remote pod's HTTPS panel page can't open a plain `ws://127.0.0.1` bridge — browsers block insecure sockets from a secure origin (mixed content) and gate public→loopback (Private Network Access). Chrome allows it *with a permission prompt*; **Safari / Firefox / Comet hard-block it**.

The fix is the secure `wss://` bridge (comfyui-mcp **0.23.4** on npm + panel **0.4.11** on the registry): when the orchestrator is started with the pod URL, it opens a token-gated Cloudflare `wss://` tunnel and advertises it to the pod, and the panel fetches + uses it — works in every browser.

The catch: the orchestrator can only learn the pod URL from the CLI (the panel's hello rides the ws bridge, which those browsers block). Full zero-input inference is impossible for Safari/Comet.

## What this PR does

The panel now **prefills this pod's own hostname** (`window.location.origin`) into the `connect` command it displays — help dropdown, onboarding step, and the "no agent" hint:

```
npx -y comfyui-mcp@latest connect https://<this-pod>
```

One copy-paste line, URL already filled — no guessing, no manual typing. On `http`/`localhost` it shows the bare form (auto-targets local). Implemented via a single `connectCommand()` helper.

## Testing
- Server-side secure path verified live against a 0.4.11 pod: tunnel up, advertise succeeded, pod `/bridge_url` returned the wss URL.
- Panel JS syntax-checked; `connectCommand()` returns the URL form on https, bare on http.
- ⚠️ Live browser test (Safari/Comet connecting over wss) to be confirmed.

## Coordination
- Server **0.23.4** already published to npm — `connect <URL>` works today.
- After merge (registry publishes **0.4.12**), **rebuild the `docker/runpod` image** so fresh pods bake the new panel.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+## [0.4.12] - 2026-07-01
+
+### Changed
+
+- **The panel prefills THIS pod's own URL into the `connect` command it shows** (help
+  dropdown, onboarding step, and the "no agent" hint) when served over https — e.g.
+  `npx -y comfyui-mcp@latest connect https://<this-pod>`. Running it with the URL lets
+  the local orchestrator open a secure `wss://` bridge that works in **every browser**
+  (Safari / Firefox / Comet included), not just Chrome-with-a-permission-prompt. The
+  URL is read from `window.location`; on http/localhost the bare form is shown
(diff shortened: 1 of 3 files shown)


artokun:
Superseded — reapplied on top of the 0.4.12 per-shell-copy structure (makeShellCommandBlock + connectCommand) and shipped as 0.4.13 directly.

PR #56, opened 2026-07-04
fix(panel): sync Bridge URL to ComfyUI settings on save

### LINKED ISSUE #55: [panel] Backend chip switch silently reverts the Advanced Bridge URL to a stale value
## Summary

Setting a custom "Bridge URL" under Settings → General → Advanced and clicking Reconnect works as expected. But the very next time the user switches the bottom backend chip (claude ↔ codex ↔ gemini ↔ ollama), the Bridge URL field silently reverts to the previous (stale) value, the agent fails to reconnect, and the user's custom URL is lost.

## Reproduction

1. Install comfyui-mcp-panel + a running bridge on a non-default port/token.
2. Open ComfyUI, open the panel.
3. Settings → General → Advanced → Bridge URL: paste a custom URL like `ws://grimoire.lost.plus:9180/?token=...` and click Reconnect. Panel connects successfully. ✓
4. Click a different backend chip at the bottom of the panel (e.g. claude → codex, or codex → claude). 
5. The panel tries to reconnect and fails. 
6. Re-open Settings → General → Advanced → Bridge URL: the field is back to the previous value, not the one you typed.

Environment
- comfyui-mcp-panel: 0.5.0
- ComfyUI: 0.27.0
- Browser: Safari 18.6 & Comfy Desktop 1.0.28
- Bridge: comfyui-mcp@0.24.5 connect mode

LPFchan:
Fixes #55

### DIFF
--- a/web/js/comfyui-mcp-panel.js
+++ b/web/js/comfyui-mcp-panel.js
@@ -116,6 +116,12 @@ function saveBridgeUrl(url) {
   } catch {
     // localStorage unavailable — session-scoped settings only.
   }
+  // Mirror into ComfyUI's setting store so a backend switch (which calls
+  // configuredBridgeUrlFor → getSetting(SETTING_BRIDGE)) sees the same value.
+  // Without this, the per-backend chip switch silently reverts the Bridge URL
+  // to whatever the ComfyUI setting still holds, and the user's custom URL
+  // is lost on the very next backend pick.
+  setSetting(SETTING_BRIDGE, url);
 }
 
 // Per-TAB session id: sessionStorage is scoped to the tab and survives


artokun (approved):
Confirmed the root cause and the fix — merging-quality as-is. The chip switch re-resolves the URL via `configuredBridgeUrlFor() → getSetting(SETTING_BRIDGE)`, but `saveBridgeUrl()` only ever wrote localStorage, so any URL saved through the panel path was invisible to the store and the switch snapped back to the stale value. Mirroring into the setting store is the right fix, and I verified the two risk paths are already covered:

- no reconnect-storm regression: `setSetting()` wraps the write in `suppressSettingOnChange`, so the `SETTING_BRIDGE.onChange → applyBridgeUrl` applier doesn't re-fire (and `applyBridgeUrl` early-returns on an unchanged URL as a second layer);
- ephemeral tunnel URLs are unaffected: `setUrl(next, { persist: false })` skips `saveBridgeUrl` entirely, so per-session wss tunnel URLs still don't get persisted.

Only side effect is `seedFromSettings` now materializing the default URL as an explicit stored setting when unset — benign, it's the same value as `defaultValue`. Thanks!

Issue #54, opened 2026-07-03
Panel unreachable from a remote browser


LPFchan:
### Setup
- `comfyui-mcp-panel` installed as a ComfyUI custom node (v0.5.0 on ComfyUI-Manager, latest on the registry).
- ComfyUI running on a remote machine, accessing it from my laptop
- Orchestrator started manually on the same machine as ComfyUI: `npx -y comfyui-mcp@latest --panel-orchestrator` (port 9180, loopback).
- I access the ComfyUI UI from a laptop on the same LAN, via `http://(serverip):8188` (no tunnel / no proxy).

### What happens
When I click Connect in the Agent tab, the panel shows:

> No agent is listening on the bridge (ws://127.0.0.1:9180). This ComfyUI won't start one — run the agent on YOUR machine, then click Connect:
>     npx -y comfyui-mcp@latest connect

The orchestrator is running and healthy — I can `ss -tlnp` and see port 9180 listening, and a raw WebSocket handshake from `127.0.0.1:9180` succeeds with proper MCP responses. But the panel can't reach it.

### Diagnosis
The panel resolves the bridge URL to `ws://127.0.0.1:9180` (the panel's `_BRIDGE_HOST` constant in `__init__.py`). On my laptop, `127.0.0.1` is the laptop itself, not the server. So the panel is trying to reach a service that only exists on the comfyui server box, and the browser's 127.0.0.1 has nothing listening on 9180.

### Versions
- comfyui-mcp-panel: 0.5.0 (ComfyUI-Manager registry)
- comfyui-mcp: 0.24.2
- ComfyUI: 0.27.0
- Panel JS: `web/js/comfyui-mcp-panel.js` (the `showExternalHintOnce` function on line ~8810)

artokun:
@LPFchan can you confirm the browser and if it works with Chrome? Thanks for providing the other versions

EDIT:
Can you also provide your ComfyUI Manager Version?

artokun:
took a closer look — your diagnosis is right, and it's actually browser-independent (ignore my question above): the panel JS runs in YOUR laptop's browser, so \`ws://127.0.0.1:9180\` resolves to the laptop, where nothing is listening. The orchestrator you started on the ComfyUI box is healthy but unreachable by design — the bridge binds loopback-only.

**the supported topology is the other way around: the orchestrator runs on the machine with the browser**, driving the remote ComfyUI:

\`\`\`bash
[...]

PR #57, opened 2026-07-04
fix(panel): polyfill crypto.randomUUID for non-secure contexts

LPFchan:
crypto.randomUUID is gated to SECURE contexts; plain http://<serverip> (ComfyUI's default bind) is not one, so on a LAN connection the panel's getTabId() throws, the outer try/catch in sendHello() silently swallows it, the hello frame is never sent, the bridge never spawns an agent, and the panel shows 'waiting for the panel agent…' forever (then reconnects every 20s).

The polyfill installs a spec-compliant randomUUID() on the existing window.crypto via Object.defineProperty. crypto.getRandomValues (the CSPRNG) is available in non-secure contexts, so the polyfill produces a proper RFC 4122 v4 UUID. If defineProperty fails (e.g. crypto is a frozen host object), we leave it alone — behavior reverts to the pre-polyfill state rather than risk clobbering getRandomValues / subtle on Crypto.prototype. The IIFE is placed after the file's full header block, not inside it.

Follow-up to #54.

### DIFF
--- a/web/js/comfyui-mcp-panel.js
+++ b/web/js/comfyui-mcp-panel.js
@@ -1,6 +1,6 @@
 // =============================================================================
 // ComfyUI Agent Panel — sidebar driven by an autonomous background agent.
-//
+// =============================================================================
 // Shipped as a UI-only custom node pack (served via WEB_DIRECTORY). The panel
 // connects to the loopback WebSocket bridge owned by the comfyui-mcp panel
 // orchestrator, started with:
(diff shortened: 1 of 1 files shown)


artokun (changes requested):
Confirmed the bug is real and I want this fix — two changes requested before merge.

First, the actual trigger is simpler (and more common) than the sandboxed-iframe/CSP framing: **`crypto.randomUUID` only exists in secure contexts**, and plain `http://<serverip>:8188` LAN access — the exact topology 0.24.5 just made first-class — is not one. `crypto.getRandomValues` IS available in insecure contexts, so your getRandomValues path is the one that will actually run. And `getTabId()`'s own `catch` block calls `crypto.randomUUID()` again, so it rethrows into `sendHello()`'s silent catch → the eternal 'waiting for the panel agent…'. Diagnosis solid.

**Requested changes:**

1. **The last-resort fallback corrupts `window.crypto`.** `Object.assign({}, window.crypto, ...)` only copies *own enumerable* properties — `getRandomValues` and `subtle` live on `Crypto.prototype`, so the replacement object silently loses them, breaking any other extension (or our own getRandomValues branch!) that touches crypto afterward. Either bind the real methods through:
   ```js
   const c = window.crypto;
   window.crypto = {
     getRandomValues: c?.getRandomValues?.bind(c),
     subtle: c?.subtle,
     randomUUID: _polyfill,
   };
   ```
   …or just drop that branch entirely (if `defineProperty` on the crypto instance throws, we're better off leaving the global alone).

2. **Placement:** the block is currently inserted mid-way through the file's banner comment (between the title line and the 'Shipped as a UI-only custom node pack…' text), splitting it. Move the IIFE to just below the full header block.

Optional, take it or leave it: a local `uuid()` helper used at the 5 call sites (lines 129, 134, 3942, 3975, 6840) would avoid patching a global shared with every other ComfyUI extension. I'm fine with the polyfill approach given it's spec-compliant — but only with the fallback in (1) fixed so it can never make crypto *worse*.

Thanks for the third solid contribution this week 🙏

artokun (approved):
Both requested changes addressed, and the revision is better than what I asked for: dropping the crypto-replacement branch entirely (rather than patching it) means the polyfill can never make `window.crypto` worse, and removing the Math.random path is right since `getRandomValues` isn't secure-context-gated. Placement below the banner + the corrected non-secure-context framing — all good. Merging. Thanks again 🙏

PR #84, opened 2026-07-16
feat(civitai): Creator filter — top-creators leaderboard picker + username search across every feed

artokun:
Adds a **Creator filter** to the CivitAI explorer's filter sheet — feature parity with the mobile app (comfyui-mcp-mobile PR #2), requested in Discord help thread `1527122588189327421` and by the maintainer.

## What it does

- **Creator section in the filter sheet**: an empty search field shows the site's **top-creators leaderboard** (tRPC `leaderboard.getLeaderboard` via the existing header-injecting Python proxy), ranked with compact download/like counts. Typing swaps to a **debounced (300ms) `GET /v1/creators` username search** (with model counts).
- **Picked creator = removable pill**, threaded through every query the explorer makes:
  - image/video feed: `/v1/images?username=<creator>`
  - media keyword search: Meilisearch filter `user.username = "<escaped>"` (backslashes escaped **before** quotes so a trailing `\` or crafted `\"` can't break out of the filter string)
  - model tabs: `/v1/models?username=<creator>`
- **Reset filters** clears it (`username` is now part of `DEFAULT_FILTERS` / `filtersDirty`, so the header's filter dot lights up while a creator is active).

## Known API quirks handled (discovered during the mobile pass)

- `/v1/models` with **both** `query` and `username` returns an empty page — with a creator picked we send only `username` and match the keyword client-side; the models loader chases up to 4 extra pages when a page filters down to nothing so the infinite-scroll sentinel can't stall.
- The leaderboard endpoint intermittently 401s bare user agents — the proxy's browser-shaped headers usually pass, and on failure the picker degrades to *"Top creators unavailable right now."* without breaking the rest of the sheet.

## Review findings from the mobile pass, avoided here

- Meili escaping order: backslashes first, then quotes (unit-tested).
- Debounced creator lookups carry a **generation counter** (bumped on every keystroke, not just at debounce fire) so a slow stale response can't repopulate the dropdown under a newer query. The counter lives outside the sheet's re-render closure so re-renders can't resurrect stale responses either.
- **Favorites tab shows the filter visibly inert** (dimmed pill + "Ignored on the Favorites tab…") instead of silently no-oping — favorites are *your* likes, from every creator; the tRPC favorites feed has no creator param. Matches the mobile semantics.

## Verified

- `node --check` on `cmcp-civitai.js`, `cmcp-civitai-ui.js`, and `comfyui-mcp-panel.js` as `.mjs` copies — clean.
- `npm run test:unit` — 35 pass (9 new in `browser_tests/unit/civitai-creator.test.mjs`: escaping order, the models quirk, leaderboard parsing incl. deleted-user drop and non-array degrade, `/v1/creators` shaping + blank-query short-circuit, username threading into feed/search).

## Owed

- **Live browser verification** (open the explorer, pick a leaderboard creator, confirm each tab narrows, confirm the 401 degrade path) — not run in this environment; needs a ComfyUI + Chrome pass before release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,15 @@ All notable changes to this project are documented here. This project adheres to
   level/base-model toggles no longer mutate the frozen module defaults
 
 ### Added
+- CivitAI browser: **Creator filter** in the filter sheet (parity with the
+  mobile app) — an empty field shows the site's top-creators leaderboard
+  (ranked, with download/like counts; degrades to a friendly note when the
+  endpoint balks), typing runs a debounced username search, and the picked
+  creator becomes a removable pill that narrows every feed: images/videos
+  (`/v1/images?username=`), media search (Meili `user.username` filter,
+  escaped), and model tabs (`/v1/models?username=` — keyword+creator matches
+  the keyword client-side around an API quirk). Favorites shows the filter as
+  visibly ignored (your likes come from every creator); Reset clears it
 - CivitAI browser: **like/unlike toggle** — heart on card hover and in the
   lightbox (signed out, the heart opens sign-in); likes mirror into a **default
   likes collection** picked (or created) in the new account sheet
--- a/browser_tests/unit/civitai-creator.test.mjs
+++ b/browser_tests/unit/civitai-creator.test.mjs
@@ -6,18 +6,27 @@ import { test } from "node:test";
 import assert from "node:assert/strict";
 import { CivitaiClient, DEFAULT_FILTERS, filtersDirty } from "../../web/js/cmcp-civitai.js";
 
-/** Client whose proxy calls are captured; each call resolves `payload`. */
+/** Client whose proxy calls are captured. `payload` may be a single response
+ *  (every call resolves it) or an array consumed one per call (last repeats). */
 const stub = (payload) => {
   const calls = [];
+  const pages = Array.isArray(payload) ? [...payload] : null;
   const client = new CivitaiClient({
     fetchApi: async (_path, opts) => {
       calls.push(JSON.parse(opts.body));
-      return { ok: true, json: async () => payload };
+      const body = pages ? (pages.length > 1 ? pages.shift() : pages[0]) : payload;
+      return { ok: true, json: async () => body };
     },
     apiURL: (p) => p,
   });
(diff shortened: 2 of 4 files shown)


artokun:
Addressed the review findings in c6f4970:

1. **Stale in-flight loads** — `reload()` now bumps the request generation (and resets the guarded loading flag) before clearing the grid, so changing/removing the creator (or any filter/tab/query) mid-load discards the old response and its cursor instead of letting it repopulate the fresh grid.
2. **No dead-end pagination** — a models load that appends nothing while pages remain renders an explicit "No matches yet — keep searching" button into the scroll sentinel; a finished empty grid shows "No results."
3. **Chase gated + relocated** — the page-chase moved into `fetchModels` and only runs for the keyword×creator combo (the sole client-side-filtered path). Every other call issues exactly one request, identical to pre-PR behavior — unit tests now assert the request counts (1 on all default paths; chase stops on first match; cap = initial + 4 hops, cursor returned for explicit continuation).
4. **Sheet debounce race** — the creator debounce timer is hoisted to sheet scope and cancelled on every re-render (plus an `isConnected` guard), so a pre-rerender timer can't strand the new sheet on "Looking up creators…".

Verified: `node --check` (.mjs copies) clean, `npm run test:unit` 37 pass / 0 fail, `npm run typecheck` clean. Live browser pass still owed.

artokun:
Round-2 nits addressed in 0057c3f:

1. **Sheet teardown** — `openSubModal` gained an optional `onClose` fired on every dismiss path (x button, backdrop, `sheet.close()`); the filter sheet uses it to cancel the pending creator debounce and bump the lookup generation, so an already-started lookup's completion early-returns instead of mutating detached nodes. The direct-call path (focus) now clears a pending debounce tick too, so blur/refocus inside the 300ms window no longer double-fires the same text.
2. **Searching overlay ownership** — `reload()` captures a `state.searchSeq` generation (same pattern as the grid's `reqId`); a superseded reload's `finally` no longer hides the spinner belonging to the newer search that replaced it.

Verified: `node --check` (.mjs copy) clean, `npm run test:unit` 37 pass / 0 fail, `npm run typecheck` clean. Awaiting the live browser pass.

PR #87, opened 2026-07-16
feat(civitai): "See more from @creator" + GitHub-style @creator search syntax

artokun:
Implements artokun's spec from Discord thread 1527122588189327421 — two entry points into the just-merged creator filter (#84), reusing its exact filter path (no second mechanism).

## What

**1. "See more from @creator" in the detail views**
- **Lightbox** (image/video): when the item has an author, an action button in the side pane sets `filters.username`, closes the lightbox, and reloads. The sheet's pill and the filter dot reflect it, and the pill removes it — same slot the filter sheet sets.
- **Model detail sheet**: same action from `detail.creator`, closing the sheet and reloading the current model tab.
- From the **Favorites** tab the click lands on the matching Images/Videos tab first, because the favorites tRPC feed has no creator param (the sheet already renders the filter as visibly inert there).

**2. GitHub-style `@creator` search syntax**
`@bab0zi cyberpunk city rain` → creator=`bab0zi`, search=`cyberpunk city rain`.
- New pure `parseCreatorQuery()` export in `web/js/cmcp-civitai.js`; the search bar (and the agent-seeded `open_civitai` query) parse through it.
- First `@token` wins; later `@tokens` are consumed (as literals they'd only pollute the keyword search). A lone `@` and mid-word `@` (emails like `me@example.com`) stay plain text. `@name` alone (no terms) just sets the filter. Typing a new `@name` replaces the previous creator.
- The `@token` is stripped from the visible box after parsing — the removable pill owns the creator from there, so removing the pill can't be undone by a stale token re-parse.
- For a pre-seeded query, an explicit `opts.filters.username` wins over the parsed token.

## Meilisearch combination (maintainer asked)
Checked the Meilisearch docs via context7 ("Search and filter together"): sending `q` + `filter` in ONE request is the canonical combination — the engine first restricts to documents matching the filter, then ranks that set by relevancy to the multi-term `q`. That is exactly what the existing `searchMedia` does (`q` + escaped `user.username = "…"` filter via `escapeMeili`), so no client change was needed — the parser just feeds the two halves into the existing paths. REST feeds/models keep `username=` plus the existing keyword handling, including the `query`×`username` empty-page quirk workaround from #84.

## Tests
New `browser_tests/unit/civitai-at-syntax.test.mjs` (10 tests): the example from the spec, token-anywhere ordering, `@name`-alone, no-token passthrough, first-token-wins, lone `@`, mid-word `@`, whitespace collapsing, raw-username return, and an integration check that a parsed `we"ird\name` username lands escaped in the Meili filter next to the multi-term `q`.

## Verification
- `node --check` (as `.mjs`): both edited files pass
- `npm run test:unit`: 47/47 pass (37 existing + 10 new)
- `npm run typecheck`: clean
- Registry-YARA: no inline event attributes (all `addEventListener`), no new colors, no new external hosts

## Risks / notes
- PR #85 (feat/civitai-load-workflow) also touches `cmcp-civitai-ui.js` — whichever lands second rebases; my lightbox change is a small insertion in `render()` next to the actions row, so conflicts should be trivial.
- Consuming *later* `@tokens` (rather than leaving them as literal search terms) is a judgment call — literals would hit Meili as junk terms; easy to flip if unwanted.
- Search-box whitespace runs now collapse to single spaces when re-parsed (token join) — cosmetic only.
- No visual browser check on this box (no local Chrome) — logic is unit-covered; the lightbox/detail buttons reuse existing `cmcp-btn` styling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,19 @@ All notable changes to this project are documented here. This project adheres to
   level/base-model toggles no longer mutate the frozen module defaults
 
 ### Added
+- CivitAI browser: **"See more from @creator"** in the lightbox and in the
+  model detail sheet — one click sets the creator filter (the same removable
+  pill + filter-dot slot the sheet uses), closes the view, and reloads; from
+  Favorites it lands on the matching Images/Videos tab (the favorites feed has
+  no creator param)
+- CivitAI browser: **GitHub-style `@creator` search syntax** — typing
+  `@bab0zi cyberpunk city rain` in the search bar sets the creator filter to
+  `bab0zi` and searches the remaining terms (Meili gets `q` + escaped
+  `user.username` filter in one request; REST feeds get `username=` plus the
+  existing keyword handling). The first `@token` wins and a new `@name`
+  replaces the previous creator; `@name` alone just sets the filter; a lone
+  `@` and mid-word `@` (emails) stay plain text. Agent-seeded queries
+  (`open_civitai`) honor the same syntax
 - CivitAI browser: **Creator filter** in the filter sheet (parity with the
   mobile app) — an empty field shows the site's top-creators leaderboard
   (ranked, with download/like counts; degrades to a friendly note when the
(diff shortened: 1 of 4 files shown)


artokun:
Superseded by #86, which merged this exact feature (parseCreatorQuery + See-more buttons) on 07-15 with three rounds of codex review — this parallel branch went CONFLICTING when it landed. The @-syntax tests here are covered by browser_tests/unit/civitai-creator.test.mjs on main.

PR #91, opened 2026-07-19
feat(tools): view_selected / view_nodes_in_viewport / view_errored_nodes

artokun:
## What

Three read-only live-canvas executors backing the new `panel_view_*` tools, so the agent can scope to the user's actual intent instead of skimming a big graph.

### `graph_view_selected`
Reads `canvas.selectedItems` (the modern `Set`, which also holds groups/reroutes), falling back to the legacy `selected_nodes` object. `current_node` is deliberately **ignored** — verified live that it tracks the pointer, not the selection (it read 66 while the selection was 9). Returns the empty case with a clear hint rather than guessing.

### `graph_view_nodes_in_viewport`
Nodes intersecting the visible rect (any overlap counts, so half-on-screen nodes aren't dropped).

**Bug found by live testing:** it derives the rect from the canvas element's **CSS size ÷ `ds.scale`**, *not* `canvas.visible_area`. LiteGraph only refreshes `visible_area` when it **draws**, and sizes it from the **backing store** — an unpainted/background canvas still reports the `300×150` HTML default:

| source | measured | correct? |
|---|---|---|
| `visible_area` | `333 × 167` | ❌ → returned **0 nodes** |
| CSS rect ÷ scale | `1712 × 1478` | ✅ |

`visible_area` is kept as the fallback for a detached/headless canvas.

### `graph_view_errored_nodes`
LiteGraph only sets a boolean and paints a red outline — the reason lives elsewhere, which is exactly why users report *"red node, no error message"* (confirmed live: `badge.text` was `""` and `lastNodeErrors` was `null`). This joins the reason onto the node from four sources: the `missingModel` store (file, directory, **download URL**), `missingNodesError`, `app.lastNodeErrors` (per-input validation), and `executionError`.

## Testing — verified live against ComfyUI on :8188

On a canvas with 3 missing-model errors:

- **errored** → all 3 red nodes with exact causes, e.g. node 62 `CLIPLoader` → `qwen_3_4b.safetensors` missing from `text_encoders` + HF download URL.
- **selected** → empty case returns the hint; selecting node 62 returned it with its widgets.
- **viewport** → `1712×1478 @ 0.9` with all 10 nodes in view; after panning 6000 away, correctly **0** in view.

`npm run typecheck` clean, `npm run test:unit` **67 passed**. A temporary test scaffold used to exercise the executors was removed (verified absent).

## Pairs with

`comfyui-mcp` PR (the tool declarations). Cmd names verified matching across the bridge.

> **Dev-loop note:** ComfyUI must be restarted to pick up panel JS edits — it serves a snapshot from boot.

### DIFF
--- a/web/js/comfyui-mcp-panel.js
+++ b/web/js/comfyui-mcp-panel.js
@@ -3449,6 +3449,234 @@ const GRAPH_TOOL_EXECUTORS = {
     };
   },
 
+  // WHICH NODE(S) THE USER HAS SELECTED. Without this the agent can only guess
+  // what "this node" / "the highlighted one" means and ends up skimming the whole
+  // graph for clues (a user with a 700-node canvas burned a lot of tokens doing
+  // exactly that). Selection is the cheapest possible scope: read it FIRST.
+  // LiteGraph keeps two views of it — `selectedItems` (a Set, the modern one, and
+  // the only one that also holds groups/reroutes) and `selected_nodes` (an
+  // id→node object). Prefer the Set, fall back to the object. `current_node` is
+  // NOT the selection (it tracks the last node under the pointer), so it's ignored.
+  graph_view_selected() {
+    const { graph, canvas } = getGraphCtx();
+    const inGraph = new Set(graph._nodes ?? []);
+    const picked = [];
+    const others = [];
+    const items = canvas?.selectedItems;
+    if (items && typeof items.forEach === "function") {
+      for (const it of items) {
+        if (inGraph.has(it)) picked.push(it);
+        else if (it) others.push(it.constructor?.name ?? "item");
+      }
+    }
+    // Fallback for frontends that only maintain the legacy id→node object.
+    if (!picked.length && canvas?.selected_nodes) {
+      for (const n of Object.values(canvas.selected_nodes)) {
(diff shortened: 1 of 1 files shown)


artokun:
### Added: "why is this red?" UI affordance (d99d6ad)

Surfacing the errored-node tool in the UI rather than relying on the user knowing to ask.

A small amber **⚠** button appears in the composer row **only while the canvas has flagged nodes** (count when >1, node ids in the tooltip). Clicking it seeds a diagnostic ask pointing the agent at `panel_view_errored_nodes`, explicitly telling it to read the real cause instead of guessing from widget values.

Detection is a 1.2s poll over `node.has_errors`, diffed against the last id list so the DOM is only touched when the set changes; the interval is cleared with the panel's other timers on unmount.

**Verified live on :8188** against the 3 missing-model errors:

- rendered `⚠ 3`, titled `3 nodes with errors (id 62, 63, 66) — ask why`
- seeded: *"Why are nodes 62, 63, 66 highlighted red on my canvas? Use panel_view_errored_nodes to read the actual reason…"*
- hid when the flags were cleared, reappeared when restored

`npm run typecheck` clean, `npm run test:unit` 67 passed.

PR #96, opened 2026-07-20
feat(training): LoRA training wizard — dataset gather/label + launch/monitor (panel)

artokun:
# Training wizard: gather/label datasets + launch/monitor local LoRA training from the panel

Replaces the coming-soon Training modal with a working 4-step wizard for the
**Image Character** flow (FLUX.1-dev). Companion to **comfyui-mcp PR #237** (the
trainer backend) — the wizard drives it through the existing bridge `call_tool`
channel. Other flow cards stay "Coming soon (P2)". Mobile translation follows
once this UX settles.

## Steps

1. **Dataset (gather)** — name + trigger word; pick from recent ComfyUI **outputs**
   (structured grid via new `GET /comfyui_mcp_panel/training/list-outputs`, newest-first,
   heap-bounded, off the event loop) or **upload** images from disk (drag-drop/picker
   → existing `uploadBlobToInput`; navigation blocked until the batch settles).
2. **Label** — per-image caption fields + bulk tools (prefix trigger to all,
   apply-first-to-all, clear). Live caption count.
3. **Launch** — summary, param presets (Smoke 200-step / Standard 2000 / Custom with
   strict full-token validation that survives navigation), GPU label, and a
   `train_doctor` preflight (docker/gpu/image/HF token/localFs — Launch stays
   disabled while pending and when the orchestrator targets a remote ComfyUI).
   Then `resolve-paths` (new py route: ComfyUI image refs → containment-checked
   absolute paths) → `train_prepare_dataset` → `train_start`.
4. **Monitor** — chained (non-overlapping, generation-guarded) `train_status` polls:
   progress bar, loss, ETA, log tail, sample thumbnails (new `/training/file` route,
   contained to the training roots), cancel with confirm, completion/failure cards.
   **Jobs** view lists past runs and reopens them.

Also: capability gate at wizard entry (orchestrators predating the trainer release
get a clear "trainer backend not available" card instead of a doomed walkthrough).

## Verification

- **Playwright spec** `browser_tests/training-wizard.spec.ts` (MockBridge + intercepted
  routes): walks the full wizard, asserts the exact `train_prepare_dataset`/`train_start`
  payloads and the completion card. Full Tier-1 suite green.
- **Py routes** verified standalone (resolve-paths traversal/missing, list-outputs
  bounding/pattern/extension filter, /file containment) and live via curl.
- **Full live pass on the rig**: a second orchestrator running the trainer branch +
  this wizard in a real browser → launched a 200-step smoke run from the panel UI,
  watched live step/loss/ETA in the monitor, completed → `panel_e2e.safetensors`
  handed off to `models/loras` + LoRA catalog with preview.
- codex-reviewed (5 rounds, 15 findings fixed: scanner-safe env access, base-path URLs,
  event-loop blocking, remote-FS gating, poll races, upload race, param validation).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,16 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+## [0.9.7] - 2026-07-20
+
 ### Changed
 - Codex/ChatGPT reasoning-effort scale extends to `max` and `ultra` for the
   GPT-5.6 family (per-model ceilings still intersect with the model list —
   Luna tops out at `max`); pairs with comfyui-mcp >= 0.41.0
+
+### Added
+- max + ultra on the Codex scale for GPT-5.6 (#94)
+
 
 ## [0.9.6] - 2026-07-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,17 +6,6 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
-## [0.9.7] - 2026-07-20
-
-### Changed
-- Codex/ChatGPT reasoning-effort scale extends to `max` and `ultra` for the
-  GPT-5.6 family (per-model ceilings still intersect with the model list —
-  Luna tops out at `max`); pairs with comfyui-mcp >= 0.41.0
-
-### Added
-- max + ultra on the Codex scale for GPT-5.6 (#94)
-
-
 ## [0.9.6] - 2026-07-19
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Changed
+- Codex/ChatGPT reasoning-effort scale extends to `max` and `ultra` for the
+  GPT-5.6 family (per-model ceilings still intersect with the model list —
+  Luna tops out at `max`); pairs with comfyui-mcp >= 0.41.0
+
 ## [0.9.6] - 2026-07-19
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


artokun:
Adversarial review (codex) — **FIX FIRST**, two functional findings plus three smaller ones. Landing stays with this branch's owner; posting findings only.

1. **MAJOR — `py/training_routes.py:66`: `list-outputs` does unbounded recursive work per request.** Results are bounded, but `os.walk(root)` visits and stats the ENTIRE output tree on every call — `asyncio.to_thread` keeps it off the event loop, yet with no traversal cap, cache, timeout, or concurrency guard, repeated calls can saturate the shared executor on a big outputs dir. Cap the walk (entries visited, or depth + mtime-sorted top-level dirs) or memoize per-mtime.

2. **MAJOR — `web/js/cmcp-training-ui.js:541`: one hanging upload permanently deadlocks wizard navigation.** The batch is counted pending, then uploaded sequentially via `uploadBlobToInput` (`comfyui-mcp-panel.js:11676`) which has no abort signal or timeout — a request that never settles disables **Next** forever. Add a per-file timeout/AbortController and surface the failure (skip + toast) instead of blocking.

3. MINOR — `training_routes.py:74`: `os.stat(full)` follows file symlinks with no `islink`/realpath-containment check — the endpoint inventories targets outside the output root (later dataset resolution does reject escapes, so exposure is metadata + listing only).

4. MINOR — `cmcp-training-ui.js:665-707` (and `src/tools/train.ts:45-51` backend-side): custom params validate positive-only with no upper bounds — a fat-fingered steps/rank/resolution passes straight through to an effectively-unbounded run or instant OOM.

5. MINOR — `cmcp-training-ui.js:537`: extension-regex-only file validation; renamed non-images upload fine and fail later during training. Sniff magic bytes or at least decode-probe before counting the image toward the dataset.

Verified clean: no path traversal via the `list-outputs` query (limit + substring only); launch requires two explicit clicks (nothing starts on modal open); owner-death/cancel honesty matches the #237 backend semantics; CI on this head passed node syntax, Bandit, and both registry YARA parity scans; no duplicate top-level declarations; modal conventions intact. Note: codex's sandbox couldn't run typecheck/unit locally — re-run both before merge.

PR #100, opened 2026-07-21
feat(runpod): RunPod control panel + honest host indicator

artokun:
Companion to orchestrator PR artokun/comfyui-mcp#251 and mobile PR artokun/comfyui-mcp-mobile (feat/runpod-control).

In-panel RunPod control modal + honest host indicator, driven by the orchestrator's `runpod_status` + `comfyui_target` broadcasts.

## What
- **cmcp-runpod-ui.js** — control modal: live pod status (GPU/VRAM/uptime/$hr/idle-countdown) and one-tap **deploy / start / stop / connect / use-local** over the whitelisted `runpod_*` tools via `callTool` (no agent turn). Deploy is confirm-armed (it bills). gpu-cli.sh credit.
- **Toolbar 'RunPod' button doubles as the honest host pill** — label + tint reflect where renders run (Local vs the pod name); tooltip shows GPU/$hr.
- Bridge client dispatches the two new frames to the pill + open modal.

The pod runs our template, so once connected the agent installs the user's exact custom nodes / LoRAs → full canvas parity remotely.

## Notes
- Stacks on `feat/training-panel-ux` (shares the toolbar append line); merge that first or rebase.
- ESM parse verified; needs a live panel smoke test (reload ComfyUI).

### DIFF
--- a/web/js/cmcp-runpod-ui.js
+++ b/web/js/cmcp-runpod-ui.js
@@ -26,6 +26,7 @@ function injectStyle(root) {
   if (styleInjected) return;
   styleInjected = true;
   const css = `
+.cmcp-rp-modal{max-width:min(480px,92vw)!important;width:auto;}
 .cmcp-rp-body{display:flex;flex-direction:column;gap:0.75rem;min-width:min(440px,90vw);max-width:520px;}
 .cmcp-rp-host{display:flex;align-items:center;gap:0.5rem;padding:0.55rem 0.7rem;border-radius:8px;
   font-size:0.85rem;font-weight:600;border:1px solid var(--p-content-border-color,#3f3f46);}
@@ -53,9 +54,15 @@ function injectStyle(root) {
 `;
   const el = document.createElement("style");
   el.textContent = css;
-  (root.getRootNode?.() || document).appendChild
-    ? (root.getRootNode?.() || document).appendChild(el)
-    : document.head.appendChild(el);
+  // Append into the same root the modal mounts in: a ShadowRoot when the panel
+  // lives in shadow DOM, else document.head. (A bare `document.appendChild`
+  // throws — the document may hold only one element child.)
+  const r = root?.getRootNode?.();
+  const target =
+    r && typeof r.appendChild === "function" && r.nodeType === 11 /* DOCUMENT_FRAGMENT (ShadowRoot) */
+      ? r
+      : document.head;
+  target.appendChild(el);
 }
 
 /** Pull human text out of a tool_result frame (result = MCP content array). */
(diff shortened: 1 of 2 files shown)


artokun:
Review-only pass (this branch's session owns landing). Land **after** mcp #251 (this consumes its frames + whitelisted tools; against an old orchestrator it degrades to a "Local" pill and "tool not permitted" modal errors — no crash). Gates on this head: `npm run typecheck` clean, `npm run test:unit` 67/67.

0. **Scope note:** this branch stacks the LoRA training-wizard commit (53ab720 — `py/training_routes.py`, ~1000 lines of `cmcp-training-ui.js`) under the two runpod commits; that's most of the +1768. If training is meant to be PR #96's payload, rebase this onto main after #96 lands; otherwise retitle so it's reviewed here. (FWIW `training_routes.py` path handling looks sound — realpath + containment on all three routes, fail-closed.)

1. **MEDIUM (money) — `web/js/cmcp-runpod-ui.js:286-303`: deploy confirm-arming defeats itself on a double-click.** Click 1 arms, click 2 confirms — an accidental double-click IS the confirmation, and Enter/Space keyboard repeat is even faster. Add a short post-arm cool-down (~500ms ignoring confirms) or a separate confirm button, so billing requires two distinct human decisions.

2. MEDIUM — the arming is client-side only: any bridge-authed client (mobile, mirrored tab) can call `runpod_pod_create` unconfirmed — the real guard must be server-side (flagged on mcp #251 as a `confirm: true` param); pass it here on the armed click once it exists.

3. LOW/MEDIUM — `cmcp-runpod-ui.js:60-66` vs `:166`: style injection contradicts the body-mount fix — `injectStyle()` targets the panel's ShadowRoot, but the overlay moved to `document.body` (outside it), so shadow-DOM mounts render the modal unstyled and `styleInjected` latches. Inject into `document.head` unconditionally.

4. LOW (honesty) — "Stop ends billing" overstates: stop ends GPU-time billing; the volume bills until terminated (console-only). Suggest "Stop ends GPU billing (disk storage still bills until you terminate in the console)".

5. LOW (honesty) — the `onPod` fallback (`cmcp-runpod-ui.js:187` + `reflectRunpodHost()`): with no `comfyui_target` frame yet, it infers "on pod" from `watching && RUNNING` — but `runpod_watch` explicitly watches WITHOUT retargeting, so the pill can claim "Rendering on RunPod" while renders are local. Default to local when the target is unknown; it's the dishonest branch of an honesty feature.

6. LOW (UX) — `podInput` has no Enter handler; and Stop uses `currentPodId()` which prefers the watched pod over the typed one (type pod B, press Stop → watched pod A stops). Use only the watched id for Stop, or comment the intent.

Frame handling is additive and clean (explicit type checks, per-tab handlers, the RunPod key never reaches the panel). Landing reminders: plugin cache + marketplace copies need syncing per CLAUDE.md, and this pairs with an mcp release per the coordinated-release convention.

Issue #98, opened 2026-07-21
Add durable workflow-scoped chat history with restart and multi-tab safety


JusticeWay:
## Status — 2026-07-28

This proposal began against Panel 0.8.2. The implementation now follows the four focused stages requested by the maintainer and is based on the 0.11.2 line.

| Stage | Status | Result |
|---|---|---|
| PR 1 — workflow UUID identity + strict scope guard | ✅ Merged: #101 (`525116a`) | Stable workflow identity, clone/fork handling, exact scope checks, transcript identity separated from bridge routing |
| PR 2 — IndexedDB canonical storage + migration | ✅ Merged: #106 (`acd910e`) | Durable schema-v3 store, localStorage startup shadow, causal multi-tab merge, tombstones/checkpoint fencing, canonical clear-all |
| PR 3 — archive UX + transcript hydration | 🟠 Open: #153 (`cb07566`) | Multi-chat archive, scope modes, search/pin/rename/delete, safe portable import/export, workflow-version context, stale/foreign-session hydration |
| PR 4 — optional server backup | ⏳ Deferred | Will be rebased and security/Registry-audited only after PR 3 merges |

PR #153 is rebased onto current `main` (`6aa98c1`, Panel 0.11.2). Its branch also contains the #154 reconnect fix and the GLM provider changes from #155/#156.

## Goal and invariant

Agent Panel transcripts should survive full browser, ComfyUI, and orchestrator restarts without ever displaying or resuming another workflow's conversation through a stale pointer.

The default remains panel-owned continuity: the orchestrator rebinds the live agent as workflow tabs change. A workflow UUID is a transcript/archive scope key only; bridge routing keeps its independent `wf:` / `tmp:` tab identity.

### Workflow identity — #101

- `extra.comfyui_mcp.workflow_uuid` provides stable transcript identity.
- Rename and Save As preserve continuity for the same live workflow.
- Copies/forks receive independent identities, including restart-safe unsaved forks.
- Opening a workflow does not dirty it; the UUID rides the next user save.
- Scope checks use exact UUID equality only.
- Panel-owned default behavior stays a zero-delta path.

### Durable browser store — #106

- IndexedDB `comfyui-mcp-panel-history`, schema v3, is canonical.
- A bounded localStorage shadow provides instant paint and degraded-storage operation.
- Canonical limits are 500 chats × 5,000 records; the normal shadow is 20 × 200.
- Concurrent tabs merge causally at message, thread, and metadata levels.
[...]

PR #101, opened 2026-07-21
Add workflow-scoped chat identity

JusticeWay:
## Summary

Introduce a stable workflow UUID for transcript scoping and enforce an exact-match guard before any workflow-owned chat/session is restored or resumed.

This is PR 1 of the focused series requested in #98. It remains limited to the identity and safety foundation; IndexedDB storage, archive UX, and optional server backup are kept in separate follow-up branches.

## What changed

- persist `extra.comfyui_mcp.workflow_uuid` plus the last known workflow path only when the first chat entry is recorded;
- embed identity silently, so merely opening/switching a workflow never dirties the canvas;
- keep transcript identity (`workflow:<uuid>`) separate from bridge routing (`wf:<path>` / `tmp:<uuid>`);
- keep default panel-owned behavior unchanged and preserve its ride-along workflow provenance;
- apply strict scope guards only in per-workflow mode, including load and reload restore;
- avoid destructive thread re-keying during the settings-hydration race;
- preserve identity across rename/Save As and reuse a copied workflow's path alias across restarts until it is saved;
- migrate an exact legacy `wf:<full-path>` transcript once, with no fuzzy matching;
- retain the live tab `SESSION_KEY` during a scope correction;
- use a weak/bounded workflow owner registry and locale-independent path normalization;
- dim disabled foreign-workflow history rows.

## Review follow-up

Commit `984d3e9` addresses every item from the maintainer review. It also adds regressions for opening without dirtying, default-mode opening of pre-upgrade threads, hydration safety, stable unsaved fork identity, and foreign-row presentation.

## Validation

- `npm run typecheck`
- `npm run test:unit` — 74/74 passed on this branch
- ES module parse checks — passed
- `npm audit --omit=dev` — 0 vulnerabilities
- targeted Playwright (`workflow-chat-identity.spec.ts`) — 4/4 passed

Part of #98.

### DIFF
--- a/browser_tests/unit/workflow-chat-identity.test.mjs
+++ b/browser_tests/unit/workflow-chat-identity.test.mjs
@@ -4,7 +4,8 @@ import test from 'node:test'
 import {
   isThreadInScope,
   normalizedWorkflowPath,
-  shouldForkEmbeddedWorkflowUuid
+  shouldForkEmbeddedWorkflowUuid,
+  workflowAliasForPath
 } from '../../web/js/lib/workflow-chat-identity.js'
 
 test('normalizes Windows paths for stable identity comparisons', () => {
@@ -51,6 +52,13 @@ test('keeps the canonical path when stale aliases still mention the same UUID', 
       'workflows/current.json': 'same-uuid'
     }
   }), false)
+})
+
+test('reuses the path alias minted for an unsaved fork after a browser restart', () => {
+  assert.equal(workflowAliasForPath({
+    'workflows/original.json': 'embedded-original',
+    'Workflows\\Copy.JSON': 'stable-fork'
+  }, 'workflows/copy.json'), 'stable-fork')
 })
 
 test('scope guard authorizes only an exact workflow UUID key', () => {
--- a/browser_tests/unit/workflow-chat-identity.test.mjs
+++ b/browser_tests/unit/workflow-chat-identity.test.mjs
@@ -1,0 +1,62 @@
+import assert from 'node:assert/strict'
+import test from 'node:test'
+
+import {
+  isThreadInScope,
+  normalizedWorkflowPath,
+  shouldForkEmbeddedWorkflowUuid
+} from '../../web/js/lib/workflow-chat-identity.js'
+
+test('normalizes Windows paths for stable identity comparisons', () => {
+  assert.equal(
+    normalizedWorkflowPath('Workflows\\Portrait.JSON'),
+    'workflows/portrait.json'
+  )
+})
+
+test('forks a copied workflow with an embedded UUID on a clean browser', () => {
+  assert.equal(shouldForkEmbeddedWorkflowUuid({
+    embeddedUuid: 'same-uuid',
+    embeddedPath: 'workflows/original.json',
+    currentPath: 'workflows/copy.json'
+  }), true)
+})
+
+test('keeps identity for the same live object during rename or Save As', () => {
+  assert.equal(shouldForkEmbeddedWorkflowUuid({
+    objectUuid: 'same-uuid',
+    embeddedUuid: 'same-uuid',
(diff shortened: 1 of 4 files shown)


artokun (changes requested):
Thanks for turning the split around this fast — and for keeping PR 1 genuinely scoped to identity + guard. I ran the gates locally: typecheck clean, 73/73 unit green as you said, and both files parse clean as ES modules. The core direction is exactly what we agreed in #98: `isThreadInScope` is strict everywhere (no fuzzy anywhere, verified), the legacy `wf:<path>` migration is exact-full-path and provably one-shot, bridge routing (`wf:`/`tmp:`, bind-before-re-hello) is untouched — I traced it against the orchestrator's `migrated_from` path and it composes. The extracted pure lib + unit tests are the right shape. Nice work.

Four things need to change before I can take this out of draft, all fixable without redesign:

1. **Opening a workflow must never dirty it.** In per-workflow mode, `onWorkflowMaybeChanged` computes `historyKey` with `embed: true`, so merely switching to a workflow writes `extra.comfyui_mcp` under `beforeChange/afterChange` and sets `isModified` — a user who opens someone's workflow to look at it gets a dirty asterisk and a save prompt. Please drop `embed` from the switch handler entirely, and when you do embed (first `record()` is fine), write `extra` *silently* — no `beforeChange/afterChange`, no `setDirtyCanvas`, no `isModified` — so the UUID simply rides the next save the user initiates. The alias map already covers identity until that save happens.

2. **The restore path can destroy a per-workflow thread via the settings-hydration race** you yourself defend against in `record()`: at mount, `sessionFollowsPanel()` reads `true` before settings hydrate, so `restoreLastThread` rewrites a per-workflow user's pointed thread to `panel:global` and persists it. Never make a destructive key rewrite based on a possibly-unhydrated setting.

3. **Default mode must be a zero-delta path.** Right now panel-owned mode (a) re-keys threads to `panel:global` instead of the ride-along `wfid` (that provenance is exactly what PR 3's archive grouping will want), and (b) the new `loadThread` guard blocks opening any pre-upgrade thread from the history popup — the rows are still clickable in default mode, they just error. Simplest fix that also solves (2): keep main's behavior byte-for-byte when `sessionFollowsPanel()`, and make the scope guard + new keys active only in per-workflow mode.

4. **Fork identity isn't stable across restarts.** `workflowStableUuid` prefers the embedded UUID over `aliases[path]`, and the fork branch always mints a fresh UUID — so a copy that's opened, forked, chatted-in but not yet saved re-forks to a *new* UUID on every browser restart, orphaning its own transcript. When forking, reuse `aliases[currentPath]` if it exists and differs from the embedded UUID.

Smaller asks: don't null `SESSION_KEY` in `record()`'s scope reset (it's the live tab session pointer; clearing `CURRENT_THREAD_KEY` alone is right); `_workflowUuidOwners` holds workflow objects strongly forever — prune or use `WeakRef`; `toLocaleLowerCase` → `toLowerCase` in `normalizedWorkflowPath` (Turkish-I makes identity locale-dependent); a dimmed style for the disabled history rows; and once (3) lands, a small default-mode Playwright case asserting old threads still open would lock the invariant in.

None of this shakes the foundation — the UUID-as-transcript-scope/path-as-routing split is right and PRs 2–4 can build on it once the embed timing and the default-mode path are nailed down. Looking forward to the next revision.

JusticeWay:
I agree with you and have made edits based on your feedback.

artokun (approved):

artokun:
@JusticeWay merged — thanks again for the clean turnaround. Green light to move on to the rest of the series from #98: PR 2 (IndexedDB transcript storage), then the archive UX, then the optional server backup. Same shape as this one please — focused scope, the pure-lib + unit-test split, and rebase each on current main (this merge is `525116a`) so the identity foundation is under you.

PR #106, opened 2026-07-21
Add durable IndexedDB transcript storage

JusticeWay:
## Summary

Make Agent Panel transcripts survive full ComfyUI/browser restarts by moving canonical chat storage to IndexedDB while retaining a bounded localStorage startup/fallback shadow.

This is PR 2 of the focused series requested in #98. It is rebased on current `main` (`add9920`, Agent Panel 0.11.0), contains the merged workflow identity foundation from #101, and composes with the current tabbed side-panel UI.

Archive UX and optional server backup remain separate follow-up PRs.

## What changed

- Added a pure `ChatHistoryStore` library with schema-v3 normalization and a fenced one-way legacy-message migration.
- IndexedDB is the canonical atomic snapshot store.
- A bounded localStorage shadow provides synchronous startup, fallback, and migration from existing installs.
- Concurrent thread and message updates merge with observed monotonic causal revisions.
- Mutable session, TODO, provenance, title/pin, and card state have independent revisions, so stale appends cannot roll them back.
- Causal message and thread tombstones prevent stale tabs from resurrecting deleted content.
- Canonical post-merge limits are 500 conversations and 5,000 messages per conversation.
- Tombstone and metadata operation maps are bounded through canonical checkpoints.
- Committed history synchronizes across tabs, with BroadcastChannel invalidation when the localStorage shadow cannot be written.
- Dirty history is retained and retried when both persistence backends fail.
- Reload restores the active panel-owned conversation without switching to a newer background conversation.
- Per-workflow mode keeps exact UUID-only selection and clears transcript, TODO, and session state after strict scope rejection.
- Workflow alias set/delete operations publish immediately and survive stale-tab rename and compaction cases.
- Backend switch and hard restart durably invalidate the previous agent session before reconnect.
- Canonical and shadow eviction preserve the current tab's active thread.
- Synchronous reload restore remains paint-only until settings and IndexedDB hydrate.
- Initial hello/connect waits for final settings-aware thread and session binding.
- Pending persistence flushes on visibility and page lifecycle boundaries.
- The history popover explains where transcripts persist and provides a confirmed **Clear all history** action.

## Clear-all semantics

Clear All performs an atomic canonical reset rather than a bare `indexedDB.deleteDatabase()`:

1. Re-read and merge the latest canonical state inside one IndexedDB `readwrite` transaction.
2. Write an empty transcript checkpoint with an incremented generation.
3. Preserve materialized `workflowAliases`, because stable workflow identity is metadata rather than chat content.
4. Clear transcripts, active chat pointers, chat tombstones, and operation maps.
5. Write an empty localStorage shadow.
6. Broadcast a reset invalidation so other open ComfyUI tabs reload the canonical checkpoint.

Each affected tab also clears visible transcript, TODO, and session state and sends `new_session`, so backend memory cannot continue a conversation the user deleted.

If the canonical IndexedDB transaction cannot commit, the action fails closed and leaves the shadow untouched. It does not briefly claim success and then rehydrate old history on reload.

A plain database deletion would erase workflow UUID aliases and allow a still-open stale tab to recreate the database and republish deleted transcripts. The checkpoint reset prevents both failure modes.

## Complexity decision

The merge/checkpoint implementation is intentionally retained for PR 2.

Replacing it with per-thread last-write-wins would be a semantic redesign rather than a mechanical simplification. It would remove the reviewed guarantees for:

- concurrent messages in the same thread;
- durable deletes;
- independently revised mutable fields;
- backward clocks;
- bounded compaction;
- stale-tab fencing.

Those guarantees match #98's durable and correctly scoped multi-tab history requirement and now have deterministic regressions.

The implementation surface is substantial, so this is a conscious tradeoff. A later behavior-preserving extraction or startup-write optimization can be reviewed separately after the storage semantics land. The existing startup `load()` migration write is unchanged in this PR.

## Storage bounds

| Store | Limit |
|---|---:|
| Canonical IndexedDB history | 500 conversations |
| Canonical transcript | 5,000 messages per conversation |
| localStorage startup shadow | 20 conversations / 200 messages each |
| Tombstone and metadata operation maps | Bounded through canonical checkpoints |

## Invariants retained from PR 1

- Workflow UUID is a transcript scope key, not a bridge routing key.
- Bridge `wf:` / `tmp:` routing and the orchestrator rebind path are unchanged.
- Default panel-owned mode remains compatible with pre-upgrade threads and keeps ride-along workflow provenance.
- Strict exact-match scope selection is active only in per-workflow mode.
- Opening a workflow does not dirty it.
- Clearing chat history does not remove workflow UUID identity.

## Scope

This PR intentionally does not add archive search/pin/export controls, workflow snapshot hydration, or Python backup routes. Those remain PR 3 and PR 4.

## Validation

Rebased HEAD: `d5f56d1`

- `npm run test:unit` — **139/139 passed**.
- `npm run typecheck` — passed.
- `node --check` for both reviewed modules — passed.
- `git diff --check` — passed.
- `npm audit --omit=dev` — **0 vulnerabilities**.
- `conversation-persistence.spec.ts` — **7/7 passed** against live ComfyUI at `127.0.0.1:8188` with one Chromium worker.
- GitHub Actions lint/security run `29992435822` — passed.

The new browser case covers the complete destructive path: confirmation UI and persistence note, atomic IndexedDB empty state, local tab/session reset, preserved workflow alias, reload persistence, and rejection of a stale tab's attempt to republish the pre-clear snapshot.

The rest of the browser suite covers full restart, IndexedDB-only recovery, pointed-thread/session hydration, strict workflow detachment, individual-delete resurrection, and alias rename/compaction echo prevention.

Part of #98.

### DIFF
--- a/browser_tests/conversation-persistence.spec.ts
+++ b/browser_tests/conversation-persistence.spec.ts
@@ -225,6 +225,57 @@ test('restores from IndexedDB after the localStorage shadow is lost', async ({
   await expect(panel.agentBubbles.filter({ hasText: 'indexeddb-only reply' }).last()).toBeVisible()
 })
 
+test('panel delete remains final when a stale tab republishes the removed thread', async ({
+  page,
+  context,
+  panel,
+  mockBridge
+}) => {
+  await panel.goto()
+  await panel.setBridgeUrl(mockBridge.url)
+  await panel.openSidebar()
+  await panel.connect()
+
+  const received = mockBridge.waitForUserMessage()
+  await panel.sendMessage('delete me causally')
+  await received
+  await expect.poll(() => indexedHasText(page, 'delete me causally')).toBe(true)
+  const removedThreadId = await page.evaluate((key) => sessionStorage.getItem(key), CURRENT_THREAD_KEY)
+  expect(removedThreadId).not.toBeNull()
+
+  const staleTab = await context.newPage()
+  await staleTab.goto(page.url())
+  const staleSnapshot = await staleTab.evaluate((snapshotKey) =>
+    JSON.parse(localStorage.getItem(snapshotKey) || '{}'), LOCAL_HISTORY_SNAPSHOT_KEY)
+  expect(staleSnapshot.threads?.some((thread: any) => thread.id === removedThreadId)).toBe(true)
+
+  await panel.root.getByTitle('Chat history').click()
+  await panel.root.locator('.cmcp-hist-row', { hasText: 'delete me causally' })
+    .getByTitle('Delete this chat')
+    .click()
+  await expect.poll(() => indexedHasThread(page, removedThreadId!)).toBe(false)
+  await expect.poll(() => page.evaluate(({ snapshotKey, threadId }) => {
+    const snapshot = JSON.parse(localStorage.getItem(snapshotKey) || '{}')
+    const tombstone = snapshot.meta?.deletedThreads?.[threadId]
+    return tombstone?.deleted === true && Number.isFinite(tombstone?.revision?.updatedAt)
+  }, { snapshotKey: LOCAL_HISTORY_SNAPSHOT_KEY, threadId: removedThreadId! })).toBe(true)
+
+  await staleTab.evaluate(async (snapshot) => {
+    const { ChatHistoryStore } = await import('/extensions/comfyui-mcp-panel/lib/chat-history-store.js')
+    const staleStore = new ChatHistoryStore({ writerId: 'stale-panel-test' })
+    staleStore.persist(snapshot.threads || [], snapshot.meta || {})
+    const result = await staleStore.flush()
+    if (result !== true && result?.ok !== true) throw new Error(`stale write failed: ${JSON.stringify(result)}`)
+    await staleStore.close?.()
+  }, staleSnapshot)
+
+  await page.reload()
+  await panel.openSidebar()
+  await expect(panel.userBubble('delete me causally')).toHaveCount(0)
+  await expect.poll(() => indexedHasThread(page, removedThreadId!)).toBe(false)
+  await staleTab.close()
+})
+
 test('reload keeps the pointed conversation and live tab session during durable hydration', async ({
   page,
   panel,
--- a/browser_tests/conversation-persistence.spec.ts
+++ b/browser_tests/conversation-persistence.spec.ts
@@ -48,26 +48,26 @@ async function indexedHasText(page: import('@playwright/test').Page, text: strin
   }, text)
 }
 
-async function seedReloadRace(
+async function seedReloadEvictionRace(
   page: import('@playwright/test').Page,
   currentThreadId: string
 ): Promise<void> {
   await page.evaluate(async ({ pointedId }) => {
     const future = Date.now() + 60_000
-    const background = {
-      id: 'newer-background-thread',
+    const background = Array.from({ length: 500 }, (_, i) => ({
+      id: `newer-background-thread-${i}`,
       schemaVersion: 2,
-      workflowKey: 'workflow:background-race',
-      createdAt: future - 100,
-      updatedAt: future,
-      ts: future,
+      workflowKey: `workflow:background-race-${i}`,
+      createdAt: future + i,
+      updatedAt: future + i,
+      ts: future + i,
       msgs: [{
-        id: 'newer-background-message',
+        id: `newer-background-message-${i}`,
         role: 'user',
-        text: 'newer background transcript',
-        createdAt: future
+        text: i === 0 ? 'newer background transcript' : `background transcript ${i}`,
+        createdAt: future + i
       }]
-    }
+    }))
     const rewrite = (snapshot: any) => {
       const threads = Array.isArray(snapshot?.threads) ? snapshot.threads : []
       const pointed = threads.find((thread: any) => thread.id === pointedId)
@@ -79,7 +79,10 @@ async function seedReloadRace(
       return {
         ...snapshot,
         updatedAt: future,
-        threads: [...threads.filter((thread: any) => thread.id !== background.id), background],
+        threads: [
+          ...threads.filter((thread: any) => !thread.id?.startsWith('newer-background-thread-')),
+          ...background
+        ],
         meta: {
           ...(snapshot?.meta || {}),
           updatedAt: future,
@@ -220,7 +223,7 @@ test('reload keeps the pointed conversation and live tab session during durable 
   )
   expect(currentThreadId).not.toBeNull()
   await expect.poll(() => indexedHasText(page, 'conversation selected by this tab')).toBe(true)
-  await seedReloadRace(page, currentThreadId!)
+  await seedReloadEvictionRace(page, currentThreadId!)
 
   await page.evaluate(() => localStorage.removeItem('comfyui-mcp.panel.autoConnect'))
   await page.reload()
@@ -232,10 +235,20 @@ test('reload keeps the pointed conversation and live tab session during durable 
     () => page.evaluate((key) => sessionStorage.getItem(key), SESSION_KEY)
   ).toBe('live-tab-session')
 
-  // Cross the settings-hydration boundary and the async IndexedDB load. Neither
-  // is allowed to repaint another conversation or replace the live tab session.
-  await page.waitForTimeout(3_000)
+  // The active metadata rewrite occurs only after settings + IndexedDB hydration,
+  // so this is a deterministic completion signal for the final binding.
+  await expect.poll(() => page.evaluate(() => {
+    const meta = JSON.parse(localStorage.getItem('comfyui-mcp.panel.historyMeta') || '{}')
+    return meta.activeByScope?.['panel:global'] || null
+  })).toBe(currentThreadId)
   await expect(panel.userBubble('conversation selected by this tab')).toBeVisible()
   await expect(panel.userBubble('newer background transcript')).toHaveCount(0)
   expect(await page.evaluate((key) => sessionStorage.getItem(key), SESSION_KEY)).toBe('live-tab-session')
+  expect(await page.evaluate((pointedId) => {
+    const shadow = JSON.parse(localStorage.getItem('comfyui-mcp.panel.threads') || '[]')
+    return {
+      count: shadow.length,
+      hasPointed: shadow.some((thread: any) => thread.id === pointedId)
+    }
+  }, currentThreadId)).toEqual({ count: 20, hasPointed: true })
 })
--- a/browser_tests/conversation-persistence.spec.ts
+++ b/browser_tests/conversation-persistence.spec.ts
@@ -380,6 +380,16 @@ test('workflow rename publishes alias tombstones and a stale tab cannot echo the
   panel,
   mockBridge
 }) => {
+  await page.route(
+    (url) => /\/(api\/)?settings\/?$/.test(url.pathname),
+    async (route) => {
+      if (route.request().method() !== 'GET') return route.continue()
+      const response = await route.fetch()
+      const settings = await response.json() as Record<string, unknown>
+      settings['comfyui-mcp.sessionFollowsPanel'] = false
+      await route.fulfill({ response, json: settings })
+    }
+  )
   await panel.goto()
   await panel.openSidebar()
   await page.evaluate(() => {
--- a/browser_tests/conversation-persistence.spec.ts
+++ b/browser_tests/conversation-persistence.spec.ts
@@ -48,6 +48,77 @@ async function indexedHasText(page: import('@playwright/test').Page, text: strin
   }, text)
 }
 
+async function seedReloadRace(
+  page: import('@playwright/test').Page,
+  currentThreadId: string
+): Promise<void> {
+  await page.evaluate(async ({ pointedId }) => {
+    const future = Date.now() + 60_000
+    const background = {
+      id: 'newer-background-thread',
+      schemaVersion: 2,
+      workflowKey: 'workflow:background-race',
+      createdAt: future - 100,
+      updatedAt: future,
+      ts: future,
+      msgs: [{
+        id: 'newer-background-message',
+        role: 'user',
+        text: 'newer background transcript',
+        createdAt: future
+      }]
+    }
+    const rewrite = (snapshot: any) => {
+      const threads = Array.isArray(snapshot?.threads) ? snapshot.threads : []
+      const pointed = threads.find((thread: any) => thread.id === pointedId)
+      if (pointed) {
+        pointed.sessionId = 'stale-stored-session'
+        pointed.updatedAt = future - 1
+        pointed.ts = future - 1
+      }
+      return {
+        ...snapshot,
+        updatedAt: future,
+        threads: [...threads.filter((thread: any) => thread.id !== background.id), background],
+        meta: {
+          ...(snapshot?.meta || {}),
+          updatedAt: future,
+          activeByScope: {
+            ...(snapshot?.meta?.activeByScope || {}),
+            'panel:global': 'missing-active-pointer'
+          }
+        }
+      }
+    }
+
+    const localThreads = JSON.parse(localStorage.getItem('comfyui-mcp.panel.threads') || '[]')
+    const localMeta = JSON.parse(localStorage.getItem('comfyui-mcp.panel.historyMeta') || '{}')
+    const local = rewrite({ threads: localThreads, meta: localMeta })
+    localStorage.setItem('comfyui-mcp.panel.threads', JSON.stringify(local.threads))
+    localStorage.setItem('comfyui-mcp.panel.historyMeta', JSON.stringify(local.meta))
+
+    const db = await new Promise<IDBDatabase>((resolve, reject) => {
+      const request = indexedDB.open('comfyui-mcp-panel-history', 2)
+      request.onsuccess = () => resolve(request.result)
+      request.onerror = () => reject(request.error)
+    })
+    try {
+      await new Promise<void>((resolve, reject) => {
+        const tx = db.transaction('snapshots', 'readwrite')
+        const store = tx.objectStore('snapshots')
+        const get = store.get('state')
+        get.onsuccess = () => store.put(rewrite(get.result || {}), 'state')
+        get.onerror = () => reject(get.error)
+        tx.oncomplete = () => resolve()
+        tx.onerror = () => reject(tx.error)
+        tx.onabort = () => reject(tx.error)
+      })
+    } finally {
+      db.close()
+    }
+  }, { pointedId: currentThreadId })
+}
+
 test('restores the latest panel-owned conversation after sessionStorage is lost', async ({
   page,
   panel,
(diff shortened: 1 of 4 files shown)


artokun:
Reviewed read-only (didn't touch the branch). The core is solid: IndexedDB is canonical with a localStorage startup shadow, `idbMergeWrite` re-reads and merges under the same readwrite transaction (concurrent-tab safe), tombstones stop deleted threads resurrecting, and message-level UUID union-merge handles two tabs writing the same thread. ESM-valid, no duplicate top-level decls, 81/81 unit tests pass including the new store suite, and it composes correctly with #101 — `selectThreadForScope` still routes through the exact-match `isThreadInScope` guard and the `panel:global` selection is metadata-only, so the #98 scope guard holds. Nice work.

A few things before this comes out of draft:

1. **Panel reload-restore now rewrites session keys.** The synchronous restore path used to deliberately avoid writing keys ("Settings may not be hydrated yet, so this path must never rewrite keys"), but it now does `ssSet(CURRENT_THREAD_KEY, …)` and `ssSet(SESSION_KEY, panelThread.sessionId || null)` even on the normal `pointed`-exists path. Can you confirm writing `SESSION_KEY` from the thread's stored `sessionId` before settings hydrate can't bind a stale session — this is the one documented invariant the PR changes (it's the same hydration-race class that the PR-1 review flagged).

2. **Async `historyStore.load()` can switch the visible conversation on reload.** After the sync restore paints, the background load merges IndexedDB and may call `loadThread(durableActive)`; in `sessionFollowsPanel()` mode `selectPanelThread` returns newest-by-`updatedAt` across all threads, so if a background workflow thread is newer than the one just painted, the panel could repaint to a different conversation (and flash). Is that intended for the global-panel session, and is there a guard against surprise-switching?

3. **E2E not run in review.** `conversation-persistence.spec.ts` (the full-restart + IndexedDB-only recovery cases) is Playwright and isn't in `test:unit` — please run the browser specs before this comes out of draft; they're the real proof of the restore path.

4. Minor: the DB opens at `version === CHAT_HISTORY_SCHEMA` (2) for a brand-new store with no in-`onupgradeneeded` data migration (migration happens at the app layer via `mergeHistorySnapshots`) — fine as-is, but worth a comment that a future schema bump will auto-fire `onupgradeneeded`. And `MAX_THREADS`/`MAX_THREAD_MSGS` jumped 20→500 / 200→5000 in the panel; confirm that's the intended durable cap.

With #1 and #2 answered and the Playwright specs green, this looks close to ready — really strong PR 2 of the series.

JusticeWay:
Addressed all four points in `9fb3cc8` and rebased the branch onto current `main` (`fcad938`).

1. **No pre-hydration key rewrite:** the synchronous reload path is now paint-only. It reads the exact tab pointer, but never infers scope, writes `CURRENT_THREAD_KEY`/`SESSION_KEY`, updates active metadata, persists, or sends a session frame. Final binding waits for both IndexedDB and the existing settings-hydration boundary. For the pointed conversation, the live `sessionStorage` value is authoritative even when empty; a stored `thread.sessionId` is used only for full-restart recovery where no tab pointer survived. Initial hello/connect also waits for that reconciliation.
2. **No surprise conversation switch:** the new pure `selectRestoreThread` prefers the conversation already selected by this tab. In per-workflow mode that preference still passes the strict UUID scope guard. Only a full restart with no surviving tab pointer uses durable active/newest fallback, so a newer background-workflow thread cannot repaint the visible chat.
3. **Browser proof:** `conversation-persistence.spec.ts` is 3/3 green against a live ComfyUI. The third case seeds both failure modes directly: a newer background transcript plus a stale stored session id, then crosses the settings/IndexedDB hydration boundary and verifies the visible chat and live tab session remain unchanged. The four workflow identity/default-mode browser cases also pass in the serial suite. Unit is now 104/104, typecheck and ESM parse are clean, and `npm audit --omit=dev` reports 0 vulnerabilities.
4. **Schema/caps:** added the requested `onupgradeneeded` comment. The 500-thread / 5,000-message limits are intentional canonical IndexedDB caps; the synchronous localStorage shadow remains bounded at 20/200.

For transparency, the whole current-main serial suite is 26/32 in this local setup; the remaining failures are confined to unrelated CivitAI/Training drive fixtures and Kimi provider discovery. The persistence and workflow-identity suites are green. Marking the PR ready for review.

artokun (changes requested):
Reviewed the four points. **1 and 3 hold up under inspection** — nice work on those:

- The synchronous pass really is paint-only. The IIFE at `comfyui-mcp-panel.js:14724` does nothing but `paintThread()` — no `ssSet`, no `setActiveThread`, no persist, no session frame. I went looking for indirect writes (helpers that persist as a side effect, handlers firing during paint) and found none.
- `:14774` uses `reloadSessionId` directly with no `||` fallback when the thread is the tab's own, so an empty-but-present value stays authoritative and can't silently resurrect a stale session id.

Also confirmed independently: unit is **104/104** on the branch, and the new storage keys (`comfyui-mcp.panel.*`, IndexedDB `comfyui-mcp-panel-history`) don't collide with the panel's existing `cmcp.*` keys.

**Claims 2 and 4 don't hold, and they share one root cause.**

`comfyui-mcp-panel.js:14741`:

```js
threads = merged.threads.slice(-MAX_THREADS);
```

`mergeHistorySnapshots` returns threads sorted **oldest-first**, so this keeps the newest 500 on pure recency — **nothing pins the thread this tab is currently showing**. The same shape appears at `:10215` and `:10271`.

Once the pointed thread is sliced away, `selectRestoreThread`'s `preferred` lookup returns `null` and it falls through to `selectPanelThread` → newest thread. So a **surviving** tab pointer still reaches the fallback, which is what point 2 says can't happen — and eviction *can* drop the active thread, which is what point 4 says it can't.

Reproduced against the real module:

```js
import { selectRestoreThread } from "./web/js/lib/chat-history-store.js";
const threads = [...Array(501)].map((_, i) => ({ id: "t" + i, updatedAt: 1000 + i, msgs: [{ id: "m" + i }] }));
const kept = threads.slice(-500);              // what line 14741 does
selectRestoreThread(kept, { activeByScope: {} },
  { panelOwned: true, scopeKey: "panel:global", preferredThreadId: "t0" });
```

```
tab was showing : t0
survived slice  : false
repainted to    : t500      <-- wrong conversation
```

Needs >500 threads, so it's not an everyday path — but it's the exact failure this PR exists to prevent, and the fix looks small: keep the pointed/active thread when slicing instead of trusting recency alone.

**One thing worth checking that may be much more reachable.** The shadow is capped at `LOCAL_SHADOW_THREADS = 20`, and the catch at `:14748` deliberately supports running on the localStorage shadow alone when IndexedDB is unavailable (private browsing, blocked storage). If the initial `threads` value is shadow-capped in that mode, the same logic would repaint users off any thread outside the newest **20** — no 500-thread collection required. I haven't confirmed that path, so treat it as a question rather than a second finding; you're better placed to say.

Everything else looks good, and the design is sound — this is a targeted fix, not a rethink.

<sub>Review by Claude. Disclosure: a Codex pass was started on this first and wedged after 26 minutes without producing findings, so the analysis above is mine.</sub>

JusticeWay:
Good catch — both the 500-thread canonical cap and the 20-thread shadow had the same pure-recency flaw.

Fixed in `211045d` with a shared pure `retainBoundedThreads` helper. Every cap site now keeps a strict maximum while prioritizing:

1. the current browser tab's pointed/painted thread;
2. durable `activeByScope` pointers;
3. the newest remaining threads.

That replaces all three `slice(-MAX_THREADS)` shapes you called out. `ChatHistoryStore.persist/load` now accepts the tab pointer as protected input, and the shadow writer also automatically protects metadata-active threads before filling the remaining 20 slots by recency.

The shadow concern is confirmed: before this change, a pure-recency shadow could omit an older active thread. Going forward the 20-thread shadow retains it, so an IndexedDB-unavailable reload can recover it. If an older build has already evicted the record from localStorage *and* IndexedDB is unavailable, those missing bytes cannot be reconstructed; when IndexedDB is available again, load recovers the canonical record and rewrites a correctly pinned shadow.

Regression proof:

- the exact 501-thread shape now retains `t0`, evicts `t1`, retains `t500`, and `selectRestoreThread(... preferredThreadId: t0)` returns `t0`;
- protected ids exceeding the cap still obey the strict limit and priority order;
- an IndexedDB-null store persists and reloads the active old thread in a 20/21 shadow case;
- the Playwright race now injects 500 newer background threads, reloads the panel, waits for final settings-aware binding, and verifies the pointed conversation/session stay unchanged and that the 20-thread shadow still contains the pointed thread.

Validation: unit **106/106**, `conversation-persistence.spec.ts` **3/3**, typecheck clean, both ES modules parse clean.

JusticeWay:
I apologize for any mistakes. The thing is, I don't have enough time to devote myself 100% to this project, but I'm doing my best to actively participate in the development of this wonderful and important project.

In addition, I'm ready to provide you with ongoing support and work on interesting and important features.

JusticeWay:
Thanks for the careful counterexample. You were right: retaining the pointed thread had to happen inside every canonical/shadow eviction path, and the 20-thread IndexedDB-unavailable fallback needed the same guarantee.

I fixed those cases first, then ran three independent deep review/fix passes over the complete PR 2 diff. Those passes uncovered additional stale-tab, deletion, checkpoint, migration, and persistence-failure interleavings that the earlier green suite did not cover. The branch is now updated through `5fcda5e`.

Highlights of the resulting hardening:

- active-aware canonical and shadow eviction retains the current tab thread, then durable active pointers, then newest records;
- causal message/thread tombstones prevent stale writers from resurrecting deleted content;
- post-merge IndexedDB limits are enforced atomically;
- per-field and per-message HLC-style revisions prevent stale appends or backward clocks from rolling back session/TODO/provenance/card state;
- malformed local tombstones/ops and forged local checkpoints cannot delete canonical IndexedDB history;
- workflow alias deletes publish immediately and remain deleted through cross-tab sync and compaction;
- legacy id-less messages use a fenced one-way schema-v3 migration, including stale pre-migration writers;
- strict per-workflow detach clears transcript/TODO/session before the next record;
- backend switch/hard restart durably invalidate the old session before reconnect;
- total localStorage + IndexedDB failure returns a retryable failure and retains the dirty snapshot rather than claiming a successful flush;
- bounded checkpoints, BroadcastChannel fallback, and channel cleanup are covered.

Validation on the pushed HEAD:

- unit: **127/127**;
- Playwright persistence: **6/6** against live ComfyUI (`127.0.0.1:8189`, Chromium, one worker);
- typecheck and both module syntax checks: pass;
- `git diff --check`: pass;
- `npm audit --omit=dev`: 0 vulnerabilities.

The new browser cases explicitly cover panel deletion followed by stale-tab republication, alias tombstone compaction followed by a stale unrelated append, strict workflow session detachment, and the original pointed-thread hydration/eviction invariants.

The PR body is updated with the current schema, invariants, bounds, and validation. No archive UI or Python surface was added; PR 2 remains limited to the same four files and storage/migration scope requested in #98.

JusticeWay:
Rebased PR 2 onto current `main` / Agent Panel **0.10.0** (`971a1ba`) as requested for the focused series.

- The 0.10.0 dock/slide changes and version bump are present.
- The PR diff remains limited to the same four history/storage files.
- Rebase completed without conflicts.
- New pushed HEAD: `835d134`.
- A recovery branch preserves the exact pre-rebase state: `JusticeWay/backup/pr2-pre-010-rebase-20260722` at `5fcda5e`.

The current live ComfyUI settings endpoint exposed one test-fixture edge during verification: it can transiently return an empty response body. Production code was unaffected; the Playwright interceptor now safely supplies the single deterministic per-workflow setting instead of calling `response.json()` unconditionally (`835d134`).

Post-rebase validation:

- unit: **127/127**;
- Playwright persistence: **6/6** against live ComfyUI `127.0.0.1:8188`;
- typecheck and both module syntax checks: pass;
- `git diff --check`: pass;
- `npm audit --omit=dev`: 0 vulnerabilities;
- Codex live readiness/catalog still reports signed-in Sol/Terra/Luna through the official `comfyui-mcp 0.45.0` runtime.

The PR body has been updated to the new base and validation endpoint.

artokun (commented):
Reviewed the full store module and the entire wiring surface. This is high-quality work — thank you for the thoroughness (and the heavier test coverage than implementation).

## Strengths (verified, not assumed)
- **Well isolated** — the CRDT logic lives in `web/js/lib/chat-history-store.js`; the monolith only gains wiring.
- **No new XSS surface** — no `innerHTML`/`insertAdjacentHTML` added; restored transcripts still render through the existing `paintThread`/`renderTodo`.
- **Sync→async handled** — instant paint from the localStorage shadow, then `historyRestoreReady`-gated durable hydration (and there's an explicit reload-during-hydration test).
- **Graceful degradation** — `openDb`'s 2s timeout + null fallback means a disabled/private-mode IndexedDB or a blocked cross-version upgrade degrades to shadow-only instead of hanging.
- **Durable deletes are correct** — per-thread delete writes a `deletedThreads` tombstone that merges into canonical IDB (covered by the "stale tab republishes the removed thread" test). Prototype-pollution guarded via `Object.create(null)` maps.

## Please address before merge
1. **No "clear all history" escape hatch now that history is durable.** Clearing browser localStorage no longer clears chat history — the canonical IndexedDB copy re-hydrates on next load — and there's no bulk-clear affordance in the panel, while individual deletes are now durable. That leaves a user with no in-panel way to get a clean slate (they'd have to clear site data manually). Could you add a **"Clear all history"** action that `indexedDB.deleteDatabase`s (or tombstones every thread) and clears the shadow? A one-line note on where transcripts now persist would also help.

## Non-blocking notes
2. **Complexity vs. requirement.** ~1200 lines of merge/checkpoint/fence logic is a lot of surface area to maintain for browser chat history. It's defensively written and well-tested, so it's fine — but it's only worth it if durable cross-tab *conflict-free* sync is a real requirement. If per-thread last-write-wins would satisfy the actual need, this is more machinery than the problem strictly demands. Flagging for a conscious call, not asking you to change it.
3. `load()` triggers a `persist()` (a write) on every startup for auto-migration — harmless at chat frequency, just noting it.

I did not hand-trace every branch of the CRDT merge — leaning on the 127-test unit suite for those, which looks thorough. Nice work overall; once the clear-all path lands I think this is good to go.

JusticeWay:
Thanks — agreed on the missing escape hatch, and I also took the complexity note as a request for an explicit design decision rather than as a request to rewrite the store during this review.

Addressed on rebased HEAD `d5f56d1` (current `main` `add9920`, Agent Panel 0.11.0).

### Clear All

The history popover now includes:

- a one-line persistence note: transcripts are stored in this browser using IndexedDB;
- a **Clear all history** action, disabled when history is empty;
- an explicit destructive confirmation that says it affects every workflow/open tab, cannot be undone, and does not delete workflows or their stable identities;
- an actionable failure message if the canonical reset cannot be committed.

The implementation deliberately does **not** call `indexedDB.deleteDatabase()` directly. That superficially clears the data, but it has two correctness problems here:

1. `workflowAliases` are stable workflow identity metadata from PR 1, not transcript content; deleting the database would remove them too.
2. Any open stale tab still holding the pre-clear snapshot could recreate the database and republish the deleted conversations.

Instead, Clear All performs one atomic canonical reset:

1. open a single IndexedDB `readwrite` transaction;
2. re-read and merge the latest canonical snapshot inside that transaction;
3. write an empty transcript checkpoint with `generation + 1`;
4. preserve the materialized workflow aliases while clearing threads, active chat pointers, chat tombstones, and operation maps;
5. write an empty localStorage shadow;
6. broadcast a `history-reset` invalidation to the other tabs.

The advanced checkpoint fences every snapshot created before the clear, so a stale tab cannot resurrect old history. Each affected live panel also clears its visible transcript/TODO/session state and sends `new_session`; that prevents the next prompt from continuing backend memory for a conversation the user believes was deleted.

The operation fails closed if IndexedDB cannot commit. In that case it leaves the localStorage shadow untouched and reports failure instead of briefly looking empty and then rehydrating the old canonical history on reload.

### Complexity call

I agree that ~1,200 lines is substantial surface area. I am consciously retaining the current merge/checkpoint implementation for PR 2.

Changing it to per-thread last-write-wins now would not be a mechanical simplification: it would remove the already-reviewed guarantees for two tabs appending to the same thread, durable deletes, independently revised session/TODO/provenance/card fields, backward clocks, bounded compaction, and stale-tab fencing. Those are the failure modes uncovered while implementing #98's durable, correctly-scoped multi-tab history, and they now have deterministic regressions.

So the tradeoff is: keep the larger, isolated, heavily tested pure module for the semantics already under review; do not mix a storage-model redesign into the final correctness fix. A later behavior-preserving extraction or simplification can be reviewed separately after the semantics land. Likewise, I left the harmless startup `load()` migration write unchanged; making it conditional would be an optimization with its own migration/compaction edge cases and is better kept out of this PR.

### Regression proof

Added three store regressions plus one full browser scenario covering:

- reset creates an empty checkpoint while preserving workflow aliases;
- a second store receives the reset broadcast;
- a stale pre-clear tab cannot republish deleted transcripts;
- alias identity survives that stale write;
- IndexedDB-unavailable Clear All fails closed without mutating the shadow;
- the actual UI confirmation, persistence note, canonical empty state, transcript/session reset, `new_session`, reload, and stale-tab attempt.

Post-rebase validation:

- unit: **139/139**;
- Playwright persistence: **7/7** against live ComfyUI, Chromium, one worker;
- typecheck: clean;
- both ES modules parse clean;
- `git diff --check`: clean;
- `npm audit --omit=dev`: 0 vulnerabilities.

The PR body now documents the reset semantics and this complexity decision. For transparency, GitHub briefly kept `refs/pull/106/head` on the pre-rebase SHA even though the fork ref had updated; I refreshed the PR metadata after restoring its link. It is now open, mergeable, and correctly reports head `d5f56d1` on base `add9920`; CI is running.

artokun (approved):
Verified the clear-all escape hatch on rebased HEAD `d5f56d1` — the blocking item from my last review is fully addressed. Approving.

## What I checked (ran, not assumed)

**Store (`clearAll` / `createHistoryResetSnapshot`)**
- The reset happens in a single IndexedDB readwrite transaction that re-reads canonical, merges the tab's local snapshot, then writes the empty checkpoint — no read-then-write race window.
- The checkpoint **generation advances**, so a stale pre-clear tab that later persists its old snapshot gets fenced instead of resurrecting threads (confirmed by both the unit test and the live e2e).
- `workflowAliases` survive the reset while `activeByScope`/`activeOps`/`deletedThreads` are emptied — clearing transcripts doesn't destroy workflow identity, exactly the right call.
- Fails **closed** with `history-clear-canonical-unavailable` when IndexedDB is unavailable, leaving the shadow untouched rather than diverging shadow from canonical.

**Panel wiring**
- Persistence note + destructive confirm + disabled-when-empty all present; on success the tab pointers (`CURRENT_THREAD_KEY`/`SESSION_KEY`/`CTX_KEY`) are cleared and a `new_session` frame is sent, so backend agent memory doesn't outlive the visible clear. The remote-delete path sending `new_session` too is a nice catch beyond what I asked for.
- All identifiers the click handler touches resolve inside the `buildPanel` closure — no TDZ/scope hazards at click time.

**Tests**
- Unit: **139/139** pass on the branch.
- E2E: ran the full 7-spec `conversation-persistence` suite against live ComfyUI 0.28.2 / frontend 1.45.21 — **7/7 pass**, including the clear-all fence test… with one caveat below.
- Branch merges clean with current `main` (`660a45f`).

## Non-blocking notes

1. **Hardcoded extension mount in the new spec.** `conversation-persistence.spec.ts` imports `/extensions/comfyui-agent-panel/js/lib/chat-history-store.js` in 5 places. That matches the registry name and the documented prerequisite in `playwright.config.ts`, so you followed convention — but my dev junction is named after the repo (`comfyui-mcp-panel`), and 2 of 7 tests fail with a module-fetch 404 until the path is substituted. A follow-up that derives the mount at runtime (probe the candidates, or resolve the panel's own script URL from `performance.getEntriesByType("resource")`) would make the suite portable across junction names. Happy to take that as a tiny separate PR.
2. **Shadow-only mode can't clear.** When IndexedDB is permanently unavailable (private browsing), `clearAll` fails closed and the panel says "Close other ComfyUI tabs, then try again" — advice that can't succeed in that mode. Cosmetic; history there is localStorage-only and clearable via browser site data, but the message could acknowledge it.

Thank you for the sustained care across this whole review cycle — PR 2 of the #98 series is ready. Looking forward to PR 3 (archive UI).

<sub>Review by Claude. Verification (unit run, live e2e run, merge-tree check, scope-resolution audit) performed independently on the rebased head.</sub>

artokun:
Merged 🎉 — thank you for the persistence (pun intended) across this whole cycle: three rebases, a deep self-review pass that caught interleavings our suite didn't cover, and a clear-all implementation that landed exactly on the invariant. This is PR 2 of the #98 series done, and the foundation for everything after it.

One transparency note: `main` moved again right before merge (#143's working-indicator fix touched `newChat`/`paintThread`), so rather than ask you for a fourth rebase I pushed the conflict resolution to your branch myself (`159d831`). Two things worth your eyes post-hoc:

- `paintThread` keeps your `persist: false` todo render and gains #143's `if (agentWorking) bumpThinking()` re-pin.
- Your clear-all handler gained `endTurnLocally()` before `resetFeed()` — same invariant #143 added to `newChat`, since clearing history abandons any in-flight turn. Without it, a live turn's indicator would rebuild onto the freshly-cleared empty chat.

Verified on the merge result before landing: 139/139 unit, 30/30 e2e against live ComfyUI. If you disagree with any part of the resolution, say so and we'll adjust in a follow-up.

**Next: PR 3 — the archive UI** (multi-conversation per workflow, search/pin/rename, export/import, the "current workflow only" filter, and your "Ask on workflow switch" scope mode). That's where the UX review happens, so screenshots/GIFs in the PR description will speed things up a lot. The two non-blocking notes from my review (runtime-derived extension mount in the spec, shadow-only clear messaging) are welcome as tiny riders or a separate PR — your call.

Genuinely great work. Looking forward to it.

artokun:
### codex:rescue review (round 5) — 4 findings, all fixed in `715b2c0` and merged with latest main

After the clean auto-merge with main (#141 etc.) and a fresh `codex exec review --base main`:

1. **[P1] A2UI record identity broken** — `record()` spread-cloned entries, so card resolve/dismiss/ui_update mutations landed on a detached clone and never persisted. `record()` now mutates in place; card records stay the same objects as `thread.msgs` entries.
2. **[P2] Missing causal creation stamps** — new threads/messages synthesized `createdRevision` from `createdAt`, so a checkpoint at/ahead of wall clock could classify genuinely new history as pre-checkpoint and drop it. `reviseThread`/`touchMessage` now stamp write-once causal `createdRevision`s.
3. **[P2] Live card bindings stale after remote merge** — cross-tab merges cloned every message; cards kept mutating the old records. The merge now rebinds `liveA2uiCards` by message id.
4. **[P2] Dead-panel callbacks** — a `readCanonical` started before unsubscribe could still deliver into a destroyed panel. Delivery is now gated on the subscription still being active.

Verification: 142 unit (+3 regressions) and the full Playwright suite green against a live ComfyUI — including all 7 conversation-persistence specs. The only remaining failures are the two pre-existing sets (agent-drive parallel flake + workflow-chat-identity, both reproducing on plain main).

@JusticeWay — pushed these on top of your branch to land it; the store design held up well through five rounds. Thanks for the patient iteration.

PR #114, opened 2026-07-22
feat(apps): micro-apps — workflow→app conversion, local/RunPod runs, publish/explore registry

artokun:
## The Apps epic (panel side)

A WeChat-mini-program-style micro-app layer on top of ComfyUI APP mode. Convert any workflow into a named one-click **app** (name, description, thumbnail, chosen input/output endpoints), run it headless locally or on a connected RunPod pod, package it with best-effort hidden workflow, and publish it to a Cloudflare-backed registry with stars/trending/explore — with full parity in the mobile app.

**Sibling PRs:** comfyui-mcp (apps_* bridge tools + mobile whitelist) and comfyui-mcp-mobile (Apps tab + detail/run + Explore). Monetization (hosted runs, per-gen pricing) is design-notes-only — schema hooks reserved (`pricing_json`, `hosted_only`), nothing built. See `docs/design/apps-monetization.md`.

### What's here

- **`py/apps_routes.py`** — bundle storage under `user/comfyui-mcp-panel/apps/<uuid>/` (deliberately NOT the workflows dir, so hidden apps never surface there), CRUD routes, run engine (patch `<nodeId>.<widget>` values into the stored API snapshot → queue), `dry:true` mode for pod runs, `/bundle` export for publish/install.
- **`web/js/cmcp-apps.js`** — `AppBuilder` (defensive ComfyUI APP-mode config import, heuristic input/output selection, widget classification, deps scan), `AppsClient`, `RegistryClient`, creator/star identity keys.
- **`web/js/cmcp-apps-ui.js`** — the Apps modal (toolbar button next to Civitai): My Apps grid, convert flow over the live canvas, generated run form, local run + polling + output gallery, honest hide-workflow toggle (deletes the graph; copy says plainly it's obfuscation, not security), **Run on RunPod** (dry-patch → `enqueue_workflow` on the connected pod; CivitAI-pinned models pushed first, unpinned deps warned not silently skipped), Publish/Update, Explore tab, install with deps-consent dialog, `ran` reporting.
- **`registry/`** — dependency-free Cloudflare Worker (D1 + R2): publish with sha256-keyed creator identity (DB stores only the hash), keyset pagination, trending = 7d stars×3 + runs, star/unstar, run marks (1/app/day), reports. Excluded from the pack via `.comfyignore`.
[...]

PR #125, opened 2026-07-22
feat(apps): transfer image inputs to the pod before RunPod runs

artokun:
Follow-up to the Apps epic (#114) — removes the honest-refusal guard from the codex round: **pod runs with image inputs now work end to end**.

- The panel bridge client gains `uploadMedia(blob, name)`: the same cid-correlated `upload_media` frame the mobile app uses. The orchestrator's existing handler writes the bytes to the CONNECTED ComfyUI's `input/` — the pod when the session is on a pod — so **no new server surface** (`media_uploaded` replies route into the same pending-call map as `tool_result`).
- The Apps detail form keeps image inputs as raw Files and the run path picks the destination: local runs upload same-origin to `/upload/image` (unchanged); pod runs transfer through the bridge FIRST, then dry-patch with the pod-side filename and enqueue.

**Tests:** pod e2e spec seeds a `LoadImage` input and asserts the bytes go through the bridge and the enqueued prompt carries the pod-side filename; full apps spec 6/6 green against a live ComfyUI serving this branch; 104 unit green; `tsc` clean.

Drive-by fixture fix: `connect()` dismisses the connection popover (main's #116/#117 dropdown semantics leave it open over the toolbar, which was blocking toolbar clicks in every bridge-connecting spec).

Note: verified against a worktree-serving spare ComfyUI instance (the junctioned checkout is on another agent's branch); no real pod run — the pod path is mocked-bridge e2e.

### DIFF
--- a/browser_tests/apps.spec.ts
+++ b/browser_tests/apps.spec.ts
@@ -288,7 +288,10 @@ test('run on RunPod: dry-patches locally, enqueues on the pod via the bridge', a
       version: 1,
       hideWorkflow: false,
       appMode: {
-        inputs: [{ nodeId: 6, widget: 'text', label: 'Prompt', kind: 'text', default: 'a cat' }],
+        inputs: [
+          { nodeId: 6, widget: 'text', label: 'Prompt', kind: 'text', default: 'a cat' },
+          { nodeId: 5, widget: 'image', label: 'Face', kind: 'image' }
+        ],
         outputs: [],
         importedFromFrontend: false
       },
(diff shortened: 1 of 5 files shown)


artokun:
### codex:rescue review — 2 findings, both fixed in `c5cf48a`

Ran `codex exec review --base main` on the rebased branch (post-registry-deploy commit). Both findings were real; fixes pushed, full apps spec 6/6 green against a live worktree-serving ComfyUI, 104 unit green, `tsc` clean.

1. **[P1] Filename collisions on pod uploads** — two image inputs sharing a basename (or a repeat run with `image.png`) would overwrite each other in the pod's `input/`, silently substituting the wrong image into one or both nodes. Fixed: uploads now go under a uniquified remote name (`cmcp-app-<app>-<rand>-<orig>`); e2e asserts the shape.
2. **[P2] 60s upload timeout abandoned valid uploads** — large images over slow pod links legitimately exceed a minute. Raised to 180s, matching the mobile client's budget for the same frame.

PR #129, opened 2026-07-23
fix(apps): convert flow offered zero inputs on current frontends

artokun:
Two bugs found **dogfooding** the Apps convert flow end to end (real ComfyUI, real registry):

1. **Widget-socket guard treated every widget as link-driven.** Modern frontends materialize an input socket (`link: null`) for EVERY widget, so "has a widget input" excluded all candidates — a plain EmptyImage→Preview graph offered **zero** inputs. Now only a CONNECTED widget-input (`link != null`) excludes a widget.
2. **`INPUT_HINT_TYPES` missed the empty-latent/image generators** (EmptyImage, EmptyLatentImage, EmptySD3LatentImage, LTXV/Hunyuan/Mochi latents) — their size/batch widgets are among the most-exposed app endpoints.

**Verification:** e2e fixture now carries an unconnected widget socket (stays a candidate) and a connected one (excluded) with pick-list assertions — 6/6 green against a live ComfyUI. Plus the full real drive: convert → run (256px swatch, real generation) → publish → explore → star, all against the production registry (which now hosts its first app, `artokun/solid-color-swatch-4b041650` ★1).

### DIFF
--- a/browser_tests/apps.spec.ts
+++ b/browser_tests/apps.spec.ts
@@ -182,11 +182,11 @@ async function stubAppsBackend(page: Page) {
           version: 0.4
         }
       })
-      const node = (id: number, type: string, widgets: unknown[], outputNode = false) => ({
+      const node = (id: number, type: string, widgets: unknown[], outputNode = false, inputs: unknown[] = []) => ({
         id,
         type,
         title: type,
-        inputs: [],
+        inputs,
         widgets,
         constructor: { nodeData: { output_node: outputNode } }
       })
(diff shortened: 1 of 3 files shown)


artokun:
### codex:rescue review — clean, no findings

`codex exec review --base main` verified the fix against current ComfyUI workflow slot semantics (`link: null` for unconnected widget inputs vs a link id when connected) and the expanded node set against the live node sources: *'No actionable regression was found.'* Also re-ran the 104 node unit tests successfully on its side. Ready to merge.

PR #140, opened 2026-07-24
feat(civitai): rating badges + level-narrowed favorites (catchall)

artokun:
Follow-up to #134.

- **Colour-coded rating badges** on every card — PG green, PG-13 blue, R orange, X/XXX red.
- **Favorites now narrow by browsing level.** Previously the tab fetched the whole likes collection and showed out-of-level items as black "gated" placeholders, so picking PG showed all 18 (4 real + 14 black). Now it shows **only** the selected level(s). **Catchall:** if the exhausted feed has nothing at the chosen level, it falls back to showing all favorites (out-of-level as gated) so a strict pick never dead-ends on an empty grid.

Live-verified against a real 18-item collection: PG→4, R→1, XXX→11; PG-13 (0 owned) → catchall shows all 18.

### DIFF
--- a/web/js/cmcp-civitai-ui.js
+++ b/web/js/cmcp-civitai-ui.js
@@ -86,6 +86,15 @@ function injectCss() {
     font-size: .7rem; color: #fff; background: linear-gradient(transparent, rgba(0,0,0,.75)); }
   .cmcp-cv-badge { position: absolute; top: .3rem; left: .3rem; background: rgba(0,0,0,.6); color:#fff;
     font-size: .6rem; padding: .1rem .3rem; border-radius: 4px; }
+  /* Rating badge (top-right), colour-coded by browsing level. */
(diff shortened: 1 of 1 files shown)


artokun:
### codex:rescue review — 2 findings, both fixed in 6eb7101 + merged with latest main

1. **[P2] Rating badge hidden on Favorites** — the persistent like button shared the badge's exact top-right coordinates and covered it on every favorite card. Like button (and the owned marker, same collision) now sit one row below the badge.
2. **[P2] Badge missing on model-result tabs** — Checkpoints/LoRAs/Workflows render via `modelCard`, which never got the badge. Added; model results carry `nsfwLevel` already.

Full e2e on the merged branch: 28 passed; the 7 failures are the two pre-existing sets (agent-drive parallel flake + workflow-chat-identity, both reproducing on plain main).

PR #141, opened 2026-07-24
feat(apps): registry app view — inputs immediately, star icon, requirements panel

artokun:
Requested UX changes to the registry app view:

- **No Install button / consent gate.** Opening a registry app silently installs the bundle (idempotent) and lands straight in the detail — the input form is visible immediately. First Run just runs.
- **Star is a ☆/★ icon next to the title** (with live count), initialized from the new `GET /v1/apps/{id}/starred` endpoint — no more action-row Star button.
- **New Requirements side panel** (local + registry apps): models show ✓ when installed, ⬇ to download via the bridge when pinned (CivitAI version id or URL), or a 'manual' note; custom node class_types show ✓ when present in `/object_info`, else 'missing' with Manager/agent guidance. Replaces the old install-time confirm dialog — requirements are visible *before* the first run.

**Worker:** `/starred` lookup + tests, **already redeployed to production**.
**Tests:** the explore spec covers the no-gate flow (inputs visible, no Install button), star icon toggle + count, and the requirements rows including a download click flipping to ✓. 6/6 spec + 104 unit green against a live ComfyUI; `tsc` clean.

### DIFF
--- a/browser_tests/apps.spec.ts
+++ b/browser_tests/apps.spec.ts
@@ -525,12 +525,18 @@ async function stubRegistry(page: Page) {
               outputs: [],
               importedFromFrontend: false
             },
-            deps: { models: [], customNodes: [] }
+            deps: {
+              models: [{ name: 'flux.safetensors', widget: 'ckpt_name', civitaiVersionId: 12345 }],
+              customNodes: ['FancyCustomNode']
+            }
           },
           prompt: FIXTURE_PROMPT,
           workflow: FIXTURE_WORKFLOW
         }
       })
+    }
+    if (req.method() === 'GET' && url.pathname.endsWith('/starred')) {
+      return route.fulfill({ json: { starred: false } })
     }
     if (req.method() === 'GET' && url.pathname.endsWith('/thumbnail')) {
       return route.fulfill({ status: 404, json: { error: 'no thumbnail' } })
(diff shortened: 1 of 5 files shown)


artokun:
### codex:rescue review — 6 findings, all fixed in `52bc156`

**P1s**
1. **60s default abandoned real downloads** — model downloads routinely exceed `callTool`'s default timeout; the UI reported failure while the backend kept going, inviting duplicates. Both download calls now use a 20-minute budget (matching the CivitAI downloader).
2. **Check/download target mismatch** — `/models` checked the local ComfyUI while `download_civitai_model` wrote to the connected pod, so a row could show ✓ while the local Run still lacked the model. Install-state checks now go through the SAME channel as downloads (`list_local_models` over the bridge), local `/models` only as offline fallback.

**P2s**
3. Dep `targetSubfolder` now wins over the widget heuristic (pinned LoRAs no longer land in `checkpoints`).
4. Model matching compares complete normalized paths, not basenames (`foo/x.safetensors` ≠ `bar/x.safetensors`).
5. Star button disabled until `/starred` resolves — a click during the window toggled from the wrong default and corrupted the count.
6. Folder model-list cache invalidated after a download — reopening the app no longer shows the fresh model as missing.

Spec updated (star init-gate assertion, bridge-path model checks); 6/6 spec + 104 unit green.

PR #149, opened 2026-07-24
fix(history): stop dropping legacy shadow-only threads; update identity specs to the #106 contract

artokun:
Bisected the red workflow-chat-identity specs to #106 (acd910e) and found a **real data-loss bug** plus two specs asserting pre-#106 behavior.

**Store fix (the bug):** a legacy-idless local shadow (any pre-v3/idless message present) was rejected wholesale by the schema-3 canonical fence — and `persist()` then erased those shadow-only threads from localStorage too. Foreign/legacy conversations vanished on hydration. Rejected threads are now retained flagged `legacyShadow`: visible in history, scope-disabled, and excluded from the canonical write in `idbMergeWrite` (the fence's canonical protection is unchanged). +1 unit regression (143 green).

**Spec updates (behavior contracts):**
- The hydration test asserted pre-#106 behavior (loose sessions adopted cross-workflow). Rewritten to #106's explicit contract: a workflow-scoped conversation NEVER adopts a loose tab session — the key is cleared instead of mis-bound; the old thread keeps its own scope binding.
- The foreign-thread injection now goes through the store API with proper message ids — the real cross-tab write path. Direct localStorage injection was this tab's own cache and is legitimately overwritten by the owning panel's flush.

workflow-chat-identity 4/4 green, conversation-persistence 7/7 green, full suite green apart from the pre-existing agent-drive parallel flake (verified isolated).

### DIFF
--- a/browser_tests/unit/chat-history-store.test.mjs
+++ b/browser_tests/unit/chat-history-store.test.mjs
@@ -1458,3 +1458,46 @@ test('unsubscribe suppresses a pending readCanonical delivery (codex: dead-panel
   await new Promise((resolve) => setTimeout(resolve, 0))
   assert.equal(calls, 0, 'a read started before unsubscribe must not deliver')
 })
+
+test('legacy-idless shadow-only threads are retained in the shadow, fenced out of canonical (no data loss)', async () => {
+  const now = Date.now()
+  const canonical = {
+    schemaVersion: CHAT_HISTORY_SCHEMA,
+    updatedAt: now - 1000,
+    meta: {
+      checkpoint: { generation: 3, revision: { updatedAt: now - 1000, writerId: 'w1', sequence: 1 } },
+      activeByScope: {},
+      workflowAliases: {}
+    },
+    threads: [
+      { id: 'marker', workflowKey: 'workflow:a', createdAt: now - 500, updatedAt: now - 500,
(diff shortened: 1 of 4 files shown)


artokun:
### codex:rescue round 2 — 3 findings, all fixed in `d2d9c22` (plus a laundering hole found while fixing)

1. **[P1] Canonical commits erased the quarantine** — `idbMergeWrite` resolved a canonical-only snapshot and `persist()` rewrote the local shadow from it, deleting every retained thread on the next cycle. It now writes canonical-only but resolves the full merged set.
2. **[P1] Shadow cap could evict the only copy** — the 20-thread shadow cap exempts `legacyShadow` threads.
3. **[P2] Non-durable editing** — `legacyShadow` rows are read-only in every mode, not just per-workflow.
4. **Laundering (found during the fix):** `persist()` normalizes ids *before* the fence evaluates, so quarantined threads would pass as id-ed and merge into canonical on the next persist. The fence now keys on the `legacyShadow` flag itself (persist + readLocal). The per-thread partition from round 2 was reverted — it admitted exactly the shifted-duplicate content the fence exists to reject.

145 unit green (fence test intact), workflow-chat-identity 4/4, conversation-persistence 7/7, full suite green apart from the pre-existing agent-drive parallel flake.

PR #153, opened 2026-07-26
feat(history): add workflow chat archive UI

JusticeWay:
## Summary

PR 3 of the durable chat-history series tracked in #98, following merged #101 and #106.

- multiple conversations per workflow with search, current-workflow filtering, pin, rename, delete, and active-chat selection
- portable JSON merge import/export with a 25 MB input limit and bounded workflow snapshots
- explicit Panel / Workflow / Ask-on-switch scope modes, with migration from the legacy boolean
- archive grouping by stable workflow identity across saves and renames
- provider/model/effort metadata plus workflow-version context for each user turn
- bounded transcript hydration when a stored provider session is missing, stale, or belongs to another provider

## Current base

Rebased onto current `main` at `6aa98c1` (Panel 0.11.2). This includes the merged PR 2 store, its post-merge rescue fixes, the #154 reconnect fix, and the GLM provider work from #155/#156.

Workflow UUID remains a transcript/archive scope key only. Bridge routing and the panel-owned live-agent rebind path are unchanged.

## Data-safety response to the latest review

The four findings from the maintainer's adversarial pass are fixed and regression-tested:

1. **Import cannot evict local chats or messages.** Before changing state, import computes all genuinely new thread/message IDs. If the result would exceed 500 chats or 5,000 entries in any chat, the entire import fails with an actionable error. It never relies on recency capping after the merge, so no existing local transcript is silently displaced.
2. **Same-ID import is add-only.** Existing local messages and every existing thread field win regardless of foreign timestamps or raw `fieldOps`. An archive may add a missing message or workflow version, but cannot replace/delete local content, title, todos, workflow metadata, provider/session provenance, or an existing workflow-version record. Duplicate imported records share one remaining-version budget, so they cannot crowd local graph versions out.
3. **IndexedDB fallback no longer reports truncated durability as success.** A complete 20×200 localStorage shadow can still serve as the degraded durable store. Once that shadow is incomplete, persistence returns a retryable failure, retains the full snapshot in memory, and displays a clear warning that overflow is not durable. Recovery retries on the next mutation.
4. **Workflow aliases are bounded.** Both the causal operation map and the materialized alias map are compacted to 512 entries behind a checkpoint. Import fills only free alias slots, preserves all local aliases, and reports how many portable aliases were skipped.

The review also exposed two adjacent only-copy cases, fixed in the same data-safety pass:

- quarantined `legacyShadow` transcripts are exempt from the 20-chat/200-entry shadow caps because the IndexedDB fence deliberately excludes them;
- failure to save those only-copy transcripts to localStorage is no longer masked by a successful canonical write for the rest of history.

The import policy is deliberately fail-closed rather than “protect locals and evict imported rows”: users get an atomic, explainable result, and the success message remains truthful. Colliding IDs are deliberately add-only because an imported archive is untrusted portable data, not an authority allowed to rewrite local causal history.

## Other deep-audit fixes retained

- portable export omits browser-local sessions, active pointers, checkpoints, CRDT operations, and tombstones;
- future schemas are rejected and raw payloads are reduced to the portable allowlist;
- active thread and live A2UI records are rebound after import;
- provider-mismatched or rejected stale sessions start fresh with bounded transcript hydration;
- archive switching ends the previous local in-flight turn before repaint/resume;
- identical workflow display titles remain separate UUID/path groups;
- workflow snapshot limits use UTF-8 bytes;
- clear-all remains canonical and cross-tab safe.

## Verification

Current head: `cb07566`.

- `npm run test:unit` — **162/162 passed**
- `npm run typecheck` — passed
- relevant Playwright history/persistence/identity suite — **15/15 passed** serially against live ComfyUI 0.28.3
- `node --check web/js/comfyui-mcp-panel.js` — passed
- `node --check web/js/lib/chat-history-store.js` — passed
- `git diff --check` — passed
- `npm audit --omit=optional --audit-level=high` — **0 vulnerabilities**

Part of #98.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Added
+- **Chat archive UI** — multiple named or pinned conversations per workflow,
+  workflow-grouped search and filtering, JSON merge import/export, explicit
+  Panel / Workflow / Ask-on-switch scope modes, provider/model metadata, and
+  per-user-message workflow-version context. Missing provider sessions continue
+  from a bounded transcript replay. Portable imports cannot carry foreign
+  sessions, checkpoints, tombstones, or active pointers into local history;
+  provider-mismatched sessions start fresh with bounded transcript hydration.
+  Import is add-only on colliding IDs and fails closed at thread/message caps
+  instead of evicting local history. Materialized metadata maps are bounded, and
+  an IndexedDB outage now warns when the small localStorage shadow is incomplete.
+  Legacy shadow-only transcripts are never silently truncated by those caps.
+
 ## [0.11.0] - 2026-07-22
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Each provider runs its own orchestrator on its own loopback port (Claude on
 | **Reasoning effort + model selector** | A per-provider effort/model picker — Claude's models, or ChatGPT's GPT-5-class models via Codex — moved into the model selector; a chosen effort survives a provider switch by mapping to the nearest valid level. |
 | **Pending-message tray + reconnect durability** | Queue messages while the agent is busy (edit / send-now / reorder), and reclaim a wedged orchestrator on Connect. |
 | **Apps (micro-apps)** | Convert any workflow into a named, one-click **app** (name, description, thumbnail, chosen input/output endpoints) from the toolbar's **Apps** button — runs headless (the canvas is never touched), locally or on a connected RunPod pod, with best-effort hidden-workflow packaging and a publish/explore registry (`registry/`, Cloudflare Worker + D1 + R2) shared with the mobile app. |
+| **Durable workflow chat history** | Keep multiple named or pinned chats per workflow, search across them, export/import JSON, survive ComfyUI/browser restarts through IndexedDB, and track the graph version used by every user turn. |
 
 ## Quickstart
 
--- a/browser_tests/chat-history-v2.spec.ts
+++ b/browser_tests/chat-history-v2.spec.ts
@@ -1,0 +1,386 @@
+import { test, expect } from './fixtures/panelTest'
+import { resolveHistoryStoreModuleUrl } from './fixtures/historyStoreModule'
+
+const THREADS_KEY = 'comfyui-mcp.panel.threads'
+const META_KEY = 'comfyui-mcp.panel.historyMeta'
+
+async function forceWorkflowScope(page: import('@playwright/test').Page) {
+  await page.evaluate(() => {
+    const w = window as any
+    const app = w.comfyAPI?.app?.app || w.app
+    const settings = app.ui.settings
+    if (!w.__cmcpOriginalGetSettingValue) {
+      w.__cmcpOriginalGetSettingValue = settings.getSettingValue.bind(settings)
+    }
+    settings.getSettingValue = (id: string) =>
+      id === 'comfyui-mcp.chatScope'
+        ? 'workflow'
+        : w.__cmcpOriginalGetSettingValue(id)
+  })
+}
+
+async function indexedThreadCount(page: import('@playwright/test').Page): Promise<number> {
+  return page.evaluate(async () => {
+    const db = await new Promise<IDBDatabase>((resolve, reject) => {
+      const req = indexedDB.open('comfyui-mcp-panel-history', 3)
+      req.onsuccess = () => resolve(req.result)
+      req.onerror = () => reject(req.error)
+    })
+    try {
+      return await new Promise<number>((resolve, reject) => {
+        const req = db.transaction('snapshots', 'readonly').objectStore('snapshots').get('state')
+        req.onsuccess = () => resolve(Array.isArray(req.result?.threads) ? req.result.threads.length : 0)
+        req.onerror = () => reject(req.error)
+      })
+    } finally {
+      db.close()
+    }
+  })
+}
+
+async function indexedHasMessage(
+  page: import('@playwright/test').Page,
+  text: string
+): Promise<boolean> {
+  return page.evaluate(async (wanted) => {
+    const db = await new Promise<IDBDatabase>((resolve, reject) => {
+      const req = indexedDB.open('comfyui-mcp-panel-history', 3)
+      req.onsuccess = () => resolve(req.result)
+      req.onerror = () => reject(req.error)
+    })
+    try {
+      return await new Promise<boolean>((resolve, reject) => {
+        const req = db.transaction('snapshots', 'readonly').objectStore('snapshots').get('state')
+        req.onsuccess = () => resolve(
+          (req.result?.threads || []).some((thread: any) =>
+            thread.msgs?.some((message: any) => message.text === wanted))
+        )
+        req.onerror = () => reject(req.error)
+      })
+    } finally {
+      db.close()
+    }
+  }, text)
+}
+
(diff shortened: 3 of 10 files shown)


JusticeWay:
Rebased this onto current `main` (`55535ce`) and did a fresh end-to-end audit against all of the post-merge notes in #106 before asking for review.

That audit found and fixed three correctness issues in my original #153: a raw imported checkpoint could delete local history, the active thread/A2UI records were detached after import, and an archive row could pass a session ID to a different provider. I also fixed title-based workflow grouping, in-flight archive switching, UTF-8 snapshot sizing, and import count reporting. The PR body now documents the exact portable-import boundary and every reconciliation point with #106.

Verification on head `33b3a25`: 155/155 unit, 15/15 relevant Playwright E2E (serial, live ComfyUI 0.28.3), typecheck, JS syntax, diff check, npm audit, and upstream CI are all green. Ready for another review.

artokun:
## Review — data-safety (codex adversarial pass + verification)

Strong PR overall: the export allowlist, future-schema rejection, `meta`-reduced-to-aliases on import, active-thread + A2UI rebind, provider-mismatch session handling, and UTF-8-measured byte limits all check out. The core "foreign checkpoint/tombstone can't enter via `meta:{}`" guard is sound.

**But an adversarial pass surfaced real data-loss edges that a first read missed** — flagging before merge:

### P1 — Import can silently evict existing local chats at the 500-thread cap
After the merge, `capHistoryThreads(imported.threads, currentThreadId)` runs and persists the reduced set (`comfyui-mcp-panel.js:12335`). Retention protects only active/current ids and otherwise keeps newest; imported threads get fresh (newest) revisions (`chat-history-store.js:1587-1591`), so they push out the oldest **local** threads. **Repro:** 500 local chats incl. one old inactive → import 1 chat → the old local chat is silently dropped. Worse, the toast says *"existing history was preserved."* (`:12345`) which is then false. Fix: protect existing local ids across an import cap, or don't cap-evict non-imported threads on import.

### P1 — Same-ID imported records can overwrite local content/metadata
Colliding message ids: a later portable `updatedAt` replaces the local message body (`chat-history-store.js:1594-1608`). A newer imported thread also revises local `title`/`todos`/workflow fields, and missing values become deletion ops (`:1623-1636`). Raw `fieldOps` are normalized before the portable allowlist applies (`:396-449`, `:1002-1023`), so a foreign field tombstone can still influence the materialized local record. **Repro:** import raw JSON reusing an existing thread/message UUID with a future `updatedAt` → local body replaced, omitted title/todos deleted. Fix: for existing ids, never let import delete/overwrite local fields — treat import as add-only for colliding ids, or gate field updates behind the same allowlist as ops.

### P1 — IndexedDB-unavailable silently truncates durable storage
When IDB is down, the fallback shadow write caps at 20 chats / 200 messages (`chat-history-store.js:20-21`, `:1315-1332`) yet reports persistence success and discards the full snapshot (`:1384-1413`). **Repro:** block IDB, keep 21+ chats, reload → the overflow is gone with no warning. Fix: surface a clear "history not durably saved" state rather than silently truncating-and-succeeding.

### P2 — Imported workflow aliases grow unbounded
Per-alias length is bounded but not the count (`:977-985`); import adds every missing alias (`:1647-1652`) and compaction bounds only `aliasOps`, not the materialized map. Repeated imports with unique paths grow persistent metadata without limit. Fix: cap the materialized alias map.

**Verified-clean:** export field/alias allowlist (`:1002-1023`, `:1548-1560`), schema rejection + meta→aliases (`:949-975`), post-import active-thread/A2UI rebind (`comfyui-mcp-panel.js:10711-10720`, `:12338-12343`), provider-mismatch clear + bounded 8K hydration (`:11914-11960`, `:14239-14255`), UTF-8 byte caps (`:53-55`, `:111-115`).

Tests are green (155/155 unit, typecheck clean) and it correctly preserves the #154 reconnect fix it's rebased on — so the remaining work is the four data-loss edges above, all in the import path.

JusticeWay:
Addressed all four data-safety findings and rebased the PR onto current `main` (`6aa98c1`, Panel 0.11.2). New head is `cb07566`.

What changed:

- Import now performs an atomic preflight for both the 500-thread cap and the 5,000-message-per-thread cap. It fails with an actionable error before mutating state, so imported recency can never evict an existing local chat/message and the success toast remains truthful.
- Same-ID import is strictly add-only. Every existing local message and thread field wins regardless of foreign timestamps or raw `fieldOps`; only missing messages and workflow versions may be added. Duplicate source records share one workflow-version budget, so they cannot indirectly evict local versions either.
- IDB-unavailable persistence reports success only when the complete snapshot fits the 20×200 shadow. If it does not, the store returns a retryable failure, keeps the complete write intent in memory, and the panel explains exactly what is and is not durable.
- The materialized workflow-alias map is bounded alongside `aliasOps` (512). Import preserves all local aliases, fills only free slots, and reports skipped portable aliases.

I chose fail-closed import rather than silently retaining some imported rows because it gives an atomic and explainable result: local data is never selected for eviction, partial imports are avoided, and “existing history was preserved” is always true after success. Same-ID collisions are add-only because portable JSON is untrusted archive input, not a causal authority over the current browser store.

The deeper pass also found and fixed two adjacent only-copy cases: `legacyShadow` transcripts are now exempt from both local shadow caps (they are fenced out of IDB), and failure to write their only localStorage copy can no longer be masked by a successful canonical commit for other data.

New regression coverage includes the exact 500+1 repro, per-thread message overflow, future-timestamp/raw-tombstone collisions, duplicate-record workflow-version pressure, 600 imported aliases, materialized-map stale-resurrection fencing, IDB-down shadow truncation, and legacy only-copy quota failure.

Verification on `cb07566`: 162/162 unit, 15/15 relevant Playwright E2E against live ComfyUI 0.28.3, typecheck, both JS syntax checks, diff check, and npm audit (0 vulnerabilities) all green. The PR description now documents the policy and rationale in full.

artokun:
## Re-review (cb07566) — all four data-loss findings resolved ✅

Verified the fixes against the four issues from the prior pass (codex adversarial re-review + local test run):

1. **Cap-eviction on import — FIXED.** The import now counts existing + new thread ids and **aborts before mutation** if the cap would be exceeded (`chat-history-store.js:1640-1658`); the "existing history was preserved" toast only shows on a *successful* import, and a cap failure surfaces "History import failed" (`comfyui-mcp-panel.js:12371-12389`). An import can no longer silently drop a pre-existing local chat.
2. **Same-id overwrite — FIXED.** Existing threads are cloned and imported messages whose ids already exist are skipped (`:1690-1723`); imported title/todos/fields apply only to genuinely new threads (`:1754-1769`). Foreign field tombstones can't clear or replace local fields.
3. **IDB-unavailable truncation — FIXED.** An incomplete capped shadow is now detected, returns `ok:false`, preserves the full dirty snapshot for retry, emits `history-canonical-unavailable-shadow-truncated`, and the panel surfaces a clear "older history is not durably saved" warning (`:1365-1387`, `:1456-1489`, `comfyui-mcp-panel.js:10595-10602`).
4. **Unbounded aliases — FIXED.** Imported aliases are capped to remaining `maxMetadataOps` capacity (excess skipped, local aliases untouched) and the materialized map is capped in compaction (`:1777-1796`, `:516-526`).

**Also verified:** `npm run typecheck` clean, **162/162 unit tests pass** (7 new since the last pass), both files syntax-valid, and the #154 unsaved-tab reconnect fix it's rebased on is intact.

No remaining data-safety blockers from my side — this is mergeable. Nice work addressing all four cleanly.

artokun (approved):
Re-reviewed — all four data-loss findings resolved (see comment), tests green, mergeable. Clearing my earlier review.

PR #208, opened 2026-07-29
Registry listing: fix mojibake description + claim local/offline tags

artokun:
The `[project] description` in `pyproject.toml` carried a double-encoded em dash (`â€"`), rendering as "sidebar â€" runs locally" **on the Comfy Registry listing and in ComfyUI-Manager search** — the highest-visibility text this project has, shown at the moment someone decides whether to install.

Also adds the tags the listing never claimed: `local`, `local-first`, `offline`, `self-hosted`, `ollama`, `local-llm`, `no-api-key`, `gemini`. It advertised claude/chatgpt/gpt-5 but nothing about running locally — the differentiator against a cloud-only official agent.

Verified: `pyproject.toml` parses via `tomllib`, no mojibake remains.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,21 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+## [0.9.7] - 2026-07-24
+
+### Fixed
+- **Registry listing description rendered mojibake.** The em dash in
+  `pyproject.toml`'s `description` had been decoded as CP1252 and re-encoded as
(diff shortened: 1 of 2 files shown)


artokun:
Closing — I built this against a stale local checkout. The mojibake in `pyproject.toml`'s description is **already fixed on main** (0.11.4 is clean), so that half is redundant, and the 0.9.7 bump in here would have been a downgrade from 0.11.4.

The Registry `Tags` addition is still valid — main still carries the original 14 with nothing claiming local/offline. Reopening that piece alone against current main.

Issue #176, opened 2026-07-29
Render structured backend failure as [object Object] in panel chat


artokun:
### What happened / root cause
After a Claude/Antigravity backend turn failed with an individual quota error, the sidebar chat rendered an additional `[object Object]` message instead of readable text. The panel appears to stringify a structured error/message object with implicit JavaScript object coercion.

### Steps to reproduce
1. Use a backend that returns a structured failure object, observed with Claude Opus via Antigravity.
2. Exhaust the provider quota so the turn fails.
3. Observe the sidebar chat after the readable quota error.
4. An extra `[object Object]` bubble is displayed.

### Exact error (scrubbed)
`agy exited with code 1. Error: Individual quota reached. Please upgrade your subscription to increase your limits.`
Then: `[object Object]`

### Fix
Applied locally: no. Upstream-only: yes.
Normalize structured chat payloads before rendering: prefer a string message/error field; otherwise serialize safely instead of implicit object coercion. Add a regression test for structured backend failures.

### Environment
darwin / ComfyUI 0.29.0 / MPS 128 GB / comfyui-mcp 0.48.5 / panel 0.11.4.

artokun:
Fixed in **panel 0.11.6** (WS-9, PR #228): chat now serializes structured backend/user payloads (errors, attachments) as readable text instead of `[object Object]`, and inline chat media persists across a panel reload. Please update and reopen if it persists. (Note: a DISTINCT [object Object] site in the sidebar output payload is tracked separately as #238.)

Issue #175, opened 2026-07-29
User message attachments/objects stringify as [object Object] into agent prompt


artokun:
### What happened / root cause
Panel/orchestrator handoff to the agent turned a non-text user payload into the literal string `[object Object]` in the model prompt, so the agent lost the prior message/attachment context. User then replied agreeing something was strange (`Ja det er da mærkeligt`) without recoverable prior content.

Likely `String(obj)` / template coercion on a structured chat part (attachment, card click, media, or multi-part message) instead of JSON/text extraction.

### Steps to reproduce
1. Open ComfyUI agent panel chat.
2. Send a message that includes a non-plain-text part (attachment, structured UI payload, or multi-part content) — or continue a thread after such a part.
3. Observe agent turn receives user content as literally `[object Object]` plus any following plain text.

### Exact error (scrubbed)
Prompt user section ended as:
```
The user's first message follows.

[object Object]

Ja det er da mærkeligt
```
No exception; silent context loss.

### Fix
<applied locally: no>  <upstream-only: yes>
Serialize each chat part by kind: text → string; image → panel media ref / description; card click → stable label; never default Object.prototype.toString.

### Environment
OS darwin · GPU mps 48GB · ComfyUI 0.29.0 · comfyui-mcp 0.48.5 · panel 0.11.4 · Backend label in ENV: Claude (session actually Grok Build) · torch 2.13 · python 3.12

artokun:
Fixed in **panel 0.11.6** (WS-9, PR #228): chat now serializes structured backend/user payloads (errors, attachments) as readable text instead of `[object Object]`, and inline chat media persists across a panel reload. Please update and reopen if it persists. (Note: a DISTINCT [object Object] site in the sidebar output payload is tracked separately as #238.)

Issue #168, opened 2026-07-29
Live canvas context becomes [object Object] and panel tools are missing


artokun:
### What happened / root cause
A sidebar-agent turn received the live-canvas context as the literal string `[object Object]`, and the session tool inventory contained no `panel_*` tools (including `panel_graph_outline` and `panel_query_graph`). The panel was installed and ComfyUI was reachable, so the likely root cause is failed panel-context serialization and/or live-panel tool registration during the panel/orchestrator handshake.

### Steps to reproduce
1. Open ComfyUI with the comfyui-mcp sidebar panel.
2. Keep any workflow tab active.
3. Ask the sidebar agent: `?????????`
4. Observe the agent input begins with `[object Object]`.
5. Inspect the agent tool inventory: no `panel_*` tools are registered, so the live canvas cannot be queried.

### Exact error (scrubbed)
No thrown error. Silent/malformed behavior:
- Injected context: `[object Object]`
- Missing expected tools: `panel_graph_outline`, `panel_query_graph`, and all other `panel_*` calls.
- Panel status: installed and applicable.
- ComfyUI API reachable; queue empty.

### Fix
applied locally: no  upstream-only: yes

Ensure structured canvas/session context is JSON-serialized before composing the user turn instead of relying on JavaScript object string coercion. Also surface/register live `panel_*` tools whenever the sidebar handshake is active; if registration fails, emit an explicit diagnostic rather than silently falling back to headless-only tools.

### Environment
Windows 10
ComfyUI 0.28.0
NVIDIA GeForce RTX 5060 Ti 16 GB
comfyui-mcp 0.48.5
comfyui-mcp-panel 0.11.3

artokun:
The '[object Object]' live-canvas context is fixed in **panel 0.11.15** (#276/#313) — you're on 0.11.3. The `{workflow,subgraph}` context object was string-coerced; it's now rendered to readable text (live-verified). Update to 0.11.18. If the *separate* 'panel tools are missing' symptom persists on 0.11.18 (a session-init issue, tracked as #296), reopen or comment on #296.

Issue #163, opened 2026-07-29
Open inline chat media in an in-app gallery lightbox


artokun:
### What happened / root cause
Inline image cards produced by `panel_show_media` display `cursor: zoom-in`, but clicking them opens a new browser tab instead of a modal/lightbox inside ComfyUI.

The panel route is `show_media` -> `onShowMedia` -> `paintImage`. `paintImage` binds clicks to `openMediaUrl`, and `openMediaUrl` is hard-coded to `window.open(...)`.

ComfyUI's Assets Sidebar already provides the expected UX in `MediaLightbox.vue`: body teleport, image/video support, previous/next navigation, Escape close, backdrop close, and focus restoration. However, it is an internal Vue component mounted by `AssetsSidebarTab.vue`, not an exposed public viewer service, so importing a hashed frontend chunk from the vanilla panel extension would be brittle.

### Steps to reproduce
1. Ask the panel agent to show an image via `panel_show_media`.
2. Click the inline image card in chat.
3. Observe that the image opens in a new browser tab instead of the ComfyUI gallery lightbox.

### Exact error (scrubbed)
No thrown error. The UI behavior is inconsistent with the Assets Sidebar and the zoom cursor implies an in-app preview that does not happen.

### Fix
Applied locally: no. Upstream-only: no.

Add an in-panel media lightbox matching the Assets Sidebar contract, with a gallery array per `show_media` payload, previous/next navigation, keyboard controls, backdrop close, and image/video support. Keep `window.open` only as a fallback. If ComfyUI exposes a stable public media viewer service later, prefer that service over duplicated UI.

### Environment
Windows 11 / ComfyUI 0.28.3 / NVIDIA RTX 3060 12 GB / comfyui-mcp 0.48.5 / comfyui-mcp-panel 0.6.8

PR #218, opened 2026-07-29
fix(chat): media persistence + object serialization (WS-9)

### LINKED ISSUE #177: Chat history drops all image/video cards — paintImage/paintVideo never record(), loadThread has no media branch

Every image and video card rendered in the panel chat (render results, `panel_show_media`, storyboards) is **painted to the DOM only and never written into the thread**. Reload the ComfyUI page, or switch chats in History and come back, and the conversation replays with all its text but **zero media** — the reader is left with an agent describing images that aren't there.

Two independent gaps, both in `web/js/comfyui-mcp-panel.js`:

1. `paintImage()` (~L7379) and `paintVideo()` (~L7459) never call `record(...)`. Every other feed renderer has an `append*` sibling that records (`appendUser` -> `role:"user"`, `appendAgent` -> `role:"agent"`, `appendActivity`/`paintQuestion`/`paintSecret` -> `role:"card"`). Media has no such path — there is no `role: "media"` anywhere in the file.
2. `loadThread()` (~L8137) only replays three roles:
   ```js
   if (m.role === "user") paintUser(m.text, { attachments: m.attachments });
   else if (m.role === "agent") paintAgent(m.text);
   else if (m.role === "card") paintCard(m);
   ```
   so even if media were recorded, it would be silently skipped on restore.

Notably this is **cheap to fix correctly**: the urls involved are already short and re-servable. Render results use `imageViewUrl(ref)` -> a ComfyUI `/view?filename=...` link (L9105/L9108), and `panel_show_media` items carry `item.url` (L8486-8487). Only user-pasted attachments use a heavy `data:` URI (L10883), and those are the one case that must NOT be persisted (localStorage quota).

### Steps to reproduce

1. In the panel, run any workflow that returns an image (or call `panel_show_media`). The card renders correctly.
2. Hard-refresh the ComfyUI page (or open History, switch to another chat, switch back).
3. The thread replays with all text intact; **every image/video card is gone**.

### Exact error (scrubbed)

None — this is a **silent failure**. Nothing throws, nothing warns; the media simply never reaches localStorage and is never replayed.

### Fix

<applied locally: yes>  <upstream-only: no>

Minimal patch: record servable media urls, guard replay against double-recording, and add the missing `media` branch to `loadThread`.

```diff
@@ record() / bindSession()
+  // True while loadThread() is replaying a stored conversation, so the paint*
+  // helpers repaint without re-recording what they are replaying.
+  let replaying = false;
+
+  /** Record an image/video card so history can replay it (see loadThread).
+   *  Only SERVABLE urls are stored (ComfyUI /view links, http(s)) — a data: URI
+   *  from a user attachment can be megabytes and would blow the localStorage
+   *  quota, taking the whole thread down with it. */
+  function recordMedia(kind, url, caption) {
+    if (replaying) return;
+    if (typeof url !== "string" || url.startsWith("data:") || url.startsWith("blob:")) return;
+    record({ role: "media", kind, url, caption: caption || "" });
+  }
+
   /** Bind the agent's current session id to the open thread (for reload/resume). */

@@ paintImage()
   function paintImage(url, name, gallery = null, galleryIndex = 0) {
     clearEmpty();
+    recordMedia("image", url, name);

@@ paintVideo()
   function paintVideo(url, name, gallery = null, galleryIndex = 0) {
     clearEmpty();
+    recordMedia("video", url, name);

@@ loadThread()
     resetFeed();
-    for (const m of t.msgs) {
-      if (m.role === "user") paintUser(m.text, { attachments: m.attachments });
-      else if (m.role === "agent") paintAgent(m.text);
-      else if (m.role === "card") paintCard(m);
-    }
+    replaying = true;
+    try {
+      for (const m of t.msgs) {
+        if (m.role === "user") paintUser(m.text, { attachments: m.attachments });
+        else if (m.role === "agent") paintAgent(m.text);
+        else if (m.role === "card") paintCard(m);
+        else if (m.role === "media") {
+          if (m.kind === "video") paintVideo(m.url, m.caption);
+          else paintImage(m.url, m.caption);
+        }
+      }
+    } finally {
+      replaying = false;
+    }
```

`node --check web/js/comfyui-mcp-panel.js` passes.

Known residual (deliberately out of scope for this minimal patch): media whose url is a `data:`/`blob:` URI (user-pasted attachments, L10883) still won't survive a reload. Persisting those properly needs an out-of-band store (IndexedDB or an upload to `input/`) rather than localStorage — worth a separate issue if you want attachment echo to persist too.

### Environment

- ComfyUI: 0.28.3 (local desktop, `--enable-manager --listen 0.0.0.0 --port 8188`)
- GPU: NVIDIA GeForce RTX 3060, 12 GB VRAM (CUDA 13.0, torch 2.10.0+cu130, python 3.13.12)
- comfyui-mcp: 0.48.5
- comfyui-mcp-panel: 0.6.8 (`PANEL_VERSION`, L96)
- frontend: comfyui-frontend-package 1.47.10

artokun:
WS-9 chat/media serialization + persistence. Two clusters — media persistence (panel-side) and the "[object Object]" family (panel-side defense; the orchestrator root fixes for #175/#176/#168 ship in a companion comfyui-mcp PR).

## Closes
- Closes #177 — chat history dropped every image/video card. `paintImage`/`paintVideo` never recorded, and the replay loop had no media branch. Now both record servable urls as `role:"media"` messages and `paintThread` replays them, so cards survive reload / thread switch.
- Partially addresses #176 — panel-side last line of defense: a structured backend-failure object now renders as readable text (message/error field, else safe JSON) instead of the literal `[object Object]`. Root fix (orchestrator forwarding proper text) is in the companion comfyui-mcp PR.

## What changed
- `paintImage`/`paintVideo` call a new `recordMedia(kind, url, caption)` — records only SERVABLE urls (ComfyUI `/view`, http(s)); `data:`/`blob:` attachment urls are skipped to protect the localStorage quota (known residual, unchanged from #177's note).
- `paintThread` gains a `role:"media"` replay branch wrapped in a `replaying` guard so replays repaint without re-recording.
- New `web/js/lib/display-text.js` `toDisplayText()` helper; `appendSystem`/`appendAgent` normalize structured payloads before paint (and record).

## Tests
- `browser_tests/unit/media-persistence.test.mjs` — media records survive normalize / merge / portable round-trip (reload path).
- `browser_tests/unit/display-text.test.mjs` — structured failures never yield `[object Object]`.
- `npm run test:unit` (175 pass) + `npm run typecheck` green; `node --check` valid as ES module.

Note: #175 and #168 (canvas-context serialization + missing `panel_*` tools) are orchestrator-side and handled in the companion comfyui-mcp PR. #163 (lightbox) deferred — not clean to include here.

[reports ≤panel 0.11.5]

### DIFF
--- a/browser_tests/unit/display-text.test.mjs
+++ b/browser_tests/unit/display-text.test.mjs
@@ -1,0 +1,48 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { toDisplayText } from "../../web/js/lib/display-text.js";
+
+test("passes plain strings through unchanged", () => {
+  assert.equal(toDisplayText("hello"), "hello");
+  assert.equal(toDisplayText(""), "");
+});
+
+test("nullish coerces to empty string, not the literal 'null'/'undefined'", () => {
+  assert.equal(toDisplayText(null), "");
+  assert.equal(toDisplayText(undefined), "");
+});
+
+test("primitives stringify readably", () => {
+  assert.equal(toDisplayText(42), "42");
+  assert.equal(toDisplayText(true), "true");
+});
+
+test("structured backend failure never renders as [object Object] (#176)", () => {
+  const out = toDisplayText({
+    error: "Individual quota reached. Please upgrade your subscription to increase your limits.",
+  });
+  assert.equal(out, "Individual quota reached. Please upgrade your subscription to increase your limits.");
+  assert.ok(!out.includes("[object Object]"));
+});
+
+test("prefers a message field, then serializes unknown objects as JSON", () => {
+  assert.equal(toDisplayText({ message: "boom" }), "boom");
+  const json = toDisplayText({ code: 1, kind: "quota" });
+  assert.ok(!json.includes("[object Object]"));
+  assert.deepEqual(JSON.parse(json), { code: 1, kind: "quota" });
+});
+
+test("unwraps a nested { error: { message } } payload", () => {
+  assert.equal(toDisplayText({ error: { message: "nested detail" } }), "nested detail");
+});
+
+test("Error instances render their message", () => {
+  assert.equal(toDisplayText(new Error("kaboom")), "kaboom");
+});
+
+test("a cyclic object degrades to a constructor hint, never [object Object]", () => {
+  const cyclic = {};
+  cyclic.self = cyclic;
+  const out = toDisplayText(cyclic);
+  assert.ok(!out.includes("[object Object]"));
+});
--- a/browser_tests/unit/media-persistence.test.mjs
+++ b/browser_tests/unit/media-persistence.test.mjs
@@ -1,0 +1,87 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import {
+  CHAT_HISTORY_SCHEMA,
+  mergeHistorySnapshots,
+  normalizeThread,
+  parseHistoryImport,
+} from "../../web/js/lib/chat-history-store.js";
+
+// Regression guard for panel issue #177: image/video cards are recorded as
+// `role: "media"` messages so they survive a reload / thread switch. The store
+// must preserve those records intact through normalization, merge, and the
+// portable export/import path — otherwise media would silently vanish again.
+
+const mediaThread = () => ({
+  id: "t-media",
+  schemaVersion: CHAT_HISTORY_SCHEMA,
+  createdAt: 1000,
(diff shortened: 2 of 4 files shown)


artokun:
Superseded by #228 (current WS-9 branch fix/chat-media-ws9e). Closing this stale branch from an aborted run.

Issue #226, opened 2026-07-29
panel_save_workflow({name}) RENAMES/MOVES the original file instead of saving a copy -- silent data loss, reports success


artokun:
### What happened / root cause

`panel_save_workflow({name: "..."})` on an ALREADY-SAVED workflow **renames/moves the existing file on disk** instead of writing a new one. There is no "save a copy" behaviour available, and nothing in the tool result signals that the original was consumed -- it just returns `{"saved": true, "workflow": "<new name>"}`.

Concretely, in this session:

- Open workflow: `user/default/workflows/_a_exporter/LTX-2.3_-_V2V_Extend_Any_Video_Multi-Extend_long_video.json` (758 KB)
- Called `panel_save_workflow({name: "LTX EROS Extend"})`
- Result: `{"saved": true, "workflow": "LTX EROS Extend"}`
- Disk after: `user/default/workflows/LTX EROS Extend.json` (488 KB) exists at the workflows ROOT, and the original file **is gone from `_a_exporter/`** -- both renamed AND relocated out of its subfolder.

### Impact -- data loss risk, and it defeated an explicit user instruction

The user had asked, in as many words, to save **without overwriting** the original ("enregistre sans �craser"). The tool reported success, and the original was destroyed anyway. I only caught it because I verified the filesystem afterwards; the tool result gives an agent no reason to suspect anything. Recovery was possible here purely by luck -- the user happened to keep a master copy in a separate `dev/comfyui-workflows` checkout, which I found and copied back. Without that, a 758 KB expert workflow would simply have been gone, with no backup and no undo (Ctrl+Z does not cover filesystem operations).

This is the sharpest possible edge for an agent-facing tool: the safe-sounding call is the destructive one, and the destruction is silent.

### Steps to reproduce

1. Open a workflow that is already persisted, ideally inside a subfolder (e.g. `_a_exporter/Foo.json`).
2. `panel_save_workflow({name: "Bar"})`
3. Observe: `Bar.json` is created at the workflows root; `_a_exporter/Foo.json` no longer exists.

### Fix

applied locally: no
upstream-only: yes

Three things, in priority order:

1. **Never consume the source file.** When `name` differs from the current workflow's name, this should be a SAVE-AS: write the new file, leave the original untouched. Renaming should require a dedicated tool (`panel_rename_workflow` already exists for exactly that intent).
[...]

Issue #215, opened 2026-07-29
panel_open_workflow times out when target workflow tab is backgrounded


artokun:
### What happened / root cause
`panel_open_workflow` timed out while switching from a newly loaded temporary workflow to an already-open unsaved workflow tab. The target tab was open but backgrounded. Pinning the target with `panel_set_workflow_target(mode:"pinned")` worked immediately, so the graph/session remained healthy.

### Steps to reproduce
1. Keep an unsaved workflow tab open.
2. Open another blank tab and load a workflow into it.
3. Call `panel_open_workflow` for the first unsaved tab.
4. Observe a 15-second `workflow_open` timeout.
5. Pin the same workflow and graph tools work.

### Exact error (scrubbed)
`Panel tab tmp:. did not reply to "workflow_open" within 15000 ms - the ComfyUI tab may be backgrounded or frozen`

### Fix
Applied locally: no. Upstream-only: yes.
Route the open command through the active responsive panel connection or wake/activate the target tab before waiting for its acknowledgement; fall back to pinning when the target is already registered.

### Environment
Windows 11; ComfyUI 0.28.3; RTX 5070 Ti Laptop 12 GB; comfyui-mcp 0.48.5; panel version not exposed.

artokun:
Note from the #226 live-test: this false-negative ack timeout leaves a freshly-opened on-disk workflow flagged temporary ("drift"). The #226/0.11.6 fix now SAFELY refuses a save-as in that drift state (protecting the original) rather than copying — so until #215 is fixed, save-as after an open that timed out will refuse-and-need-retry instead of copying. Fixing #215 (make the open ack reliable) restores smooth save-as copy behavior.

artokun:
Fixed in **comfyui-mcp 0.48.18** (#502) — same fix as #319. A backgrounded/frozen target tab that doesn't ACK workflow_open in time no longer produces a false timeout: the tool polls the authoritative active-workflow signal and returns success if the open actually happened. Update to 0.48.18.

PR #239, opened 2026-07-30
fix(save): save-as must never move a persisted source — #226 (reopened)

artokun:
## Root cause (confirmed against the reporter's frontend, comfyui-frontend-package 1.45.21)

Issue #226 was reopened because #231 does not work. #231 trusted `svc.saveWorkflowAs` to copy. It does not always copy.

In the installed 1.45.21 bundle (`dialogService-B7T2yww0.js`):

- `saveWorkflowAs(e, {filename})` branches on `e.isTemporary`:
  - **persisted** source → `t.saveAs(e, a)` (copy; original untouched) ✅
  - **temporary** source → `renameWorkflow(e, a)` — a **MOVE** that consumes the source ❌
- `isTemporary` is a **derived getter**: `get isTemporary(){return this.size===-1}` (and `isPersisted = !isTemporary`). It cannot be reassigned and drifts purely from `size`.

So the data-loss path: after a `panel_open_workflow` ack-timeout race (#215) the active doc can carry the real on-disk path yet be flagged temporary (`size===-1`). The panel delegates to `saveWorkflowAs`, the frontend takes its `renameWorkflow` branch, and `orig → copy` is a move — the original is gone, only the copy remains. #231 never verified the source survived.

## Fix (defense-in-depth; invariant: never destroy a file that exists on disk)

`web/js/lib/workflow-save.js`:
- Classify the source by **actual persisted state** (a real backing file via `getWorkflowByPath` / known-workflow lists / `isPersisted`), not the volatile in-memory flags.
- **Prevent**: refuse a save-as when an on-disk source is mis-flagged temporary, instead of delegating a move. Fires only in the drift state; a correctly-opened persisted workflow (`isTemporary===false`) still copies normally.
- **Backstop**: after `saveWorkflowAs`, assert the source still exists on disk (only when a reliable oracle exists — no false alarms on minimal doubles); throw loudly on a phantom success.
- The `renameWorkflow` fallback is now gated on `!sourcePersisted`, so a persisted source is never moved even if `wasUnsaved` wrongly reports unsaved.

## Tests

`browser_tests/unit/workflow-save.test.mjs` — added a faithful 1.45.21 service double (copy vs move branches, `getWorkflowByPath` backed by disk) plus:
- persisted-but-mis-flagged-temporary save-as ⇒ refused, source preserved, no `renameWorkflow`, no delegated move;
- correctly-flagged persisted ⇒ still copies;
- rogue moving `saveWorkflowAs` ⇒ caught by the disk-existence backstop.

227/227 unit tests pass (`npm run test:unit`). No version bump.

## NOT yet live-verified
Static + unit only. Needs the Chrome re-test: open a persisted workflow → single `save_workflow({name})` → assert the original survives on disk.

### DIFF
--- a/browser_tests/unit/workflow-save.test.mjs
+++ b/browser_tests/unit/workflow-save.test.mjs
@@ -407,6 +407,50 @@ test('disk-existence backstop catches a saveWorkflowAs that moves a persisted so
   )
 })
 
+test('P0 round-5: no-name save of a persisted workflow with an EMPTY filename refuses — never moves Orig.json → .json (#226)', async () => {
+  // Flow edge: an on-disk workflow whose in-memory filename is empty/unresolved,
+  // saved with NO name. effectiveName/finalTargetPath come out empty; the old
+  // code called saveInPlace unconditionally and the frontend's saveWorkflow
+  // recomputed the target from the empty name and MOVED "Orig.json" → ".json".
+  const active = {
+    path: 'workflows/Orig.json',
+    filename: '',
+    directory: 'workflows',
+    isPersisted: true,
+    isTemporary: false
+  }
+  const svc = makeService({ files: [active.path], active })
+
+  await assert.rejects(
+    () => saveActiveWorkflow(svc, undefined, {}),
+    /cannot resolve a target filename/
+  )
+  // Source preserved; NOTHING was called that could relocate it.
+  assert.ok(svc.disk.has('workflows/Orig.json'), 'source preserved')
+  assert.ok(!svc.disk.has('workflows/.json'), 'never moved to a bare .json')
+  assert.equal(svc.calls.length, 0, 'no service call moved it')
+})
+
+test('no-name save of a persisted workflow with a normal filename saves IN PLACE to its own file (#226)', async () => {
+  // The healthy counterpart: filename resolves, so the target equals the current
+  // path and it is a true in-place save (no relocation) — the source file is the
+  // one written, never moved.
+  const active = {
+    path: 'workflows/Orig.json',
+    filename: 'Orig.json',
+    directory: 'workflows',
+    isPersisted: true,
+    isTemporary: false
+  }
+  const svc = makeService({ files: [active.path], active })
+
+  await saveActiveWorkflow(svc, undefined, {})
+
+  assert.deepEqual(svc.calls, [['saveWorkflow', 'workflows/Orig.json']], 'saved to its own path')
+  assert.ok(!svc.calls.some((c) => c[0] === 'renameWorkflow'), 'never renamed')
+  assert.equal(svc.disk.size, 1)
+})
+
 test('P0 round-4: oracle returns the DRIFTED non-persisted wf for an on-disk path ⇒ unknown ⇒ REFUSE (#226)', async () => {
   // Source IS on disk, but its in-memory object is mis-flagged temporary. The
   // store's getWorkflowByPath returns that same non-persisted object. A RETURNED
--- a/browser_tests/unit/workflow-save.test.mjs
+++ b/browser_tests/unit/workflow-save.test.mjs
@@ -377,6 +377,33 @@ test('P0-a: no existence oracle + mis-flagged temporary + real name ⇒ REFUSE, 
   assert.equal(svc.calls.length, 0, 'nothing was called')
 })
 
+for (const realName of ['Untitled 2026-07-24', 'Unsaved Workflow 3']) {
+  test(`P0 name-vs-proof: a REAL on-disk "${realName}" drifted temporary + NO oracle ⇒ REFUSE, never moved (#226)`, async () => {
+    // The exact collision: a user really can have "workflows/Untitled ….json"
+    // (or "Unsaved Workflow 3.json") ON DISK. If its flags drift temporary and
+    // there is no existence oracle, the placeholder NAME must NOT be treated as
+    // proof of absence — that would classify a real file as never-persisted and
+    // then MOVE (destroy) it. With no oracle to confirm absence ⇒ unknown ⇒ refuse.
+    const active = {
+      path: `workflows/${realName}.json`,
+      filename: `${realName}.json`,
+      directory: 'workflows',
+      isPersisted: false, // drifted
+      isTemporary: true // drifted
+    }
+    // makeService ⇒ NO getWorkflowByPath, NO lists ⇒ no existence oracle.
+    const svc = makeService({ files: [active.path], active })
+
+    await assert.rejects(
+      () => saveActiveWorkflow(svc, 'a-copy', {}),
+      /cannot be proven absent from disk/
+    )
+    assert.ok(svc.disk.has(`workflows/${realName}.json`), 'real source preserved')
+    assert.ok(!svc.calls.some((c) => c[0] === 'renameWorkflow'), 'never renamed')
+    assert.ok(!svc.calls.some((c) => c[0] === 'saveWorkflowAs'), 'never delegated a move')
+  })
+}
+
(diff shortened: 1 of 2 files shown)


artokun:
## Round 2 — closed two more P0 data-loss paths (codex round 1)

**P0-a — No-oracle drift.** `isPersistedOnDisk()` returned FALSE when `isPersisted` was false and no existence API/list existed, so the temporary→move branch (and the fallback rename) proceeded and 1.45.21 moved the on-disk source. Replaced with a **tri-state** `classifySource()` → `persisted` / `never-persisted` / `unknown`. **UNKNOWN now fails safe (refuse).** A move/rename is only ever taken when the source is *provably* never-persisted (a placeholder `Unsaved Workflow …` / `Untitled …` temp tab). Both the `saveWorkflowAs` guard and the fallback-rename branch use this.

**P0-b — No-name / in-place relocation.** `isSaveAs` ignored a mode-derived path change, so a plain `saveWorkflow` relocated e.g. an on-disk `Foo.json` opened `initialMode:"app"` (no name) → `Foo.app.json`, consuming the source. The save is now classified by **relocation** (mode-correct target path vs current path, using the effective name = explicit/auto else current). Any relocation of a non-never-persisted source is routed down the safe copy/refuse path; `saveWorkflow` runs only when the target equals the current on-disk path.

**Tests** (229/229 pass): genuine-temp grounding retested against the faithful 1.45.21 double (temporary→`renameWorkflow` branch); added P0-a (unknown existence + temporary ⇒ refuse, never move) and P0-b (app-mode no-name ⇒ copy, source survives).

Still needs the Chrome live-re-test before merge.

artokun:
## Round 3 — closed the name-vs-proof P0 (codex round 2)

`classifySource()` returned `never-persisted` from the placeholder NAME alone (`Untitled …` / `Unsaved Workflow …`) plus temporary flags. But a real `workflows/Untitled 2026-07-12.json` can exist on disk, so a drifted-temporary REAL file with no oracle was classified never-persisted and moved (destroyed).

**Fix:** `never-persisted` now requires an existence **oracle to affirmatively confirm no backing file**. No oracle ⇒ `unknown` ⇒ fail-safe refuse. The name is no longer consulted for classification at all. Net: the only path that may rename is a source the oracle *proves* has no file on disk.

**Tests:** added real on-disk `Untitled 2026-07-24` and `Unsaved Workflow 3`, drifted `isTemporary=true`/`isPersisted=false`, NO oracle ⇒ classify `unknown` ⇒ REFUSE, source preserved, no rename / no saveWorkflowAs-move. 231/231 pass. All prior cases stay green.

Ready for codex round 3. Still needs the Chrome live-re-test before merge.

artokun:
## Round 4 — codex round 3 items closed

1. **P0 — oracle exception misread as absence.** A `getWorkflowByPath` lookup that THROWS was caught and treated as "no file" ⇒ never-persisted ⇒ move. Now only a SUCCESSFUL lookup returning no persisted workflow sets `confirmedAbsent`; an exception leaves it `unknown` ⇒ refuse. A list MISS also no longer counts as absence (an unlisted file may exist on disk) — only a positive persisted hit is trusted.

2. **Regression fixed — genuine new workflow grounds again.** A temporary doc with NO backing path was being refused. A doc with no path has nothing on disk to lose ⇒ classified `never-persisted` (no oracle required), so the everyday "save my brand-new workflow" path grounds/saves. Only a doc that HAS a path but whose on-disk status can't be established is `unknown` ⇒ refuse.

3. **Copy guarantee made provable.** The persisted happy-path test now asserts the SOURCE FILE still exists after `saveWorkflowAs` (proving copy semantics on the modeled 1.45.21 frontend), and the post-hoc backstop stays as the regression detector. (Explicit-copy-without-touching-source isn't cleanly available from the lib's `svc` on 1.45.21; saveWorkflowAs+backstop is safe since persisted→`t.saveAs` copies.)

**Tests:** added oracle-throws ⇒ refuse, and no-path new-workflow ⇒ saves. 233/233 pass; all prior cases green.

Invariant holds: no path moves a persisted-or-unknown source; genuine no-path/oracle-confirmed-absent temps still save. Ready for codex round 4; still needs the Chrome live-re-test before merge.

artokun:
## Round 5 — codex round 4 P0 closed (the last classification edge)

**P0 — a returned oracle object was misread as absence.** A successful `getWorkflowByPath` set `confirmedAbsent` for every result except `isPersisted===true` — including when it returned the *same drifted temporary `wf`* (`isPersisted=false`). So an on-disk source whose in-memory object is mis-flagged temporary was classified `never-persisted`, `saveWorkflowAs` took its temporary→rename branch and moved it, and the backstop was skipped (`cls!=="persisted"`). Data loss.

**Fix:** only a `null`/`undefined` return (truly nothing at the path) sets `confirmedAbsent`. Any returned object that is not affirmatively persisted (incl. the drifted temp itself) leaves the result `unknown` ⇒ refuse. The list oracle already trusts only a positive persisted hit; a list MISS never implies absence.

**Tests:** the faithful double's `getWorkflowByPath` now returns the DRIFTED active non-persisted object for an on-disk path (not always a fresh persisted stub). Added: on-disk source + oracle returns drifted `wf` ⇒ `unknown` ⇒ REFUSE (source preserved, no rename/move); and list-miss ⇒ REFUSE. 235/235 pass; genuine no-path/empty-disk temps still ground.

Net invariant intact: the only path that renames is one where the oracle returns *nothing* at the path (or the doc has no path); a returned object — persisted or drifted-temp — never authorizes a move. Ready for codex round 5; Chrome live-re-test remains the merge gate.

artokun:
## Round 6 — codex round 5 flow edge closed

**Save-in-place with an unresolved target.** For `{path:"workflows/Orig.json", filename:"", isPersisted:true}` saved with no name, `effectiveName`/`finalTargetPath` came out empty, `relocates` was false, and `saveInPlace` ran unconditionally — the frontend's `saveWorkflow` then recomputed the target from the empty name (→ bare `.json`) and renamed `Orig.json` → `.json`. A persisted-source MOVE with no absent-oracle result.

**Fix:** require a RESOLVED, non-empty `finalTargetPath` before either branch runs. If none can be derived while the workflow HAS a current path ⇒ refuse rather than let `saveWorkflow` recompute to a different path. `relocates` no longer short-circuits on an empty target (an empty target vs a non-empty current path is treated as a relocation, and the empty-target guard fires first).

**Tests:** empty-filename + no-name ⇒ refuse (`cannot resolve a target filename`), source preserved, no service call moves it; plus the healthy counterpart — a normal-filename no-name save is a true in-place save to the workflow's own path (`saveWorkflow` targets `Orig.json`, never renamed). 237/237 pass; all prior cases green.

That's 6 distinct data-loss paths closed. Ready for codex round 6; Chrome live-re-test remains the merge gate.

artokun:
## Round 7 — codex round 6 false-alarm closed (final)

**Backstop false alarm (not data loss).** The post-save backstop used a boolean `pathExists()` that swallowed a `getWorkflowByPath` throw and, with no lists, concluded the source vanished — reporting a SUCCESSFUL save-as copy as `save moved the original`.

**Fix:** the probe now uses the SAME tri-state rule as `classifySource` — only a SUCCESSFUL `null`/`undefined` return proves absence; a getter throw or a list-miss is UNKNOWN and never alarms. Replaced `hasExistenceOracle` + `pathExists` with `confirmedAbsentAt()`.

**Tests:** throwing post-save probe on a valid copy ⇒ reported SUCCESS (source survived), not a false "moved" error; the real-destruction backstop (oracle returns null after a rogue move ⇒ throws) stays green. 238/238 pass.

All destructive paths were confirmed closed in round 6; this removes the last false-positive. Ready for codex round 7 and the Chrome live-re-test.

Issue #184, opened 2026-07-29
panel_install_node silently no-ops for rgthree-comfy repository


artokun:
### What happened / root cause
panel_install_node reported queued success and panel_node_queue_status reached done/idle for the official rgthree-comfy repository, but panel_list_nodes did not contain the pack and no directory was created. Retrying with both repository=https://github.com/rgthree/rgthree-comfy.git version=nightly and id=rgthree/rgthree-comfy version=latest produced the same silent no-op. Manual git clone populated the pack immediately.
### Steps to reproduce
1. Call panel_install_node with repository=https://github.com/rgthree/rgthree-comfy.git and version=nightly.
2. Poll panel_node_queue_status until is_processing=false.
3. Call panel_list_nodes and inspect custom_nodes.
4. The pack is absent despite the task being counted as done.
5. Retry with id=rgthree/rgthree-comfy; same result.
### Exact error (scrubbed)
No explicit error. Install response queued=true; queue done_count advanced; panel_list_nodes remained unchanged.
### Fix
Applied locally: yes, workflow dependency restored by manually cloning the official repository to ~/ComfyUI/custom_nodes/rgthree-comfy. Upstream-only: no for the user's immediate workflow, yes for the tool behavior.
The install route should surface Manager task failure/stderr and must not treat a drained queue as successful when the destination pack is absent. The tool already warns callers to verify; returning the underlying failure would avoid shell fallback.
### Environment
Windows 11 / ComfyUI 0.29.0 / RTX 3080 10 GB / comfyui-mcp 0.48.5 / panel 0.11.4

PR #246, opened 2026-07-30
fix(install): verify node actually installed, no silent success (#232)

### LINKED ISSUE #232: panel_install_node silently reports success without installing GitHub repository
panel_install_node reported queued=true and Manager queue completed successfully, but the requested GitHub custom-node repository was not installed and did not appear in panel_list_nodes. The same silent failure occurred with both repository URL and owner/repo id. A direct git clone into custom_nodes succeeded immediately, suggesting the panel/Manager install wrapper does not surface a clone/install failure.

### Steps to reproduce
1. Call panel_install_node with repository=https://github.com/TenStrip/10S-Comfy-nodes.git and version=nightly.
2. Poll panel_node_queue_status until is_processing=false; it reports done_count incremented.
3. Call panel_list_nodes: 10S_Nodes is absent.
4. Repeat with id=TenStrip/10S-Comfy-nodes; same result.

### Exact error (scrubbed)
No error was returned. Both calls returned queued=true and the queue later reported done, but nothing was installed.

### Fix
applied locally: no (tool code not patched)  upstream-only: yes
Workaround: cloned the same repository directly to ComfyUI/custom_nodes/10S_Nodes, which populated correctly. The tool should verify the target directory/installed pack after Manager completion and return the Manager task failure details instead of queued/done success.

### Environment
Windows 11 / ComfyUI 0.28.0 / NVIDIA RTX 5070 Ti 16 GB / comfyui-mcp 0.48.6 / panel 0.11.0

artokun:
Fixes #232.

## Root cause
`panel_install_node` (`nodes_install`) queued the Manager install, called `queue/start`, and returned `queued:true` **immediately**. ComfyUI-Manager marks every queued task `"done"` even when the underlying clone/resolve fails, and the queue/status surface exposes only aggregate counts (no per-task result). So a GitHub-repo install that never landed (v4 "install by id" not resolving, a bad ref, security_level gating, etc.) reported the same `queued:true` + queue `done` as a real success. #184 only inspects the **immediate** batch `failed` array, which the async clone failure never populates — and the `v2` task path had no failure check at all.

## Fix
After `queue/start`, wait for the queue to drain, then **verify** the pack is actually present in `/customnode/installed` (which reflects `custom_nodes` on disk before a reboot). If it's absent, throw a clear error naming the target instead of returning silent success. Applies to all three dialects (`v2` task, `v2-batch`, `legacy`).

- `web/js/lib/manager-install.js`: new pure, unit-testable `parseInstalled` + `nodeInstalledMatches` helpers, mirroring the mcp orchestrator's proven `parseInstalled` / `nodeInstalledMatches` (`src/services/node-management.ts`). Tolerant of the v4 map shape, the legacy array shape, and matches on module / cnr_id / aux_id and their basenames.
- `web/js/comfyui-mcp-panel.js` (install handler only): `waitForQueueDrain` + `verifyInstalled` helpers; `nodes_install` now verifies before returning `{ queued, installed:true, ... }`.

Scope: panel-only, edits localized to the install handler. No version bump.

## Tests
Extended `browser_tests/unit/manager-install.test.mjs`: landed ⇒ success, did-not-land ⇒ error, across all three dialects, for both registry ids and git URLs, plus `parseInstalled` shape coverage. Full panel unit suite green (285 tests).

### DIFF
--- a/browser_tests/unit/manager-install.test.mjs
+++ b/browser_tests/unit/manager-install.test.mjs
@@ -188,19 +188,28 @@ test("queueFailureSignal fires only on explicit evidence (status or batch)", () 
 });
 
 // --- classifyInstallOutcome: TRI-STATE, no false success / no false failure --
-// Exercises the EXACT status/list shapes codex round 2 named, per dialect. The
-// handler verifies buildInstallRequest's id (already the repo NAME for a git URL).
+// Exercises the EXACT status/list shapes codex named, per dialect. The handler
+// verifies buildInstallRequest's id (already the repo NAME for a git URL) and
+// computes renameProne the same way we mirror here.
 const DRAINED = { is_processing: false, done_count: 1, total_count: 1 };
+const FAIL_STATUS = { is_processing: false, done_count: 1, total_count: 1, error_count: 1 };
+// Mirror the handler's renameProne derivation exactly.
+const renameProneOf = (args) => !!installGitUrl(args) || String(args.id ?? "").includes("/");
+
 for (const dialect of ["v2", "v2-batch", "legacy"]) {
-  const targetOf = (args) => {
+  // Build the classifier input the handler would pass for these args.
+  const inputFor = (args) => {
     const req = buildInstallRequest(dialect, args, "uid-1");
-    return dialect === "v2" ? req.params.id : req.body.id;
+    return {
+      target: dialect === "v2" ? req.params.id : req.body.id,
+      dialect,
+      renameProne: renameProneOf(args),
+    };
   };
 
   test(`${dialect}: drained + pack present ⇒ installed (registry)`, () => {
     const o = classifyInstallOutcome({
-      target: targetOf({ id: "rgthree-comfy" }),
-      dialect,
+      ...inputFor({ id: "rgthree-comfy" }),
       status: DRAINED,
       installed: { "rgthree-comfy": { ver: "1.0.0", cnr_id: "rgthree-comfy" } },
     });
@@ -209,20 +218,20 @@ for (const dialect of ["v2", "v2-batch", "legacy"]) {
 
   test(`${dialect}: drained + pack present ⇒ installed (git URL, repo-name dir)`, () => {
     const o = classifyInstallOutcome({
-      target: targetOf({ repository: "https://github.com/rgthree/rgthree-comfy.git" }),
-      dialect,
+      ...inputFor({ repository: "https://github.com/rgthree/rgthree-comfy.git" }),
       status: DRAINED,
       installed: { "rgthree-comfy": { ver: "nightly", aux_id: "rgthree/rgthree-comfy" } },
     });
     assert.equal(o.state, "installed");
   });
 
-  test(`${dialect}: drained + absent + explicit failure ⇒ failed`, () => {
+  // Identifiable (claimed registry id) + drained + definitively absent + failure
+  // evidence ⇒ the ONE case that hard-fails.
+  test(`${dialect}: identifiable id, drained + absent + explicit failure ⇒ failed`, () => {
     const o = classifyInstallOutcome({
-      target: targetOf({ repository: "https://github.com/TenStrip/10S-Comfy-nodes.git" }),
-      dialect,
-      status: { is_processing: false, done_count: 1, total_count: 1, error_count: 1 },
-      installed: {}, // nothing landed — exactly the #232 report
+      ...inputFor({ id: "no-such-registry-pack" }),
+      status: FAIL_STATUS,
+      installed: { "rgthree-comfy": { cnr_id: "rgthree-comfy" } }, // real list, our pack absent
     });
     assert.equal(o.state, "failed");
     assert.match(o.message, /FAILED/);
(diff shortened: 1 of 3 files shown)


artokun:
## Round 2 — addressed all 3 HIGH (tri-state outcome)

Pushed `48ba591`. The install path now resolves a **tri-state** outcome instead of a boolean — never a false success (the original #232 bug) and never a false failure.

**HIGH 1 — >120s / inconclusive falsely reported failed.** `waitForQueueDrain` now returns `{ drained, status }` — `drained:true` only when a drained state was actually observed. The decision moved into a pure `classifyInstallOutcome`:
- pack positively present ⇒ **installed**
- queue truly drained **AND** installed-list readable **AND** explicit Manager failure evidence **AND** pack absent ⇒ **failed** (throws)
- everything else — still processing at deadline, status/list unreadable, or drained-but-no-evidence ⇒ **unverified** (honest `{ installed:false, verified:false, pending:true, note: "…still in progress / could not confirm — check the server log" }`, not thrown, not success). We never print "queue reported done" unless it did.

**HIGH 2 — not truly bounded.** `managerV2`/`managerCall`/`managerGet` now accept a `signal`; every status and installed-list fetch is wrapped in `AbortSignal.timeout(...)` so a stalled request can't outlive the deadline, and the initial settle-sleep uses `boundedDelay` (respects the remaining budget).

**HIGH 3 — bare-string renamed-dir false-negative.** A bare module name that doesn't positively match (e.g. `10S-Comfy-nodes` → dir `10S_Nodes`) now falls to **unverified**, never a hard fail. `queueFailureSignal` fires only on an explicit error/failed count/array — never inferred from a missing pack.

**Tests:** `classifyInstallOutcome` + `queueFailureSignal` are pure and unit-tested. Added end-to-end outcome tests across all 3 dialects (v2 / v2-batch / legacy): installed (registry + git), drained+absent+evidence ⇒ failed, drained+absent+no-evidence ⇒ unverified, still-processing ⇒ unverified, **renamed-dir ⇒ unverified (asserts NOT failed)**, unreadable-list ⇒ unverified. Panel unit suite green (295).

Ready for codex round 2.

artokun:
## Round 3 — closed all 5 tri-state gaps

Pushed `803671d`.

**#1 over-reports drained.** New pure `queueDrained(status)` requires POSITIVE evidence: a well-formed object with `is_processing===false` AND numeric `done_count`/`total_count` AND `done>=total`. `null` / `{}` / missing-counts / `{error_count:1}` ⇒ NOT drained. `waitForQueueDrain` now returns the raw status and the classifier re-derives drained, so a malformed status is never a false drain (nor a false failure).

**#2 still-processing → success.** `classifyInstallOutcome` checks `drained` FIRST — not-drained ⇒ unverified/pending regardless of pack presence. Never `verified:true` unless drained was positively observed (a present pack mid-clone could be a stale/partial dir).

**#3 malformed list treated as readable.** New `isReadableInstalledList(raw)` — only a real array/map counts. `null` / primitive / malformed ⇒ `listReadable:false` ⇒ unverified, even with a failure status. Handler sets `listError` on throw; classifier also re-checks shape.

**#4 v2-batch throws outside the gate.** Removed the early `assertBatchOk` throw from install. The synchronous `failed[]` is captured and FED to `classifyInstallOutcome` as failure EVIDENCE (`queueFailureSignal` gained `batchFailed`/`target`), but the outcome still passes the tri-state gate (drained + readable + absent). Present pack still wins over stale batch evidence.

**#5 not every fetch bounded.** `MANAGER_FETCH_TIMEOUT_MS` now wraps EVERY Manager fetch — dialect-detection probes, install, start, status, and list — via `AbortSignal.timeout`.

**Tests (315 green):** direct coverage of `queueDrained` (null/{}/`{error_count:1}`/primitive/array/missing-counts/still-processing/done<total), `isReadableInstalledList` (array/map/null/primitive), `queueFailureSignal` (status + batch), and per-dialect classifier cases for every named shape — including still-processing-with-pack-present ⇒ unverified, `{error_count:1}`-not-drained ⇒ unverified, null/malformed list + failure status ⇒ unverified, renamed-dir ⇒ unverified, and the v2-batch `failed[]` path through the gate (failed / unverified-if-not-drained / installed-if-present).

Ready for codex round 3.

artokun:
## Round 4 — closed the false-failure edge

Pushed `bbf86b3`.

**Root fix.** A failure signal alone no longer hard-fails when pack presence is INCONCLUSIVE. `classifyInstallOutcome` now returns `failed` ONLY when the pack is **definitively absent** — i.e. the target is **identifiable** (a claimed registry id the matcher recognizes via `cnr_id`/`aux_id` if present) AND not matched, on a drained + readable + failure-evidence run. A rename-prone install (git URL or owner/repo id like `TenStrip/10S-Comfy-nodes` → dir `10S_Nodes`) can install under a dir name that differs from the repo name, so "not name-matched" is not "absent" — those go to **unverified** even with `error_count` / a stale `batchFailed[target]`.

Precedence: present ⇒ installed; else drained + readable + definitively-absent + failure-evidence ⇒ failed; everything else ⇒ unverified.

The handler derives `renameProne = !!installGitUrl(args) || String(id).includes("/")` and threads it through `verifyInstalled` → `classifyInstallOutcome`.

**Exact codex cases added (all 3 dialects):**
- drained + `error_count:1` + renamed-dir pack present-but-unmatched + git/owner-repo target ⇒ **NOT failed** (unverified).
- drained + `error_count:1` + git target + empty list (can't rule out a rename) ⇒ **NOT failed** (unverified).
- stale `batchFailed:[target]` + renamed pack present ⇒ **NOT failed** (unverified).
- owner/repo id (`TenStrip/10S-Comfy-nodes`) renamed-dir + `error_count` ⇒ **NOT failed**.
- The single hard-fail path retargeted to an **identifiable** registry id (`no-such-registry-pack`) absent from a real list + failure evidence ⇒ **failed** (the useful case still works).

Full panel unit suite green: **325 pass, 0 fail** (64 in `manager-install.test.mjs`). Panel JS validated as ES-module syntax.

Ready for codex round 4.

artokun:
## Round 5 — installed-list shape validation

Pushed `9ae80cf`.

`isReadableInstalledList` now validates SHAPE, not just container type. A payload is readable ONLY when it is:
- an EMPTY array or map (legitimately "nothing installed"), OR
- an ARRAY whose every element is a node-entry object (or a legacy bare non-empty string), OR
- a MAP whose every value is a node-entry object — a plain object carrying at least one expected key (`module`/`cnr_id`/`aux_id`/`name`/`title`/`ver`/`version`).

An error envelope (`{error|detail|message:…}`), an array with `null`/primitive/non-entry items, or a no-entry-shape object ⇒ NOT readable ⇒ `listReadable:false` ⇒ **unverified**. This closes the false-`failed` path where a `200`-but-malformed body coerced to a trustworthy empty list under a drained failure status + identifiable target.

**Tests (340 green):** `isReadableInstalledList` now covers `{error}`/`{detail}`/`{message}`, `[null]`, `[123]`, `{foo:"bar"}` ⇒ false, plus real array-of-entries, real map, empty array/map, and bare-string arrays ⇒ true. Classifier: each malformed payload + drained `FAIL_STATUS` + identifiable id ⇒ **unverified, NOT failed**, across all 3 dialects.

Expecting PASS on round 5. Ready for your live-verify + merge.

artokun:
## Round 6 — fixed the critical wiring bug + added guards

Pushed `f3b2d1a`. Good catch — this was a guaranteed total false-failure.

**Bug:** `waitForQueueDrain()` calls `queueDrained(status)`, but the panel's `import { … } from "./lib/manager-install.js"` omitted `queueDrained`. The first successful status poll of every install would throw `ReferenceError: queueDrained is not defined`, failing even genuine installs. The direct-import unit tests import `queueDrained` themselves, so they never exercised the panel's own import statement.

**Fix:** added `queueDrained` to the panel import (line 69).

**Two new guards so the class can't recur:**
1. **Source-level wiring test** — reads `comfyui-mcp-panel.js`, parses its `manager-install.js` import list, and asserts every export it *calls* (`\bNAME\s*(`) is imported. Exactly the thing direct-import tests can't see; this test fails if any used export is missing from the import.
2. **Real-source end-to-end test** — extracts the actual `waitForQueueDrain` function text from the panel source and runs it against a mock `managerGet` (poll 1 processing, poll 2 drained), asserting it resolves a positively-drained status with no `ReferenceError`.

Full panel unit suite green: **342 pass, 0 fail** (83 in `manager-install.test.mjs`). Panel JS validated as ES-module syntax.

Ready for codex round 6 — then your live-verify + merge.

Issue #366, opened 2026-07-31
panel_set_widget on a promoted subgraph widget silently renders with the OLD value (parent widget left stale, wins at queue time)


artokun:
### What happened / root cause

`panel_set_widget` on a **subgraph node's promoted widget** reports success and the new
value is readable on the inner node — but the **render still uses the OLD value**. The
edit is silently discarded at queue time.

Root cause is in `web/js/lib/widget-write.js`. `resolveWidgetWrite()` correctly resolves a
promoted widget to its inner `(node, widget)` and `applyWidgetWrite()` writes + verifies it
there. But for a ComfyUI subgraph, the inner node's widget is **converted to an input fed
from the subgraph input rail (`-10`)** — the *authoritative* value is the **parent subgraph
node's own promoted widget**, which is what serializes into the rail when the prompt is
queued.

So after the write:

- inner node `257 PrimitiveInt "Width"` → `value = 704`  ✅ (what the tool reports and what
  `panel_query_graph` reads back)
- parent node `267` promoted widget `value_2` → still `1280` ❌ (never touched)

At queue time the stale parent pushes `1280` back down the rail. **Both the tool result and
a follow-up read confirm the new value, yet the render uses the old one** — the failure is
completely invisible until you look at the output.

Impact: any agent editing a subgraph-based workflow (very common — model-family packs and
community graphs wrap everything in one subgraph node with promoted widgets) silently
renders with stale prompt/resolution/duration. In this session a 10s vertical PSA rendered
as the *previous* prompt at the *previous* landscape resolution.

### Steps to reproduce

1. Open a workflow whose main node is a **subgraph** with promoted widgets backed by inner
   `PrimitiveInt` / `PrimitiveStringMultiline` nodes (repro used an LTX-2.3 T2V subgraph:
   parent widgets `value` → inner `266 Prompt`, `value_2` → inner `257 Width`,
   `value_3` → inner `258 Height`, `value_4` → inner `225 Duration`).
2. `panel_set_widget(node_id=<subgraph node>, widget="value_2", value=704)`
   → returns `{ set: { node_id: "257", widget: "value", previous: 1280, value: 704,
   promoted_from: { subgraph_node_id: "267", inner_node_id: "257" } } }`.
3. `panel_enter_subgraph` + `panel_query_graph({ids:[257]})` → `value=704` ✅
4. `panel_query_graph({ids:[267], fields:'detail'})` on the PARENT → `value_2` is still
   `1280` ❌
[...]

Issue #347, opened 2026-07-30
panel_restart_comfyui hangs until 300 second tool timeout


artokun:
### What happened / root cause
panel_restart_comfyui timed out after 300 seconds even though the panel and ComfyUI became reachable again. The call did not return a terminal reboot result, leaving the agent unable to distinguish a completed restart from a failed one.

### Steps to reproduce
1. Start from a connected live workflow with an idle queue.
2. Call panel_restart_comfyui.
3. Wait for the server to restart and reconnect.
4. The tool call remains pending until the 300 second MCP timeout.

### Exact error (scrubbed)
tool call failed for panel/panel_restart_comfyui: timed out awaiting tools/call after 300s

### Fix
applied locally: no  upstream-only: yes
The restart tool should return immediately after reboot initiation, or resolve when the server reconnects, rather than waiting beyond the MCP timeout.

### Environment
Windows 11 / ComfyUI 0.28.0 / NVIDIA RTX 5070 Ti 16 GB / comfyui-mcp 0.48.5 / panel 0.11.0

artokun:
Fixed in comfyui-mcp 0.48.22 via #536/#509 (commit a570792). The tool-side `panel_restart_comfyui` reboot path no longer runs an unbounded poll that could hang to the 300s MCP timeout. It now certifies reboot readiness via a concurrent, post-write-gated boot-endpoint observer with a HARD-CAPPED budget (settle + budget ≤ 250s, well below the ~300s tools/call timeout), so a completed reboot returns a terminal ready:true and even a FAILING wait always returns a clean ready:false in time instead of being killed as a bare 300s timeout.

Version cross-check: this was reported on comfyui-mcp 0.48.5; the fix landed across the 0.48.x restart-timeout cluster (#493/#263/#222/#266/#306/#307) and was completed by the reboot-readiness observer in 0.48.22 (#536/#509).

If you still see a 300s hang on 0.48.22+, please reopen with the ComfyUI/Manager version and the restart logs.

Issue #240, opened 2026-07-30
Combo widget set by panel_set_widget silently changes to another enum after run


artokun:
### What happened / root cause
`panel_set_widget` reported that an AnimaLLLiteApply combo widget was set to the exact valid enum `ANIMA\anima-lllite-pose-1.safetensors`, and immediate panel_query_graph detail showed that value. After executing the workflow, panel_graph_outline showed the same node silently changed to `ANIMA\anima-lllite-any-test-like-v2.safetensors`. The rendered result behaved like the wrong model. This appears to be combo-index/value persistence drift between panel mutation and frontend serialization/execution.

### Steps to reproduce
1. Add an AnimaLLLiteApply node.
2. Set widget `lllite_name` through panel_set_widget to `ANIMA\anima-lllite-pose-1.safetensors` (valid /object_info enum).
3. Confirm panel_query_graph reports the pose file.
4. Queue/run the workflow.
5. Re-read with panel_graph_outline; the widget has changed to a different enum (`any-test-like-v2`).

### Exact error (scrubbed)
No exception is surfaced. panel_set_widget returns success, but the value silently changes after execution.

### Fix
applied locally: no  upstream-only: yes
Persist combo widgets by exact string value and verify the frontend widget value after serialization/queue submission; avoid reinterpreting the stored value as a stale dropdown index when node option lists refresh.

### Environment
Windows 11; ComfyUI 0.21.1; NVIDIA GeForce RTX 4080 SUPER 16 GB; comfyui-mcp 0.48.6; panel 0.11.5.

artokun:
Fixed in **panel 0.11.7** (PR #244): combo widgets are now set by EXACT current option only — an invalid value is rejected with an error instead of silently coercing to a neighboring enum. Live-verified (valid value sets exactly; invalid rejected). Please update and reopen if it persists.

Issue #179, opened 2026-07-29
panel_set_widget cannot set Power Lora Loader (rgthree) composite lora_* widgets


artokun:
### What happened / root cause
panel_set_widget on Power Lora Loader (rgthree) widgets lora_N accepts a JSON string value but applies a broken partial update: turns the slot on:true, sets lora:null, and ignores strength. Object values are rejected by schema (string|number|boolean only). Agent cannot add/enable a LoRA row in Power Lora Loader programmatically.

### Steps to reproduce
1. Open a graph with Power Lora Loader (rgthree) that has lora_1..lora_13 slots
2. panel_set_widget node_id=PowerLora, widget=lora_10, value='{\"on\":true,\"lora\":\"some.safetensors\",\"strength\":0.6,\"strengthTwo\":null}'
3. Observe result: on:true, lora:null, strength:1

### Exact error (scrubbed)
No hard error on first call — silent wrong write. Object value returns validation error invalid_union expected string|number|boolean.

### Fix
<applied locally: no>  <upstream-only: yes>
Allow composite widget values (object/JSON) for Power Lora Loader slots, or document a supported scalar format. After set, preserve lora filename and strength fields.

### Environment
Windows 11 / ComfyUI 0.26.2 / RTX 5070 12GB / comfyui-mcp 0.48.5 / panel (sidebar)

PR #441, opened 2026-08-01
fix(set_widget): refuse scalar-to-composite corruption (#560) + warn on control_after_generate (#558)

### LINKED ISSUE #558: panel_set_widget reports success on a seed it cannot hold — control_after_generate is invisible to panel_set_widget and panel_query_graph

I set both seed widgets to a fixed value to run a controlled experiment (one variable: the prompt text). `panel_set_widget` returned success with a clean before/after:

```json
{ "set": { "node_id": "30", "widget": "seed", "previous": 507892738009130, "value": 777777 } }
```

Every subsequent render used a **different random seed**. Both widgets had `control_after_generate: randomize`, which silently overwrote the value after each generation. I only discovered this ~22 renders later, when I read the graph back for an unrelated reason and saw `seed: 373929663537205`.

**Nothing in the tool surface reveals this:**

- `panel_set_widget` reports success and echoes the value I asked for — it has no idea the value won't survive.
- `panel_query_graph` with `fields:'detail'` returns `"seed": 777777` in `widgets` but **does not include `control_after_generate`** anywhere in the node detail. Reading the node back does not reveal the problem either.
- `panel_graph_outline` likewise shows `seed=777777` with no mode indicator.

So the only way to discover it is to notice the value changed later, by accident.

### Why this matters beyond seeds

This silently invalidates any controlled comparison run through the panel — A/B tests, prompt comparisons, LoRA-strength sweeps, sampler comparisons. I published a set of conclusions in-session asserting "the seed is fixed, so the only variable is X." That claim was false for every render, and the tool surface actively supported the false belief by reporting success.

Same class applies to any widget with a `control_after_generate` (seed, and any `INT`/`COMBO` with increment/decrement/randomize modes).

### Steps to reproduce

1. On a node whose seed widget has `control_after_generate: randomize`, call `panel_set_widget(node_id, 'seed', 777777)` → returns success.
2. `panel_run`, wait for completion.
3. `panel_query_graph({ids:[node_id], fields:'detail'})` → seed is now a random value. No warning was ever emitted, and no field in the detail output would have predicted it.

### Fix

`applied locally: no` · `upstream-only: no`

Two changes, either of which would have saved this:

1. **Surface the mode.** Include `control_after_generate` in `panel_query_graph`'s `detail` widget output (and mark it in `panel_graph_outline`, alongside how `[bypass]`/`[mute]` are already flagged). Node modes are surfaced precisely because they silently change behaviour — this is the same category.

2. **Warn on write.** When `panel_set_widget` targets a widget whose `control_after_generate` is not `fixed`, add a note to the result, e.g.:

```json
{ "set": { "...": "..." },
  "warning": "control_after_generate='randomize' on this widget — this value will be overwritten after the next generation. Set it to 'fixed' to hold the value." }
```

Optionally a `hold: true` convenience arg that flips `control_after_generate` to `fixed` in the same call, since "set this and keep it" is the overwhelmingly common intent when an agent writes a seed explicitly.

### Related, smaller: `batch_count` semantics are misleading

`panel_run`'s `batch_count` is documented as "Times to queue (default 1)". I used `batch_count: 2` expecting two samples and got **two byte-identical images** — it re-queues the same seed rather than resampling. Combined with the above, "queue N runs for N variations" fails in two different ways at once.

Worth a description clarification: *"Re-queues the SAME prompt and seed N times — this does NOT produce variations. For different results, set a different seed per run."* (Also note that `control_after_generate` advances only **after a generation completes**, so queuing several runs back-to-back before any finish will reuse one seed across all of them — which is what bit me a second time.)

### Environment

Windows 11 26200 · ComfyUI 0.29.2 (Desktop, local) · RTX 3070 8GB · CUDA 13.0 · torch 2.10.0 · Python 3.13 · **comfyui-mcp 0.48.21** · panel: Manager v4

### LINKED ISSUE #560: panel_set_widget silently CORRUPTS composite object widgets (rgthree Power Lora Loader) — reports success while destroying the value

`panel_set_widget`'s `value` accepts only `string | number | boolean`. Some widgets are **composite objects** — rgthree's `Power Lora Loader` stores each slot as `{ on, lora, strength, strengthTwo }`. Writing a scalar to one of those slots **overwrites part of the object and destroys the rest**, then reports success.

Reading is fine. `panel_query_graph {fields:'detail'}` returns the composite correctly:

```json
"lora_1": { "on": true, "lora": "LTX2.3_2d_NSFW_motion_enhancer.safetensors", "strength": 1, "strengthTwo": null }
```

Writing corrupts it. Attempting to toggle the slot off:

```
panel_set_widget(node_id=128, widget='lora_1', value=false)
-> { "set": { "previous": { "on": true,  "lora": "LTX2.3_2d_NSFW_motion_enhancer.safetensors", "strength": 1, "strengthTwo": null },
              "value":    { "on": true,  "lora": null, "strength": 1, "strengthTwo": null } } }
```

The boolean did **not** set `on: false` — it nulled `lora`, silently discarding the filename. The call returned a normal success payload.

**It is not recoverable through the same tool.** Passing the filename back does nothing:

```
panel_set_widget(node_id=128, widget='lora_1', value='LTX2.3_2d_NSFW_motion_enhancer.safetensors')
-> { "set": { "previous": { "on": true, "lora": null, ... },
              "value":    { "on": true, "lora": null, ... } } }   # unchanged, still reports success
```

Two successive calls both reported success; the widget ended in a broken state. The user had to Ctrl+Z. This destroyed real configuration on a live workflow.

Dotted paths are not supported either — `widget='lora_1.strength'` errors with the available-widget list, which is at least a *loud* failure:

```
Error: Node 128 (Power Lora Loader (rgthree)) has no widget "lora_1.strength"
(available: divider, PowerLoraLoaderHeaderWidget, lora_1 ... lora_8, divider, "➕ Add Lora")
```

### Why this matters

`Power Lora Loader` is extremely common in expert workflows (it's in the rgthree pack that ships with most video templates). An agent asked to "turn off the FaceID LoRA" or "set the DMD LoRA to 0.5" will reach for `panel_set_widget`, get a success response, and have silently destroyed the slot. There is no error to react to and no way to undo it from the tool side.

### Steps to reproduce

1. Any graph with an rgthree `Power Lora Loader` that has at least one configured slot.
2. `panel_set_widget(node_id=<loader>, widget='lora_1', value=false)`
3. Re-read with `panel_query_graph {ids:[<loader>], fields:'detail'}` → `lora` is now `null`, `on` is unchanged.

### Fix

`applied locally: no` · `upstream-only: no`

Options, roughly in order of preference:

1. **Refuse the write.** If the existing widget value is a non-null object and the incoming `value` is a scalar, return an error naming the sub-fields rather than coercing:
   `"widget 'lora_1' is a composite {on, lora, strength, strengthTwo} — pass an object, or use lora_1.on / lora_1.strength"`.
   This alone converts silent data loss into a loud, safe failure.
2. **Support dotted sub-paths** — `lora_1.on`, `lora_1.strength`, `lora_1.lora`. The error message already enumerates widget names, so the plumbing to resolve a path is close by.
3. **Widen `value` to accept an object** and merge (rather than replace) into the existing composite, so unspecified keys are preserved.

(1) is the minimum safety fix; (2) is what makes the node actually drivable, which is the underlying user request — there is currently **no way for a panel agent to toggle or restrength a LoRA** in the most widely used LoRA loader in the ecosystem.

### Related

Same family as #558 (`panel_set_widget` reporting success on a seed it cannot hold). The common thread: **`panel_set_widget` reports success without verifying the write took effect.** A post-write read-back comparison would have caught both.

### Environment

Windows 11 26200 · ComfyUI 0.29.2 (Desktop, local) · RTX 3070 8GB · CUDA 13.0 · torch 2.10.0 · Python 3.13 · **comfyui-mcp 0.48.21** · panel: Manager v4 · rgthree-comfy

artokun:
## Summary

Fixes two `panel_set_widget` honesty bugs on composite widgets and seed-control widgets. Both filed against the same root cause: **`panel_set_widget` reported success without the value actually holding.**

### #560 — scalar-to-composite corruption (data loss)

rgthree's **Power Lora Loader** stores each slot as a composite object `{on, lora, strength, strengthTwo}`. Writing a bare scalar to `lora_1` used to overwrite one field and null the rest, reporting success.

- The silent corruption itself was already refused by #179's composite branch on `origin/main` (a bare scalar now throws), but the node was **not drivable**: dotted addressing errored and the refuse message did not explain the fix.
- This PR adds explicit **sub-field addressing** — `panel_set_widget(node_id, 'lora_1.on', false)` merges **only** that field and preserves `lora`/`strength`/`strengthTwo`. Resolution is **exact-name-first** (a real widget whose own name contains a dot always wins; an empty suffix like `lora_1.` is refused).
- Both the dotted path **and** the existing full-JSON-object path now go through one `mergeCompositeFields()` that **fails closed on an unknown field** (no junk members added) and **validates each value against the existing field's type** (no storing the string `"false"` where a boolean is expected).
- On a **subgraph parent**, a dotted sub-field is refused rather than writing the projected rail directly (which would bypass the #366 atomic inner+rail path) — promoted composites are driven with a full JSON object instead.

### #558 — `control_after_generate` is invisible (silent failure)

A seed governed by a non-`fixed` `control_after_generate` is silently overwritten after each run. The control combo (`serialize:false`/`canvasOnly`) rendered unassociated from the seed it governs, with no write-time signal.

- `panel_query_graph {fields:'detail'}` now surfaces an explicit, seed-associated `control_after_generate` map (e.g. `{ "seed": "randomize" }`); `panel_graph_outline` marks the governed widget (`seed=… [after_gen=randomize]`).
- `panel_set_widget` returns a **`warning`** when the written value will not persist, pointing at the exact control widget to set to `fixed`. Wording is mode-accurate (randomize = new random value; increment/decrement = ±1 each run). Warnings are computed on the **concrete** node/widget through nested promotions, only for genuinely governed value widgets (never the control widget itself).

## Verification

- Detection extracted to `web/js/lib/control-after-generate.js` (unit-tested); composite sub-field merge + type validation + warning tested end-to-end through `runSetWidget`, including a nested `A→B→KSampler` promotion.
- **795 unit tests pass** (`node --test browser_tests/unit/*.test.mjs`); panel module + libs parse clean as ES modules.
- Preserves #423 fail-closed, #347 empty-string-clear, #240 strictness, and #179/#366 composite-merge/promoted-rail behavior.
- Passed an adversarial Codex self-review (5 rounds; each of 6 findings fixed and re-verified).

Closes artokun/comfyui-mcp#560
Closes artokun/comfyui-mcp#558

### DIFF
--- a/browser_tests/unit/control-after-generate.test.mjs
+++ b/browser_tests/unit/control-after-generate.test.mjs
@@ -1,0 +1,289 @@
+/**
+ * Unit tests for web/js/lib/control-after-generate.js (#558) — run with `node --test`.
+ *
+ * control_after_generate SILENTLY rewrites a seed/INT/COMBO value after each
+ * generation, and the ComfyUI frontend adds it as a serialize:false/canvasOnly
+ * widget that renders UNASSOCIATED from the seed it governs — so a value the agent
+ * explicitly set never holds, with nothing in the read/write surface revealing it.
+ * These tests pin the detection that summarizeNode / graph_outline / runSetWidget use.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import {
+  CONTROL_AFTER_GENERATE_MODES,
+  isControlAfterGenerateWidget,
+  controlAfterGenerateEntries,
+  controlAfterGenerateModes,
+  controlEntryForWidget,
+  controlAfterGenerateWarning,
+} from "../../web/js/lib/control-after-generate.js";
+
+// A KSampler-shaped node: seed value widget followed by its control combo, exactly
+// as the ComfyUI frontend builds it (control widget serialize:false/canvasOnly).
+function seedNode(mode = "randomize", { linked = true, prefix = "" } = {}) {
+  const seed = { name: "seed", type: "INT", value: 777777 };
+  const control = {
+    name: `${prefix}control_after_generate`,
+    type: "combo",
+    value: mode,
+    options: { values: [...CONTROL_AFTER_GENERATE_MODES], serialize: false, canvasOnly: true },
+  };
+  if (linked) seed.linkedWidgets = [control];
+  return { id: 30, type: "KSampler", widgets: [seed, control, { name: "steps", type: "INT", value: 20 }] };
+}
+
+test("detects the control_after_generate combo by name + option shape", () => {
+  const { widgets } = seedNode("randomize");
+  assert.equal(isControlAfterGenerateWidget(widgets[1]), true);
+  assert.equal(isControlAfterGenerateWidget(widgets[0]), false); // the seed itself
+  assert.equal(isControlAfterGenerateWidget({ name: "sampler", options: { values: ["euler"] } }), false);
+});
+
+test("detection requires BOTH control markers (serialize:false AND canvasOnly:true) plus the mode options", () => {
+  // No marker at all (an unrelated combo that merely shares the option strings) ⇒ NOT a control.
+  assert.equal(
+    isControlAfterGenerateWidget({ name: "mode", type: "combo", options: { values: [...CONTROL_AFTER_GENERATE_MODES] } }),
+    false,
+  );
+  // Only ONE marker (serialize:false, no canvasOnly) ⇒ NOT a control (P2 false-positive guard).
+  assert.equal(
+    isControlAfterGenerateWidget({ name: "mode", options: { values: [...CONTROL_AFTER_GENERATE_MODES], serialize: false } }),
+    false,
+  );
+  // Only canvasOnly, no serialize:false ⇒ NOT a control.
+  assert.equal(
+    isControlAfterGenerateWidget({ name: "mode", options: { values: [...CONTROL_AFTER_GENERATE_MODES], canvasOnly: true } }),
+    false,
+  );
+  // BOTH markers present but options are NOT the modes ⇒ NOT a control.
+  assert.equal(
+    isControlAfterGenerateWidget({ name: "x", options: { values: ["a", "b"], serialize: false, canvasOnly: true } }),
+    false,
+  );
+  // BOTH markers AND the mode options ⇒ detected (the real frontend shape).
+  assert.equal(
+    isControlAfterGenerateWidget({
+      name: "control_after_generate",
+      type: "combo",
+      value: "increment",
+      options: { values: [...CONTROL_AFTER_GENERATE_MODES], serialize: false, canvasOnly: true },
+    }),
+    true,
+  );
+});
+
+test("associates the mode with the seed via linkedWidgets", () => {
+  const node = seedNode("randomize", { linked: true });
+  assert.deepEqual(controlAfterGenerateModes(node), { seed: "randomize" });
+  const entries = controlAfterGenerateEntries(node);
+  assert.equal(entries.length, 1);
+  assert.deepEqual(entries[0], { widget: "seed", control: "control_after_generate", mode: "randomize" });
+});
+
+test("associates the mode with the seed by POSITION when linkedWidgets is absent", () => {
+  const node = seedNode("increment", { linked: false });
+  assert.deepEqual(controlAfterGenerateModes(node), { seed: "increment" });
+});
+
+test("handles a control_prefix-renamed control widget", () => {
+  const node = seedNode("decrement", { linked: true, prefix: "Base " });
+  const entries = controlAfterGenerateEntries(node);
+  assert.equal(entries[0].widget, "seed");
+  assert.equal(entries[0].mode, "decrement");
+  assert.equal(entries[0].control, "Base control_after_generate");
+});
+
+test("P2: a combo-governed control with a 5th 'increment-wrap' option IS detected (superset, not exact)", () => {
+  // ComfyUI appends 'increment-wrap' when the governed widget is a combo — detection must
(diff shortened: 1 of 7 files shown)


artokun:
Independent adversarial codex review (gpt-5.6-terra/high, 3 rounds) — CLEARED for merge.

Each round caught a real defect the author's own multi-round self-gate had passed:
- R1: **P0** — null-valued composite fields inferred type from the (null) current value → accepted any primitive (`lora_1.lora=5` silently made a `string|null` field numeric, reported success).
- R2: **P0** — a partially-corrupt row disabled its own schema and fell back to inferring from the corrupt value, deepening corruption; plus a detection side-effect that could mutate the just-written seed post-verification.
- R3: clean — no P0. Round-2 P0 verified fixed (partial-corrupt rows now enforce the declared schema and repair forward); control detection is side-effect-free and matches the upstream ComfyUI control shape.

**Accepted known limitation (documented):** classification is key-set-shape-only (`{on,lora,strength,strengthTwo}`), so a *foreign* composite that happens to share those exact keys would also be schema-enforced — a **fail-closed refusal, never corruption**, on a practically-nonexistent input (the `lora` field name is rgthree-specific). A node-type discriminator was considered and rejected because a detection miss would re-open the R2 corruption fall-through; shape-only fails safe. Merge-approved.

Issue #435, opened 2026-07-31
panel_set_widget on a promoted subgraph widget doesn't fire the inner node's widget callback (LoadImage image never reloads; MaskEditor opens the stale image)


artokun:
### What happened / root cause

`panel_set_widget` targeting a **promoted subgraph widget** updates the value but does not run the *inner* node's widget callback, so callback-driven side effects never happen. For a `LoadImage` inside a subgraph this means `showImage()` is never called and `node.imgs` is never repopulated — the node keeps displaying the previous image, and **MaskEditor (which reads `node.imgs`) opens the OLD image**.

`graph_set_widget` (web/js/comfyui-mcp-panel.js, ~line 4981) does the right thing for ordinary nodes:

```js
w.value = value;
// Fire the widget's own callback so combo/number side effects run —
// the same path a manual UI edit takes.
w.callback?.(value, app.canvas, node, node.pos, undefined);
```

But when `node` is a **subgraph host node**, `w` is the *promoted view* of an inner widget. Firing the promoted view's callback forwards the value down to the inner widget's `.value`, but never invokes the **inner** widget's own `callback`. Value propagation works (reading the inner node afterwards shows the new value) while every callback side effect is silently skipped.

This is silent — the tool returns `{ set: { previous, value } }` with correct values, so the agent believes the swap succeeded.

### Steps to reproduce

1. Have a subgraph containing a `LoadImage` whose `image` widget is promoted to the wrapper node.
   (Here: node `76` = "LAB 07 — ANIMA INPAINT" subgraph, inner `LoadImage` id `68`.)
2. Inner LoadImage currently holds `clipspace/clipspace-painted-masked-<A>.png`.
3. Call `panel_set_widget(node_id=76, widget="image", value="SHOWCASE_11a_cinematic_grade_00012_.png")`.
   → returns success, `previous` and `value` both correct; `panel_query_graph` on inner node 68 also shows the NEW value.
4. Right-click the node → **Open in MaskEditor**.
   → MaskEditor opens the **OLD** image (`<A>.png`), not the newly-set one.
5. Paint a mask and save → the resulting clipspace PNG is derived from the OLD image.

Confirmed by pixel dimensions: the newly-set source is 832x1216, but the saved mask PNG came out 1344x1472 (the dimensions of the previously-loaded image), and visually matched the old image.

### Exact error (scrubbed)

[...]

Issue #233, opened 2026-07-30
panel_set_widget on a subgraph node's PROMOTED widget writes by shifted position and corrupts the inner node (official SeedVR2 template unqueueable)


artokun:
### What happened / root cause

Loading the official ComfyUI template `utility_seedvr2_3b_int8_upscale_video.json` with `panel_load_workflow(path:...)` yields a subgraph node (`Upscale Video (SeedVR2 3B Int8)`) whose PROMOTED widget values are misaligned by 2 positions against the inner `KSampler`, so the graph cannot be queued at all.

The template's subgraph node carries `widgets_values: []`; the inner KSampler carries `widgets_values: [959948902156062, "fixed", 1, 1, "euler", "simple", 1]` with `seed` / `sampler_name` / `scheduler` / `denoise` promoted to the parent as rails (`-10.5`..`-10.8`). `panel_graph_outline` and `panel_query_graph` INSIDE the subgraph read the inner KSampler correctly (`steps=1 cfg=1 sampler_name=euler scheduler=simple denoise=1`), but the parent's promoted widgets read `sampler_name=1, scheduler=1, denoise="euler"` — i.e. shifted by two.

The part that is squarely ours: **`panel_set_widget` on the parent's promoted widget resolves the name on the parent but writes into the shifted inner slot, silently corrupting the inner node** — it reports success (`previous`/`value` look right) while making the graph worse. There is no name-safety check and no type validation that the written value matches the target widget's declared type, so an INT widget ends up holding `"euler"`.

Note: the panel load path itself is just `app.loadGraphData(clone)` (web/js/comfyui-mcp-panel.js, `graph_load`), so the initial 2-slot misalignment may well originate upstream in the frontend's promoted-widget derivation when the subgraph node's `widgets_values` is empty. Filing here because (a) it reproduces through `panel_load_workflow` and (b) the `panel_set_widget` corruption + false success is ours regardless of who causes the initial shift.

### Steps to reproduce

1. `panel_load_workflow(path: "<site-packages>/comfyui_workflow_templates_json/templates/utility_seedvr2_3b_int8_upscale_video.json")`
2. `panel_run(to_node_id: <SaveVideo id>)` -> validation error on KSampler (see below).
3. `panel_set_widget(node_id: <subgraph node>, widget: "sampler_name", value: "euler")` (and `scheduler`, `denoise`) -> each returns success.
[...]

Issue #426, opened 2026-07-31
panel_search_nodes fails generically with legacy ComfyUI-Manager instead of falling back


artokun:
### What happened / root cause
`panel_search_nodes({query:"ControlNet"})` returned `Error: ComfyUI-Manager not reachable (is the built-in Manager enabled?)` on a connected ComfyUI 0.28.3 instance using Manager 3.x LEGACY. The panel otherwise works. The search tool appears to require the newer Manager API and has no fallback to ComfyUI `/object_info` for already-installed node discovery.

### Steps to reproduce
1. Run ComfyUI 0.28.3 with ComfyUI-Manager 3.x legacy.
2. Connect comfyui-mcp panel.
3. Call `panel_search_nodes({query:"ControlNet"})`.

### Exact error
`Error: ComfyUI-Manager not reachable (is the built-in Manager enabled?)`

### Fix
Applied locally: no. Upstream-only: yes.
Suggested change: detect legacy Manager separately and either fall back to `/object_info` for installed-node search or return a precise `Manager v4+ required for registry/install search` response while still listing installed matches.

### Environment
Windows 11; NVIDIA RTX 5070 Ti Laptop GPU 12 GB; ComfyUI 0.28.3; Manager 3.x LEGACY; comfyui-mcp 0.48.21; panel version unavailable in startup environment.

PR #479, opened 2026-08-01
fix(set_widget): frontend-only rgthree nodes (#475) + outer promoted display-proxy sync (#477)

### LINKED ISSUE #475: panel_set_widget rejects frontend-only rgthree Fast Bypasser nodes
panel_set_widget refuses to edit widgets on frontend-only rgthree nodes because it requires every live canvas class_type to exist in the ComfyUI backend object-info registry. "Fast Bypasser (rgthree)" is a frontend graph/control node and legitimately has no backend definition, so the safety check blocks a reversible canvas edit that the frontend can perform.

### Steps to reproduce
1. Open a workflow containing rgthree "Fast Bypasser (rgthree)".
2. Call panel_set_widget on its boolean bypass toggle.
3. The tool rejects the edit before sending it to the live canvas.

### Exact error (scrubbed)
Error: Cannot set widget on node 302 ("Fast Bypasser (rgthree)"): the ComfyUI backend does not provide node type "Fast Bypasser (rgthree)" (not installed, or its pack was removed) — refusing to write to a node the live backend no longer defines (#458).

### Fix
applied locally: no
upstream-only: yes
The runtime package source could not be located in the available workspace/install paths within one focused attempt. Suggested change: allow panel_set_widget for a node that exists on the live canvas even when object-info lacks its class_type, at least for known frontend-only node classes, while retaining the stale-node guard for executable backend nodes. Alternatively validate frontend-only nodes against the live LiteGraph node/widget instance instead of backend object-info.

Temporary workaround: bypass the underlying VAEEncode node directly so rgthree Any Switch selects the EmptyLatentImage txt2img input.

### Environment
Windows 11; ComfyUI 0.18.0; RTX 3090 24 GB; comfyui-mcp 0.48.27; panel 0.11.22.

### LINKED ISSUE #477: Subgraph promoted widget write leaves outer widget stale
panel_set_widget on a subgraph promoted widget reported success and updated the inner widget, but the parent subgraph widget remained stale. applyWidgetWrite writes only the resolved inner widget and never synchronizes the outer proxy widget.

### Steps to reproduce
1. Promote an inner widget to a subgraph node.
2. Call panel_set_widget on the parent subgraph node.
3. Query the parent node: its widget retains the old value although the tool reports success with promoted_from.

### Exact error
Silent stale state; no exception.

### Fix
applied locally: yes
upstream-only: no

After inner write verification, synchronize the matching outer widget:
if (promotedFrom) {
  const outerWidget = (node.widgets ?? []).find(
    (cand) => cand?.name?.toLowerCase() === String(widgetName).toLowerCase(),
  );
  if (outerWidget && outerWidget !== w) outerWidget.value = coerced;
}

### Environment
Windows 11 / ComfyUI 0.29.2 / RTX 4090 24GB / comfyui-mcp 0.48.27 / panel 0.11.19

artokun:
## Summary

Two `panel_set_widget` correctness fixes in the widget-write cluster, built on the shipped #458 fail-closed / #560 composite / #366 promoted-rail / #435 inner-callback work.

### #475 — frontend-only rgthree nodes were wrongly rejected
`panel_set_widget` refused a genuine frontend-only rgthree node (`Fast Bypasser (rgthree)`) because its `class_type` is absent from the authoritative `/object_info`. The #458 fresh-backend guard (`assertTypeAgainstFreshBackend`) fails closed on any type the backend doesn't define — but a frontend-only / native node (rgthree, `Note`, `Reroute`, `PrimitiveNode`) is registered in LiteGraph with **no** backend `nodeData` and legitimately has no `/object_info` entry.

**Root cause:** the guard treated "absent from object_info" as "removed/unverifiable" for *all* types, with no way to recognize a genuine frontend-only node.

**Fix:** new `isFrontendOnlyRegisteredType` (registered in the live registry with a **defless** class — no `nodeData`). `assertTypeAgainstFreshBackend` permits such a type **only when object_info was fetched but simply lacks it**. The #458 hole stays closed:
- a **removed** backend node keeps a stale-positive class that **still carries `nodeData`** → not frontend-only → still refused;
- an **unavailable** object_info (null) still fails closed for **every** type ("can't verify at all").

Same `nodeData` probe `assertResolvedTargetRegistered` already uses to trust native/defless types, so the downstream write guards stay consistent.

### #477 — promoted write left the OUTER display proxy stale
After a promoted subgraph widget write, the parent-facing display widget stayed stale even though the tool reported success with `promoted_from`.

**Root cause:** #366 synced only the **primary rail projection** (`input._widget`). In ComfyUI 0.29.2 a single promoted host input also references a **distinct display proxy** (`input.widget`, a real member of `node.widgets`) — left unsynced, so a query/render of the outer node saw the OLD value.

**Fix:** `resolveHostPromotedWidget` → `resolveHostPromotedWidgets` returns **every identity-authenticated projection** the host input references (rail + display proxy), by **identity, never by name** (a same-named decoy is never touched — #233 preserved). `applyWidgetWrite` syncs the display proxies **atomically** with the rail (snapshot → verify-stuck → roll back on failure). The post-write **topology recheck** and rollback partial-state diagnosis compare the **full projection set**, so a callback that swaps in a new same-named live proxy holding the old value is caught as drift — never a false success. Single-projection shape (all existing #366/#233 fixtures) is byte-identical.

## Tests
- `node --check` + `npm run test:unit` green (**1119** tests).
- #475: frontend-only writes; removed-type & unavailable-oracle fail-closed; frontend-vs-removed distinction; unit coverage for `isFrontendOnlyRegisteredType` / `assertTypeAgainstFreshBackend`.
- #477: dual-projection sync, display-proxy drift + full rollback, topology-swap hard-fail, single-projection regression.

## Review
Codex `gpt-5.6-terra` / high adversarial review: round 1 caught a real P1 (topology recheck ignored the display-proxy set → a proxy swap could report false success); fixed (full-set drift detection + replacement-proxy partial-state diagnosis) and re-reviewed → **PASS**, no P0/P1.

Closes #475
Closes #477

### DIFF
--- a/browser_tests/unit/node-resolve.test.mjs
+++ b/browser_tests/unit/node-resolve.test.mjs
@@ -717,3 +717,123 @@ test("#475 set_widget: a frontend-only node with object_info UNAVAILABLE ⇒ FAI
   );
   assert.equal(node.widgets[0].value, true, "unverifiable ⇒ no write even for a frontend-only type");
 });
+
+// ---- #475 P0 (adversarial, #458 fail-closed): absence of nodeData ALONE must NOT
+//      authorize a write. A REMOVED backend pack can leave a DEFLESS HUSK registered;
+//      it is NOT on the frontend-only allowlist, so it must STILL fail closed. --------
+
+test("#475 P0: isFrontendOnlyRegisteredType — a DEFLESS husk NOT on the allowlist is NOT frontend-only (removed-backend husk stays refused)", () => {
+  // A removed backend pack's frontend registration left a bare (defless) class named
+  // "RemovedBackendNode". It is defless AND registered — but NOT allowlisted, so it is
+  // NOT treated as frontend-only. Only genuinely-known frontend types are.
+  const reg = loadedRegistry([], ["RemovedBackendNode", "Note"]);
+  assert.equal(isFrontendOnlyRegisteredType(reg, "RemovedBackendNode"), false, "defless husk is NOT frontend-only without a positive allowlist marker");
+  assert.equal(isFrontendOnlyRegisteredType(reg, "Note"), true, "an allowlisted native IS frontend-only");
+});
+
+test("#475 P0: assertTypeAgainstFreshBackend — a DEFLESS husk absent from object_info STILL fails closed (allowlist is the positive marker, not nodeData-absence)", () => {
+  const reg = loadedRegistry([], ["RemovedBackendNode"]); // registered, defless, NOT allowlisted
+  const fresh = objectInfo(); // backend no longer provides it
+  assert.throws(
+    () => assertTypeAgainstFreshBackend(fresh, "RemovedBackendNode", 9, { registry: reg }),
+    /backend does not provide/i,
+    "a defless husk must not be exempted merely for lacking nodeData (#458)",
+  );
+});
+
+test("#475 P0 set_widget: a REMOVED backend node left as a DEFLESS HUSK ⇒ FAIL CLOSED (no fabricated success)", async () => {
+  // The exact adversarial #458 case the naive `!nodeData` exemption reopened: the pack
+  // was uninstalled but its frontend registration survives as a bare defless class, and
+  // a live node of that type sits on the canvas. object_info no longer lists it. The
+  // write MUST be refused, not authorized+reported as success.
+  const reg = loadedRegistry();
+  const ctor = function DeflessHusk() {}; // NO nodeData — but also NOT allowlisted
+  reg["RemovedBackendNode"] = ctor;
+  const node = { id: 405, type: "RemovedBackendNode", widgets: [{ name: "steps", type: "INT", value: 20 }], constructor: ctor };
+  await assert.rejects(
+    () =>
+      runSetWidget(node, "steps", 30, {
+        registry: reg,
+        getRegistry: () => reg,
+        getFreshObjectInfo: async () => objectInfo(), // backend does NOT provide it
+        ...HOOKS,
+      }),
+    (err) => err instanceof Error && /backend does not provide/i.test(err.message),
+  );
+  assert.equal(node.widgets[0].value, 20, "a removed-backend defless husk is never mutated (#458 stays closed)");
+});
+
+test("#475 P0 provenance: an ALLOWLISTED name whose class carries BACKEND provenance (comfyClass, even with no nodeData) is NOT frontend-only (name-collision husk stays refused)", () => {
+  // Codex-adversarial: a removed backend pack that used a reserved allowlisted name
+  // (e.g. "MarkdownNote") and left a class WITHOUT nodeData but still bearing the
+  // registerNodesFromDefs `.comfyClass` marker must NOT be exempted by name alone.
+  const reg = loadedRegistry();
+  const husk = function MarkdownNoteHusk() {};
+  husk.comfyClass = "MarkdownNote"; // backend-registration provenance (no nodeData)
+  reg["MarkdownNote"] = husk;
+  assert.equal(
+    isFrontendOnlyRegisteredType(reg, "MarkdownNote"),
+    false,
+    "a backend-provenance class is not frontend-only even under an allowlisted name",
+  );
+  // And end-to-end: the write is refused, not fabricated as success.
+  const node = { id: 406, type: "MarkdownNote", widgets: [{ name: "text", type: "string", value: "x" }], constructor: husk };
+  return assert.rejects(
+    () =>
+      runSetWidget(node, "text", "y", {
+        registry: reg,
+        getRegistry: () => reg,
+        getFreshObjectInfo: async () => objectInfo(), // backend does not provide MarkdownNote
+        beforeChange() {},
+        afterChange() {},
+        setDirty() {},
+      }),
+    /backend does not provide/i,
+  );
+});
+
+test("#475 P0 instance-provenance: a STALE BACKEND INSTANCE under an allowlisted name (bare registry class, but the INSTANCE constructor carries a backend def) ⇒ FAIL CLOSED", async () => {
+  // Codex round-3 path: the registry entry for "MarkdownNote" is a bare native class
+  // (passes the registry-class check), but the actual write-target NODE was created from
+  // a backend def and its OWN constructor still carries nodeData/comfyClass. That is a
+  // removed backend node, not a frontend-only one — the instance-provenance gate refuses
+  // it even though the registry class looks frontend-only.
+  const reg = loadedRegistry();
+  reg["MarkdownNote"] = function BareNative() {}; // bare registry class (looks frontend-only)
+  const backendCtor = function MarkdownNoteBackend() {};
+  backendCtor.nodeData = { input: { required: {} } }; // INSTANCE carries backend provenance
+  const node = {
+    id: 407,
+    type: "MarkdownNote",
+    widgets: [{ name: "text", type: "string", value: "x" }],
+    constructor: backendCtor,
+  };
+  await assert.rejects(
+    () =>
+      runSetWidget(node, "text", "y", {
+        registry: reg,
+        getRegistry: () => reg,
+        getFreshObjectInfo: async () => objectInfo(), // backend does NOT provide MarkdownNote
+        ...HOOKS,
+      }),
+    (err) => err instanceof Error && /backend does not provide/i.test(err.message),
+  );
+  assert.equal(node.widgets[0].value, "x", "a stale backend instance is never mutated (#458 stays closed)");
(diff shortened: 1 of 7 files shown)


artokun:
## Hardening round (P0 + P1 from adversarial re-review)

An independent adversarial review found a **P0** (the #475 exemption reopened the #458 fail-closed hole) and a **P1** (incomplete #477 rollback). Both fixed in `aef4ce5`.

### P0 — frontend-only exemption now requires positive provenance, not just "defless"
The original exemption keyed "safe to write" solely on a defless registered class (`!nodeData`). A **removed** backend pack can leave a defless husk registered → that alone would fabricate success on a node the backend no longer defines. The exemption now requires **three** independent positive signals, failing closed on any doubt:
1. **Reserved-namespace allowlist** membership (`FRONTEND_ONLY_NODE_TYPES`: core `Note`/`MarkdownNote`/`Reroute`/`PrimitiveNode` + vendor-namespaced `… (rgthree)` frontend controls) — a removed pack's arbitrary class_type is not reserved.
2. **No backend provenance on the REGISTERED class** — `hasBackendProvenance = nodeData || comfyClass` (ComfyUI's `registerNodesFromDefs` stamps both).
3. **No backend provenance on the WRITE-TARGET INSTANCE's own constructor** (`set-widget` passes `node: authTarget`) — closes the reserved-name-collision + stale-backend-instance path (a node named `MarkdownNote` whose own constructor retains a backend def is refused even if the registry class is bare).

`object_info` unavailable still fails closed **before** the exemption, for all types.

### P1 — #477 rollback now restores the promotion topology
Rollback restored values + `widgetId` but not `hostInput.widget`/`_widget`, so a callback that swapped in a replacement proxy left it wired after a "clean" rollback. Now snapshots `promotedHostWidgetRef`/`promotedHostProjectionRef` before the write, restores them in the rollback envelope, and read-back-verifies both (else honest partial-state failure).

### Tests
`npm run test:unit` green (**1126**). New guards, each **mutation-verified** (they fail against the pre-fix logic): defless-husk refused; allowlisted-name-with-`comfyClass` refused; stale-backend-instance under a bare native class refused; genuine frontend node still writable; topology restored after a rolled-back proxy-swap.

### Codex
`gpt-5.6-terra`/high adversarial review reached **PASS** after surfacing three real findings across rounds (defless-husk P0 → registry-vs-instance provenance P0 → topology-restore P1), each fixed. Flagging for your re-review as the owner of this #458 fail-closed / security-adjacent decision — in particular the reserved-namespace allowlist as the trust anchor for "genuine frontend-only".

artokun:
## Hardening round 2 (nested-intermediate P0 + rollback-topology P1)

Independent adversarial review found two more issues on the nested-promotion + rollback paths. Both fixed in `a0649cf`.

### P0 — authorize EVERY driven-through promotion container, not just the terminal
A nested promotion A→B→concrete fresh-authorized only the terminal, but `applyWidgetWrite` mutates + reports success on the immediate inner **and** drives the value through every intermediate container — each was trusted merely for being "defless with subgraph metadata."
- `isVirtualSubgraphContainer(node)` — positive structural proof of a **real** nested graph (`.subgraph` with `_nodes`/`getNodeById`), not a bare `subgraph:{}`. This matches the codebase's own universal subgraph-node convention (grep: the panel identifies subgraph nodes solely by `.subgraph` everywhere — no `isVirtualNode` marker exists).
- `assertMutatedNodeAuthorized()` — a mutated node absent from fresh `/object_info` is permitted **only** as a provenance-clean virtual-subgraph container (no `nodeData`/`comfyClass` on class **or instance** constructor) or a frontend-only leaf; present ⇒ authorized by presence; null map ⇒ fail closed.
- `collectPromotionIntermediates()` walks the same chain as `followPromotionToConcrete` and returns **every** intermediate; set-widget authorizes the outer parent + each intermediate. A removed-backend husk as immediate **or** deeper container, or a fake-`subgraph:{}` husk, is refused; a genuine N-level virtual chain still succeeds.

**Why the provenance gate + structural container check is sufficient (not just "defless"):** the only way a stale instance is provenance-clean (per `asset-staleness.js`) is a generic-fallback placeholder, which has **no functioning `.subgraph`** and no `_subgraphSlot` inputs — so it fails `isVirtualSubgraphContainer` and can never be a *resolving* intermediate (`collectPromotionIntermediates` only yields nodes whose subgraph genuinely resolves the chain; a non-resolving husk fails the pre-existing concrete-resolution guard). The intersection (provenance-clean) ∩ (functioning resolving container) is exactly genuine virtual subgraphs.

### P1 — restore + verify the outer promotion topology on rollback
Rollback restored host refs + values but not `node.widgets` membership, nor verified the live host-input identity. Fixes:
- snapshot `node.widgets`; restore membership/order **only when the original host input is still wired** into `node.inputs` (else a wholesale-replaced promotion is left for honest partial-state reporting, not masked into a false clean rollback);
- read-back now requires every captured proxy to remain a live member, the live projection set to match the captured set, **and** the live host input to be `===` the captured one — so a replaced input (same rail proxy, different `widgetId`) is reported as partial state.

### Tests
`npm run test:unit` green (**1133**). New guards, each **mutation-verified** (fail against pre-fix logic): removed-backend husk as immediate B and as deeper C (3-level) refused; fake-subgraph refused; genuine 3-level succeeds; proxy-swap + `parent.widgets` replacement rolled back to original; host-input replacement reported as partial state.

### Codex
`gpt-5.6-terra`/high reached **PASS** after probing the nested/intermediate mutation path specifically. It surfaced (and I fixed) chain-coverage for deeper intermediates and the live host-input identity gap; its final residual (structural checks are forgeable) resolved once shown the checks are stricter than the codebase's own universal `.subgraph` convention and that the generic-fallback placeholder — the only provenance-clean stale instance — cannot be a functioning resolving container. Flagging for your re-review as the #458 fail-closed / security-adjacent owner.

artokun:
## Hardening round 3 — observed-backend-history (ever-seen) trust root

Implemented the architectural direction: client-side signals (subgraph shape, name allowlist, `nodeData`/`comfyClass` provenance) are all client-forgeable and cannot be the fail-closed trust root for #458. The non-forgeable anchor is **what the backend actually reported over time**. Commit `4cd61ef`.

### Panel (`comfyui-mcp-panel.js`)
- **`seenObjectInfoTypes`** — session-scoped Set of every `class_type` ever reported in any backend `/object_info` this session, populated at every fetch point (`registerComfyNodeDefs` → reconnect + refresh_nodes + add_node + downloads; the set_widget + add_node fresh oracles; the #426 full-object_info fetch).
- **Startup seed** — `setup()` seeds the full baseline (`getNodeDefs`, up to 5×/150ms backoff), recording every pack present at page load before it can be removed mid-session. `graph_set_widget` **awaits** the seed (retries once if unseeded) before authorizing.
- **Reliability gate** (closes codex's failure-path finding) — `objectInfoHistorySeeded` is true only once a non-empty baseline was actually recorded. The predicate is `!seeded || seenObjectInfoTypes.has(t)` — while not reliably seeded it **fails closed** (every absent type refused), so a failed/empty seed never opens a hole.

### Logic (`node-resolve.js` + `set-widget.js`)
For a type **absent from current `/object_info`**: **ever-seen ⇒ removed backend ⇒ refuse**; never-seen ⇒ genuinely frontend-only ⇒ the prior allowlist + class/instance provenance + container-shape checks apply as defense-in-depth. This closes both previously-forgeable cases you/codex identified: the allowlisted pure-JS husk (`MarkdownNote` from a removed pack) and the provenance-clean real-subgraph node posing as a virtual container. `isVirtualSubgraphContainer` now also honors upstream `isVirtualNode`/`isSubgraphNode()`.

### P1
Rollback read-back now requires **exact `node.widgets` identity + order** vs the pre-write snapshot (same array ref, members, order) **and** the live host input `===` the captured one — so a stateful `afterChange` that re-adds a replacement proxy, or a replaced input, is reported as partial state, never a falsely-clean rollback.

### Residual (inherent, agreed with codex)
The only case that slips is a type **never observed before its removal** (a tab opened after removal, or a pack removed during a sub-second transient seed failure) — inherent unobservability; such a node is a placeholder that `assertResolvedTargetRegistered` refuses at write time. While unseeded, every absent type is refused.

### Verification
`npm run test:unit` green (**1139**). New guards mutation-verified: ever-seen husk refused (leaf); ever-seen provenance-clean container intermediate refused; never-seen native/rgthree allowed; genuine chain succeeds; stateful-afterChange re-add reported as partial state. Panel verified as ES module; no `onload`/`onerror` tokens added; svg count unchanged (66); no process-spawn literals. Codex `gpt-5.6-terra`/high **PASS** after probing the ordering race (startup seed) and the failure-path (reliability gate).

Issue #186, opened 2026-07-29
panel_new_workflow collides with an existing Unsaved Workflow tab and routes edits to wrong graph


artokun:
### What happened / root cause
`panel_new_workflow` was called while another unsaved workflow tab already existed. The tool returned `active: Unsaved Workflow`, and subsequent `panel_paste_nodes` calls successfully reported node ids from the newly created graph. Before the next edit, panel routing silently snapped back to the pre-existing `workflows/Unsaved Workflow.json` tab. All returned ids then referred to different node types or did not exist, so 20 valid `panel_connect` calls failed. `panel_list_workflows` exposed only one `Unsaved Workflow.json`, indicating a key/path collision between unsaved tabs.

### Steps to reproduce
1. Keep an existing unsaved workflow tab open.
2. Call `panel_new_workflow`.
3. Paste nodes and record the returned ids.
4. Issue further graph edits using those ids.
5. Observe graph tools targeting the older unsaved tab; `panel_graph_outline` shows the wrong graph and ids no longer match.

### Exact error (scrubbed)
Examples: `Could not connect node 3 (KSampler) ? node 11 (CheckpointLoaderSimple)` although the immediately preceding paste reported node 3 as `LoraLoader` and node 11 as `CLIPTextEncode`; other calls returned `No node with id 13 in the current graph`.

### Fix
Applied locally: no. Upstream-only: yes.
Suggested minimal fix: give every unsaved tab a unique opaque routing key instead of `workflows/Unsaved Workflow.json`, and have `panel_new_workflow` return that key. Keep the session bound to that key until an explicit tab switch. Workaround: save/rename the new tab immediately to a unique filename, then use `panel_set_workflow_target(mode:'pinned')` before pasting/editing.

### Environment
Windows 11; ComfyUI 0.28.0; NVIDIA RTX 4070 SUPER 12 GB; comfyui-mcp 0.48.5; panel version unavailable.

artokun:
Partially addressed in **panel 0.11.5** (WS-1): the pinned-target mismatch guard reduces the blast radius, but the residual collision (two "Unsaved Workflow" tabs sharing a `wf.key`) still needs instance-based key disambiguation. Keeping open for that follow-up.

PR #481, opened 2026-08-01
fix(session): emit durable per-instance workflow_uuid in hello (#570)

artokun:
Paired producer for **artokun/comfyui-mcp#570** (orchestrator PR: artokun/comfyui-mcp#614).

## Why

#587 keyed an unsaved workflow's durable resume on `origin + title + backend`. That is not unique — every unsaved workflow shares the default `"Unsaved Workflow"` title — so after a reset/restart the orchestrator resumed an **unrelated same-title workflow's conversation** (the reopened #570).

## What

Advertise the panel's durable, reload-surviving, globally-unique **per-instance workflow uuid** — `workflowStableUuid()`, the same id that already keys the durable chat transcript and is forked on clone/import (#186/#386) — as `workflow_uuid` on **every** hello. The orchestrator now keys the stable resume index on that uuid instead of the non-unique title, so two distinct unsaved workflows can never cross-resume, and the id survives the `tmp:<uuid>` tab-id churn a reload causes.

- `buildHelloPayload()` gains a `workflowUuid` field -> emits `frame.workflow_uuid`, **omitted when blank** so old orchestrators are unaffected.
- `sendHello()` passes `workflowUuid: workflowStableUuid()`.

`workflowStableUuid()` returns `crypto.randomUUID()`-format ids in every branch, which the orchestrator validates against a strict uuid regex (and folds together with the server-observed trusted origin).

## Tests

`session-rebind.test.mjs`: `workflow_uuid` rides the hello when present; omitted when blank/unknown. Full panel unit suite green.

Draft — do not merge; land together with the orchestrator PR after human re-review (session-loss is high-stakes).

### DIFF
--- a/browser_tests/unit/session-rebind.test.mjs
+++ b/browser_tests/unit/session-rebind.test.mjs
@@ -257,6 +257,19 @@ test("#296: resume rides the hello only when present", () => {
   assert.equal(buildHelloPayload({ tabId: "t", resume: "sess-1" }).resume, "sess-1");
 });
 
+test("#570: the durable per-instance workflow uuid rides the hello when present", () => {
+  // Lets the orchestrator key an UNSAVED workflow's resume on a globally-unique id
+  // that survives the tmp:<uuid> tab-id churn — so a same-title sibling can't
+  // cross-resume its conversation (the reopened #570 residual).
+  const frame = buildHelloPayload({ tabId: "tmp:abc", title: "Unsaved Workflow", workflowUuid: "wf-uuid-1" });
+  assert.equal(frame.workflow_uuid, "wf-uuid-1");
+});
+
+test("#570: workflow_uuid is omitted (never blank) when unknown, so old behavior stands", () => {
+  assert.equal("workflow_uuid" in buildHelloPayload({ tabId: "t" }), false);
+  assert.equal("workflow_uuid" in buildHelloPayload({ tabId: "t", workflowUuid: "   " }), false);
+});
+
 // ---- #379 soft-reload recovery: never leave the bridge dead -----------------
 
 test("#379: a REFUSED reload (by-design 503) still reconnects, dropping the interlock", () => {
--- a/browser_tests/unit/session-rebind.test.mjs
+++ b/browser_tests/unit/session-rebind.test.mjs
@@ -257,6 +257,17 @@ test("#296: resume rides the hello only when present", () => {
   assert.equal(buildHelloPayload({ tabId: "t", resume: "sess-1" }).resume, "sess-1");
 });
 
+test("#570 P0c: hello ALWAYS advertises enforces_workflow_stamp so the orchestrator won't gate this build's graph edits", () => {
+  // A current build unconditionally enforces the per-command workflow stamp; the flag must
+  // ride every hello (not gated behind any optional field) so the server never fails closed
+  // on our mutating graph commands.
+  assert.equal(buildHelloPayload({ tabId: "t" }).enforces_workflow_stamp, true);
+  assert.equal(
+    buildHelloPayload({ tabId: "wf:x.json", title: "x", backend: "codex" }).enforces_workflow_stamp,
+    true,
+  );
+});
+
 test("#570: the durable per-instance workflow uuid rides the hello when present", () => {
   // Lets the orchestrator key an UNSAVED workflow's resume on a globally-unique id
   // that survives the tmp:<uuid> tab-id churn — so a same-title sibling can't
--- a/browser_tests/unit/workflow-chat-identity.test.mjs
+++ b/browser_tests/unit/workflow-chat-identity.test.mjs
@@ -6,8 +6,23 @@ import {
   isWorkflowCreationLoad,
   normalizedWorkflowPath,
   shouldForkEmbeddedWorkflowUuid,
+  trustedUnsavedEmbeddedUuid,
   workflowAliasForPath
 } from '../../web/js/lib/workflow-chat-identity.js'
+
+// #570 P0b — trust an unsaved workflow's embedded uuid ONLY when it carries the fork marker.
+test('#570 P0b a MARKED embedded uuid is trusted (fork-minted → unique per-instance)', () => {
+  assert.equal(trustedUnsavedEmbeddedUuid({ embeddedId: 'uuid-A', embeddedOwned: true }), 'uuid-A')
+})
+
+test('#570 P0b a LEGACY unmarked embedded uuid is NOT trusted → fork (null)', () => {
+  // A pre-rollout copy carries the source uuid with no marker → must not be adopted.
+  assert.equal(trustedUnsavedEmbeddedUuid({ embeddedId: 'uuid-A', embeddedOwned: false }), null)
+  assert.equal(trustedUnsavedEmbeddedUuid({ embeddedId: 'uuid-A' }), null)
+  // No embedded id at all → null (mint fresh).
+  assert.equal(trustedUnsavedEmbeddedUuid({ embeddedOwned: true }), null)
+  assert.equal(trustedUnsavedEmbeddedUuid({}), null)
+})
 
 // Fakes for the ComfyWorkflow 4th arg: a REUSE passes a ComfyWorkflow OBJECT; a CREATION
 // passes null/undefined/string.
--- a/browser_tests/unit/workflow-chat-identity.test.mjs
+++ b/browser_tests/unit/workflow-chat-identity.test.mjs
@@ -42,16 +42,30 @@ test('#570 fence applies to every graph_* op and all four workflow mutators (act
   assert.equal(activeWorkflowFenceApplies({ cmd: 'workflow_close', targetsNonActive: false }), true)
 })
 
-test('#570 fence does NOT apply to pinned / navigation / resolved-non-active workflow ops', () => {
-  // Pinned → the #349 workflow_path guard is the authority.
-  assert.equal(activeWorkflowFenceApplies({ cmd: 'graph_add_node', hasPin: true }), false)
-  assert.equal(activeWorkflowFenceApplies({ cmd: 'workflow_close', hasPin: true }), false)
+test('#570 fence ALSO applies to PINNED commands (the pin guard authorizes by path, not uuid)', () => {
+  // A pinned command (workflow_path present) is NOT exempt: after an in-place replacement at the
+  // SAME saved path the pin still matches but the uuid does not, and only the uuid fence catches
+  // it. activeWorkflowFenceApplies no longer takes a pin — the fence fires for pinned ops too.
+  assert.equal(activeWorkflowFenceApplies({ cmd: 'graph_add_node' }), true)
+  assert.equal(activeWorkflowFenceApplies({ cmd: 'workflow_save' }), true)
+  assert.equal(activeWorkflowFenceApplies({ cmd: 'graph_set_widget' }), true)
+})
+
+test('#570 fence does NOT apply to navigation / resolved-non-active workflow ops', () => {
   // Navigation / creation.
   assert.equal(activeWorkflowFenceApplies({ cmd: 'workflow_open' }), false)
   assert.equal(activeWorkflowFenceApplies({ cmd: 'workflow_new' }), false)
   // Selector RESOLVED to a genuinely non-active open workflow → deterministic target → runs.
   assert.equal(activeWorkflowFenceApplies({ cmd: 'workflow_close', targetsNonActive: true }), false)
   assert.equal(activeWorkflowFenceApplies({ cmd: 'workflow_rename', targetsNonActive: true }), false)
+})
+
+test('#570 pinned command after in-place replacement at the same path: uuid mismatch ⇒ REFUSED', () => {
+  // A pinned graph write stamped for A (uuid-A) whose workflow_path still matches the active
+  // canvas — but the workflow at that path was replaced in place by B (uuid-B). The fence APPLIES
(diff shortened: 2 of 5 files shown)


artokun:
## Update — combined-diff adversarial review (P0a / P0b / P1 + hardening)

An independent adversarial pass on the **combined** (mcp + panel) diff found the uuid keying correct but surfaced two more P0 wrong-resume paths and a P1 durability gap that bypassed it. All are now fixed, then further hardened across several review rounds. The **wrong-resume invariant is fully closed** (verified: no cross-workflow / cross-origin / cross-provider / reset-resurrection / cold-import / path-reuse resume remains).

### P0a — same-socket switch rebound one workflow's agent onto another (mcp)
The bridge stamps `migrated_from` on **any** same-socket re-hello under a new tab id (socket continuity, not workflow continuity). A workflow **switch** A→B on one browser tab was rebinding A's live agent onto B — B inherited A's conversation, B's own resume was refused, and A's session persisted under B's key.
- Compute the trusted `{server-origin, uuid}` identity **before** the migration decision; rebind **only** when both identities exist and are **equal**, else **retire** (new `PanelAgentManager.retire()`: stops the agent, **preserves** its durable session) — fail closed (old/degraded/malformed panels never rebind an unprovable continuation).
- `handleEvent()` drops all events once `closed` — a buffered backend event can't leak the retired workflow's reply/session into the switched-to tab (`retire-no-leak` test).
- Render-queued (`heldDuringGen`) messages: **explicit cancellation notice**, not a silent drop.
- `UiBridge.revokeTabMigration()` fences the old tab's route on a switch, so a stale `panel_*` call can't resolve onto the new canvas (ui-bridge regression test).

### P0b / P1 — unsaved uuid durability + cold-import fork (panel)
The uuid was resolved lazily from the graph `extra` embedded value — not durable in default panel-scoped mode (**P1**), and carried by a **copied** graph (**P0b**: a cold import/duplicate resumed the source's session).
- The unsaved uuid is now keyed on the pair **(ComfyUI tab path, reload-stable graph id)** in the panel's own localStorage, not `extra`. The path is reused on a reload (durable — P1) but **deduped** on a fresh import (`Unsaved Workflow (2).json` → new path → fresh uuid → fork — P0b). The graph id (`activeState.id`) is a second guard so a new workflow reusing a **freed path slot** can't inherit a closed one's uuid; reuse requires **both** ids present and equal (fail closed). Pure `resolveUnsavedWorkflowUuid` (unit-tested).

### Known residual (for your review — NOT a wrong-resume)
Codex flags one remaining **pre-existing** wrong-**canvas** race: a `panel_*` command dispatched to the browser **microseconds before** a same-socket switch can still execute on the new workflow after re-hello — the orchestrator cannot un-send a frame already delivered, and the bridge command frames carry no workflow generation the panel could verify. My change **improves** on the prior behavior (the route-fence closes the post-switch window; the pre-fix rebind masked this inside the larger wrong-resume bug) but cannot fully close the pre-switch in-flight window orchestrator-side. Fully closing it needs a **generation-bound panel-command protocol** (every command carries a workflow generation the panel checks before executing + invalidate pending on re-hello) — a separate cross-cutting workstream beyond this session-resume fix. Flagging for a follow-up decision; not reintroducing a wrong-resume to buy an approval.

Tests: mcp full suite green (2773 pass; 1 pre-existing unrelated `node-dev` flake excluded). Panel full unit suite green (1140). Draft — do not merge.

artokun:
## Update 2 — producer-side create-boundary fork + codex APPROVE

Two more wrong-resume/delivery paths found on the combined diff, both fixed; the combined fix now passes independent adversarial codex review (**approve** — "No ship-blocking wrong-resume or post-switch wrong-delivery path found").

### P0 (producer / panel) — a paste/duplicate copy could inherit the source's session
The unsaved identity was keyed on (ComfyUI path, graph id), but a PASTE/DUPLICATE copy **preserves both** the source's graph id and can share the synthesized `workflows/Unsaved Workflow.json` path (native temporaries share it) — so the copy presented the source's identity and the orchestrator rebound the source's live agent onto it (and resolved its stable session after restart). Path-dedup didn't hold.

Fixed by forking at the **creation boundary** instead: `installCreateBoundaryFork` wraps `app.loadGraphData` and mints a fresh `extra.comfyui_mcp.workflow_uuid` on every creation, identified by `isWorkflowCreationLoad` — ComfyUI passes the open `ComfyWorkflow` **object** as `loadGraphData`'s 4th arg for every REUSE (reload-restore, tab-switch, undo, reroute-migration, in-place repaint) and a non-object (null/string) for every CREATION (import/open/drag/template/shared-url/**duplicate**/**paste**/new-blank; external imports also set `openSource`). So a copy gets a fresh uuid; a reload round-trips the same uuid (durable — P1). The panel's own snapshot-revert opts out (`__cmcpNoFork` + passes the active workflow object). Verified against ComfyUI 1.42.15 (installed) + 1.47.2 (source): all 8 creation paths route through `loadGraphData` with a non-object 4th arg or `openSource`; every reuse passes a workflow object. The old (path,graph-id) localStorage machinery is removed; pure `isWorkflowCreationLoad` unit-tested.

### P0a refinement (consumer) — cross-provider render-queued work
`heldDuringGen` is keyed by `tab::backend`, so the workflow-switch cancellation only cleared the current provider's queue. Queue on A → switch provider → switch workflow left held work under the old provider (`onRunEnd` would respawn it for the switched-away workflow). Now the retire branch sweeps **every** backend for `migratedFrom` and emits one aggregate cancellation notice.

### Known residual (unchanged — separately tracked, NOT a wrong-resume)
A `panel_*` command already delivered to the browser microseconds **before** a switch can still act on the new canvas after re-hello — a wrong-**canvas** boundary issue needing a generation-bound panel-command protocol (a separate cross-cutting workstream). Codex confirms this is out of scope for the session-resume fix.

Tests: mcp full suite 2826 pass (1 pre-existing unrelated `node-dev` flake excluded); panel full unit suite 1162 pass. Both rebased onto the latest main (mcp 0.48.29 / panel post-0.11.29). Draft — do not merge.

artokun:
## Update 3 — the load-bearing resume-ownership P0 + saved-workflow identity binding

The independent review's core P0 (default panel-scoped chat resuming across workflows via an untrusted `hello.resume`) and the deeper saved-workflow identity edges are now fixed. **Every wrong-RESUME path is closed** (final codex round found no resume path); the only remaining codex finding is the separately-tracked in-flight wrong-**canvas** residual (below).

### Resume bound to trusted identity (server-side, load-bearing)
`hello.resume` is honored ONLY when the session it names is OWNED by the tab's trusted `{origin, uuid, backend}` — its exact-tab record (now identity-bound, see below) or its trusted stable-key entry. An unowned resume is dropped and no longer short-circuits the stable-key fallback. So the panel's global session can't cross-resume a fresh/copied workflow.

### Exact session bound to the FULL identity (origin+uuid), proven-ownership reset
The exact record (`wf:<path>::backend`) was keyed only on the tab id — a saved workflow **overwritten in place** (same path, new uuid), a **cross-origin** copied uuid on the same bridge port, or a **pre-upgrade/legacy** record could resume the prior/other workflow (via `hello.resume` *and* spawn's exact-store hit). Now each exact record is bound to the durable canonical identity `origin::uuid` (`SessionStore.Entry.u` via a manager `identityForKey` dep). On hello, an exact record is kept + resumable ONLY when **proven** — bound identity present, hello identity present, and equal. Every other case (in-place overwrite, unbound legacy record, no trusted identity, different origin) triggers a **full session boundary** (`manager.reset`: stops the live agent + clears pending resume/held mail + the durable session), before the ownership check and spawn.

### Creation-boundary fork made fail-safe
The `isWorkflowCreationLoad` classifier now recognizes a REUSE only when the `loadGraphData` 4th arg is genuinely a ComfyWorkflow (a non-array object with a string `path`); any primitive/array/mis-shaped object forks (bias to fork).

### Deliberate behavior changes (disclosed for your call)
- **Panel-scoped global chat no longer follows across workflows** — resume is bound to identity (per the #570 invariant; a persistent panel-global chat would be a separate authorized feature).
- **One-time reset on upgrade** for a saved-workflow chat whose on-disk record predates the identity binding (a lost resume, once).
- **Identity-less clients** (old panels / no trusted server origin) lose *saved-workflow* resume (reset per reload) — modern browser panels have a trusted handshake origin and prove out, so they're unaffected; unsaved tabs are unaffected (their tmp: id churns).

### Remaining residual (separately tracked — NOT a wrong-resume)
The in-flight wrong-**canvas** race: a `panel_*` command already delivered to the browser microseconds **before** a same-socket workflow switch can act on the new canvas. The orchestrator can't retract a delivered frame; the route-fence closes the not-yet-dispatched window but the full fix needs a **generation-bound panel-command protocol** (every command carries a workflow generation the panel verifies) — a separate cross-cutting workstream, out of scope for this session-resume fix.

Tests: mcp full suite 2831 pass (1 pre-existing unrelated `node-dev` flake excluded); panel full unit suite 1163 pass. Both rebased on latest main. Draft — do not merge.

artokun:
## Update 4 — full identity-boundary teardown; convergence on the deferred residual

The resume-ownership work is complete and the identity boundary now tears down **every** kind of per-tab state on an unproven workflow transition. Across the review cycle codex confirmed no wrong-**resume** remains; its final finding is the separately-tracked in-flight frame/command residual (below).

### Identity boundary — fully fail-closed, all state kinds
On hello, a tab's state is kept **only when positively proven** to be this workflow's identity: the state's identity **and** the hello's identity both present **and** equal (`origin::uuid`). Otherwise (in-place saved-workflow overwrite = same tab id, new uuid; cross-origin copied uuid; unbound pre-`u`/legacy record; or an identity-less client) everything is torn down **before** any replay/spawn/resume:
- live agent + armed pending resume + **failed-start held mail** (`manager.hasAnyState` → `manager.reset`),
- durable exact session (now bound to the full canonical `origin::uuid` via `Entry.u`),
- `heldDuringGen` render-held queue (swept across **all** providers),
- and the bridge's **buffered deliveries** — the bridge now runs `onPanelMessage` first and the reset calls `UiBridge.dropQueuedDeliveries` so a prior workflow's buffered `say`/`session`/`stream`/`show_media` can't replay into the replacement.

State that **is** proven this workflow's (a normal reload/reconnect) is untouched. A proven same-workflow **migration** (tab-id change) rebinds the agent and carries all held queues (all providers); a same-socket **workflow switch** retires + route-fences + cancels held work.

### Scope note (disclosed, for your call)
Browser panels always carry a trusted handshake WS Origin, so they prove out and are fully covered — the modern-panel norm. Only genuinely identity-less clients (non-browser/relay, or transient pre-`u`/old panels) pay a **lost resume** across reload/reconnect (never a cross-workflow leak) — the fail-closed trade-off.

### Remaining residual (separately tracked — the deferred workstream)
An **already-dispatched** frame or `panel_*` command — written to the socket microseconds **before** a same-socket switch — can still land in the newly selected workflow. Bridge frames carry no per-workflow generation, so the orchestrator cannot retract a frame already delivered to the browser; the route-fence + buffered-delivery drop close everything the server controls, but the pre-dispatched window needs a **generation-bound panel-command protocol** (every frame/command tagged with the workflow generation, validated panel-side) — a separate cross-cutting panel+bridge workstream, out of scope for this session-resume fix.

Tests: mcp full suite 2834 pass (1 pre-existing unrelated `node-dev` flake excluded), incl. new identity-transition, spawn-window live-agent, failed-start held-mail, cross-origin, pre-upgrade, and buffered-delivery-drop regression tests. Panel 1163 pass. Draft — do not merge.

artokun:
## Update 5 — structural inversion + legacy-uuid invalidation → codex **APPROVE**

The root pattern behind every prior leak ("teardown gated on an incomplete state-set") is fixed structurally, and independent adversarial codex now **approves**: *"no remaining queued or replayed cross-workflow path found."*

### P0a — inverted, unconditional-and-complete teardown (orchestrator)
The identity boundary now decides **only whether to KEEP** a tab's state, on positive proof (state identity **and** hello identity both present **and** equal `origin::uuid`). When **not** proven, it tears down **everything** for the tab id **unconditionally** — no longer gated on any state-presence check, so a channel the gate didn't enumerate (e.g. a bridge-buffered `show_media`/frame with no manager state) can't slip through. The complete, idempotent sweep: `manager.reset` across **all** providers (live agent + pending resume + failed-start held mail + durable session) + `heldDuringGen` (all providers) + `bridge.dropQueuedDeliveries` (missed frames + mailbox). A future channel is covered automatically.

### P0b — legacy copied-uuid invalidation (panel)
The create-boundary fork only forks *creation* loads, so a graph copied **before** the rollout, reloaded as a ComfyWorkflow (reuse), kept its duplicated embedded uuid. Fixed with a **fork-ownership marker** stamped whenever our fork mechanism mints a uuid; on unsaved load the embedded uuid is trusted **only when marked** — a legacy/unmarked embed is forked to a fresh identity and re-stamped (one-time strip post-upgrade, a lost resume never a leak), durable thereafter. A pre-rollout copied unsaved graph can no longer derive the source's `origin::uuid`.

### Codex final verdict: **approve**
> Ship for the targeted wrong-resume/delivery risk: no remaining queued or replayed cross-workflow path found.

### Sole remaining residual (separately tracked)
A frame/command **already written to the socket** before a workflow switch cannot be recalled by the server — the deferred **generation-bound panel-command protocol** (tag every frame/command with the workflow generation, validated panel-side). Out of scope for this session-resume fix.

Tests: mcp 2834 pass (1 pre-existing unrelated `node-dev` flake excluded); panel 1165 pass. Draft — do not merge.

Issue #314, opened 2026-07-30
panel_set_widget cannot drive LTXDirectorCS timeline UI (local_prompts/segments)


artokun:
## What happened
Agent used panel_set_widget on LTXDirectorCS (node type) widgets timeline_data, local_prompts, segment_lengths. Tool returned success and panel_query_graph showed the new values, but the node's custom JS timeline UI still displayed the previous/default global+segment prompts. User confirmed UI unchanged; generation behaved as if only character-sheet Ghost Mask refs drove the clip (still-like CS image as video).

## Root cause (custom node)
WhatDreamsCost-CSGlide/js/ltx_director.js:
- Hides timeline_data, local_prompts, segment_lengths as HIDDEN_WIDGET_NAMES
- Constructor parses timeline_data once into this.timeline (source of truth)
- commitChanges() rewrites local_prompts/segment_lengths FROM in-memory this.timeline segments
- Agent widget writes do not re-parse into this.timeline or refresh the DOM prompt fields

## Expected
Either panel_set_widget on timeline_data/local_prompts should force the custom UI to re-sync, OR the tool/docs should report that custom-timeline nodes are not agent-writable for prompts.

## Environment
ComfyUI 0.28.1, Windows, GTX 1070, comfyui-mcp 0.48.14, LTXDirectorCS from WhatDreamsCost-CSGlide

PR #509, opened 2026-08-01
fix(panel): reconcile local_prompts when timeline_data updates on PromptRelayEncodeTimeline (#506)

### LINKED ISSUE #506: PromptRelayEncodeTimeline timeline_data update leaves local_prompts stale
panel_set_widget updated PromptRelayEncodeTimeline.timeline_data successfully, but did not synchronize the derived local_prompts widget. The tool contract says timeline_data updates should re-sync the editor and regenerate derived widgets. The executable prompt therefore remained stale until local_prompts was manually set.

### Steps to reproduce
1. Open a workflow containing PromptRelayEncodeTimeline.
2. Call panel_set_widget on timeline_data with a JSON string containing a changed segment prompt.
3. Re-read the node with panel_query_graph fields=detail.
4. Observe timeline_data contains the new prompt while local_prompts still contains the prior prompt.
5. Workaround: call panel_set_widget on local_prompts directly; re-read confirms it updates.

### Exact error
Silent state desynchronization; no error was returned.

### Fix
applied locally: no
upstream-only: yes
Extend the panel's special timeline_data synchronization path to PromptRelayEncodeTimeline (not only LTXDirector variants), invoking the node callback/rebuild logic or deriving and writing local_prompts and segment_lengths atomically from timeline_data.

### Environment
Windows 11; ComfyUI 0.28.2; RTX 5080 16 GB; comfyui-mcp 0.48.31; panel 0.11.0.

artokun:
Closes #506

## Root cause

`PromptRelayEncodeTimeline` (pack `ComfyUI-PromptRelay`) has three hidden widgets:

| widget | role |
| --- | --- |
| `timeline_data` | JSON `{"segments":[{prompt,length,color}]}` — the state the node's in-browser `TimelineEditor` parses **once** (constructor / `onConfigure`) into `this.timeline` |
| `local_prompts` | **derived** — `segments.map(s => s.prompt).join(" \| ")` |
| `segment_lengths` | **derived** — `segments.map(s => s.length).join(", ")` |

The editor regenerates both derived widgets from `this.timeline` on every commit
(`syncWidgetsFromTimeline`). And the node's **Python `execute()` reads only `global_prompt`,
`local_prompts` and `segment_lengths` — it never reads `timeline_data`.**

So a raw `panel_set_widget` on `timeline_data` wrote the widget (so `panel_query_graph` showed
the new value and the tool "succeeded"), but never reached the editor, left both derived widgets
stale, and therefore **rendered with the previous prompts** — then got silently reverted on the
next UI touch. This is worse than the sibling #314 (LTXDirector) bug: there the desync was
cosmetic-then-reverted; here it silently produces the wrong image.

## Reconcile semantics chosen, and why

**Regenerate, not reject.** The derived widgets are a total, pure function of `timeline.segments`
— the node defines them that way — and since `execute()` never reads `timeline_data`, rejecting
the write would leave no way to drive this node at all. A `timeline_data` write now recomputes
both derived values from the same segments and applies all three **atomically** (everything is
computed before any mutation, so no path can leave `timeline_data` updated with the prompts
stale), then re-hydrates the live editor so its next commit is a no-op instead of a revert.

Unlike LTXDirector, this pack exposes no re-hydration entry point, so the derivation is mirrored
in `web/js/lib/prompt-relay-timeline.js`. A useful consequence: the write is correct **even with
no live editor** (node never rendered, pack JS not loaded) — the widgets are consistent and the
editor's constructor re-parses what we wrote.

Direct writes to `local_prompts` / `segment_lengths` are now refused with a redirect (they are
reverted on the next commit and desync the node the other way round).

## Can user-authored prompt text be lost?

No path loses it silently. Refused outright (anything the pack would coerce or reset):

- absent / empty / non-array `segments` — the pack resets to a blank two-segment default, wiping every prompt
- a non-object segment — the pack's parser throws and falls back to that same blank default
- a missing or non-string `prompt` — the pack coerces to `""`, destroying that segment's text
- a `length` the pack would clamp or truncate (`0`, `12.5`, `"24.7"`, `"2e3"`, past `MAX_SAFE_INTEGER`)
- a present non-string `color`

Disclosed rather than dropped:

- `replaced_out_of_band` — derived text that no timeline could have produced (e.g. written with #506's own workaround)
- `overwrote_uncommitted_edit` — text the user was typing that an explicit `segments` write replaced
- `superseded_timeline_data` — the `timeline_data` copy that lost the merge-base decision

Plus warnings for every state where PromptRelay's Python would encode something other than what
the timeline shows: a blank prompt (dropped by `if p.strip()`, shifting every later segment), a
literal `|` (splits into extra prompts), and padding whitespace (stripped before encoding — the
check uses Python's exact `str.isspace()` set, not JS `trim()`, which disagrees in both
directions).

**Merge base.** The trickiest part: `timeline_data` lags typed text by the editor's 120 ms
debounce, while the editor's own copy lags a workflow load by `onConfigure`'s 10 ms re-parse —
each candidate can be the stale one. The rule is widget-first: `timeline_data` wins whenever it
is still self-consistent with both derived widgets; the editor is used only when it is not,
which is exactly the debounce window. Either way the losing copy is reported.

## Verification

- 51 new unit tests in `browser_tests/unit/prompt-relay-timeline.test.mjs`; full suite 1295 passing, `tsc --noEmit` clean.
- Two of those extract the real PromptRelay branch out of `graph_set_widget` in `comfyui-mcp-panel.js` and execute it with stubs, pinning master → apply, derived → refusal, other → fall-through, and its ordering (after the LTX route, before `runSetWidget`).
- Self-gated with `codex -m gpt-5.6-terra`, 8 rounds to PASS. Rounds 1–7 each found a real defect: the debounce race (would have destroyed in-flight typed text), a stale-editor merge-base collision, Python-vs-JS whitespace divergence, unsafe-integer lengths serializing as `"1e+21"`, and two undisclosed-overwrite paths. All fixed; the only remaining note is a deliberate over-strict ceiling on absurd segment lengths.

### Still needs rig verification

Everything above is unit-level against a faked editor. On a real rig, with a workflow containing
`PromptRelayEncodeTimeline`: `panel_set_widget` `timeline_data` with a changed segment prompt →
re-read with `panel_query_graph fields=detail` and confirm `local_prompts` **and**
`segment_lengths` match the new timeline; confirm the node's canvas blocks and prompt box show
the new text; touch the timeline in the UI and confirm nothing reverts; then run the workflow and
confirm the render uses the new prompt.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+- `panel_set_widget` on `PromptRelayEncodeTimeline.timeline_data` no longer leaves the node rendering the PREVIOUS prompts. That node's Python `execute()` reads only `local_prompts` + `segment_lengths` — never `timeline_data` — and both are derived by the in-browser timeline editor, so a raw `timeline_data` write reported success while the render silently kept the old prompts (and was reverted on the next UI touch). The write now regenerates both derived widgets from the new timeline and applies all three atomically, re-hydrates the live editor so its next commit is a no-op, merges onto the node's current timeline so unmentioned fields are preserved, and REFUSES any value the node would silently coerce or reset to a blank default. Direct writes to the derived widgets are refused with a redirect, and a node found already out of sync returns its previous prompt text rather than dropping it silently (#506)
+
 ## [0.11.32] - 2026-08-01
 
 ### Fixed
--- a/browser_tests/unit/prompt-relay-timeline.test.mjs
+++ b/browser_tests/unit/prompt-relay-timeline.test.mjs
@@ -287,6 +287,36 @@ test("MID-TYPING: the live editor wins over a timeline_data widget lagging by th
   assert.equal(res.replaced_out_of_band, undefined);
 });
 
+test("MID-TYPING: an explicit segments write that DISCARDS in-flight text says so", () => {
+  // Replacing `segments` is the caller's stated intent, so it is applied — but the prompt text
+  // the user had typed and not yet committed is handed back rather than vanishing silently.
+  const { node, widgets, editor } = makeRelayNode();
+  editor.timeline.segments[0].prompt = "user was typing this";
+  widgets.local_prompts.value = derivePromptRelayWidgets(editor.timeline.segments).local_prompts;
+
+  const res = relay(applyPromptRelayTimelineWrite(node, { segments: [seg("agent set", 20)] }));
+  assert.equal(res.merge_base, "editor");
+  assert.equal(res.overwrote_uncommitted_edit, "user was typing this | b");
+  assert.ok(res.warnings.some((w) => w.includes("UNCOMMITTED prompt edit")));
+  assert.equal(widgets.local_prompts.value, "agent set");
+});
+
+test("MID-TYPING: a write that PRESERVES the in-flight text reports no overwrite", () => {
+  const { node, widgets, editor } = makeRelayNode();
+  editor.timeline.segments[0].prompt = "user was typing this";
+  widgets.local_prompts.value = derivePromptRelayWidgets(editor.timeline.segments).local_prompts;
+  const res = relay(applyPromptRelayTimelineWrite(node, { zoom: 4 }));
+  assert.equal(res.overwrote_uncommitted_edit, undefined);
+  assert.equal(widgets.local_prompts.value, "user was typing this | b");
+});
+
+test("a normal (non-debounce) write never reports overwrote_uncommitted_edit", () => {
+  const { node } = makeRelayNode();
+  const res = relay(applyPromptRelayTimelineWrite(node, { segments: [seg("totally different", 3)] }));
+  assert.equal(res.merge_base, "timeline_data");
+  assert.equal(res.overwrote_uncommitted_edit, undefined);
+});
+
 test("MID-TYPING: a pending debounce commit AFTER our write is a no-op (no rollback)", () => {
   const { node, widgets, editor } = makeRelayNode();
   editor.timeline.segments[0].prompt = "in flight";
--- a/browser_tests/unit/prompt-relay-timeline.test.mjs
+++ b/browser_tests/unit/prompt-relay-timeline.test.mjs
@@ -299,6 +299,25 @@ test("MID-TYPING: a pending debounce commit AFTER our write is a no-op (no rollb
   assert.ok(before);
 });
 
+test("POST-LOAD: a stale editor is rejected even when its derived strings COLLIDE with the widgets", () => {
+  // The dangerous near-miss: after a load the restored widgets agree with each other, but the
+  // old editor's timeline happens to derive the SAME prompt join and length list while
+  // differing in per-segment data. Preferring the widget whenever it is self-consistent means
+  // the stale editor can never win, so its timeline is not resurrected.
+  const { node, widgets, editor } = makeRelayNode({
+    timelineSegments: [seg("a", 24, { color: "#new" }), seg("b", 36, { color: "#new2" })],
+  });
+  editor.timeline = {
+    stalePreviousWorkflow: true,
+    segments: [seg("a", 24, { color: "#old" }), seg("b", 36, { color: "#old2" })],
+  };
+  const res = relay(applyPromptRelayTimelineWrite(node, {}));
+  assert.equal(res.merge_base, "timeline_data");
+  const written = JSON.parse(widgets.timeline_data.value);
+  assert.equal(written.stalePreviousWorkflow, undefined);
+  assert.deepEqual(written.segments.map((s) => s.color), ["#new", "#new2"]);
+});
+
 test("POST-LOAD: the timeline_data widget wins over an editor still holding the OLD workflow", () => {
   // onConfigure restores the widgets first and re-parses the editor ~10ms later. Merging onto
   // the editor in that window would resurrect the previous workflow's timeline.
@@ -476,6 +495,25 @@ test("WARNS about leading/trailing whitespace — the python side strips each en
   // A fully-blank prompt is reported by the stronger blank-prompt warning, not this one.
   const clean = relay(applyPromptRelayTimelineWrite(makeRelayNode().node, { segments: [seg("   ", 24)] }));
   assert.equal(clean.warnings.filter((w) => w.includes("leading/trailing whitespace")).length, 0);
+});
+
+test("the whitespace notice covers characters PYTHON strips but JS trim() does not", () => {
+  // python's str.strip() also removes U+001C…U+001F and U+0085; JS trim() does not. A prompt
+  // padded with one of those is dropped/shifted by the encoder, so it must still be reported.
+  for (const pad of ["\u001c", "\u001d", "\u001e", "\u001f", "\u0085", "\ufeff"]) {
+    const res = relay(
+      applyPromptRelayTimelineWrite(makeRelayNode().node, { segments: [seg(pad + "fox" + pad, 24)] }),
+    );
+    assert.ok(
+      res.warnings?.some((w) => w.includes("leading/trailing whitespace")),
+      `no whitespace warning for U+${pad.codePointAt(0).toString(16)}`,
+    );
+  }
+  // A prompt made only of those characters counts as BLANK (python drops it entirely).
+  const blank = relay(
+    applyPromptRelayTimelineWrite(makeRelayNode().node, { segments: [seg("\u001c\u0085", 24), seg("b")] }),
+  );
+  assert.ok(blank.warnings.some((w) => w.includes("EMPTY prompt")));
 });
 
 test("a UI-refresh failure does NOT fail the write, and is reported", () => {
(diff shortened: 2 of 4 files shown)


artokun:
## Addressed the HIGH finding — the join was load-bearing for a data-loss disclosure

You were right, and the same defect was in the sibling disclosure too. Both are fixed structurally, plus two more P0s my own re-gate surfaced while verifying the fix.

**The defect.** `overwrote_uncommitted_edit` and `superseded_timeline_data` were both gated on the *joined* `local_prompts` differing. `" | "` is not an injection, so equal joins never implied equal segments — your exact scenario (persisted `["old","c"]`, in-flight edit making segment 0 the valid prompt `"a | b"`, incoming write of `["a","b | c"]`) collapses to the same `"a | b | c"` on both sides and the disclosure was omitted.

**The fix.** New `sameSegmentContent(a, b)` compares segment arrays element-by-element on prompt (strict) and length; every "did the content change?" decision now goes through it. I swept the file as asked: the only remaining join comparisons are timeline-vs-*widget* (`derivedMatchesWidgets`, the `replaced_out_of_band` loop), which are legitimate — the widget stores nothing but the join, so "is this widget consistent with this timeline?" is the right question there. `derivePromptRelayWidgets` now carries an explicit comment that the join must never decide content change.

**Recovery data.** Both disclosures now return `{ prompts: [...], lengths: [...] }` per segment instead of a joined string — handing back `"a | b | c"` for a timeline whose real segments were `["a | b","c"]` would have told the caller their text was lost without telling them what it was.

**Two more P0s found re-gating the fix:**
1. The first gate required *both* segment arrays, so an editor copy paired with an **unreadable `timeline_data`** was discarded with no disclosure — the case where the editor is the *only* record of that content. The gate is now: anomalous when the two readable copies differ, **or** when an editor copy exists with no readable master. A headless node (no editor, readable widget) is deliberately not anomalous, so ordinary headless writes stay quiet.
2. Length equality used a plain `Number()` coercion, making `"2e3"` equal `2000` — but the pack's `parseInt` reads `"2e3"` as `2`, so a live 2000-frame segment could be replaced by a 2-frame one with the disclosure suppressed. `sameLength` now delegates to `normalizeSegmentLength`, the same lossless rule the write path uses, and anything it cannot *prove* equal counts as different (erring toward disclosing).

**Tests** (54 in this file, 1302 suite-wide, `tsc` clean): your exact pipe-collision scenario asserting the disclosure **must** fire despite identical joins; the mirror with genuinely identical segments asserting it must **not**; a length-only change under an identical prompt join; corrupt-master discloses / corrupt-master-reproduced stays quiet / headless stays quiet; and unit coverage of `sameSegmentContent` including `"2e3"` vs `2000`, `"0x10"` vs `16`, `null` vs `"null"`.

Rebased on `origin/main`. My codex gate returned **PASS with no findings** on the third round of this follow-up. Nothing you listed as passing was touched — the derive-then-assign ordering, separators, no-editor path, and refusal set are unchanged.

Left panel #510 (the pre-existing NUL map-key sentinels) alone as instructed.

artokun:
## Fixed — the join no longer decides authority, only consistency

You and codex were right, and my sweep was wrong: I checked every join comparison for whether it *reported* a change and concluded the survivors were fine, without asking what those survivors were being used to *decide*. `derivedMatchesWidgets` can only answer "could this widget hold this timeline's serialization?" — asking it "which of two colliding timelines is current?" is not a question a lossy value can answer, and that is exactly what `chooseMergeBase` was doing by checking the widget first on join equality.

**`chooseMergeBase` rewritten.** The derived widgets are now a *filter*, never a tie-break:

1. only one candidate readable → it wins;
2. both readable and **structurally identical** (`sameSegmentContent`) → no authority question exists, the master field is canonical;
3. they differ and exactly **one** is consistent with the derived widgets → that one is current (this is what separates the debounce window from the post-load window);
4. they differ and the filter **cannot** separate them — both consistent (the collision) or neither — → prefer the **live editor**, per your rule: a superseded master is recoverable from the saved workflow, an uncommitted keystroke is not.

Case 4 is reported as `merge_base_ambiguous: true` with a warning naming the cause, and the losing copy still comes back via `superseded_timeline_data` / `overwrote_uncommitted_edit` — a tie is declared, never resolved silently.

I considered probing the editor's private `_textCommitTimer` as a dirty-bit and deliberately did **not** make it load-bearing: it would be a feature-detected private field that, if ever set stale, could resurrect a post-load editor. The preference rule gets the same safety without depending on pack internals.

**Regression test** drives your exact scenario: persisted `["a","b | c"]` / `[24,36]`, editor at `["a | b","c"]` with identical lengths (both derive byte-identical `"a | b | c"` and `"24, 36"`), overlay `{ "zoom": 4 }` requesting no segment change. It asserts the editor's content **survives** into both `timeline_data` and the editor, that `merge_base_ambiguous` is set, that `superseded_timeline_data` returns `["a","b | c"]`, and that `overwrote_uncommitted_edit` is absent. A second test pins that a genuinely-desynced tie is flagged while an ordinary write is not.

**Mutation-verified** rather than assumed: I reverted `chooseMergeBase` to the join-first rule in a scratch copy and ran the scenario — `merge_base = timeline_data`, surviving prompts `["a","b | c"]`, the edit destroyed. The test genuinely pins the invariant.

Nothing on your confirmed-good list was touched: the round-1 scenario, `sameLength`, `contentSnapshot` being built pre-mutation, the compute-then-assign atomicity, and the anomaly gate are all unchanged — 1304 tests pass, `tsc` clean, YARA/vocabulary gates clean.

My codex gate returned **PASS with no findings**, including an explicit branch-by-branch walk of all nine states you'd expect (in-sync, debounce, post-load, post-load collision, metadata collision, neither-consistent, unreadable master, headless, empty). It also confirmed the residual you'd predict: a post-load stale editor whose joins collide *is* still selectable, that's unavoidable from the available signals, and it's disclosed and recoverable from the saved workflow.

Rebased on `origin/main`, head `941dfe4`.

artokun:
## Fixed — the classification now follows which branch settled authority

You're right that this is more than cosmetic. The disclosure is the safety net for the residual we accepted in case 4, and a net that cries wolf on every post-load write is not a net.

**Root of it:** the gate asked "do the two copies differ?" and then labelled any discarded editor copy an *uncommitted edit*. But branch 3 rejecting the editor via the derived-widget filter is **positive evidence it is stale** — that information was available and thrown away.

**Fix.** `chooseMergeBase` now reports which branch decided, surfaced as `merge_base_reason` (`only-readable-copy` / `copies-identical` / `filter-rejected-editor` / `filter-rejected-master` / `unresolved-tie`), and the disclosures are gated on it:

- `overwrote_uncommitted_edit` — only when the editor actually **held authority** and the write did not reproduce it (the real debounce case, the branch-4 tie, and the unreadable-master case).
- `superseded_timeline_data` — only when the master did **not** hold authority. I tightened this symmetrically: it had the same defect, firing when branch 3 selected the master and the caller then replaced its segments, which is plain caller intent rather than a copy set aside behind their back.
- `discarded_stale_editor` — **new**, for a copy the filter rejected. Payload only, no warning, as you specified: the content stays useful, the alarm doesn't.

The generic `anomalous` check now only guards the stale-editor payload, so an ordinary write on a node whose copies agree is completely silent.

**Mutation-tested, per your instruction.** Reverted the classification to the anomaly-based form in a scratch copy and ran the post-load `{zoom: 4}` case: it emitted `overwrote_uncommitted_edit: {prompts:["previous workflow"],lengths:[99]}` plus the full "this node held an UNCOMMITTED timeline edit... write those back if you meant to keep them" warning. The new test asserts `overwrote_uncommitted_edit === undefined` **and** `warnings === undefined`, so it fails against the old code — load-bearing.

Three tests added: the post-load metadata-only case (master selected, no uncommitted-edit warning, quiet payload returned); the post-load explicit-segments case (no false `superseded_timeline_data`); and a stale editor whose content the write happens to reproduce (reports nothing at all).

Nothing on your confirmed-good list was touched — the four `chooseMergeBase` branches, the case-2 boundary, `derivedMatchesWidgets` as filter-only, the round-2 collision behaviour, `merge_base_ambiguous` firing only on genuine ties, pre-mutation snapshots, and the contiguous three-assignment write are all unchanged. 1307 tests pass, `tsc` clean, YARA/vocabulary gates clean.

Codex gate: **PASS, no findings**, with a ten-state table of exactly which disclosure fires in each and whether it's honest, plus an explicit check that the tightened gating cannot suppress a genuine live-editor loss — the regression class that bit this review twice.

Rebased on `origin/main`, head `dc7b216`.

artokun:
Maintainer gate adjudication (final round). The last review finding — a metadata overlay on an UNREADABLE timeline_data rehydrates the live editor and could lose an in-flight edit (test at browser_tests/unit/prompt-relay-timeline.test.mjs:328) — is accepted as a documented residual:

- Mid-typing text is in `editor.timeline` IMMEDIATELY (the textarea handler updates it synchronously; only the timeline_data JSON write is debounced), so `fromEditor` already carries in-flight edits and the discard guard fires on any write that doesn't reproduce them. The remaining window is keystrokes pending in the DOM before the input event — a sub-frame interval no application code can observe.
- The pinned test covers a metadata-only overlay that discards NOTHING (segments reproduced) — the refusal path for actual discards exists and is tested (`CORRUPT MASTER: discarding the SOLE live editor copy FAILS CLOSED`).

Across six gate rounds this PR eliminated: unresolved-tie overwrites, explicit-push-mid-edit overwrites, index-keyed metadata carry onto wrong segments, active reorder/drag stale-index splices, and sole-copy unreadable-master overwrites. The residual above is filed for a future pass. Merging on CI green.

Issue #387, opened 2026-07-31
panel_set_widget rejects image uploaded moments earlier because LoadImage combo options stay stale


artokun:
### What happened / root cause
upload_image successfully registered boswellia_ad_label_repair_source.jpg through ComfyUI /upload/image. Immediately afterward, panel_set_widget on an existing LoadImage node rejected that filename because its cached combo options contained only example.png. panel_get_errors continued to report the old missing clipspace image.

### Steps to reproduce
1. Open a workflow with a LoadImage node.
2. Call upload_image for a new filename; observe success.
3. Call panel_set_widget({node_id:<LoadImage>, widget:"image", value:<uploaded filename>}).
4. It refuses the valid uploaded input because the combo options are stale.

### Exact error
panel_set_widget refused "image" on node 191 (LoadImage): Value "boswellia_ad_label_repair_source.jpg" is not a valid option for combo widget "image". Valid options (1): "example.png"

### Fix
applied locally: no
upstream-only: yes
Suggested change: refresh dynamic COMBO choices from live /object_info (or specifically refresh input assets) before validation, and allow a LoadImage filename that the server confirms exists in its input directory. upload_image could also invalidate the panel combo cache.

### Environment
Windows 11; ComfyUI 0.28.0; frontend 1.45.21; comfyui-mcp 0.48.12; panel 0.11.10.

artokun:
Additional report: on Windows 11 / ComfyUI 0.29.2, an `upload_image` result under a nested path (`xyr_canvas/<asset>.png`) was rejected by `panel_set_widget` for `LoadImage.image` even after it refreshed combo options; the valid-options list omitted the nested path. The reporter is on comfyui-mcp 0.48.22 and panel 0.11.20; their temporary workaround is uploading at the input root. Updates are available: comfyui-mcp 0.48.23 and panel 0.11.22.

Issue #510, opened 2026-08-01
web/js/comfyui-mcp-panel.js carries two literal NUL-byte map-key sentinels on main — risks the panel file diffing as binary


artokun:
`web/js/comfyui-mcp-panel.js` on `main` contains **two literal NUL bytes** (`\0`) used as composite map-key separators. Verified with `tr -dc '\000' < file | wc -c` → `2`, at byte offsets **632745** and **634735**:

```js
const key = bk + "\0" + m.id;
if (seen.has(key)) ...
```

```js
const key = m.provider + "\0" + m.id;
if (pushed.has(key)) ...
```

### Why this matters

Functionally the sentinel works — `\0` is a legal JS string character and won't collide with a provider name or model id. The problem is **file integrity and reviewability**:

- Git treats a file as **binary** if it finds a NUL in its sampled window. This file is large enough (~630 KB before the first sentinel) that the sample window misses them today, so it still diffs as text. That is luck, not design — it depends on the sentinels staying far enough into the file.
- If they ever fall inside the sampled window (or a tool samples differently), the diff collapses to `Bin X -> Y bytes`, `--numstat` reports `- -`, and **the whole panel file becomes unreviewable in PRs**. That is a serious problem for a file this central.
- This is a recurring pattern in this codebase: agent-authored code reaches for `\0` as a "can't collide" separator, tests and builds pass, so self-review gates miss it. These two are the residue of that.

### Fix

Replace both with a printable separator that cannot appear in either component — `␟` (SYMBOL FOR UNIT SEPARATOR), a literal `"|"` if the components are known not to contain it, or simply key the map on a nested `Map`/composite object and avoid string concatenation entirely. Then assert the file is NUL-free so it cannot regress:

```sh
test "$(tr -dc '\000' < web/js/comfyui-mcp-panel.js | wc -c)" -eq 0
```

Worth wiring that assertion into the existing lint/YARA gate step, since that gate already runs over this file and is the natural place to catch the next one.

### Notes

- Found while reviewing PR #509 (unrelated); that PR does **not** touch these lines and is not blocked by this.
- Low severity — no user-visible misbehaviour. This is a maintainability/reviewability cleanup.

artokun:
**Update — there is a third NUL byte, in a second file.** Verified the same way:

- `web/js/comfyui-mcp-panel.js` — **2** (offsets 632745, 634735), `bk + "\0" + m.id` and `m.provider + "\0" + m.id`
[...]

PR #511, opened 2026-08-01
fix(panel): share one frontend-only-node allowlist between the sibling widget guards (#496) + writable empty dynamic combo (#507)

### LINKED ISSUE #496: panel_set_widget fails closed on frontend-only node types (MarkdownNote/Note/Reroute) — #458 guard lacks the allowlist its sibling guard has

After updating the panel to 0.11.30, `panel_set_widget` refuses to write to
**frontend-only node types**. Setting the `text` widget on a `MarkdownNote`
(a README note in a workflow) fails with:

> Cannot set widget on node 11 ("MarkdownNote"): the ComfyUI backend does not
> provide node type "MarkdownNote" (not installed, or its pack was removed) —
> refusing to write to a node the live backend no longer defines (#458).
> Check the exact class_type via panel_search_nodes.

This worked before the update.

**Root cause:** `assertTypeAgainstFreshBackend()` in
`web/js/lib/node-resolve.js:175-192` authorizes a write only if the type is a
key in a freshly-fetched `/object_info`:

```js
if (typeof type !== "string" || !Object.prototype.hasOwnProperty.call(freshDefs, type)) {
  throw new Error(... `the ComfyUI backend does not provide node type "${type}"` ...);
}
```

`MarkdownNote`, `Note`, `Reroute` and `PrimitiveNode` are registered purely in
the LiteGraph frontend and are **never** present in `/object_info` by design.
They therefore always fail closed, even on a perfectly healthy backend. The
guard cannot distinguish "frontend-only node" from "uninstalled pack".

**This looks like an oversight rather than intent**, because the sibling guard
`assertResolvedTargetRegistered()` in the same file already accounts for exactly
this class of node and says so in a comment at `node-resolve.js:236-237`:

```js
// (Native/defless types have no registered nodeData to compare and are trusted,
// so this never false-negatives Note/Reroute/etc.)
```

So one guard handles frontend-only types correctly; `assertTypeAgainstFreshBackend`
— added later to close the set_widget gap found in review of #375 — does not.

**Impact:** no agent can edit a Note / MarkdownNote on an updated panel. That is
a common operation (documenting a workflow before sharing it). It fails with a
message that misleadingly blames a missing/removed custom node pack.

### Steps to reproduce

1. Update panel to 0.11.30.
2. `panel_add_node` with `class_type: "MarkdownNote"` (this still succeeds).
3. `panel_set_widget` on that node, widget `text`, any string value.
4. Fails closed with the error above. Same for `Note`.

### Exact error (scrubbed)

```
Error: Cannot set widget on node 11 ("MarkdownNote"): the ComfyUI backend does
not provide node type "MarkdownNote" (not installed, or its pack was removed) —
refusing to write to a node the live backend no longer defines (#458). Check the
exact class_type via panel_search_nodes.
```

### Fix

applied locally: **no**  upstream-only: **yes**

Not patched locally — this machine's policy is to not modify third-party packs.
Precise change needed, in `web/js/lib/node-resolve.js`:

```diff
+// Registered purely frontend-side by LiteGraph; never present in /object_info
+// by design, so the fresh-backend oracle can never authorize them. Same class of
+// node that assertResolvedTargetRegistered() already trusts (see its comment).
+const FRONTEND_ONLY_TYPES = new Set(["Note", "MarkdownNote", "Reroute", "PrimitiveNode"]);
+
 export function assertTypeAgainstFreshBackend(freshDefs, type, nodeId = "(unknown)") {
   const label = typeof type === "string" ? ` ("${type}")` : "";
+  // Frontend-only types need no backend verification — exempt BEFORE the
+  // unreachable-backend check so they stay editable while ComfyUI is down.
+  if (typeof type === "string" && FRONTEND_ONLY_TYPES.has(type)) return;
   if (!freshDefs || typeof freshDefs !== "object") {
```

Placing the exemption before the `!freshDefs` branch also keeps notes editable
when the backend is unreachable, which is correct: a frontend-only node's
writability does not depend on ComfyUI being up.

Worth confirming the same allowlist is not also needed in `graph_add_node`'s
fresh-backend authorization path — `panel_add_node("MarkdownNote")` currently
still succeeds, so it presumably takes a different route, but the asymmetry
(can add, cannot edit) is itself surprising.

### Workaround used

Save the workflow, patch `widgets_values` in the saved JSON externally, then
`panel_load_workflow` the file back. Works, but heavy for a text edit.

### Environment

- OS: Windows 11 Enterprise 10.0.26200
- ComfyUI: 0.27.1 (local)
- GPU: RTX 5080, 16 GB VRAM
- comfyui-mcp: **0.48.30**
- panel: **0.11.30** (fresh `git pull` to a7b5bf7)
- ComfyUI-Manager: 3.x legacy

### LINKED ISSUE #507: panel_set_widget can never write a dynamic combo whose server-side option list is empty (StarNodes Ollama "model")

Some custom nodes declare a **client-populated combo** — the server-side
`INPUT_TYPES` entry is an EMPTY option list, and the node's own frontend JS fills
the dropdown at runtime (a "🔄 Refresh Models" button).

Concrete case: StarNodes `StarOllamaPromptHelper`
(`comfyui_starnodes/image_tools/star_ollama_prompt_helper.py`) declares

```python
"model": ((), {"tooltip": "Select a model. Click refresh to load available models from the server."}),
```

`/object_info` therefore reports `"model": [[], {...}]` — zero options.

In `web/js/lib/widget-write.js`, the combo branch does:

```js
const options = comboOptions(widget);
if (!options) { /* refuse: no readable option list */ }
if (options.includes(value)) return value;
throw new WidgetWriteError(`Value ... is not a valid option for combo widget ...`);
```

`comboOptions()` returns `[]` (truthy), so the `!options` guard doesn't fire and
strict membership rejects **every** value. Worse, `set-widget.js`'s recovery path
calls `refreshCombos()` first, which overwrites whatever the node's own refresh
button had populated client-side with the server's empty list — so the retry can
never succeed either, and the user's freshly-loaded model list gets wiped.

Net effect: such a widget is **permanently unwritable by the agent**. The only
way out is to tell the user to set it by hand in the node UI.

### Steps to reproduce

1. Install StarNodes (`comfyui_starnodes`) and open the bundled example workflow
   `Starnodes Krea2 Unbound HQ`.
2. Call `panel_set_widget(node_id=<StarOllamaPromptHelper>, widget="model", value="qwen3-vl:8b")`
   (any value; the value is a real model on the user's Ollama server).

### Exact error (scrubbed)

```
Error: panel_set_widget refused "model" on node 9 (StarOllamaPromptHelper) after
refreshing combo options: Value "qwen3-vl:8b" is not a valid option for combo
widget "model". Valid options (0):
```

### Fix

applied locally: **yes** · upstream-only: **no**

Treat an EMPTY option list as "nothing to validate against" rather than "nothing
is valid". With zero options there is no index-ambiguity to fail open on (the
#240 concern was a numeric value being reinterpreted as an index into a real
list), so a scalar can be accepted as-is. The existing `!options` guard (missing
`options.values` / throwing fn) is untouched.

```diff
--- a/web/js/lib/widget-write.js
+++ b/web/js/lib/widget-write.js
@@ -298,6 +298,14 @@
+    // A DYNAMIC combo declared with an EMPTY option list (e.g. StarNodes'
+    // `"model": ((), {...})`, filled in client-side by the node's own refresh
+    // button) has nothing to validate against — strict membership would reject
+    // EVERY value and leave the widget permanently unwritable. With zero options
+    // there is no index ambiguity to fail open on, so accept a scalar as-is.
+    if (options.length === 0 && (typeof value === "string" || typeof value === "number")) {
+      return value;
+    }
     // STRICT typed membership: no numeric<->string coercion. Numeric options
     // [0,1,2] accept numeric 1; string options ["0","1","2"] require "1", never
     // the number 1 (which would otherwise behave like an index).
     if (options.includes(value)) return value;
```

Possible follow-up (not patched here): `set-widget.js`'s `refreshCombos()`
recovery clobbers a client-populated option list with the server's empty one —
it should probably skip the overwrite when the fresh server list is empty but the
in-canvas widget has options.

Note: `browser_tests/unit/widget-write.test.mjs` does not run on this machine
(pre-existing, unrelated to the patch): node 18 resolves
`web/js/lib/widget-write.js` as CommonJS →
`SyntaxError: Named export 'WidgetWriteError' not found`. Same failure on the
pristine file (verified via `git stash`).

### Environment

OS: Linux 6.8.0-136-generic · ComfyUI 0.29.0 · GPU: RTX 3090 24GB · CUDA 12.8 ·
torch 2.10.0 · python 3.10 · comfyui-mcp 0.48.32 · panel 0.11.32

artokun:
Two guards that fail CLOSED on a legitimate case, plus the hardening the self-gate demanded.

Closes #496
Closes #507

## #496 — one shared frontend-only allowlist across the whole guard family

The `#458` guard family authorizes a node type against a freshly-fetched `/object_info`. Genuinely **frontend-only** types (`Note`, `MarkdownNote`, `Reroute`, `PrimitiveNode`, the rgthree control nodes) are registered purely by the LiteGraph frontend and are **never enumerated by `/object_info` by design**, so that oracle can never authorize them and the guard refuses them on a perfectly healthy backend.

**State of the issue as filed:** the `panel_set_widget` half was already fixed by #479 — `FRONTEND_ONLY_NODE_TYPES` exists and both set_widget guards consult it. What #479 did **not** do is teach `graph_add_node`'s guard, and each set_widget guard still spelled the exemption out inline, clause for clause. So `MarkdownNote` was **writable but not addable**: `assertAddNodeResolvableRefreshing` had no exemption at all. Verified against this machine's live ComfyUI — `/object_info` does not list `Note`/`MarkdownNote`, so `panel_add_node("MarkdownNote")` fails closed with *"Unknown node type … the ComfyUI backend does not provide it"*. A duplicated predicate is exactly how that drift happened.

Two shared predicates now carry the decision, and every guard whose oracle is `/object_info` routes through them:

- `isAuthorizedFrontendOnlyType(registry, type, node)` — live-registry membership + reserved-allowlist membership + no backend provenance on the registered class *or* the target instance's constructor. Same conjunction as before, one copy.
- `isRemovedBackendType` / `backendHistoryVerdict(type, wasTypeEverDefined)` — the non-forgeable observed-backend-history gate, previously duplicated verbatim in both set_widget guards.

| guard | oracle | consumes the shared allowlist |
|---|---|---|
| `assertAddNodeResolvableRefreshing` | fresh `/object_info` | **yes (new)** |
| `assertTypeAgainstFreshBackend` | fresh `/object_info` | yes |
| `assertMutatedNodeAuthorized` | fresh `/object_info` | yes |
| `assertResolvedTargetRegistered` | live LiteGraph registry | **no, deliberately** |

The fourth family member needs no allowlist: its oracle is the registry, where a frontend-only type *is* present — that is what "frontend-only" means. A comment now says so, so no list gets pasted there.

**The allowlist (unchanged, single copy in `node-resolve.js`):** `Note`, `MarkdownNote`, `Reroute`, `PrimitiveNode`, `Fast Bypasser (rgthree)`, `Fast Muter (rgthree)`, `Fast Groups Bypasser (rgthree)`, `Fast Groups Muter (rgthree)`, `Label (rgthree)`, `Reroute (rgthree)`, `Node Collector (rgthree)`.

**Nothing loosened.** The ever-seen gate runs FIRST in every guard, so a since-removed pack is refused before the exemption is consulted; an unavailable `/object_info` still fails closed for every type; and a defless husk not on the allowlist, a class squatting an allowlisted name with backend provenance, and an allowlisted name absent from the live registry are all still refused. `graph_add_node`'s exemption additionally **requires** the history oracle to be wired — without it nothing absent from fresh `/object_info` is exempted at all.

## #507 — a dynamic combo with an empty option list was permanently unwritable

Some custom nodes declare a client-populated combo — StarNodes' `"model": ((), {...})`, which `/object_info` reports as `[[], {...}]` — and fill the dropdown from their own frontend JS. `comboOptions()` returns `[]`, which is **truthy**, so the "no readable option list" guard never fired and `[].includes(value)` rejected every value.

Zero options means the option set is **not knowable**, not "no value is valid", and #240's reason for strict membership (a number reinterpreted as an *index* into a real list) cannot apply when there is no list to index into. Acceptance is nonetheless not automatic, because an empty *live* list can also be merely stale:

1. `applyWidgetWrite` refuses the empty case **retryably** (`emptyOptions`), driving `runSetWidget`'s authoritative `/object_info` combo refresh.
2. The last-resort attempt runs only after the refresh **and** the #387 server-asset probe have both failed, and only when `serverDeclaresEmptyComboOptions()` confirms the **server itself** declares that concrete input a zero-length combo. The live widget alone is not enough — the production refresh deliberately never clobbers a dynamic `options.values` *function*, so a source returning `[]` right now would otherwise look empty while the server publishes a real list.
3. A non-empty list — server or client-populated, whichever the live widget holds — still enforces exact membership. Object/array values fail closed and are non-retryable. An unreadable list is still refused.
4. On that path every **other** mutated combo (the promoted parent rail and display proxies, whose lists can differ from the inner's) must contain the value, and a *dynamic* sibling source is treated as unverifiable and refused outright.
5. The final attempt is not wrapped in a catch — a genuine failure from it propagates unchanged rather than being masked by the earlier combo rejection.

As a side effect the reported list-wiping goes away: the first attempt now succeeds for this shape, so the refresh that overwrote the user's client-loaded models never runs.

## Self-gate

Eight rounds of `codex exec -m gpt-5.6-terra`, scoped to the graph-corruption / fabricated-success class. Final verdict **PASS** on the #507 flow and on the shared-allowlist work. Findings and resolutions:

| # | severity | finding | resolution |
|---|---|---|---|
| 1 | SEVERE | the add exemption trusted an allowlisted name + marker-free class — both forgeable | exemption now **requires** the observed-backend-history oracle |
| 1 | MODERATE | awaiting the history seed could hang `graph_add_node` forever | bounded `awaitObjectInfoHistorySeed()`; giving up leaves the history unseeded ⇒ fail closed |
| 2 | SEVERE | a dynamic `[]`-returning source vs a real server list wrote an off-list value | gated on `serverDeclaresEmptyComboOptions()`; tests now use the **production** `refreshComboOptionsFromDefs` |
| 2 | SEVERE | a post-timeout re-seed installed a post-removal map as the baseline | retry removed; the loss is latched |
| 3 | SEVERE | a later observation (incl. reconnect) could still erase the only evidence of a removal | trust root extracted to `object-info-history.js` with an irreversible `loseBaseline()` latch |
| 3 | MODERATE | a promoted write mutates rail/proxy combos that were never validated | every mutated combo cross-checked on the empty-list path |
| 4 | SEVERE | `registerComfyNodeDefs` marked the history seeded from a post-gap payload | it now **records only**; the startup seed is the sole caller of `markSeeded` |
| 5 | MODERATE | a stateful dynamic rail source could return `[]` during the check and a real list after | a dynamic sibling source is unverifiable ⇒ refused |
| 6 | MINOR | the rail check keyed on the flag, over-reaching onto normally-validated writes | narrowed to writes the empty-list acceptance actually authorized |
| 7 | SEVERE | …but that narrowing re-read the inner list, which a stateful source could exploit | verdict now taken from **coercion time** (`emptyAcceptanceUsed`), never a re-read |

**Disclosed, not fixed.** The final round flagged that the baseline is established from an *asynchronous* `/object_info` response, so a pack removed between page load and the first successful fetch is never observed. There is no synchronous page-load oracle that would close it; exploiting it needs a pack to squat a ComfyUI-core-reserved name, strip its own frontend class's provenance markers, and be uninstalled inside that sub-second window — an adversarial construction rather than an accidental failure, in a product where panel, orchestrator and user share one machine. The residual is **unchanged from `main`** and is now documented in `object-info-history.js` so nobody "fixes" it by weakening the two fail-closed rules.

## Tests

`browser_tests/unit`, 1292 passing (was 1245). Beyond the per-issue cases:

- **Parity** — all three `/object_info`-oracle guards return the same verdict over a leaf-node matrix (frontend-only, live backend node, unknown, defless husk, squatted name, removed-this-session).
- **Single source of truth** — mutating the one exported `FRONTEND_ONLY_NODE_TYPES` Set at runtime moves all three guards together and removing it moves them back, so a future hard-coded second copy fails here. That is #496 as an executable invariant.
- **Honest diagnosis** — with no baseline the guards say *"the `/object_info` baseline never loaded — reload the ComfyUI tab"* instead of falsely blaming a removed pack, while a genuinely removed pack still gets the removed-pack message. The unseeded sentinel is asserted TRUTHY, so a consumer that only tests for truth still fails closed.

`tsc --noEmit` clean. Every changed file re-checked as an ES module (`node --check` on a `.mjs` copy) — a duplicate top-level declaration would silently break the panel and CJS `node --check` misses it. No version/`pyproject.toml` bump. No NUL sentinels (`git diff --numstat` shows real line counts throughout).

## Needs rig verification

Unit tests cannot drive LiteGraph or a live ComfyUI, so on the rig please confirm:

1. `panel_add_node` with `class_type: "MarkdownNote"` / `"Note"` / `"Reroute"` succeeds and `panel_set_widget` writes its `text` — the exact #496 repro. This is the one behaviour change to `graph_add_node`, and it depends on ComfyUI's real frontend classes being provenance-clean (no `.nodeData`/`.comfyClass`), which only the browser can show.
2. Adding a genuinely unknown type still fails closed, and normal backend nodes still add — `graph_add_node` now awaits the history seed first, so confirm there is no added latency in the steady state.
3. `panel_set_widget` on a StarNodes `StarOllamaPromptHelper` `model` widget with a real Ollama model, and that the node's own client-loaded list is no longer wiped.
4. A promoted subgraph widget still writes and syncs the parent rail (the `#366`/`#477` path the new cross-check sits next to).

Also worth noting: two of the recent comments on #496 report a **different** failure — `panel_set_widget` refused on an *outer UUID subgraph* node's promoted widget (`ltx-2.3-img2vid`, node 320) as an "unverifiable virtual-subgraph node". That is `assertMutatedNodeAuthorized`'s virtual-container branch, not the frontend-only allowlist, and it is not addressed here; it needs the rig to determine whether ComfyUI's `SubgraphNode` class carries backend provenance or fails `isVirtualSubgraphContainer`. Suggest re-filing it separately rather than holding #496 open for it.

### DIFF
--- a/browser_tests/unit/node-resolve.test.mjs
+++ b/browser_tests/unit/node-resolve.test.mjs
@@ -29,8 +29,10 @@ import {
   isAuthorizedFrontendOnlyType,
   isRemovedBackendType,
   HISTORY_UNSEEDED,
+  HISTORY_PENDING,
   backendHistoryVerdict,
 } from "../../web/js/lib/node-resolve.js";
+import { createObjectInfoHistory } from "../../web/js/lib/object-info-history.js";
 // The PRODUCTION graph_set_widget handler body — the executor and these tests
 // call it verbatim, so the tested ordering IS the shipped ordering (#458).
 import { runSetWidget } from "../../web/js/lib/set-widget.js";
@@ -1477,4 +1479,123 @@ test("#496: backendHistoryVerdict — the one classifier all three guards share"
   // refusal, but it is not a claim that the pack was removed.
   assert.equal(isRemovedBackendType("X", () => HISTORY_UNSEEDED), false);
   assert.equal(isRemovedBackendType("X", () => true), true);
+});
+
+// ---- #496 REGRESSION (coordinator review): a merely SLOW /object_info must not burn the
+//      session. A caller's bounded wait can expire while the fetch is still in flight;
+//      that is the ABSENCE of evidence, not evidence, so it yields a TEMPORARY refusal
+//      and a late-arriving seed restores normal operation with no tab reload. Latching
+//      there would make ordinary latency a permanent false refusal — the same bug class
+//      as #496 and #507 themselves. -----------------------------------------------------
+
+test("#496 REGRESSION: a seed that resolves AFTER the bound ⇒ the next MarkdownNote add SUCCEEDS", async () => {
+  // Drives the REAL history object through the real guard, so the state machine and the
+  // guard's reading of it are exercised together.
+  const history = createObjectInfoHistory();
+  const reg = registryWithNatives();
+  const fresh = objectInfo(); // healthy backend; never lists MarkdownNote (by design)
+  const opts = ADD_OPTS(fresh, (t) => history.wasTypeEverDefined(t));
+
+  // t = 0..bound — the startup getNodeDefs() is still in flight and the tool gave up
+  // waiting. Refused, but TEMPORARILY: the message must say "retry in a moment" and must
+  // NOT blame a removed pack or tell the user to reload.
+  await assert.rejects(
+    () => assertAddNodeResolvableRefreshing(() => reg, "MarkdownNote", opts),
+    (err) =>
+      /still loading|retry in a moment/i.test(err.message) &&
+      !/pack uninstalled|its backend was removed/i.test(err.message) &&
+      !/Reload the ComfyUI tab/i.test(err.message),
+  );
+  assert.equal(history.baselineLost, false, "a bounded wait must not latch the session");
+
+  // t = bound+ε — the slow response lands and seeds the baseline (this is what
+  // seedObjectInfoHistory does when its in-flight fetch finally returns).
+  history.recordTypes(fresh);
+  assert.equal(history.markSeeded(), true);
+
+  // The very SAME add now succeeds. No tab reload, nothing burned.
+  await assert.doesNotReject(
+    () => assertAddNodeResolvableRefreshing(() => reg, "MarkdownNote", opts),
+    "a late seed must restore normal operation",
+  );
+  // …and so does the write path, on the same recovered baseline.
+  const node = { id: 11, type: "MarkdownNote", widgets: [{ name: "text", type: "string", value: "" }], constructor: reg["MarkdownNote"] };
+  const { set } = await runSetWidget(node, "text", "# README", {
+    registry: reg,
+    getRegistry: () => reg,
+    getFreshObjectInfo: async () => fresh,
+    wasTypeEverDefined: (t) => history.wasTypeEverDefined(t),
+    ...HOOKS,
+  });
+  assert.equal(set.value, "# README");
+  // The recovered baseline is a REAL one, not a blanket allow: a type it DID observe is
+  // still refused once it disappears from /object_info.
+  const gone = objectInfo(); // KSampler is in `fresh` (observed) but we drop it here
+  delete gone["KSampler"];
+  assert.throws(
+    () =>
+      assertTypeAgainstFreshBackend(gone, "KSampler", 12, {
+        registry: reg,
+        wasTypeEverDefined: (t) => history.wasTypeEverDefined(t),
+      }),
+    /its backend was removed|pack uninstalled/i,
+  );
+});
+
+test("#496 REGRESSION: PENDING and LATCHED are reported DIFFERENTLY by all three guards", async () => {
+  // "hasn't loaded yet, retry" and "never loaded, reload the tab" are different problems
+  // with different user actions. Both refuse; only the diagnosis differs.
+  const reg = registryWithNatives();
+  const fresh = objectInfo();
+  const node = { id: 11, type: "MarkdownNote", widgets: [{ name: "text", type: "string", value: "" }], constructor: reg["MarkdownNote"] };
+  const PENDING = () => HISTORY_PENDING;
+  const LATCHED = () => HISTORY_UNSEEDED;
+  const temporary = (m) => /still loading|retry in a moment/i.test(m) && !/Reload the ComfyUI tab/i.test(m);
+  const permanent = (m) => /Reload the ComfyUI tab/i.test(m) && !/retry in a moment/i.test(m);
+
+  await assert.rejects(
+    () => assertAddNodeResolvableRefreshing(() => reg, "MarkdownNote", ADD_OPTS(fresh, PENDING)),
+    (err) => temporary(err.message),
+  );
+  await assert.rejects(
+    () => assertAddNodeResolvableRefreshing(() => reg, "MarkdownNote", ADD_OPTS(fresh, LATCHED)),
+    (err) => permanent(err.message),
+  );
+  assert.throws(
+    () => assertTypeAgainstFreshBackend(fresh, "MarkdownNote", 11, { registry: reg, node, wasTypeEverDefined: PENDING }),
+    (err) => temporary(err.message),
+  );
+  assert.throws(
+    () => assertTypeAgainstFreshBackend(fresh, "MarkdownNote", 11, { registry: reg, node, wasTypeEverDefined: LATCHED }),
+    (err) => permanent(err.message),
+  );
+  assert.throws(
+    () => assertMutatedNodeAuthorized(fresh, reg, node, "target", PENDING),
+    (err) => temporary(err.message),
+  );
+  assert.throws(
+    () => assertMutatedNodeAuthorized(fresh, reg, node, "target", LATCHED),
+    (err) => permanent(err.message),
+  );
+});
+
+test("#496 REGRESSION: the PENDING sentinel still FAILS CLOSED — it never authorizes anything", async () => {
+  // The recoverable state must not become a hole: while pending, NOTHING is exempted.
+  const reg = registryWithNatives();
+  const fresh = objectInfo();
+  const node = { id: 11, type: "MarkdownNote", widgets: [{ name: "text", type: "string", value: "" }], constructor: reg["MarkdownNote"] };
+  assert.equal(backendHistoryVerdict("MarkdownNote", () => HISTORY_PENDING), "pending");
+  assert.notEqual(backendHistoryVerdict("MarkdownNote", () => HISTORY_PENDING), "never-seen");
+  await assert.rejects(
+    () =>
+      runSetWidget(node, "text", "# README", {
+        registry: reg,
+        getRegistry: () => reg,
+        getFreshObjectInfo: async () => fresh,
+        wasTypeEverDefined: () => HISTORY_PENDING,
+        ...HOOKS,
+      }),
+    /still loading|retry in a moment/i,
+  );
+  assert.equal(node.widgets[0].value, "", "no write while the baseline is pending");
 });
--- a/browser_tests/unit/node-resolve.test.mjs
+++ b/browser_tests/unit/node-resolve.test.mjs
@@ -1112,9 +1112,151 @@ test("#496 SINGLE SOURCE OF TRUTH: all three guards read the SAME allowlist Set 
   assert.deepEqual(await guardVerdicts(reg, fresh, SYNTHETIC, node), [false, false, false]);
 });
 
+test("#496 add_node (codex SEVERE): the frontend-only exemption REQUIRES the ever-seen oracle — without it, nothing absent from object_info is exempted", async () => {
+  // Client-side name + provenance markers are FORGEABLE: a removed pack whose frontend
+  // class had .nodeData/.comfyClass stripped and which squats a reserved allowlisted
+  // name is indistinguishable from a genuine native by those signals alone. Only the
+  // non-forgeable observed-backend-history gate can rule it out, so with NO history
+  // oracle wired there is no exemption at all — pre-#496 fail-closed behaviour.
+  const reg = registryWithNatives();
+  const fresh = objectInfo();
+  for (const t of ["Note", "MarkdownNote", "Reroute", "PrimitiveNode"]) {
+    await assert.rejects(
+      () =>
+        assertAddNodeResolvableRefreshing(() => reg, t, {
+          getFreshObjectInfo: async () => fresh,
+          refresh: async () => {},
+          // wasTypeEverDefined deliberately OMITTED
+        }),
+      /Unknown node type|does not provide/i,
+      `#496: "${t}" must NOT be exempted without the observed-backend-history oracle`,
+    );
+  }
+  // Wiring the oracle (and only then) enables the exemption.
+  await assert.doesNotReject(() =>
+    assertAddNodeResolvableRefreshing(() => reg, "MarkdownNote", ADD_OPTS(fresh)),
+  );
+});
+
 test("#496: isRemovedBackendType — the shared ever-seen gate is inert without an injected oracle and never guesses", () => {
   assert.equal(isRemovedBackendType("KSampler", () => true), true);
   assert.equal(isRemovedBackendType("KSampler", () => false), false);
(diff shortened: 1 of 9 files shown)


artokun:
Coordinator HIGH fixed — and you were right that the irony matters. The latch turned a
timeout into a verdict, which made a *third* false refusal on top of the two this PR fixes.

**The split.** A timeout is the absence of evidence, so it no longer decides anything.
There are now three states:

| state | when | refusal | recovers |
|---|---|---|---|
| `SEEDED` | a page-load-anchored `/object_info` was observed | — | — |
| `PENDING` | no baseline **yet** — fetch in flight, or a caller bounded its wait | *"the node-type baseline is still loading — retry in a moment, no reload needed"* | **yes**, by itself |
| `LATCHED` | the seed's whole attempt sequence **finished with no data** | *"the baseline never loaded — reload the ComfyUI tab"* | tab reload |

`awaitObjectInfoHistorySeed` no longer latches, and no caller can: `loseBaseline()` now has
exactly **one** call site — inside `seedObjectInfoHistory`, after all five attempts finish
empty. That is the only place holding actual evidence. Verified by grep, and codex
confirmed it.

Your 8.1s scenario now goes: the tool gives up at 8s and refuses *that one call* as
temporary; the seed is untouched by the bound and keeps running; the response lands and
calls `markSeeded()`; the next call authorizes normally. Nothing burned, no reload.

**I kept the bound**, because it still stops a never-settling request hanging the tool —
and it is only safe to have *because* pending is recoverable, which is now stated in the
code where the constant lives.

**Regression test** — your exact scenario, driven through the real history object and the
real guard, not a mock: seed resolves *after* the bound → the next `MarkdownNote` add
**succeeds**, the write path succeeds on the same recovered baseline, and the recovered
baseline is proven **real** (a type it did observe still gets refused once it disappears
from `/object_info`, so recovery is not a blanket allow).

**One thing codex caught that I'd have missed.** The first version of that test called
`recordTypes()`/`markSeeded()` by hand, so it exercised the state machine but not the
production seam — re-adding `loseBaseline()` inside the bounded wait would still have
passed. That is exactly how this bug got in: the logic had no seam, so review caught it
instead of a test. The wait is now extracted as
`awaitHistoryBaseline(history, seedPromise, timeoutMs)` in `object-info-history.js`, the
panel delegates to it, and the tests drive the real helper with a controlled
never-settling promise. Re-adding a latch there now **fails a test**.

Also added: `PENDING` and `LATCHED` are asserted to be reported differently by all three
guards, and `PENDING` is asserted never to authorize anything (both sentinels stay TRUTHY,
so a consumer that only tests for truth still fails closed).

**Re-gate:** codex now returns **PASS with zero findings** — no SEVERE, no MODERATE, no
MINOR — and explicitly re-confirmed each surface you asked me not to disturb: no
authorization widening beyond the shared predicate, non-empty combo lists still strict with
the fallback gated on `latest.emptyOptions`, unresolved/stuck writes still failing before
mutation, the parent-rail check still keyed off coercion-time `emptyAcceptanceUsed` with no
inner-list re-read, and promoted rails/proxies still prechecked, synchronized and verified
with rollback intact. It also confirmed the extraction is behaviour-preserving (one bounded
race, `finally` timer cleanup, rejection swallowed, no replacement seed; the new boolean
return is ignored by both production callers).

1301 unit tests passing (was 1245 on `main`), `tsc --noEmit` clean, every changed file
re-checked as an ES module.

Rig verification list in the PR body is unchanged, plus one addition: with ComfyUI briefly
slow or reconnecting, confirm a refused add now says *"still loading — retry in a moment"*
and that simply retrying works without a tab reload.

Thanks for the catch on #510 and #512.

Issue #516, opened 2026-08-02
Subgraph conversion leaves stale red LoadImage outline until no-op widget write


artokun:
### What happened / root cause
Converting tested Phase 3 branches into native subgraphs with panel_create_subgraph / panel_subgraph_group left the shared upstream LoadImage node red (node.has_errors=true) even though the referenced input existed and both pre- and post-collapse executions succeeded. panel_get_errors reported errored_count:1 with reasons:[] and no validation or execution error. The existing stale-red cleanup in graph_set_widget clears this state, but subgraph conversion does not invoke equivalent adjudication.

### Steps to reproduce
1. Create a LoadImage that feeds nodes later wrapped by panel_create_subgraph or panel_subgraph_group.
2. Confirm the branch renders successfully.
3. Convert downstream branch nodes to a subgraph and mute the wrapper.
4. Call panel_get_errors.
5. Observe LoadImage red_outline:true, reasons:[], errored_count:1.
6. Call panel_set_widget on LoadImage.image with the exact same existing valid filename.
7. Call panel_get_errors again; it returns errored_count:0 and the red outline clears.

### Exact error (scrubbed)
Flagged by LiteGraph but no source explained it — it may be stale; re-run to refresh, or check the node's widget values.

### Fix
Applied locally: no code patch; verified workaround: a no-op panel_set_widget write to the same valid LoadImage.image value triggers the existing #418 nodeRedFlagIsStale cleanup and clears the false-positive state.
Upstream-only: yes.
Precise change needed: after panel_create_subgraph / panel_subgraph_group completes, reuse the same conservative nodeRedFlagIsStale adjudication used by graph_set_widget for affected surviving/root nodes, or otherwise prevent subgraph conversion from leaving node.has_errors sticky when collectMissingAssets and validation maps do not blame the node.

### Environment
Windows 11; ComfyUI 0.29.2; NVIDIA RTX 4080 SUPER 16 GB; comfyui-mcp 0.48.22; panel 0.11.32.

artokun:
[...]

Issue #418, opened 2026-07-31
Clear stale red node flag after fixing a missing asset with panel_set_widget


artokun:
### What happened / root cause
After loading a workflow with a missing LoRA, panel_set_widget successfully changed the combo to a valid installed LoRA. collectMissingAssets then correctly reported no missing_model reason, but LiteGraph node.has_errors stayed true. panel_get_errors returned the node as red with reasons: [] and a stale-warning note.

Root cause: graph_set_widget updates the widget and live asset state but does not clear LiteGraph's stale has_errors flag when the updated node has no remaining missing asset or validation error.

### Steps to reproduce
1. Load a workflow whose LoraLoaderModelOnly points to a missing LoRA.
2. Call panel_set_widget with an installed LoRA from the refreshed combo options.
3. Call panel_get_errors.
4. The node is still listed with red_outline:true and reasons:[].

### Exact error
No thrown error. Misleading result:
Flagged by LiteGraph but no source explained it - it may be stale.

### Fix
Applied locally: yes. Upstream-only: no.

Focused patch:
```diff
-    return runSetWidget(node, widget, value, {
+    const result = runSetWidget(node, widget, value, {
       ...
     });
+    if (node.has_errors) {
+      const assets = collectMissingAssets();
+      const nodeId = String(node.id);
+      const stillMissing = [...assets.models, ...assets.media].some((item) => String(item.node_id) === nodeId);
+      const stillInvalid = !!app?.lastNodeErrors?.[nodeId]?.errors?.length;
+      if (!stillMissing && !stillInvalid) node.has_errors = false;
+    }
+    return result;
```

This only clears the visual stale flag after a successful widget write when the live asset collector and current validation map find no remaining problem on that node.

Verification: npm run test:unit passed, 567 passed, 0 failed, 3 skipped (570 total).

### Environment
Windows 11; ComfyUI 0.29.1 LOCAL; RTX 5090 D 32GB; comfyui-mcp 0.48.21; panel 0.11.20.

artokun:
[...]

PR #535, opened 2026-08-02
fix(panel): re-adjudicate stale red outlines after subgraph conversion (#516)

artokun:
## What

Fixes the stale-state half of #516: `graph.convertToSubgraph` reroutes the surviving nodes' wires onto the new wrapper node but never revalidates them, so LiteGraph's sticky `has_errors` red outline (observed on an upstream LoadImage whose link was rerouted into the subgraph) lingered until a no-op `panel_set_widget` happened to trip the #418 cleanup. `panel_get_errors` meanwhile reported `errored_count: 1` with `reasons: []`.

## Root cause

The red outline is LiteGraph's boolean `node.has_errors`, which is set-once and sticky — nothing in the conversion path re-adjudicates it. The existing invalidation lives in `graph_set_widget` (#418): after a widget write it clears the flag via the conservative `nodeRedFlagIsStale` check (clear only when neither the live missing-asset collector nor any validation-error map still blames the node). Conversion simply never invoked an equivalent adjudication.

## Fix

- Extracted the #418 adjudication from `graph_set_widget` into a shared module-level `clearStaleRedFlag(node, { app, graph, rootGraph })` helper in `web/js/comfyui-mcp-panel.js` — identical semantics, including the fail-keep guarantees (distrusts live combos, OR across `app.lastNodeErrors` + the execution-error store, never clears wrappers).
- Both conversion handlers (`graph_create_subgraph`, `graph_subgraph_group`) now run that same adjudication on the wrapper's **direct link neighbours** — exactly the surviving nodes conversion just rewired (upstream sources + downstream targets) — via a new pure `collectLinkedNeighborNodeIds(node, links)` helper in `web/js/lib/asset-staleness.js` (handles live LLink objects, serialized array links, and Map-based link containers).
- No new polling or event hooks: the cleanup rides the conversion call the panel already makes.

## Tests

- `npm run test:unit`: 1490 pass, 0 fail — includes 5 new `collectLinkedNeighborNodeIds` cases in `browser_tests/unit/asset-staleness.test.mjs` (wrapper neighbour set incl. the upstream LoadImage shape, serialized/Map link containers, dangling ids, defensive shapes).
- `npm run typecheck`: clean.

Refs #516

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+- clear the stale red outline left on rewired surviving nodes after subgraph conversion (#516)
+
 ## [0.11.34] - 2026-08-02
 
 ### Added
--- a/browser_tests/unit/asset-staleness.test.mjs
+++ b/browser_tests/unit/asset-staleness.test.mjs
@@ -23,6 +23,7 @@ import {
   collectMissingNodeTypeReasons,
   graphErrorsResultIsClean,
   nodeRedFlagIsStale,
+  collectLinkedNeighborNodeIds,
   resolveMissingModelDirectory,
 } from "../../web/js/lib/asset-staleness.js";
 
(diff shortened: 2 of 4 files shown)


artokun:
Superseded by #536, which carries the same #516 fix on the current vendor-vocabulary main and adds dedicated conversion-handler coverage. Closing to avoid competing implementations.

PR #536, opened 2026-08-02
fix(panel): clear stale root outlines after subgraph conversion (#516)

artokun:
## Summary\n- re-adjudicate only direct root-graph neighbours rewired by native subgraph conversion\n- share the conservative #418 missing-asset and validation-map gate; wrappers and converted inner nodes remain untouched\n- cover both graph_create_subgraph and graph_subgraph_group paths\n\n## Validation\n- npm.cmd run test:unit (1489 passed)\n- npm.cmd run typecheck\n\nRequires independent accidental-correctness/data-integrity review before merge.

PR #542, opened 2026-08-02
fix(tests): make the clearStaleRedFlag extraction line-ending tolerant

artokun:
The #536 test suite extracts clearStaleRedFlag from the panel source with an LF-only regex; the file is CRLF on Windows (core.autocrlf), failing every Windows run. One regex, now \r?-tolerant.


artokun:
Redundant — 90683b9 on main already made the extraction CRLF-safe (beat me by minutes).

PR #543, opened 2026-08-02
fix(panel): retry identity via retry_of + session-epoch replay gating (#694)

artokun:
Panel half of artokun/comfyui-mcp#694. The orchestrator half (sibling PR in comfyui-mcp) retries a timed-out command under a FRESH rid plus `retry_of` naming the original — it never reuses rids (#683/#687) — and stamps a session `epoch` on the `models` handshake. This PR teaches the panel both halves of that contract.

## Retry identity (`retry_of`)

- `commandFingerprint` (web/js/lib/command-dedupe.js) now deletes `retry_of` alongside `rid` before canonicalizing — both are correlation metadata, not command identity. A retry therefore fingerprints **identically** to its original.
- The frame handler (web/js/comfyui-mcp-panel.js) keeps the existing `ledger.get(msg.rid, fingerprint)` first; on a miss and `typeof msg.retry_of === "string"` it tries `ledger.get(msg.retry_of, fingerprint)`.
  - **Hit** (settled reply or in-flight promise — the existing await handles both): sends `{ ...dupReply, rid: msg.rid }`. The rid rewrite matters — the orchestrator's pending map is waiting on the retry's fresh rid. Executor never re-runs: no double-applied mutation.
  - **Miss** (unknown/evicted token): falls through to `begin` + execute fresh. The ledger fails open, never closed.
- A genuinely fresh command with identical args (no `retry_of`) shares the fingerprint but misses on its new rid and executes, as it must. Same-rid verbatim replay (#521) is untouched.

## Session epoch (replay gating)

- The `models` branch stamps `thisSock.__cmcpBridgeEpoch = msg.epoch` **before** `markConnected()` fires the lost-reply replay.
- `lostReplies.record` journals the current socket's epoch; `isReplayable` gains `entry.epoch !== targetEpoch → false`, applied identically in `summaries` so `hello.lost_replies` filters the same way.
- Backward compat: epoch absent on both sides compares equal — a legacy orchestrator gets exactly today's URL+age behaviour. A restart mints a new epoch, so the old session's journal is dropped instead of replayed into its successor at the same URL (closes the residual the code comments previously flagged as needing a server-issued epoch).

## Tests

`npm run test:unit` — 1503 pass (11 new, 2 wiring regexes updated); `npm run typecheck` clean.

New coverage:
- fingerprint excludes `retry_of`; retry of a settled command → original reply, rid rewritten, executor not re-run; retry of an in-flight command → awaited, same result; unknown `retry_of` → executes fresh; identical-args fresh command after a settled one → executes again; same-rid verbatim replay (#521) unchanged
- `isReplayable`: epoch match replays, mismatch drops, both-absent replays (legacy), one-sided presence fails closed; journal records epoch and summaries filter by it
- source-wiring: epoch stamped before the replay fires, journal stamps this socket's epoch, replay compares the target socket's epoch, handler falls back to `retry_of` and rewrites the rid

### DIFF
--- a/browser_tests/unit/command-dedupe.test.mjs
+++ b/browser_tests/unit/command-dedupe.test.mjs
@@ -11,8 +11,19 @@
 // The dedupe identity is rid + payload fingerprint: the bridge's re-dispatch of
 // the SAME logical command dedupes, but a rid reused for DIFFERENT work must
 // execute fresh (never answered from the ledger) and is logged once.
+//
+// #694: the orchestrator never REUSES a rid (#683/#687) — a timed-out command is
+// retried under a FRESH rid plus `retry_of` naming the original. `retry_of` is
+// excluded from the fingerprint, so the retry dedupes against the ORIGINAL's
+// ledger entry (settled or still in flight) and is answered with the rid
+// rewritten to the retry's — never a second execution. An unknown retry_of
+// misses and executes fresh (fail-open), and a genuinely fresh command with
+// identical args (no retry_of) executes again as it should.
 import test from "node:test";
 import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
 
 import { commandFingerprint, createCommandDedupeLedger } from "../../web/js/lib/command-dedupe.js";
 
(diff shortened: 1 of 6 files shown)


artokun:
Maintainer NO-SHIP P1: retry_of can mutate the wrong workflow after a tab switch. MCP retry re-reads the live workflow target, so r2 carries B's bridge-owned workflow_uuid while retained r1 is A; panel fingerprint differs, retry_of lookup misses, then begin/execute applies on B under B's valid fence. Required: if retry_of names a retained ledger record but fingerprint differs, reject stale/mismatched retry; only unknown/evicted tokens may fail open. Add cross-repo regression r1(A) timeout → active B → r2(retry_of:r1) neither replays nor executes. This must land lockstep with its MCP companion and receive a direct data-integrity gate.

Issue #545, opened 2026-08-02
Graph-binding guard false-positives permanently: "live graph is out of sync with the active workflow" blocks every panel_* graph tool, unrecoverable by any documented remedy


artokun:
### What happened / root cause

Every `panel_*` graph tool in the session was hard-rejected with:

```
The live graph is out of sync with the active workflow: the workflow reports 27 node(s),
but the canvas is bound to a different graph. A load, tab switch, or reconnect left this
command pointed at the wrong canvas, so it was NOT applied. Re-open the active workflow tab
(panel_open_workflow) or reload the panel to rebind the graph, then retry.
```

The active workflow was correct and in view the whole time. The guard appears to compare a **cached tab-metadata node count** against the canvas serialization and treat any mismatch as a hard block. The workflow on the canvas had **unsaved user edits** (`modified: true` in `panel_list_workflows`), and its on-disk copy strips to 30 API nodes while the tab metadata reported 27 UI nodes. Unsaved canvas edits legitimately change the node count, so the guard punishes ordinary user editing and then never re-syncs.

The error's own suggested recovery steps do not work — none of these cleared it:

- `panel_open_workflow` on the same workflow
- `panel_open_workflow` to a different workflow, then back
- `panel_set_workflow_target({mode:'current'})` (three times, incl. after each reload)
- `panel_reload({scope:'frontend'})`
- `panel_reload({scope:'orchestrator'})`
- a full ComfyUI restart + panel reconnect (the socket reconnected cleanly and the guard immediately returned)

### Steps to reproduce

1. Open a workflow tab and make edits on the canvas without saving (so `modified: true` and the live node count diverges from the tab's cached count).
2. Trigger a rebind event — switch tabs / reload / reconnect.
3. Call any graph tool: `panel_graph_outline`, `panel_query_graph`, `panel_find_nodes`, or `panel_strip_workflow` with no args.
4. All fail with the out-of-sync error, and stay failing through every documented recovery path.

### Exact error (scrubbed)

```
Error: The live graph is out of sync with the active workflow: the workflow reports 27 node(s),
but the canvas is bound to a different graph. A load, tab switch, or reconnect left this command
pointed at the wrong canvas, so it was NOT applied. Re-open the active workflow tab
(panel_open_workflow) or reload the panel to rebind the graph, then retry.
```

`panel_strip_workflow` (no args) surfaces the same guard wrapped in a misleading hint:

```
[...]

Issue #411, opened 2026-07-31
panel_run cannot target output nodes inside nested subgraphs


artokun:
Environment: Windows 11, ComfyUI 0.29.0, comfyui-mcp 0.48.21.

While viewing a nested subgraph, I added root-output node ShowText|pysssss (node 359) inside that subgraph and called panel_run({to_node_id:359}).

panel_run refused:
"node 359 is not on the root graph — run-to-node targets a root-level output node (output nodes inside subgraphs aren't supported yet)"

This blocks the documented run-to-node debugging workflow for outputs/taps inside nested subgraphs and forces a whole-root render or awkward temporary boundary exposure.

Expected: panel_run resolves the current viewing scope/subgraph path and targets the nested output node, or offers a supported scoped target parameter.

Issue #370, opened 2026-07-31
Connection drop can interrupt queued panel render without delivering completion result


artokun:
### What happened / root cause
A `panel_run` to the final PreviewImage returned `{ queued: true }`. Before completion delivery, the session emitted a connection-drop/resume notice. After reconnecting to the same workflow, no render output/result was delivered and no completion state was available in the sidebar. The user cannot tell whether the run completed, was aborted, or needs requeueing.

### Steps to reproduce
1. Queue a workflow from the sidebar with `panel_run`.
2. Cause a panel/session connection drop while it is pending.
3. Resume the session on the same workflow.
4. Observe that no automatic completion image/result is delivered.

### Exact error (scrubbed)
No thrown error. Session notice: connection dropped mid-task; resumed with full context.

### Fix
applied locally: no  upstream-only: yes
Persist and reconcile prompt IDs across session reconnect, then deliver the terminal output/aborted state to the resumed session. If terminal state cannot be recovered, surface a clear “render status unknown; safe to requeue” notice.

### Environment
Windows 11; ComfyUI 0.21.1; RTX 4080 SUPER 16 GB; comfyui-mcp 0.48.21; panel 0.11.5.

artokun:
Recurrence on current releases: comfyui-mcp 0.50.36 and panel 0.11.44. A `panel_run` completion was lost after a sidebar/orchestrator reconnect and workflow-tab switches, although a later exact prompt-id queue-status query showed `done=true`, `status_str=success`, and all text outputs. Environment: Windows 11, ComfyUI 0.31.0 local, RTX 5080, Codex compact backend.

artokun:
Fixed, and unlike most of the current backlog this one is **already in a release** — 0.11.44. You reported it on panel **0.11.5**, 39 versions back.

Both halves you asked for exist, and the code names this issue:

**Reconcile across the drop** — `293fa7f` (#425, *"reconcile runs across a connection drop (#358, #370)"*), 2026-07-31:

> The live run-completion tracker … so the module-level tool handlers (e.g. `graph_run`) can register a freshly queued `prompt_id` for **#370 reconnect reconciliation**

**And the case that reconcile alone misses** — `59ff096` (#356 Bug 2), 2026-08-01. The reconnect reconcile is edge-triggered on ComfyUI's `reconnected` event, and there is a worse variant of your scenario where no edge fires at all:

[...]

Issue #358, opened 2026-07-31
panel_run returns {queued:true} when ComfyUI rejected the prompt with invalid prompt / missing_node_type


artokun:
Follow-up to #356 — a third, distinct defect on the same surface. `panel_run`'s own tool description promises:

> if ComfyUI REFUSES the prompt (validation failure on either channel — per-node `node_errors` OR a top-level error like a missing node type) it returns a FAILURE with that rejection detail, **never a false 'queued'**.

It returned a false 'queued'.

## What happened

I loaded the bundled `wan-multitalk` pack, adapted it, and called `panel_run`:

```json
{ "queued": true, "batch_count": 1 }
```

Nothing ever rendered. ComfyUI had rejected the prompt **synchronously**, at the exact timestamp of the call:

```
2026-07-30T21:01:26.653459 - [INFO] got prompt
2026-07-30T21:01:26.654227 - [WARNING] invalid prompt: {
  'type': 'missing_node_type',
  'message': "Node 'ID #159' has no class_type. The workflow may be corrupted or a custom node is missing.",
  'details': "Node ID '#159'",
  'extra_info': {'node_id': '159', 'class_type': None, 'node_title': None}
}
```

Note the rejection is ~1 ms after `got prompt` — this is exactly the top-level `missing_node_type` case the description says is handled. The user was left believing a multi-minute 14B video render was in progress; combined with the dropped-notification bug from #356, there was no signal at all that anything had gone wrong. I only found it by reading `get_logs`.

## Root cause of the underlying graph problem (useful for the packs repo too)

The `wan-multitalk` pack's `workflow.json` references two node types that don't exist on this install:

- **`AudioCrop`** — not present. This machine has `VRGDG_AudioCrop` / `VRGDG_AudioCropTime` (comfyui-vrgamedevgirl) instead.
- **`AudioSeparation`** — not present under any name (`get_node_info` returns no match).

Neither is in the pack's manifest as a dependency, and `panel_load_workflow(pack:'wan-multitalk')` loaded it without any warning. So the pack ships a graph that cannot run on an install that satisfies its documented model requirements.

**A detectable signal was already in the tooling and went unsurfaced:** both `panel_graph_outline` and `panel_get_errors` rendered these nodes with placeholder widget names —

```
159  AudioCrop  UNKNOWN=0:00 UNKNOWN=2:00
170  AudioSeparation  UNKNOWN=linear UNKNOWN=10 UNKNOWN=0.1
```

[...]

PR #559, opened 2026-08-02
fix(#556): panel_run refuses scopeless full-graph execution (strict to_node_id)

### LINKED ISSUE #556: panel_run to_node_id can silently execute the full graph
panel_run with to_node_id=14 returned queued:true and ran_to_node:14, but the unrelated SaveImage branch also executed and produced Synthetic_Adult_Male_ArtNude_00003_.png. The panel passed the third app.queuePrompt argument as a bare array; some current frontend builds appear to ignore that legacy shape, silently converting the requested partial execution into a full graph run.

### Steps to reproduce
1. Open a graph with SaveImage output plus a separate PreviewAny output.
2. Call panel_run with to_node_id set to PreviewAny.
3. Observe panel_run reports ran_to_node for PreviewAny.
4. Observe SaveImage also executes and writes an image.

### Exact error
No exception was returned; this was a silent wrong-output/full-execution failure.

### Fix
applied locally: yes
upstream-only: no
Patched the built panel JS to pass QueuePromptOptions explicitly:
- await app.queuePrompt(0, batch, partialTargets);
+ await app.queuePrompt(0, batch, partialTargets ? { queueNodeIds: partialTargets } : undefined);

The patch requires a panel reload before verification; it has not yet been claimed verified because the user's active workflow render is being completed first.

### Environment
Windows 11
ComfyUI 0.29.2
NVIDIA RTX PRO 4000 Blackwell 24 GB
comfyui-mcp 0.48.12
comfyui-mcp-panel 0.11.3

artokun:
Fixes #556.

## Root cause

`panel_run` with `to_node_id` scopes execution via ComfyUI partial execution: the resolved target must reach the `POST /prompt` body as `partial_execution_targets`. The panel passed it as a **bare array** in the 3rd `app.queuePrompt` argument — but that argument's shape differs across frontend builds:

- positional `NodeExecutionId[]` (e.g. frontend ≤1.46, the 1.48.x line)
- `QueuePromptOptions { queueNodeIds }` (1.47.x, current main — these carry an `Array.isArray` compat shim)

A build that accepts only ONE shape **silently ignores the other**: the scope never reaches the request, the FULL graph queues, and `panel_run` still returns `queued:true, ran_to_node:X` — in #556 an unrelated SaveImage branch rendered and wrote a file nobody asked for. (The issue's local patch, passing only `{queueNodeIds}`, fixes the shim-less builds but would break the positional-only ones — neither single shape is universally safe.)

Note: the resolution half (stale/unknown/non-output/`unreachable` `to_node_id`) was already strict on main — `resolveRunToNodeTarget` refuses with `queued:false` before dispatch (#411/#438/#439). The silent fall-through lived at **dispatch**.

## Fix

- **`web/js/lib/run-scope-guard.js` (new)** — `createRunFetchInterceptor`, the `api.fetchApi` wrap `graph_run` already installs for #358/#370 response capture, now also inspects every outgoing `POST /prompt` body **before** it leaves: when a scope was requested and the body doesn't carry exactly that scope, the call is refused **without touching the real fetchApi** — no unscoped prompt is ever dispatched. `graph_run` then retries the other argument shape once (a blocked attempt queued nothing, so a retry can't double-queue) and, if no shape delivers, returns `queued:false` with a truthful error naming the node and stating nothing was queued. A scoped run with no `fetchApi` to verify against is refused up front. Unscoped (full) runs are byte-identical to before; the #358 first-rejection and #370 prompt_id capture move into the interceptor unchanged.
- **`web/js/comfyui-mcp-panel.js`** — `graph_run` drives the shape-retry loop (`queuePromptScopeArgs`) and returns the truthful refusal.

## Tests

19 new unit tests in `browser_tests/unit/run-scope-guard.test.mjs`:
- dropped-scope `POST /prompt` is refused and the real fetch is **never called** (the #556 regression);
- body verification: missing / empty / wrong / extra targets and unparseable bodies all refuse (never fail open);
- both frontend build shapes end-to-end: options-only build runs scoped on the retry, positional-only on the first attempt, neither-shape build ⇒ truthful refusal with zero dispatches;
- stale `to_node_id` (graph changed after id capture) resolves `not_found` ⇒ refused before any dispatch;
- normal full run and server-rejection (#358) paths preserved.

`npm run test:unit`: **1583/1583 pass**. `tsc --noEmit` clean. Playwright e2e not run (live rig occupied).

## Overlap note

The binding/desync-guard cluster (#545/#557) is being worked concurrently on `fix/graph-binding-cluster`; this PR touches only the run-scope path and does not modify the binding guard. One shared surface: `graph_run` still calls `assertGraphBoundToActiveWorkflow` before scope resolution — unchanged. Residual limitation worth knowing: if `app.queuePrompt` defers processing (queue already busy), the deferred `POST /prompt` can happen after the interceptor is restored; the scope argument is still delivered correctly by argument normalization, but that one asynchronous path can't be body-verified. Closing it would require patching `queuePrompt` itself — out of scope here.

### DIFF
--- a/browser_tests/unit/run-scope-guard.test.mjs
+++ b/browser_tests/unit/run-scope-guard.test.mjs
@@ -28,6 +28,7 @@ import {
   scopeUnverifiedError,
   createRunFetchInterceptor,
   createScopedRunGuard,
+  cancelPendingScopedQueueItem,
 } from "../../web/js/lib/run-scope-guard.js";
 import { resolveRunToNodeTarget } from "../../web/js/lib/subgraph-scope.js";
 
@@ -125,12 +126,15 @@ test("#556 scopeDroppedError: names the node and states NOTHING was queued", () 
 });
 
 test("#556 scopeUnverifiedError: truthful — nothing CONFIRMED queued, never claims a scoped run happened", () => {
-  const msg = scopeUnverifiedError({ toNodeId: 14, timeoutMs: 5000 });
-  assert.match(msg, /node 14/);
-  assert.match(msg, /could not be verified/);
-  assert.match(msg, /CONFIRMED queued/i);
-  assert.doesNotMatch(msg, /queued:\s*true/);
-  assert.doesNotMatch(msg, /Nothing was queued/, "a deferred post may still land — must not claim otherwise");
+  const lingering = scopeUnverifiedError({ toNodeId: 14, timeoutMs: 5000, cancelled: false, lingerMs: 600000 });
+  assert.match(lingering, /node 14/);
+  assert.match(lingering, /could not be verified/);
+  assert.match(lingering, /CONFIRMED queued/i);
+  assert.match(lingering, /sentinel/, "says the guard stays installed when cancellation was impossible");
+  assert.doesNotMatch(lingering, /Nothing was queued/, "a deferred post may still land — must not claim otherwise");
+  const cancelled = scopeUnverifiedError({ toNodeId: 14, timeoutMs: 5000, cancelled: true });
+  assert.match(cancelled, /REMOVED/);
+  assert.match(cancelled, /nothing was queued/i, "after a successful cancel, 'nothing was queued' is literally true");
 });
 
 // ---------------------------------------------------------------------------
@@ -230,19 +234,60 @@ test("#556 P1: an UNRELATED full run (different signature) during the window pas
   assert.equal(guard.state.dropped, null);
 });
 
-test("#556 P1: an unrelated SCOPED run (someone else's targets) passes through untouched", async () => {
+test("#556 P1: an unrelated SCOPED run (foreign signature, its own targets) passes through untouched", async () => {
   const spy = makeFetchSpy(async () => jsonResponse(200, { prompt_id: "other-scoped" }));
   let captured = null;
   const guard = createScopedRunGuard({
     origFetchApi: spy, execIds: ["14"], signature: OUR_SIGNATURE, batch: 1, toNodeId: 14,
     onPromptId: (p) => (captured = p),
   });
-  const body = { prompt: OUR_OUTPUT, client_id: "x", partial_execution_targets: ["9"] };
+  const body = { prompt: { "1": { class_type: "LoadImage" }, "2": { class_type: "PreviewImage" } }, client_id: "x", partial_execution_targets: ["2"] };
   const res = await guard(...promptPost(body));
   assert.equal(spy.calls.length, 1);
   assert.equal(res.status, 200);
   assert.equal(captured, null, "another scope's prompt_id is never claimed as ours");
   assert.equal(guard.state.observed, 0);
+});
+
+test("#556 r2 P0-1: OUR signature with WRONG/EXTRA/PARTIAL targets is OUR corrupted dispatch — REFUSED with zero dispatch (not forwarded as 'someone else's')", async () => {
+  for (const badTargets of [["14", "9"], ["9"], ["14", "76:34"]]) {
+    const spy = makeFetchSpy(async () => jsonResponse(200, { prompt_id: "must-never-happen" }));
+    let dropped = null;
+    const guard = createScopedRunGuard({
+      origFetchApi: spy, execIds: ["14"], signature: OUR_SIGNATURE, batch: 1, toNodeId: 14,
+      onScopeDropped: (m) => (dropped = m),
+    });
+    const body = { prompt: OUR_OUTPUT, client_id: "x", partial_execution_targets: badTargets };
+    const res = await guard(...promptPost(body));
+    assert.equal(spy.calls.length, 0, `targets ${JSON.stringify(badTargets)} must never leave the tab`);
+    assert.equal(res.status, 400);
+    assert.match(dropped, /node 14/);
+    assert.match(dropped, /instead of/, "scope_mismatch names what the body carried");
+    assert.equal(guard.state.observed, 0, "a corrupted dispatch never counts as observed");
+  }
+});
+
+test("#556 r2 P0-2: a FOREIGN post carrying the SAME targets does NOT satisfy observation — our deferred corrupted post is still refused afterwards", async () => {
+  const spy = makeFetchSpy(async () => jsonResponse(200, { prompt_id: "foreign-pid" }));
+  let captured = null;
+  const guard = createScopedRunGuard({
+    origFetchApi: spy, execIds: ["14"], signature: OUR_SIGNATURE, batch: 1, toNodeId: 14,
+    onPromptId: (p) => (captured = p),
+  });
+  // Another workflow whose prompt ALSO carries ["14"] but has a different node set.
+  const foreign = { prompt: { "14": { class_type: "SaveImage" }, "2": { class_type: "LoadImage" } }, client_id: "x", partial_execution_targets: ["14"] };
+  const foreignRes = await guard(...promptPost(foreign));
+  assert.equal(spy.calls.length, 1, "the foreign scoped run passes through");
+  assert.equal(foreignRes.status, 200);
+  assert.equal(guard.state.observed, 0, "same targets + different signature is NOT our dispatch");
+  assert.equal(captured, null, "its prompt_id is never claimed as ours");
+  // graph_run would still be waiting (observed==0) — and when OUR deferred
+  // corrupted post finally surfaces, the guard is still there to refuse it.
+  assert.equal(guard.state.dropped, null);
+  const spy2 = spy;
+  const res = await guard(...promptPost(OUR_FULL_BODY));
+  assert.equal(res.status, 400, "our scope-dropped dispatch is still refused");
+  assert.equal(spy2.calls.length, 1, "and never dispatches");
 });
 
 test("#556 P1: refusal is bounded by the batch — a same-signature full body BEYOND our batch count passes through", async () => {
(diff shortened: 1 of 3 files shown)


artokun:
Gate findings addressed in f0264f0:

**P0 (deferred dispatch escapes the guard):** the scoped guard is no longer restored when `app.queuePrompt` returns — it stays installed until our dispatch is actually OBSERVED (a POST carrying exactly the resolved `partial_execution_targets` passes through) or a bounded 5s timeout. If nothing surfaces, `graph_run` returns a truthful `could not verify — nothing confirmed queued` (`scopeUnverifiedError`), never `queued:true`.

**P1 (global fetch wrap constrains strangers):** the guard is now surgical — it only refuses a targetless `POST /prompt` body that matches THIS run's prompt signature (sorted node-id|class_type set of the `graphToPrompt` output we queued; widget values excluded so `beforeQueued` seed randomization can't blur it), and only within this batch's remaining attributable posts. Unrelated full runs, other scoped runs, and unparseable bodies pass through untouched and are never captured as ours. Unscoped runs keep the historical capture path, byte-identical.

New tests: deferred-post race (guard alive until observed; deferred invalid-shape refused with zero dispatch; deferred valid shape observed with prompt_id), concurrent unrelated full/scoped runs passing through, batch-bounded refusal, unparseable-body fail-safe, timeout ⇒ truthful-unverified. `npm run test:unit` 1588/1588, `tsc --noEmit` clean.

artokun:
Round-2 gate findings addressed in 73f2d49:

**P0-1 (wrong/extra targets forwarded as foreign):** attribution is now signature-first. A body matching this run's prompt signature IS this run: exact targets ⇒ observed + dispatched; missing/wrong/extra/partial targets ⇒ OUR corrupted dispatch ⇒ refused before it leaves (`scope_mismatch` names what the body carried). Only signature-mismatched bodies are foreign (pass through, never captured, never refused).

**P0-2 (foreign same-targets post satisfied `observed`):** observation now requires signature AND exact targets. A foreign post carrying the same targets passes through but never counts as ours — our own deferred corrupted post is still refused when it surfaces. Degraded mode (no signature, `graphToPrompt` failed) attributes only an exact-targets post: enough to verify, never enough to refuse.

**P0-3 (timeout restored the guard while the deferred item stayed live):** on give-up, `graph_run` now removes the still-pending item from the frontend's `queueItems` (`cancelPendingScopedQueueItem` — the server `/queue` never saw it; only items whose stored targets normalize to exactly our scope are spliced, so a user's pending full run is never touched). When removal is impossible (hard-private `#queueItems` builds, or a scope-dropped item with no attributable targets), the surgical guard stays installed as a 10-minute sentinel so the late corrupted dispatch is still refused. The unverified error message states which path was taken.

Tests for all three: wrong/extra/partial-targets-same-signature refused with zero dispatch; same-targets-different-signature does not satisfy observation; timeout cancel asserts the exact splice (both stored shapes, foreign items untouched) and the sentinel refuses a post-timeout dispatch with zero escapes. `npm run test:unit` 1593/1593, `tsc --noEmit` clean.

artokun:
Round-3 gate findings addressed in 3e45643:

**P0-1 (sentinel never installed):** root cause was real — `keepGuardInstalled` was set after the `finally` had already restored `api.fetchApi`. The entire scoped-dispatch orchestration is now `dispatchScopedRun()` in the lib, with the give-up decision (cancel vs sentinel) made INSIDE the `try` so the `finally` honors it. The integration test drives that real control flow: asserts `apiTarget.fetchApi` is still the guard after the timeout return, that a late scope-dropped post is refused with zero dispatch, and that the sentinel self-removes at the linger bound.

**P0-2 (degraded mode failed open):** no signature now means CANNOT SAFELY DISPATCH — `dispatchScopedRun` refuses upfront (`scopeUnattributableError`) before `queuePrompt` is ever called. Test asserts `queuePrompt` was never invoked and zero posts occurred.

**P0-3 (no run identity):** this run now marks its work end to end. (a) Queue-position MARK: every scoped run queues with `number = 2^30-1`; every frontend's `api.queuePrompt` copies nonzero `number` into the POST body verbatim (verified back to v1.28.0), and it sorts to the END of the server priority queue — same append position as the historical `number=0`. A body WITHOUT the mark is foreign: passed through, never captured, never refused, never observed — even with an identical node set AND identical targets. (b) Queue-item TAG: a non-enumerable per-run Symbol on the targets array (stored by reference in `queueItems`), so timeout cancellation removes ONLY our item; an identical-scope foreign item survives. When the tag can't be found (hard-private `#queueItems`, copied array), the surgical guard stays as the long-bound sentinel.

Red-first integration tests for all three paths. `npm run test:unit` 1584/1584, `tsc --noEmit` clean.

artokun:
Round-4 gate findings addressed in 803c963:

**P0-1 (sentinel expiry re-opened the hole):** the `lingerMs` timer is gone entirely. When cancellation is impossible, the guard now stays installed for the PAGE LIFETIME — the uncancellable deferred item's late scope-dropped post is refused whenever it surfaces. This is safe by construction: the guard only ever acts on its own run's unique mark, so foreign traffic passes through untouched forever, and multiple sentinels chain through whatever wrap is current (each run restores only inside its own synchronous dispatch; an old sentinel has no cleanup that could clobber a newer run's guard).

**P0-2 (fixed mark let run A's sentinel claim run B's traffic):** the mark is now UNIQUE PER RUN (`newScopedQueueMark()` — decrementing from 2^30-1: nonzero safe integers that still always sort to the END of the server's priority queue, same append position as `number=0`). Each run's guard, observation, and cancellation key on its own mark; `cancelPendingScopedQueueItem` now requires BOTH the ownership tag AND the item's `number === this run's mark`. `dispatchScopedRun` returns the run's `queueMark`.

Tests: sentinel persists past the old linger bound (advanced clock) and still refuses the late scope-dropped post with zero dispatch; two overlapping scoped runs (A unverified+sentinel, then B to a different node) — B's marked post passes A's sentinel untouched, is observed by B's guard with prompt_id captured only by B, B's cleanup restores exactly A's sentinel, and A's sentinel still refuses A's late corrupted post; mark uniqueness/range; cancellation requires tag AND mark. Full suite 1586/1586 on BOTH node 22 and node 24, `tsc --noEmit` clean.

artokun:
Round-5 gate finding addressed in 1c9c5d3:

**P0 (batch>1 declared `dispatched` after the first verified post):** the guard's verdict is now batch-aware — it resolves only when the WHOLE batch is accounted (`observed >= batch`) or when any corrupted post ends the attempt (the frontend's batch loop breaks on the first refusal). The guard stays installed until then, so no batch post bypasses verification. Outcomes:

- `observed == batch` ⇒ `dispatched` (never before);
- corrupted with 0 verified ⇒ shape dropped ⇒ retry the other arg shape (nothing queued, no double-queue);
- corrupted with ≥1 verified ⇒ TERMINAL PARTIAL: the shape demonstrably works and the batch loop broke on the refusal, so the run returns `scopePartialBatchError` naming verified/refused/unposted counts — never `dispatched`;
- batch never completes ⇒ the same cancel-then-page-lifetime-sentinel path as the zero-post case, with the unverified error naming how many of the batch were verified (verified prompts ARE queued, scoped).

Also: `graph_run`'s scoped error return moved AFTER the #370 ledger registration, so partially-verified prompts (queued, scoped) stay reconcilable.

Tests: batch=2 first-verified/second-scope-dropped ⇒ refused with counts, no escape; batch=2 fully verified ⇒ dispatched with both prompt_ids; batch=2 second-never-arrives ⇒ unverified naming "1 of 2" with the sentinel refusing the late scope-dropped post. Full suite 1589/1589 on BOTH node 22 and node 24, `tsc --noEmit` clean.

artokun:
Round-6 gate finding addressed in bda4289:

**P1 (`observed` incremented before the fetch completed):** a batch post now counts as VERIFIED only when its attributed fetch completes with HTTP 200 + a parseable prompt_id. `classifyRunResponse` distinguishes three outcomes for an attributed post: `accepted` (200 + prompt_id, captured), `rejected` (genuine server rejection — non-200 with error/node_errors, flowing through the established #358 channel with unchanged semantics), and `malformed` (everything else). A thrown fetch or malformed response is a terminal dispatch FAILURE (`scopeDispatchError` naming what failed and how much of the batch was confirmed first) — never counted, never `queued:true`. The guard rethrows fetch errors so the frontend sees exactly what it would have seen. A failure with the batch not fully accounted keeps the page-session sentinel, because the frontend continues its batch loop on generic failures and a later corrupted post must still be refused.

Tests: fetch-throws (rethrown, not verified, not captured); malformed variants (200 w/o prompt_id, non-200 w/o rejection body, `prompt_id: null`) all fail truthfully; genuine server rejection still takes the established rejection path; batch=1 throw ⇒ `failed` + sentinel refuses a late corrupted post; batch=2 throw-then-verify ⇒ still `failed` (names the failure, '0 of 2'), post 2's id ledger-captured but the run not claimed. Full suite 1594/1594 on BOTH node 22 and node 24, `tsc --noEmit` clean.

artokun:
Round-7 gate finding addressed in 86fcbc3:

**P1 (topology-only fingerprint):** attribution now uses `promptContentHash` — a stable FNV-1a hash of the FULL queued prompt (node ids, class types, links, every input/widget value), canonicalized so key order can't blur it, computed once at queue time and compared per observed post. The ONLY excluded values are inputs of widgets that self-mutate at queue time (`beforeQueued` hooks — stock control_after_generate seeds etc.), which change between any two serializations of the SAME graph by design; `collectSelfMutatingInputNames` derives that set from the live graph (root + nested subgraphs). A marked post whose content drifted (user edit, or deferred serialization of a changed graph) is OUR corrupted dispatch with reason `graph_changed` — refused before it leaves, with a truthful error naming the drift. Drift is not an argument-shape problem, so no shape retry; terminal partials name it too. r6 accounting/rejection behavior untouched.

Tests: same-topology drift (changed widget value AND rewired link) ⇒ refused, zero dispatch, error names the graph change, exactly one attempt; a deferred post differing only in a rerolled `beforeQueued` seed still observes/dispatches; hash unit tests (widget values and links covered, key-order-insensitive, volatile exclusion); `collectSelfMutatingInputNames` recursion incl. subgraphs. All topology-era regression tests green. Full suite 1597/1597 on BOTH node 22 and node 24, `tsc --noEmit` clean.

artokun:
Round-8 gate findings addressed in 6a18d8d:

**P1-1 (global-name volatile exclusions):** exclusions are now PER-NODE pairs (`"execId inputName"`) via `collectVolatileInputs` — root nodes key as `String(id)`, nested nodes as the colon-joined subgraph-instance path (same shape as `buildNodeExecutionId`), so pairs line up with flattened API-prompt keys. An edit to a NON-hook node's same-named input is still detected as drift, and a prompt node that can't be resolved to a live node carrying the hook gets NO exclusions (fail toward detecting drift).

**P1-2 (volatility collected from `app.rootGraph` — empty in production):** `dispatchScopedRun` now reads the live root from `app?.graph ?? app?.rootGraph`, and the tests use `app.graph` (production shape) instead of masking with `rootGraph`.

Tests: pure-hash per-node exclusion (hook node's reroll excluded; non-hook same-named edit still changes the hash); `collectVolatileInputs` pair shape incl. colon paths; integration — legitimate `beforeQueued` reroll dispatches in production shape; edit to a NON-hook node's `seed` (hook node's `seed` also present) detected as drift and refused with zero dispatch. Full suite 1598/1598 on BOTH node 22 and node 24, `tsc --noEmit` clean.

Issue #584, opened 2026-08-03
Frontend reload keeps stale panel JS and blocks graph writes


artokun:
### What happened / root cause
Graph write and run calls were rejected because the connected ComfyUI tab reported panel 0.11.34, although the installed comfyui-mcp-panel files on disk are 0.11.38. Restarting ComfyUI and calling panel_reload(scope='frontend') did not update the frontend version, so graph_edit_node and graph_run remained blocked by the workflow-target safety gate. This appears to be stale frontend caching/reload behavior.

### Steps to reproduce
1. Have comfyui-mcp-panel 0.11.38 installed while an existing ComfyUI tab has loaded panel JS 0.11.34.
2. Restart ComfyUI.
3. Call panel_reload with scope frontend and allow the tab to reconnect.
4. Call panel_edit_node or panel_run.
5. The MCP still detects panel 0.11.34 and rejects the graph operation.

### Exact error (scrubbed)
graph_edit_node / graph_run could not be dispatched. The panel tab does not recheck workflow targeting at the graph write boundary after asynchronous work; detected panel 0.11.34, MCP requires panel 0.11.35+.

### Fix
applied locally: no  upstream-only: yes
The current workaround is a user hard refresh (Ctrl+Shift+R). panel_reload(scope='frontend') should reliably load the installed panel bundle, or the panel bundle URL should be cache-busted by version so stale JS cannot survive a backend restart/reconnect.

### Environment
Windows 11; ComfyUI 0.30.0 local; RTX 4070 SUPER 12 GB; comfyui-mcp 0.49.3; installed comfyui-mcp-panel 0.11.38; stale frontend 0.11.34.

artokun:
Additional reproduction on current releases: comfyui-mcp 0.49.4 and panel 0.11.38 on Windows 11 / ComfyUI 0.28.2 / RTX 4080 SUPER. `install_panel(status)` reports panel 0.11.38 on disk, but an older browser tab reports the active bundle version as `unknown`, causing `graph_add_node` to be refused (`requires panel 0.11.35+`). Read-only graph commands still work; Ctrl+Shift+R restores the tab. This corroborates stale cached frontend JS causing the disk and connected-tab versions to diverge.

artokun:
[...]

PR #596, opened 2026-08-03
fix(panel): version frontend bundle entry (#584)

artokun:
## What changed
- Replaced the auto-discovered panel entry with a small loader and moved the full runtime to a versioned module URL.
- Updated release/version, pack-content, vocabulary, syntax, and harness checks for the loader/runtime split.
- Added a regression test that locks the fresh versioned import contract.

## Why
ComfyUI discovers extension JavaScript by its fixed path. Reloading only the page could therefore execute a cached legacy panel bundle after a backend restart, leaving graph writes blocked by a stale reported panel version.

## Validation
- npm.cmd run test:unit (1706 passing)
- npm.cmd run typecheck
- node scripts/check-tool-vocabulary.mjs
- node --check web/js/cmcp-panel-loader.js
- node --check web/js/comfyui-mcp-panel.mjs


artokun:
**Holding this until the panel PR queue drains** — not a problem with the change.

This renames `web/js/comfyui-mcp-panel.js` → `comfyui-mcp-panel.mjs` and adds a loader. That rename conflicts with **every in-flight panel branch**, and there are currently ten: #622, #623, #624, #625, #626, #627, #628, #629, #630, #633, #634 — most of which edit `comfyui-mcp-panel.js` directly. Landing it now converts one clean merge into ten hand-resolved rename conflicts, each an opportunity to silently drop a hunk.

#621 (merged as `7b1b27e`) and #614 have just landed on that file, so this needs a rebase regardless.

Two things worth keeping when it does land:
- The rename to `.mjs` is the right call independently. Panel JS **loads as an ES module**, and a duplicate top-level declaration breaks the whole panel while `node --check` in CJS mode does not catch it. A `.mjs` extension makes the syntax check match how the file is actually loaded — that class of bug has bitten this repo before.
- The pyproject↔loader version-mismatch check in `publish_action.yml` is a real guard: the Comfy Registry publishes on a `pyproject.toml` version change, and a loader version that disagrees ships a mismatched bundle.

I will land this deliberately as its own step once the queue is clear, with the rebase done in one place rather than eleven.

artokun:
Closing as **superseded** — the underlying issue (#584) is already fixed on `main` by a different, less invasive approach, and that fix shipped in panel **0.11.42**.

## What landed instead

`web/js/lib/bundle-version.js` + its two call sites implement **detect-then-repair**, and it self-heals on a *normal page load* — which is precisely the scenario this PR describes ("reloading only the page could execute a cached legacy panel bundle after a backend restart"):

- `resolveBundleStaleness({running, installed})` compares the running `PANEL_VERSION` against what's installed → `current` / `stale` / `unknown`
- on `stale`, `primeModuleCache()` walks the pack's module graph and re-fetches every module with `cache: "reload"` — `collectRelativeImportSpecifiers()` covers **static imports and literal dynamic `import()` calls**, so a lazily-loaded vendor bundle can't stay stale and produce a mixed bundle
- then reloads **once**, guarded by a `sessionStorage` marker
- if `sessionStorage` is unavailable, it deliberately does **not** auto-reload (the loop guard couldn't persist) and names the manual remedy instead
- the prime is bounded at 5s so a hung module fetch can't strand the reload; if modules fail to refresh it reloads anyway and says what to do

Verified: `bundle-version.test.mjs` + `module-graph-link.test.mjs`, **18/18 passing**. Shipped in `378bf81`, an ancestor of the 0.11.42 release commit.

## Why that approach over this one

This PR replaces the auto-discovered panel entry with a loader plus a versioned module URL. That's a change to how ComfyUI discovers the extension entry point — if it's wrong, the panel doesn't load *at all*, and this repo's known failure mode there (a duplicate top-level declaration in an ES module) is invisible to `node --check`. The landed fix gets the same guarantee without touching the boot path, and additionally handles the dynamic-import case explicitly.

The branch is **not deleted** — if the versioned-URL approach turns out to be warranted later (e.g. the heuristic-cache path proves insufficient in the wild), this is the starting point. Reopen rather than rewrite.

Also note this PR is now `DIRTY` after ~4 days spanning the 0.50.0 tool-surface migration, the interactive-card fence, and #714, so it would need a substantial rebase through exactly the kind of change where a mis-resolved conflict lands silently.

PR #597, opened 2026-08-03
fix(panel): prevent restart resume from duplicating active render

### LINKED ISSUE #585: Stale restart notice causes duplicate render queue
The sidebar injected a system notice stating that ComfyUI had just restarted and instructed the agent to resume. The agent reasonably treated the previous render as aborted and queued the same final branch again. panel_run then reported that one render was still in flight and the new prompt was queued behind it. The restart notice was therefore stale or misleading and caused a duplicate render.

### Steps to reproduce
1. Queue a render with panel_run.
2. Receive the automatic waiting state.
3. A sidebar/system notice appears: ComfyUI just restarted to load newly-installed custom nodes; continue what you were doing.
4. Re-read the graph and queue the interrupted branch again.
5. panel_run reports the new prompt is behind one existing in-flight render.

### Exact error (scrubbed)
System notice: ComfyUI just restarted to load newly-installed custom nodes (now available).
panel_run response: Queued behind your own in-flight render(s) (1 running).

### Fix
Restart-resume notices should be tied to a confirmed backend lifecycle event and include whether the pre-restart prompt was aborted. Suppress stale notices after the backend is already healthy, or expose the active prompt id so the agent can distinguish an interrupted render from one still running before retrying.

### Environment
Windows 11; ComfyUI 0.30.0 local; RTX 4070 SUPER 12 GB; comfyui-mcp 0.49.3; comfyui-mcp-panel 0.11.38.

artokun:
Fixes #585.

## The bug

After a ComfyUI restart the panel nudges the agent to "continue what you were doing". The reporter's render was **still in flight** when that notice fired, so the agent reasonably concluded it had been aborted and queued the same branch again — `panel_run` then reported the new prompt queued behind "1 running" render.

## The through-line

Suppression was decided from an **in-memory, global, pre-delivery** signal, while the thing it guards against **survives reloads, spans workflows, and completes asynchronously**. Every fix below is that mismatch on a different axis.

Two failure directions are in play and they pull opposite ways. A duplicate render wastes GPU but is **visible and cancellable**; a swallowed resume leaves the user waiting forever for a turn that will never start, **silently**. Where the code genuinely cannot determine which case it is in, it takes the visible, recoverable outcome and **discloses** it.

## What changed

**Correlation, not aggression.** The reboot marker is now a versioned JSON blob in sessionStorage carrying the prompt ids that were still owed a completion frame when the reboot was armed, plus the arm time, the arming conversation, and the attempt counts. Every decision is about *those* ids — never a global "is anything pending" count, which an unrelated workflow's render would satisfy.

**Reload survival.** A restart can reload the frontend, which hands the panel an empty ledger. Persisted ids are re-adopted into the fresh tracker *before* any decision, handing the question to the existing `/history` + `/queue` reconcile. (A fresh tracker answers "settled" for an id it has never heard of — correctly, it owes nothing — so the marker alone would have reproduced the bug while looking like a fix.)

**The marker is never cleared while suppressing.** It is retired only when a resume is actually sent *and* the orchestrator acknowledges receipt. `sendUserMessage` returning true only means the socket accepted the bytes; a socket that closes before the orchestrator reads them cannot say so.

**Delivery gated on "the agent was told".** The run tracker retires a run optimistically, before the caller's async compose+send resolves. `isSettled()` now answers the delivery question honestly, and a dispatched frame that is never confirmed graduates to an explicit *unconfirmed* state — non-blocking, but flagged, so the resume discloses rather than falsely reassuring.

**Delivery targeted at the arming conversation.** The wait set stays global — a reboot restarts ComfyUI globally, so every in-flight render is exposed; this is the correct set, not an approximation of a per-workflow one (see the comment at the arm site before narrowing it). But the resume is a message into *one* agent session, so a mismatch holds the marker with a visible notice, or past the backstop is abandoned visibly — never misdelivered.

**Unknown stays unknown.** An absent arm time, a negative elapsed (clock rollback), an unrecognized marker version, an absent attempt count — each is its own answer, never substituted with a benign definite one. A 15-minute backstop resumes with an explicit "check the queue for prompt N before re-queueing" instead of holding in silence.

## Known residual — sessionStorage write failures

**A documented, deliberate stopping point, not an oversight.**

The panel's whole resume subsystem is built on best-effort `sessionStorage` (`SESSION_KEY`, `MID_TASK_KEY`, `SOFT_RELOAD_KEY`, `AUTOCONNECT_KEY` all share it). A browser that *accepts reads but refuses writes* — quota exhausted, private mode, eviction — cannot be made reliable by writing more.

**What such a user would experience:** after a frontend reload, they may see one extra restart nudge, and one render result may be reported to the agent twice. Both are visible and recoverable; neither loses work.

**What protects them:** every marker write is verified by read-back, and the attempt record is written **ahead** of the send. A failed record raises the duplicate warning — the agent is explicitly told "you may have already received this restart notice; ignore the duplicate and do NOT re-queue anything" — instead of a silent duplicate. The failures are also surfaced to the user directly: at arm, at clear, and on any failed marker update.

**The precise limit of that guarantee.** It holds for the **life of the mount**, not indefinitely. One timeline escapes it:

> the marker holds a real `0` count → an attempt's count write **fails** but its message **does** reach the orchestrator → storage later **recovers** → the frontend **reloads**.

The new mount resets its in-memory "storage was broken" state, reads a count that is a genuine zero merely gone stale, records its own attempt successfully, and computes *no prior attempt*. The next nudge is therefore an **undisclosed duplicate continuation nudge** — not a duplicate render. The agent is told to continue twice with nothing flagging it; whether that produces a re-queue depends on what it does next, and every other guard in this PR (run correlation, delivery gating, session targeting) still applies to it.

**Why this one is accepted rather than fixed.** Note what it is *not*: it is not the unknown-vs-zero fold fixed elsewhere in this PR, because no value is absent — the stored zero is real, just stale. Closing it would mean **durably recording "a write failed" at exactly the moment storage is refusing to durably record anything.** The only remaining option is to warn on *every* reboot-resume that follows *any* reload, which is the cry-wolf outcome: a warning that fires on the ordinary path trains the agent to ignore it, including in the `replaced`-session case where it genuinely matters. We took the narrow, disclosed duplicate nudge over the broad, self-defeating warning.

**Why we stopped here rather than hardening further:** each additional fix in this class *adds state to a persistence path*, and the thing being defended against is that persistence being unreliable. That is self-defeating, and it bit us once during review — a persistent storage failure was answered with a mount-local flag, which fixed the discovering mount and left the reload broken: a new bug created by adding state to work around missing state. The fix was to stop remembering and re-ask (probe writability on demand). We chose **disclosure over further state**.

Further findings in this class belong in this residual. Findings in the core class — a duplicate render reaching a live job, a swallowed resume, a misdelivered resume, or fabricated reassurance that a result was delivered — are bugs and get fixed.

## Verification

- 1765 unit tests passing; `typecheck` clean; tool-vocabulary gate clean; all three changed modules `node --check` clean **as ESM**; `git diff --check` clean; zero NUL bytes in every changed file.
- **34 mutation runs, all killed** — each fix was broken deliberately and the corresponding test confirmed to fail. Includes both directions of the unknown-elapsed error (0 never expires / infinity always expires), one that re-conflates the retry budget with the duplicate evidence, one that re-coerces an absent count to zero, one that treats any non-connected status as a bridge drop, and one that reintroduces masking parameter defaults.
- Gated independently with codex (`gpt-5.6-terra`, read-only) across nine rounds. The original three P1s plus fourteen further findings were fixed. Two proposed fixes were **declined with reasoning and upheld on re-review** — most notably a strict session-id delivery gate, which would have stranded every ordinary restart, since a session id legitimately changes across a resume.

## Still owed — live verification on the rig

Unit tests fake reload-survival and delivery-ordering most convincingly, so these are flagged rather than claimed:

1. **A real F5 mid-render** — the highest-value check. Tests simulate a reload as "a new tracker instance"; only a live reload proves the marker survives, adoption runs before the ready ack, and reconcile finds the run in `/queue`.
2. **The wait actually ends** — slow render → `comfy_reboot` → the panel holds, stays quiet, and sends the resume *after* the completion frame lands.
3. **The 15-minute backstop** end to end (the decision function is tested; the panel timer path is not).
4. **Message ordering as the agent sees it** — tests assert the decision, not that the completion precedes the nudge on the wire.
5. **Multi-workflow** — B rendering while A reboots; A's resume must not be swallowed, and must arrive once.
6. **Whether `SESSION_KEY` is stable across a real reboot+resume, and whether the `ready` ack precedes the `session` frame.** Together these decide whether the session-state disclosure reads `same` (rare) or `unknown` (routine). The tri-state logic is correct either way; if `unknown` turns out to be routine, the *wording* must not read as a warning, or it trains people to ignore the `replaced` case that actually matters.

### DIFF
--- a/browser_tests/unit/restart-resume.test.mjs
+++ b/browser_tests/unit/restart-resume.test.mjs
@@ -607,6 +607,39 @@ test("#585 P1(budget-vs-evidence): a bridge drop refreshes the BUDGET without er
   );
 });
 
+test("#585 P1(unknown-count): an absent or malformed attempt count is UNKNOWN, not a verified zero", () => {
+  // Coercing it to 0 says "no attempt was ever made" on no evidence, and that is
+  // precisely what suppresses the duplicate warning.
+  for (const bad of [{}, { t: "2" }, { t: null }, { ts: {} }, { t: NaN, ts: NaN }]) {
+    const m = decodeRebootMarker(JSON.stringify({ v: 1, at: 1000, runs: [], n: 0, ...bad }));
+    assert.equal(m.totalAttempts, null, JSON.stringify(bad));
+    assert.equal(
+      rebootResumeRepeatWarning({ totalAttempts: m.totalAttempts, attemptRecorded: true, sentThisMount: 0 }),
+      true,
+      `an unknown count must warn (${JSON.stringify(bad)})`,
+    );
+  }
+});
+
+test("#585 P1(unknown-count): re-encoding an unknown count must not launder it into a zero", () => {
+  // A retained marker is re-encoded on every wait tick, so a lossy round-trip would
+  // convert "unknown" to "verified none" within seconds of the uncertainty arising.
+  const unknown = decodeRebootMarker(JSON.stringify({ v: 1, at: 1000, runs: ["P"], n: 1 }));
+  assert.equal(unknown.totalAttempts, null);
+  const round = decodeRebootMarker(encodeRebootMarker({ ...unknown }));
+  assert.equal(round.totalAttempts, null, "still unknown after a round-trip");
+  assert.equal(round.attempts, null);
+
+  // …and the wait path, which rewrites the marker, preserves it too.
+  const held = stepRebootResume({
+    raw: encodeRebootMarker({ ...unknown }),
+    isSettled: () => false,
+    nowMs: 1100,
+  });
+  assert.equal(held.decision, "wait_for_run");
+  assert.equal(decodeRebootMarker(held.nextRaw).totalAttempts, null);
+});
+
 test("#585: a legacy marker with only the per-episode count still reports it as evidence", () => {
   // `ts` absent ⇒ fall back to `t` rather than to zero: an older marker that
   // recorded one attempt must not read as "never attempted".
--- a/browser_tests/unit/restart-resume.test.mjs
+++ b/browser_tests/unit/restart-resume.test.mjs
@@ -1,19 +1,437 @@
 import test from "node:test";
 import assert from "node:assert/strict";
 
-import { planRebootResume } from "../../web/js/lib/restart-resume.js";
+import {
+  adoptRebootRuns,
+  decodeRebootMarker,
+  encodeRebootMarker,
+  planRebootResume,
+  pruneRebootMarkerRaw,
+  rebootMarkerAfterSend,
+  stepRebootResume,
+  unsettledRebootRuns,
+  REBOOT_RESUME_MAX_WAIT_MS,
+} from "../../web/js/lib/restart-resume.js";
+import { createRunCompletionTracker } from "../../web/js/lib/run-completion.js";
 
-test("#585: a pre-restart render pending in the completion ledger suppresses the generic resume nudge", () => {
-  // The run may have survived the backend restart. Its terminal completion (or
-  // conservative reconnect reconciliation) is the signal that may wake the
-  // resumed agent; a generic "continue" prompt would invite a duplicate queue.
-  assert.equal(planRebootResume({ rebootPending: true, hasPendingRun: true }), "wait_for_run");
-});
+// A real tracker with manual timers, so these tests exercise the ACTUAL delivery
+// lifecycle rather than a hand-rolled stand-in for it. `fireTimers(ms)` runs the
+// pending timers armed for that exact delay — used to trip the delivery watchdog.
+function makeTracker(onFlush = () => {}) {
+  const timers = new Set();
+  const clock = { t: 1_000_000 };
+  const tracker = createRunCompletionTracker({
+    onFlush,
+    now: () => clock.t,
+    setTimer: (fn, ms) => {
+      const t = { fn, ms };
+      timers.add(t);
+      return t;
+    },
+    clearTimer: (t) => timers.delete(t),
+  });
+  tracker._advance = (ms) => {
+    clock.t += ms;
+  };
+  tracker._fireTimers = (ms) => {
+    for (const t of [...timers]) {
+      if (t.ms !== ms) continue;
+      timers.delete(t);
+      t.fn();
+    }
+  };
+  return tracker;
+}
 
-test("#585: an actual reboot with no pending render retains autonomous resume", () => {
-  assert.equal(planRebootResume({ rebootPending: true, hasPendingRun: false }), "resume");
-});
+const settledBy = (tracker) => (id) => tracker.isSettled(id);
+const unconfirmedBy = (tracker) => (id) => tracker.isDeliveryUnconfirmed(id);
+const DELIVERY_WATCHDOG_MS = 120000;
+
+// ── the planner's own branches ────────────────────────────────────────────────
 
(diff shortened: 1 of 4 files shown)


artokun:
Gated this independently (codex `gpt-5.6-terra`, read-only) because the handoff flagged the original review as interrupted. **FAIL — three P1s.** CI green, NUL-scan clean on all three changed files, so this would have merged on green alone.

The through-line: suppression is decided from an **in-memory, global, pre-delivery** signal, while the thing it guards against survives reloads, spans workflows, and completes asynchronously.

**P1 — duplicate render still possible after a frontend reload.**
`comfyui-mcp-panel.js:18936` bases suppression on the in-memory `runCompletionRef`, but `REBOOT_KEY` deliberately survives page reloads (`:1904-1908`). The tracker starts empty (`:297`) and reconciliation only examines existing pending entries (`lib/run-completion.js:694`).
Scenario: render P queued → reboot armed → frontend reloads while P is still running. On the new mount P isn't in the fresh ledger, `hasPendingRun` is false, the ready nudge fires and can requeue P while the original is still executing. That is the exact bug this PR exists to fix, surviving one reload.

**P1 — the predicate is global, so an unrelated render permanently swallows a required resume.**
`hasPending()` is just `pending.size > 0` (`lib/run-completion.js:729-730`), and every observed execution start is inserted (`:528`, `comfyui-mcp-panel.js:19351-19352`) with no workflow or reboot-render correlation. The suppression branch then **clears the reboot marker** (`:18939`).
Scenario: workflow A requests a reboot; workflow B has a normal render in flight. B makes the shared ledger non-empty, so on ready A's nudge is suppressed *and its marker removed*. When B drains, nothing reissues A's nudge — A's legitimate resume is swallowed permanently, with no error. This is the inverse failure, and it's worse than the duplicate: the user waits for a render that will never start.

**P1 — completion-delivery race still permits the duplicate nudge.**
`flush()` calls `markDelivered()` *before* invoking the async completion callback (`lib/run-completion.js:281-282`), while the callback awaits `composeRunCompletionFrame` (`comfyui-mcp-panel.js:19077-19100`) and only re-pends on a later failed send (`:19107-19117`).
Scenario: P hits `execution_success` right after restart, output composition still awaiting image metadata. A ready ack in that window sees nothing pending and sends the generic continuation before P's completion frame — the agent requeues P believing the prior render was lost.

**Direction that closes all three:**
1. Correlate the suppression to the **specific** reboot render — store its id alongside `REBOOT_KEY` and check that id, never a global count.
2. Make the tracking survive a reload the same way `REBOOT_KEY` does, or reconcile on mount against the live queue/history rather than an empty in-memory ledger.
3. Do **not** clear the reboot marker when suppressing. Clear it only when that specific render is confirmed complete *and delivered*.
4. Gate on completion actually **delivered to the agent**, not on `markDelivered()` fired before the async callback resolves.

Cleared and not to be disturbed: nothing here cancels or clears a queue entry that isn't ours, and no path fabricates a successful queue submission.

Not merging. Happy to take the rework if nobody else has it.

Issue #607, opened 2026-08-04
Panel permanently wedged on "workflow instance mismatch" — panel_set_workflow_target({mode:"current"}) does not clear the stale command UUID stamp


artokun:
### What happened / root cause

Every `panel_*` call in a fresh session failed with:

```
Error: workflow instance mismatch: this command targets a different workflow than the
active canvas — the workflow was switched or replaced after it was issued.
Re-select the intended workflow (panel_open_workflow) or re-target with
panel_set_workflow_target({mode:"current"}), then retry.
```

This was NOT a transient race — it was the very FIRST `panel_graph_outline` of the
session, and it persisted through every documented recovery. The session was in
`mode: "current"` the whole time (confirmed via `panel_get_workflow_target`), and
`panel_list_workflows` reported `active_confirmed: true` for
`workflows/MiniMax H3 Image.json` (`workflow_uuid 17c3ae56-…`) — i.e. the panel
could see the active canvas fine, but the UUID the orchestrator stamps onto
commands never matched `workflowStableUuid()` at the fence
(`web/js/comfyui-mcp-panel.js` ~line 11666 / `assertActiveWorkflowCommandTarget`
~line 1796).

Root cause appears to be that the orchestrator caches a workflow UUID from an
earlier hello/handshake and **`panel_set_workflow_target({mode:"current"})` does
not invalidate or re-read that cached stamp** — it returns
`{mode:"current", note:"Following the user's current workflow tab."}` (a success)
while the next command still carries the stale UUID. The error message tells the
agent to run exactly that command as the recovery, so the advertised recovery
path is a no-op and the session is unrecoverable from the agent side.

Notably `panel_list_workflows` and `panel_get_workflow_target` KEPT WORKING
(they are unfenced), which makes the failure especially confusing: the agent can
enumerate the active workflow and confirm its identity, but cannot read or edit it.

### Steps to reproduce

Not deterministically reproduced, but the preconditions were:
1. Many workflow tabs open (~31, mix of saved `wf:` and unsaved `tmp:` tabs).
2. Several tabs flagged `modified: true` (including the active one).
3. New agent session started against the already-open canvas.
4. First `panel_graph_outline` → `workflow instance mismatch`.

### Recovery attempts — all failed

- `panel_set_workflow_target({mode:"current"})` → returned OK, no effect (x2)
- `panel_get_workflow_target` → `{mode:"current"}` (so the pin was not the issue)
[...]

Issue #606, opened 2026-08-04
Binding guard wedges a BRAND-NEW empty workflow tab after a reconnect: graph_outline + workflow_open both refuse until a frontend reload


artokun:
### What happened / root cause

After the orchestrator's connection dropped and resumed (ComfyUI restart / socket drop), the panel had just created a NEW empty workflow tab (`workflow_new`, ~10s earlier, `applied:true`, routing key `tmp:<uuid>`). Every graph command against that tab then refused, permanently, with:

> The live graph is out of sync with the active workflow: the workflow reports **0** node(s), but the canvas is bound to a different graph. A load, tab switch, or reconnect left this command pointed at the wrong canvas, so it was NOT applied. Re-open the active workflow tab (panel_open_workflow) or reload the panel to rebind the graph, then retry.

Two problems:

**1. The guard fires on a genuinely-empty, brand-new tab and never clears.** `expected` in that message is `0`, so this cannot be the `baselineReadDesync` branch (`graphReadDesynced` requires a POSITIVE tracker count). It is `root-workflow-uuid-mismatch` or `root-shape-mismatch` in `assertGraphBinding` (`web/js/comfyui-mcp-panel.js` ~L4390-4418). On a fresh tmp workflow created right before/around a reconnect, the root graph carries a stale/absent workflow-uuid tag while the new workflow's ChangeTracker isn't yet *proven* empty — so the `staleTagOnEmptyCanvas` rebind path at ~L4361 (`graphRootProvenEmpty(rootGraph) && activeWorkflowProvenEmpty(activeWorkflow)`) does NOT trigger, and the guard throws forever. Nothing in the normal command path re-stamps the root, so it is self-wedging rather than transient.

**2. The remediation the error tells you to do does not work.** The message says to re-open the active workflow tab with `panel_open_workflow`. Calling it on the active tab's own key failed too:

> workflow_open could not prove that the active canvas was rebound to the requested workflow. The open may have switched the active workflow; reload the panel before graph edits.

Only `panel_reload({scope:"frontend"})` recovered it — after which `graph_outline` returned a clean `node_count: 0`. So the guard's own suggested first remedy is a dead end, and an agent that follows it burns retries before finding the one thing that works.

[...]

Issue #601, opened 2026-08-03
Dirty-tab mutation fence lets an agent make only ONE graph mutation per page load (assertGraphBoundToActiveWorkflow / #545), behind a misleading node-count error


artokun:
### What happened / root cause

During a long build session the agent could add **exactly one node** after each browser reload. Every subsequent `graph_add_node` was refused with:

> The live graph is out of sync with the active workflow: the workflow reports 339 node(s), but the canvas is bound to a different graph. A load, tab switch, or reconnect left this command pointed at the wrong canvas, so it was NOT applied.

Reads (`graph_outline`, `graph_query`) kept working throughout, which made the message look like a stale node-count cache. **It is not a count problem at all.**

Root cause is in `web/js/comfyui-mcp-panel.js`, `assertGraphBoundToActiveWorkflow()` (~L4104):

```js
const dirtyMutationBindingUnproven =
  requireDirtyMutationBinding &&
  activeWorkflow?.isModified === true &&
  !graphRootWorkflowUuidMatches({ rootGraph, activeWorkflowUuid });
```

`requireDirtyMutationBinding` is true for every non-read-only graph command (`graphCommandMayMutateWorkflow`, L11324). So once `isModified === true`, each mutation needs a POSITIVE `rootGraph.extra.comfyui_mcp.workflow_uuid` match.

The deadlock: **the first successful mutation is itself what sets `isModified = true`.** `workflowStableUuid(..., {embed:true})` writes the uuid into `workflowOwnedExtra(wf)` — i.e. `wf.extra` (L1349-1352) — while the guard reads it from the LIVE `rootGraph.extra` (`graphRootWorkflowUuidMatches`, graph-binding.js L247). After a browser reload rebuilds `app.graph` from saved JSON, that live root carries no stamp. Result:

1. F5 → tab clean → first `graph_add_node` passes, node created, tab now dirty.
2. Every later mutation → `isModified===true` + unstamped root → `dirtyMutationBindingUnproven` → refused.
3. No TTL, no self-recovery. Only another F5 clears it — for exactly one more mutation.

[...]

PR #622, opened 2026-08-04
binding wedged after reconnect or restart, with no path back — including a brand-new empty tab

artokun:
## Issues in this cluster

- #607
- #606
- #560
- #436
- #178

Grouped because they share a codepath: once the panel's binding evidence drifts (a stale root identity tag after a reconnect, or the orchestrator's cached workflow-uuid stamp going stale), every recovery the errors advertised was a dead end and the session wedged until a manual browser refresh.

Rebased onto `main` twice: after #621 merged, and again after #623/#625–#629/#632 merged. Each report re-verified against the merged tree before changing anything.

## What this PR changes

**#606 — brand-new empty tab wedged after a reconnect**
- `workflow_new` stamps the live root with the NEW tab's identity at creation, but only when the root is PROVEN content-free on every serialized surface (`graphRootProvenEmpty`). ComfyUI reuses `app.graph` across tabs and clear/configure does not reset `graph.extra`, so a fresh tab otherwise inherits the previous workflow's tag and wedges behind the binding guard while its ChangeTracker is not yet proven empty. A root with content is never re-tagged (fail closed).
- `workflow_open`'s side of this cluster is now owned by **#623's attempt-scoped `open_proof` marker** (merged into main and adopted wholesale in the integration merge): a single-use marker minted per open and written into the load payload proves THIS load landed — strictly stronger than this branch's original transition-proof re-stamp, which was dropped as superseded. The per-cause honest refusal is preserved (and improved) by `describeOpenRebindOutcome`.
- The binding refusal messages no longer say "the workflow reports 0 node(s), but the canvas is bound to a different graph" for a 0-node expectation, name the identity tag for what it is, keep the `[reason]` prefix, and name the panel reload as the certain fallback. (Coexists with #623's `root-mid-population` verdict, which merged cleanly alongside.)

**#607 — "workflow instance mismatch" with no path back**
- Server/Manager-scoped commands (`comfy_reboot`, `free_vram`, `nodes_search/list/install/queue_status`, `workflow_list`) no longer pass the workflow-instance fence at all — they never touch `app.graph`, and fencing them made `panel_restart_comfyui` itself refuse with the mismatch it was called to recover from. `refresh_nodes` is deliberately NOT exempt (it re-applies defs to live node instances — codex gate). Unknown commands still fail closed.
- A fence refusal now fires a throttled (5s) re-hello, re-advertising the panel's CURRENT workflow identity — the frame the orchestrator re-stamps its per-tab command cache from — so the advertised recovery (re-target, then retry) actually reaches the stale stamp instead of being a no-op.

## Already fixed upstream (verified, not re-fixed)

- **#560** (closed): the false-empty read guard, the dialect-normalized `graphRootMatchesState`, the both-empty heal (#560/#565/#621), and now the attempt-scoped open-rebind proof (#623).
- **#178** (closed): `tmp:` ids are WeakMap-stable per workflow object, the panel re-hellos with the fresh tab id on reconnect, and the current orchestrator rebinds on `panel_set_workflow_target({mode:"current"})`. Not reproducible panel-side on the merged tree.
- **#436**: the read-ok/write-refused flap was fixed orchestrator-side (`carryWorkflowCommandStamp`, which cites #436) — the command stamp is carried across the tab-id migration instead of being retired. The panel side (re-register via reconnect→hello) already exists. The two error strings quoted in the issue are orchestrator-side; nothing panel-side left that I could evidence.

## Deliberately NOT fixed

- **#607's "set_workflow_target({mode:'current'}) does not invalidate the cached stamp"** is orchestrator-side (comfyui-mcp). The current orchestrator already rebinds the tab and re-stamps from every hello's `workflow_uuid`; this PR's re-hello-on-refusal closes the panel half of the remaining hole (identity drift with no reconnect). No orchestrator change was made from this repo.
- **#606's "[reason] tag stripped before reaching the agent"**: the tag is present in the panel's thrown message (verified); if it does not reach the agent, the stripping is orchestrator-side and out of this repo.

- The end-to-end un-wedge: fresh `tmp:` tab + reconnect → `graph_outline` succeeds without a reload.

## Verification

- 2309/2309 unit tests pass (`npm run test:unit`) on the final integrated tree (this branch + #623/#624/#625–#629/#632); `node --check` as ES module clean; `tsc --noEmit` clean; CI lint green; GitHub reports MERGEABLE/CLEAN.
- Every test covering this branch's changes mutation-verified against the code it covers — on the integrated tree after the merge: stamp uuid, proven-empty gate, hook exactly-once/no-mask, fence exemption and its fail-closed default, refresh_nodes fenced, uuid-message branch — each break fails the matching test.
- Post-state grep confirms the new forms and the absence of the old ones; committed blobs scanned for control bytes (0).
- Adversarial self-gate (codex) on the branch's own diff: 5 rounds — refresh_nodes exemption (fixed), foreign-claimed re-stamp (fixed), unclaimed-copy hole (fixed), absent-tag/no-transition hole (fixed) — final round **PASS**. (The re-stamp those rounds hardened was later superseded by #623's marker proof in the integration merge; the gate findings are why dropping it for the stronger mechanism was safe.)
- Merge integrations: (1) #623/#625–#629/#632 — conflicts in `web/js/comfyui-mcp-panel.js` (workflow_open proof — resolved by adopting #623's marker proof) and `browser_tests/unit/graph-binding.test.mjs` (both appended suites kept). (2) #624 — conflicts in `web/js/lib/workflow-chat-identity.js` (this branch's NON_CANVAS_COMMANDS vs #624's CANVAS_INDEPENDENT_COMMANDS: adopted #624's set wholesale — `graph_update_node` exempt, `workflow_list` fenced under its sharper criterion; the #607 `comfy_reboot` deadlock case is preserved identically) and `web/js/lib/graph-binding.js` (kept this branch's #606 uuid-mismatch message, adopted #624's #601 dirty-mutation message; #624's `sealProvenRootBinding` sits complementarily after this branch's rebind heal). Lib auto-merges re-checked semantically (mid-population verdict + #606 messages coexist).

### DIFF
--- a/browser_tests/unit/binding-recovery.test.mjs
+++ b/browser_tests/unit/binding-recovery.test.mjs
@@ -1,0 +1,506 @@
+/**
+ * Unit tests for the binding-wedged-no-recovery cluster (panel #606 / #607, on the
+ * tree that absorbed #621):
+ *
+ *   #606 — a panel-created blank tab could wedge behind the binding guard after a
+ *     reconnect: ComfyUI reuses app.graph across tabs and clear/configure does NOT
+ *     reset graph.extra, so the new tab inherited the PREVIOUS workflow's root tag;
+ *     with its ChangeTracker not yet PROVEN empty the both-empty heal could not fire,
+ *     and nothing re-stamped the root. workflow_new now stamps the root at creation
+ *     (only when the root is PROVEN content-free), and workflow_open's repaint proof
+ *     re-stamps the tag directly once the content is proven (the tag riding
+ *     loadGraphData/configure is a serializer DIALECT, not workflow content).
+ *
+ *   #607 — a fence refusal ("workflow instance mismatch") meant the orchestrator's
+ *     cached stamp was stale relative to the panel's LIVE identity, yet the advertised
+ *     recovery never reached that cache: the panel re-hellos (the frame the
+ *     orchestrator re-stamps from) when the refusal fires.
+ *
+ * The harnesses extract the SHIPPING code from the panel monolith and drive it with
+ * injected doubles, so the tests are about the code that actually runs (delete the
+ * stamp / the re-stamp / the hook call and the matching test fails).
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+
+import {
+  graphRootProvenEmpty,
+  graphRootWorkflowUuidMatches,
+  graphRootMatchesState,
+} from "../../web/js/lib/graph-binding.js";
+import { commandTargetsActiveWorkflow } from "../../web/js/lib/workflow-chat-identity.js";
+
+const HERE = dirname(fileURLToPath(import.meta.url));
+const PANEL_JS = join(HERE, "../../web/js/comfyui-mcp-panel.js");
+const SRC = readFileSync(PANEL_JS, "utf8").replace(/\r\n/g, "\n");
+
+/** Balanced extraction starting at a marker's first "{", ignoring nothing fancy
+ *  (the extracted regions contain no template braces outside code). `openAt` skips
+ *  ahead when the marker itself contains braces (e.g. a `({ rid } = {})` param). */
+function balancedFrom(src, marker, openAt = null) {
+  const start = src.indexOf(marker);
+  assert.notEqual(start, -1, `missing marker: ${marker}`);
+  const open = openAt ?? src.indexOf("{", start + marker.length);
+  let depth = 0;
+  for (let i = open; i < src.length; i += 1) {
+    const ch = src[i];
+    if (ch === "/" && src[i + 1] === "/") {
+      i = src.indexOf("\n", i + 2);
+      if (i < 0) break;
+      continue;
+    }
+    if (ch === "/" && src[i + 1] === "*") {
+      i = src.indexOf("*/", i + 2);
+      if (i < 0) break;
+      i += 1;
+      continue;
+    }
+    if (ch === '"' || ch === "'" || ch === "`") {
+      const quote = ch;
+      for (i += 1; i < src.length; i += 1) {
+        if (src[i] === "\\") {
+          i += 1;
+          continue;
+        }
+        if (src[i] === quote) break;
+      }
+      continue;
+    }
+    if (ch === "{") depth += 1;
+    if (ch === "}" && --depth === 0) return src.slice(start, i + 1);
+  }
+  throw new Error(`unterminated block: ${marker}`);
+}
+
+// ---------------------------------------------------------------------------
+// #606 fix 1 — workflow_new stamps the root tag at creation (proven-empty gate)
+// ---------------------------------------------------------------------------
+
+function buildWorkflowNew({
+  rootGraph,
+  activeWorkflow,
+  stableUuid = "uuid-new-tab",
+  onStamp = () => {},
+} = {}) {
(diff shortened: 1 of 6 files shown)


artokun:
**Unblocked** — #621 merged as `7b1b27e`. The binding codepath (`graph-binding.js`, `subgraph-scope.js`, `getGraphCtx`/`resolveScope`, the mutation binding bar) is free.

Rebase on `main` before starting: #621 landed a reference-counted `acquireCanvasInteractionLock`/`releaseCanvasInteractionLock` pair with per-holder tokens, and a discriminated revert outcome (`restored`/`none`/`refused`/`failed`/`unknown`) in `lib/graph-revert.js`. If your cluster touches the canvas fence or renders a revert result, use those rather than reintroducing a bare boolean or a `null` that means two different things.

Two rules that came out of #621 the hard way and apply directly here:

artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 2 P0.** No P1, no P2. Bytes clean, no duplicate top-level declaration.

The diagnosis is right and the three-part shape (stamp at creation, re-stamp on proven repaint, re-hello on refusal) is the correct answer to a dead-end recovery. Both findings are in the *proofs* behind it.

### P0-1 — an unreadable comparison is treated as proof the root did NOT match, and that authorizes a stamp

`comfyui-mcp-panel.js:10010-10016, 10048-10053`. The catch returns `false`, and then:

```
loadTransitionProven = !rootMatchedBeforeLoad
```

So "we could not read it" and "it definitely did not match" become the same value — and the negation turns the unknown into a **positive licence to stamp**.

Concrete: an untagged, already-identical canvas belonging to a **different** workflow; its first serialization transiently fails; `loadGraphData` resolves without replacing it; the post-read succeeds and matches; the code stamps it with the **target UUID**. That blesses the wrong canvas and **reopens #349** — the hole this whole family exists to keep shut.

This is the repo's dominant fold in its most expensive position: not narrating a wrong cause, but *converting an unknown into an authorization*. `rootMatchedBeforeLoad` needs to be tri-state, and only an explicit "did not match" may license the stamp. An unreadable comparison must fall through to no-stamp — the dead end you are fixing is strictly better than a durable wrong binding.

### P0-2 — re-hello is unbounded, evidence-blind, and records its throttle before the send succeeds

`comfyui-mcp-panel.js:13533-13538`. Two problems compounding:

- **Unbounded and evidence-insensitive**: a persistent mismatch produces refusal → hello → same refusal → hello, forever at five-second spacing, even when each hello changes nothing. A recovery that retries indefinitely is a **new wedge**, not a fix — it just churns instead of sitting still.
- **It stamps the throttle timestamp before `sendHello()` has succeeded.** So if the first send never reaches the orchestrator, every retry inside the next five seconds is suppressed *on the strength of a hello that did not happen* — and recovery stays wedged while the code believes it is throttling a real attempt.

That second half is the fifth instance of one rule tonight, across both repos: **do not record that you did something before you did it.** The canvas interaction lock registered its holder before the freeze write succeeded (#621); the panel-op lock released a claim it no longer owned (#836); `agentSessionEpoch` advanced before its hello reached the wire (#633); and now this. It is worth treating as a checklist item on any state written next to an async call — *has the thing this records actually completed?*

Bound the re-hello (attempts, or stop when the evidence is unchanged), and move the throttle stamp to the success path.

### Note on merging
Verify on the actual merged tree — `main` has taken #621, #614, #643, #652, #653 and #649, all touching these files. A green check is only green for the `main` it ran against; that staleness took `main` red earlier tonight.

Issue #619, opened 2026-08-04
Intermittent root-shape-mismatch immediately after panel_enter_subgraph


artokun:
### What happened / root cause
Immediately after `panel_enter_subgraph(node_id:105)` returned success, the next `panel_graph_outline()` failed with `[root-shape-mismatch]`. The active workflow still reported the correct 9-node root, while the canvas had just navigated into a valid 16-node subgraph. This appears to be a transient scope/binding race around subgraph navigation or reconnect: the root semantic-shape guard evaluated before the live scope/tracker had settled.

### Steps to reproduce
1. Open a workflow containing a ComfyUI subgraph node.
2. Call `panel_graph_outline()` at root.
3. Call `panel_enter_subgraph({node_id:105})`; it returns success and reports subgraph scope.
4. Immediately call `panel_graph_outline()`.
5. Observed the error below once. After reconnect/re-opening the workflow, the same enter → outline → exit sequence succeeded, so the defect is intermittent.

### Exact error (scrubbed)
`Error: [root-shape-mismatch] The live graph is out of sync with the active workflow: the workflow reports 9 node(s), but the canvas is bound to a different graph. A load, tab switch, or reconnect left this command pointed at the wrong canvas, so it was NOT applied. Re-open the active workflow tab (panel_open_workflow) or reload the panel to rebind the graph, then retry.`

### Fix
Applied locally: no
Upstream-only: yes (intermittent and no longer reproducible after recovery; weakening the wrong-workflow mutation fence locally would be unsafe).

Precise change suggested: make the completion of `panel_enter_subgraph` establish a scope/binding receipt that the immediately following graph read can validate, or retry/reconcile the binding once when a root-shape mismatch occurs directly after a successful scope navigation. Do not globally disable the root-shape guard. A read-only workaround was `panel_get_subgraph(node_id)`, followed by re-opening/rebinding the active workflow.

### Environment
Windows 11; ComfyUI 0.30.0; NVIDIA GeForce RTX 5090 32 GB; Python 3.12.11; torch 2.13.0+cu130; comfyui-mcp 0.49.4; comfyui-mcp-panel 0.11.38.

Related (same root cause): #617, #618

artokun:
[...]

Issue #618, opened 2026-08-04
Post-reconnect graph outline omits existing unsaved root nodes until scope navigation


artokun:
### What happened / root cause
After a ComfyUI/sidebar restart and reconnect, the first `panel_graph_outline` for the active unsaved workflow returned a stale/incomplete root graph: 16 nodes, omitting an existing LoadImage node (id 280) and omitting its connected `reference_image` input on subgraph node 78. Entering/exiting the subgraph later rehydrated the root graph. A subsequent `panel_connect` reported `replaced_link: {node_id:"280", output:"IMAGE"}`, and `panel_query_graph` then showed node 280 had existed all along. The stale read caused creation of a duplicate LoadImage node.

This appears to be a reconnect/active-tab graph hydration race: read-only graph state was accepted before the frontend's unsaved graph was fully restored or before the binding snapshot was refreshed.

### Steps to reproduce
1. Open an unsaved workflow with a root LoadImage connected to an exposed input of a subgraph.
2. Restart ComfyUI/sidebar and let the agent reconnect.
3. Immediately call `panel_graph_outline`.
4. Observe the existing root node/input may be missing.
5. Enter then exit the subgraph, or connect another node to that exposed input.
6. Observe the missing node resurfaces; connecting reports it as the replaced link.

### Exact error (scrubbed)
No thrown error. Misleading successful output:
- initial outline: 16 nodes, node 280 absent
- later connect: `replaced_link.node_id = "280"`
- subsequent query: 18 nodes including both 280 and the duplicate 281

### Fix
applied locally: no
upstream-only: yes

Recovered the user's workflow by keeping the verified connected node and removing the disconnected duplicate. A durable fix should gate post-reconnect graph reads/mutations until the active tab's graph is fully hydrated and stable, or invalidate/reload the graph-binding snapshot before the first command after reconnect. `panel_graph_outline` / `panel_query_graph` should not return success until they reflect the same live root graph later used by mutations.

No local patch was attempted because the installed panel release worktree is broadly modified by the package/update state, and the fault spans reconnect/tab hydration timing rather than a defensible one-file change.

### Environment
Windows 11
ComfyUI 0.30.1
GPU: NVIDIA RTX 5080, 16 GB VRAM
comfyui-mcp 0.49.4
comfyui-mcp-panel 0.11.38 (verified from installed pyproject.toml after update; the session's startup ENV line still reported the pre-update 0.11.0 snapshot)

Related (same root cause): #616, #604

[...]

Issue #617, opened 2026-08-04
Frontend reconnect can loop on root-shape-mismatch for an unsaved workflow


artokun:
### What happened / root cause
After installing panel 0.11.38, restarting ComfyUI, and refreshing the frontend, the Agent panel repeatedly rebound to a fresh temporary tab id while graph commands remained associated with a different live canvas. Read-only graph_outline then failed with root-shape-mismatch even though workflow_list correctly showed Unsaved Workflow (2) active with 16 nodes. panel_open_workflow could not prove the rebind and advised another reload, which repeated the mismatch.

### Steps to reproduce
1. Keep an unsaved workflow tab open.
2. Restart ComfyUI after a panel update.
3. Reload the panel/frontend and call panel_set_workflow_target({mode:"current"}).
4. Call panel_graph_outline.
5. Observe root-shape-mismatch; panel_open_workflow on the active tmp key may also fail to prove rebind.

### Exact error (scrubbed)
"[root-shape-mismatch] The live graph is out of sync with the active workflow: the workflow reports 16 node(s), but the canvas is bound to a different graph."
"workflow_open could not prove that the active canvas was rebound to the requested workflow."

### Fix
Applied locally: workaround yes. Upstream-only: yes.
A full Comfy Desktop hard refresh followed by reopening the Agent sidebar and rebinding mode current restored graph_outline and subgraph editing. The precise upstream change needed is to invalidate stale temporary workflow routing keys on frontend reconnect and bind graph commands to the active canvas before reporting the panel connected.

### Environment
Windows 11; ComfyUI Desktop v1.0.34; ComfyUI 0.30.1; NVIDIA RTX 5080 16 GB; comfyui-mcp 0.49.4; panel 0.11.38.

artokun:
Half fixed: the 'could not prove rebound' false negative is fixed by the attempt-scoped open-proof marker (PR #653, e25e957) and the #618 mid-population fence (PR #623, b668ba6) covers the restore window. STILL OPEN: nothing invalidates stale tmp: routing keys on frontend reconnect (the 'reconnected' handler only bumps epoch/dialect cache), so the root cause of the rebind loop remains.

artokun:
Claimed: an agent is actively working on this issue in PR #677 (cluster/reconnect-recovery-hardening). Please coordinate before picking it up.

Issue #433, opened 2026-07-31
panel_list_workflows/panel_graph_outline report stale "active" workflow after ComfyUI restart, with no mismatch guard for the agent


artokun:
### What happened / root cause
During an agent session, ComfyUI restarted/reconnected mid-task (root cause of the restart itself unrelated — a background node-install flow). Before the drop, the user had switched their browser tab to workflow "homosimileXLPony_v60NAIXLEPS - Detailed Bulge - Thong 20260729 new". After the reconnect, `panel_graph_outline` and `panel_list_workflows` both kept reporting a *different* workflow ("New Template - Missing Node Packs") as the graph in scope / as `active`, for at least 3 consecutive calls, before a manual `panel_open_workflow` resynced it.

Traced the tool implementation: `workflow_list()` in `web/js/comfyui-mcp-panel.js` (~line 4811) reads directly from `app?.extensionManager?.workflow` (ComfyUI's own live frontend workflow service) with no independent state on the panel side:

```js
workflow_list() {
  const s = app?.extensionManager?.workflow;
  if (!s) throw new Error("workflow service unavailable on this frontend");
  ...
  active: active ? { path: active.path, filename: active.filename, key: active.key } : null,
  open: (s.openWorkflows ?? []).map(brief),
}
```

So the tool is faithfully reporting whatever ComfyUI's frontend currently considers `active` — meaning the actual bug (or at least contributing factor) is most likely upstream in how ComfyUI's own frontend restores/persists the active tab across a page reload triggered by a backend restart: it appears to have restored a *different* tab than the one the user had focused immediately before the restart (possibly falling back to the last-*saved* active tab rather than the last-*viewed* one, or a stale `localStorage`/session snapshot).

Separately, there IS a real gap on our side: nothing in `panel_list_workflows`/`panel_graph_outline` warns the calling agent that a reconnect just happened and the active-workflow pointer might not match what the user was last looking at. An agent that trusts `active` immediately after a reconnect notification can silently read/edit the wrong graph.

### Steps to reproduce
1. Open workflow A, then switch to workflow B in the browser tab strip.
2. Trigger a ComfyUI backend restart while an agent session is live (e.g. via a node install + restart flow).
3. Once the panel reconnects, call `panel_list_workflows` (or `panel_graph_outline`) from the agent.
4. Observe `active` may still point at workflow A (or some other previously-open tab) rather than workflow B, with no indication anything is off.

### Exact error (scrubbed)
[...]

PR #623, opened 2026-08-04
reconnect shape mismatch: a loop on unsaved workflows, an outline missing its own nodes, and a false "panel not installed"

artokun:
## Issues in this cluster

- artokun/comfyui-mcp-panel#619
- artokun/comfyui-mcp-panel#618
- artokun/comfyui-mcp-panel#617
- artokun/comfyui-mcp#888

## What landed

**#618 (and the reconnect half of artokun/comfyui-mcp-panel#617/#619): the mid-population fence.** After a ComfyUI reconnect the frontend restores the active tab's graph incrementally; for a window the live root canvas observably reads FEWER nodes than the active workflow's own current serialized state (the restore source). On a **dirty** tab every pre-existing binding predicate was blind to exactly that signature (shape guard deliberately inconclusive while dirty, artokun/comfyui-mcp-panel#545; baseline desync guard gated on a clean tracker; empty-read guard needs zero nodes, not a partial count) — so `panel_graph_outline` served an 8-of-31 mid-restore canvas as an authoritative outline, and the agent that trusted it duplicated a node it could not see and ran a whole-graph layout against the under-reported graph (#618's follow-up report).

The fix, in `web/js/lib/graph-binding.js` + one wiring point in the monolith:

- `graphRootMidPopulation()` fires only when ALL of: the artokun/comfyui-mcp-panel#433 epoch/monotonic post-reconnect window is open, ROOT scope, the workflow's CURRENT state (never the load baseline) is a well-formed read reporting N>0 nodes, and the live root reads strictly fewer than N.
- Wired into `resolveGraphBindingVerdict` as a new `root-mid-population` reason, gated on `includeBaselineReadGuard` (reads keep their lower bar at dispatch and re-assert in their executors, like the baseline desync guard) and ordered ahead of `root-shape-mismatch` inside the window — where a count-short canvas is far more often a still-restoring tab than a wrong one. UUID-conflict and dirty-mutation-binding verdicts still outrank it.
- The refusal discloses the uncertainty instead of narrating the bucket as a cause, states the command was NOT applied, and names remedies that work from where the caller is (retry once settled; `panel_open_workflow` if it persists).
- Outside the window, no pre-#618 behavior changes (mutation-verified against the existing suite, including the artokun/comfyui-mcp-panel#560 false-empty and artokun/comfyui-mcp-panel#545 dirty-tab tests).

Verification: 1993 unit tests pass (15 new). Delete-the-code check: removing the verdict wiring fails the 4 core tests. Mutation-verified (`<=` vs `<`, window-gate removal, monolith-wiring removal — each caught, two of them also by pre-existing artokun/comfyui-mcp-panel#560/#545 tests). Committed blobs scanned: 0 control bytes. Codex adversarial gate: **PASS**, no findings.

## Deliberately NOT fixed here

- **#613 (false "panel install did NOT land" on reconnect): the message is emitted by the ORCHESTRATOR repo** (`src/orchestrator/index.ts` ~L2715 wrapping `src/services/panel-installer.ts` ~L1986 in comfyui-mcp), not this repo — the panel only displays the orchestrator's `say` push. The suggested fix (re-run the authoritative disk scan before emitting the warning, suppress the stale failure when installed=true and version meets requiredPanelVersion) belongs in an orchestrator PR, adjacent to the artokun/comfyui-mcp-panel#629 manager-dialect-and-sync files. Nothing in this repo can suppress it.
- **#617's recovery loop** (`panel_open_workflow` could not prove the rebind on an unsaved tmp tab and advised a reload that repeated the mismatch): needs a live browser to pin down why the rebind proof fails for the re-minted tmp tab identity; the mid-population fence above removes the false-authoritative reads in the same window but does not claim to fix the loop itself.
- **#619's transient** (root-shape-mismatch immediately after a successful `panel_enter_subgraph`, observed once, not reproducible after recovery): the issue itself marks it intermittent and reconnect-adjacent. The fence converts the window's false-authoritative reads into honest refusals, but whether the specific enter→outline race still fires needs live confirmation.

That the post-reconnect restore of a dirty unsaved tab now surfaces `root-mid-population` refusals (instead of a partial outline) during the hydration window, and that the refusal clears once the tab settles.

### DIFF
--- a/browser_tests/unit/graph-binding.test.mjs
+++ b/browser_tests/unit/graph-binding.test.mjs
@@ -17,9 +17,11 @@ import { dirname, join } from "node:path";
 
 import {
   activeWorkflowNodeCount,
+  activeWorkflowCurrentNodeCount,
   activeWorkflowProvenEmpty,
   graphEmptyBindingUnproven,
   graphReadDesynced,
+  graphRootMidPopulation,
   graphRootMismatchesActiveWorkflow,
   graphRootProvenEmpty,
   graphRootWorkflowUuidMismatches,
@@ -123,6 +125,7 @@ function buildDirtyStaleRouteHarness({
   activeModified = true,
   activeTracker,
   rootSerializer = null,
+  postReconnectWindow = false,
 } = {}) {
   const src = readFileSync(PANEL_JS, "utf8").replace(/\r\n/g, "\n");
   const stableSource = panelFunctionSource(src, "workflowStableUuid", "workflowStorageKey");
@@ -268,6 +271,7 @@ function buildDirtyStaleRouteHarness({
     "workflowOwnsRootUuidTag",
     "rememberWorkflowUuidOwner",
     "resolveGraphRootUuidRebind",
+    "postReconnectSettleWindow",
     "WORKFLOW_META_NAMESPACE",
     "WORKFLOW_UUID_FIELD",
     `${fenceSource}\nreturn assertGraphBoundToActiveWorkflow;`,
@@ -283,6 +287,9 @@ function buildDirtyStaleRouteHarness({
     ownsTag,
     () => {},
     resolveGraphRootUuidRebind,
+    // #618 — the extracted fence reads the reconnect window through this hook;
+    // harnesses default to OUTSIDE the window and opt in explicitly.
+    () => postReconnectWindow === true,
     "comfyui_mcp",
     "workflow_uuid",
   );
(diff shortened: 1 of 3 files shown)


artokun:
**Unblocked** — artokun/comfyui-mcp-panel#621 merged as `7b1b27e`. The binding codepath (`graph-binding.js`, `subgraph-scope.js`, `getGraphCtx`/`resolveScope`, the mutation binding bar) is free.

Rebase on `main` before starting: artokun/comfyui-mcp-panel#621 landed a reference-counted `acquireCanvasInteractionLock`/`releaseCanvasInteractionLock` pair with per-holder tokens, and a discriminated revert outcome (`restored`/`none`/`refused`/`failed`/`unknown`) in `lib/graph-revert.js`. If your cluster touches the canvas fence or renders a revert result, use those rather than reintroducing a bare boolean or a `null` that means two different things.

Two rules that came out of artokun/comfyui-mcp-panel#621 the hard way and apply directly here:

artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 2 P0, 1 P1.** Bytes clean, no duplicate top-level declaration.

The diagnosis behind this PR is excellent — identifying that **every** pre-existing binding predicate was blind to the mid-restore signature on a *dirty* tab (shape guard inconclusive while dirty, baseline desync gated on a clean tracker, empty-read guard needing zero nodes rather than a partial count) is exactly the "enumerate by question asked" move. The findings are about how the new fence is *bounded*, not about that reasoning.

### P0-1 — an unreadable canvas is narrated as a known-empty one

`graph-binding.js:623` — `resolveGraphBindingVerdict` coerces an unreadable / non-finite `liveNodeCount` to **`0`** before calling `graphRootMidPopulation`, which **bypasses the predicate's own NaN fail-open rule**.

Concrete: after reconnect, 31-node active state, `liveNodeCount: NaN` or missing. The new guard returns `root-mid-population` and the message says the canvas *"shows 0 node(s)."* We never read the canvas — and we tell the user a specific number.

That is the repo's dominant fold in its most literal form, and it has a second cost: it can **wrongly refuse a legitimate command**, so the fold points at a positive here too. The predicate already had the right rule; the wrapper takes it away before it can apply.

### P0-2 — the protection expires on a TIMER, not on a proof

`graph-binding.js:628-650` with `comfyui-mcp-panel.js:4426`. The fence lifts solely when `postReconnectSettleWindow()` becomes false.

So a dirty tab whose restore **stalls permanently at 8 of 31 nodes** is protected only during that window; afterwards `graphRootMidPopulation` returns false, the dirty shape guard is still inconclusive, and **the 8-node outline is served as authoritative again** — the precise failure this PR exists to stop, on a delay. And `browser_tests/unit/graph-binding.test.mjs:2748` **codifies that as intended behaviour**, which is what makes it a P0 rather than an oversight: it is now specified.

The structural point, and it is the one worth taking from this: **a window is not a verdict.** It bounds how long we are willing to wait, never what became true. The fence should lift on a positive "the canvas settled / rebound" observation — the same principle artokun/comfyui-mcp-panel#653 is applying one repo over, where the open proof deliberately stays `unknown` rather than letting a deadline stand in for evidence. If no settled-proof is available, then the honest end state for a permanently stalled restore is a *continued* refusal that says the canvas never finished restoring, not a silent return to serving it.

### P1 — the NaN test does not exercise the deployed path

`browser_tests/unit/graph-binding.test.mjs:2937` calls `graphRootMidPopulation` **directly**, while production reaches it through `resolveGraphBindingVerdict`, which converts NaN to `0` first. So the test passes while the actual fence misclassifies unreadable input — it is testing a function the deployed code never calls in that state.

This is the standing check applied to a subtle case: if the code under test were deleted the test would fail, so it *looks* discriminating — but it never runs the path that ships. Drive it through `resolveGraphBindingVerdict`, and P0-1 turns red where it should.

### On merging
Verify on the actual merged tree — `main` has taken artokun/comfyui-mcp-panel#621, artokun/comfyui-mcp-panel#614, artokun/comfyui-mcp-panel#643, artokun/comfyui-mcp-panel#652 and the 0.11.39 release, all touching these files. A green check is only green for the `main` it ran against; that staleness took `main` red tonight.

Issue #611, opened 2026-08-04
panel_run rejects current panel when active tab bundle version is unknown


artokun:
### What happened / root cause
panel_run refused a valid live workflow because the active browser tab reported panel version/capability as unknown. Installed environment reports panel 0.11.38, indicating a stale cached frontend bundle. Read-only graph calls still succeeded.

### Steps to reproduce
1. Keep an older/cached panel tab open after updating the panel package.
2. panel_graph_outline succeeds.
3. panel_run({batch_count:10}) rejects the mutation.

### Exact error (scrubbed)
"graph_run" cannot be safely targeted to the active workflow: panel tab wf:workflows/3_Upscale_4xUltraSharp.json does not enforce per-command workflow targeting (detected panel version unknown; this MCP requires panel 0.11.35+).

### Fix
Applied locally: no. Upstream-only: yes for the already-running browser bundle.
Workaround verified: panel_strip_workflow followed by the durable batch tool queued and completed all 10 jobs.
Suggested change: version/invalidate the frontend bundle automatically or offer a one-click hard reload/reconnect action when backend version is current but tab capability is old/unknown.

### Environment
Windows 11; ComfyUI 0.30.0; RTX 3070 Ti 8 GB; comfyui-mcp 0.49.4; panel 0.11.38.

artokun:
Additional duplicate report: on a newly opened `Unsaved Workflow`, `panel_graph_outline` succeeded but `panel_add_node` for `VHS_LoadVideo` was rejected with the same unknown-version/per-command-targeting guard. Disk sync reported panel `0.11.38` installed and up to date. Environment: Windows 11, ComfyUI 0.28.2, RTX 4080 SUPER 16GB, comfyui-mcp 0.49.4, panel 0.11.38.

artokun:
Additional reproduction context: with `comfyui-mcp` 0.49.4 and panel 0.11.38 installed on disk, restarting/reconnecting ComfyUI or the agent while leaving the browser tab open can make the reconnected tab report its version/capability as unknown. Read-only graph commands and `panel_set_workflow_target` still work, but `panel_run` is safely rejected by the same workflow-targeting guard. A browser frontend reload restores operation. Environment: Windows 11, ComfyUI 0.30.0, RTX 3070 Ti.

artokun:
[...]

Issue #602, opened 2026-08-03
panel_search_nodes is incorrectly fenced to the active workflow UUID


artokun:
### What happened / root cause
`panel_search_nodes` was rejected after the active ComfyUI workflow changed. The frontend routes Manager registry discovery through `nodes_search`, but `activeWorkflowFenceApplies()` treated that non-canvas command like a graph command and compared its stamped workflow UUID to the active canvas UUID.

### Steps to reproduce
1. Open/switch to an unsaved workflow tab.
2. Call `panel_search_nodes({query:"face similarity"})`.
3. The Manager registry search is refused before it reaches `searchNodesVia`.

### Exact error (scrubbed)
```
Error: workflow instance mismatch: this command targets a different workflow than the active canvas — the workflow was switched or replaced after it was issued. Re-select the intended workflow (panel_open_workflow) or re-target with panel_set_workflow_target({mode:"current"}), then retry.
```

### Fix
applied locally: yes
upstream-only: no

```diff
--- a/web/js/lib/workflow-chat-identity.js
+++ b/web/js/lib/workflow-chat-identity.js
@@
  *  It must NOT fire only for:
+ *   • nodes_search: Manager registry discovery that never reads or mutates a canvas;
  *   • workflow_open / workflow_new: navigation/creation with their own explicit/new target;
@@
 export function activeWorkflowFenceApplies({ cmd, targetsNonActive = false } = {}) {
+  if (cmd === 'nodes_search') return false;
   if (cmd === 'workflow_open' || cmd === 'workflow_new') return false;
```

Verification:
- Imported the pure helper in Node.
- `activeWorkflowFenceApplies({cmd:"nodes_search"}) === false`.
- `activeWorkflowFenceApplies({cmd:"graph_get_state"}) === true`.
- Test exited 0.
- Frontend soft reload was scheduled. The immediate live retry saw the expected temporary “Connected: none” reload state, so the helper-level regression check is the verification available in this turn.

### Environment
macOS (darwin)
ComfyUI 0.29.0
GPU: Apple MPS, 16 GB unified memory
comfyui-mcp 0.48.18
panel 0.11.38

artokun:
[...]

Issue #600, opened 2026-08-03
Graph mutations blocked by unknown panel version and sync lock timeout


artokun:
### What happened / root cause
A live panel graph was readable, but every mutating graph_add_node request was rejected because the connected tab reported an unknown/too-old panel version and did not enforce per-command workflow targeting. The recommended install_panel(action='sync') recovery then timed out on the panel operation lock, so the workflow could not be edited.

### Steps to reproduce
1. Connect to a live sidebar panel with comfyui-mcp 0.49.3.
2. panel_graph_outline succeeds.
3. panel_add_node('LoadImagesFromFolderKJ') fails with the version/targeting guard.
4. install_panel(action='status') reports installed:false, decision:'sync', requiredPanelVersion 0.11.35.
5. install_panel(action='sync') waits 60 seconds, then errors on ~/.comfyui-mcp/panel-op.lock.

### Exact error (scrubbed)
graph_add_node could not be dispatched... panel tab does not enforce per-command workflow targeting (detected panel version unknown; this MCP requires panel 0.11.35+).
Timed out after 60000ms waiting for the panel operation lock (~/.comfyui-mcp/panel-op.lock). Another panel install/update/pin is in progress.

### Fix
applied locally: no
upstream-only: yes
The recovery path needs actionable lock-owner/status information or a safe proven-stale-lock recovery flow; the live panel should also expose its version so status and connected-tab state agree.

### Environment
Windows 11; ComfyUI 0.30.0; NVIDIA RTX 3070 Ti 8 GB; comfyui-mcp 0.49.3; panel version unknown.

artokun:
Additional recovery result: after confirming PID 36032 was gone and removing the abandoned `~/.comfyui-mcp/panel-op.lock`, `install_panel(action='sync')` no longer timed out but still did not install the pack. ComfyUI-Manager reported its queue drained, while MCP reported: “Panel install did NOT land: the pack is not present in custom_nodes after ComfyUI-Manager reported the queue drained.” The connected panel can list the live graph/custom nodes but status reports the panel absent, suggesting the recovery installer may be targeting a different saved/default workspace. Environment remains Windows 11, ComfyUI 0.30.0, comfyui-mcp 0.49.3; panel version is unknown.

artokun:
[...]

Issue #537, opened 2026-08-02
panel_enter_subgraph rejected: panel does not enforce per-command workflow targeting


artokun:
### What happened / root cause
The live-canvas panel_graph_outline read worked, but panel_enter_subgraph failed because the installed sidebar did not advertise/enforce per-command workflow targeting for graph_enter_subgraph. The orchestrator rejected the command rather than risk editing/navigating the wrong workflow tab.

### Steps to reproduce
1. Open a workflow containing a subgraph on the live ComfyUI canvas.
2. Call panel_graph_outline (works).
3. Call panel_enter_subgraph with the visible subgraph node id.
4. Observe dispatch rejection.

### Exact error (scrubbed)
graph_enter_subgraph could not be dispatched to this session's panel tab — nothing was applied. The tab may be disconnected, still reconnecting after a restart/reload, or the session's binding is stale. "graph_enter_subgraph" cannot be safely targeted to the active workflow: panel tab wf:workf does not enforce per-command workflow targeting (update the ComfyUI-MCP panel to edit the graph). Read-only graph commands still work.

### Fix
applied locally: no
upstream-only: yes
A safe fix needs the panel bridge to enforce the workflow target on graph_enter_subgraph (and ideally other navigation/edit commands), or negotiate the capability/version so the orchestrator can provide clear upgrade guidance. Current workaround: use read-only panel_query_graph / panel_graph_outline without entering the subgraph.

### Environment
Windows 11; ComfyUI 0.29.0; Python 3.12.12; torch 2.13.0+cu130; NVIDIA GeForce RTX 5060 Ti 16 GB; comfyui-mcp 0.49.0; panel 0.8.0.

artokun:
Stale panel version skew: report used 0.8.0. Current panel 0.11.34 advertises/enforces per-command workflow stamps and graph_enter_subgraph is covered; bridge correctly fails closed on old-panel hello. Panel contract tests 42/42 and MCP ui-bridge tests 122/122 pass. Update/reconnect the panel and retry.

artokun:
Duplicate recurrence of the version-skew guard in #537. The exact `does not enforce per-command workflow targeting` refusal identifies a panel build predating the capability shipped in panel 0.11.29; the panel version was not captured here, while current is 0.11.34.

[...]

PR #624, opened 2026-08-04
the write fence blocking legitimate writes: unknown bundle version, a stale JS reload, one mutation per page load

artokun:
## Issues in this cluster

- #611 — panel_run rejected: active tab bundle version unknown → **fixed** (stale-bundle self-heal, see below)
- #600 — mutations blocked by unknown panel version (+ sync lock timeout) → version side **fixed** here; the `panel-op.lock` timeout side is orchestrator-owned and landed separately
- #584 — frontend reload keeps stale panel JS → **fixed** (module-cache prime on reload paths)
- #601 — one mutation per page load behind a misleading node-count error → **fixed** (proven-clean binding seal + truthful message)
- #602 — panel_search_nodes fenced to the active workflow UUID → **fixed** (canvas-independent exemption)
- #537 — panel 0.8.0 lacks per-command targeting → **not reproducible on current panel**: the hello advertises `enforces_workflow_stamp` + `enforces_workflow_stamp_at_write`; the reporter was on panel 0.8.0 and the orchestrator's refusal already names the upgrade path (#709)

## What changed

**#601 — the dirty-tab deadlock.** After a reload, `app.graph` is rebuilt without the saved file's identity stamp; the first mutation dirties the tab, and every later mutation was refused as `dirty-mutation-binding-unproven` — one mutation per page load. `sealProvenRootBinding()` (lib/graph-binding.js) now stamps the active workflow's uuid onto an *unstamped* root only when the tab is clean, the root serializes **equal** to the workflow's own current state, *and* that proof is **exclusive** (no other clean open tab — an identical duplicate — carries the same state; codex gate). The `dirty-mutation-binding-unproven` refusal message now names the real cause and a working remedy instead of a misleading node count, and disambiguates the `multiple_active_tabs` sibling guard.

**#602 — canvas-independent commands.** The workflow-uuid fence, the dispatch-time graph-binding assert, and the pinned-target guard all exist to stop *canvas* writes landing on the wrong workflow. They now skip an explicit set of commands that never touch a canvas: `nodes_search`, `nodes_list`, `nodes_install`, `nodes_queue_status`, `graph_update_node` (a Manager pack update misnamed `graph_*`), `comfy_reboot`, `free_vram`. Everything unlisted stays fenced, fail-closed. (`refresh_nodes` was considered and deliberately kept fenced — it rewrites live node objects.) Also repaired a mojibake em dash in the workflow-instance-mismatch refusal.

**#584/#611 — stale-bundle self-heal.** A tab can run a cached older bundle after the pack updates (heuristic static caching survives restarts and plain reloads), so its hello advertises old/unknown capabilities and the orchestrator refuses every write. New `GET /comfyui_mcp_panel/version` answers the installed pack version (pyproject read at request time, strict UTF-8 + `[project]`-section parsing, `no-store`). At setup the panel awaits a bounded (2s) probe; only on a **proven** mismatch does it prime the pack's whole module graph (static *and* literal dynamic imports, `cache:"reload"`, in-pack scope, cycle-safe, bounded) and reload **once** — the sessionStorage loop guard is armed before the reload and read back (if it can't persist, no reload; hard-refresh remedy logged). Unreadable probe = "unknown" = never reload. The frontend-scope `panel_reload` primes the module cache too, time-bounded. Scope limit: this heals upgrades *from* a bundle containing this code; already-stale older bundles keep the hard-refresh remedy the orchestrator message names.

## Verification

- `npm run test:unit`: **1999 tests, 0 fail** (incl. new seal/twin/message/exemption/crawl/staleness tests; the seal is exercised through the real extracted `assertGraphBoundToActiveWorkflow`)
- `python -m unittest browser_tests.unit.test_pack_version`: 8 tests OK
- Every fix mutation-verified (seal removal, twin-exclusivity removal, message-branch removal, fence exemption/dispatch/pinned-guard skip removal, staleness verdict / cache-mode / crawl-scope / dynamic-import / minified-import mutants, version-reader + malformed-TOML + invalid-UTF-8 mutants, probe/prime timeout pins — each fails its matching test)
- Committed blobs scanned for control bytes: clean; pre-existing mojibake in the fence message repaired
- ES-module check (`node --check` as `.mjs`) passes
- Codex adversarial gate: 6 rounds, **PASS** (r1: twin-ambiguous seal, self-heal scope, dynamic imports; r2: pinned-guard skip, malformed-TOML scrape, unbounded prime; r3: sessionStorage-failure reload loop, non-strict UTF-8; r4: registration-before-probe race, minified imports, schema-invalid `project`; r5: hung `json()` body, `refresh_nodes` over-exemption, regex fallback section awareness — all addressed and re-verified)

Only a live browser can confirm: the end-to-end heal reload against a real ComfyUI cache, and that ComfyUI's session-restore path leaves the root unstamped in the #601 scenario (the seal is additive either way — it only ever fires on a strict, exclusive content proof).

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -493,6 +493,26 @@ def _register_routes():
     except Exception as _e:  # pragma: no cover - never block panel load
         _log("apps routes not registered: {}".format(_e))
 
+    @routes.get("/comfyui_mcp_panel/version")
+    async def _pack_version(_request):
+        # #584/#611 — the INSTALLED pack version, read from pyproject.toml at
+        # request time. The panel JS compares this against its running
+        # PANEL_VERSION: a mismatch proves the browser is running a CACHED stale
+        # bundle (a restart reconnects the tab but never re-downloads the
+        # extension JS), which is what leaves the tab advertising old/unknown
+        # capabilities to the orchestrator's write fence. no-store so the probe
+        # itself can never be served stale.
+        try:
+            from .py.pack_version import installed_pack_version
+
+            version = installed_pack_version()
+        except Exception:
+            version = None
+        return web.json_response(
+            {"version": version},
+            headers={"Cache-Control": "no-store"},
+        )
+
     @routes.get("/comfyui_mcp_panel/status")
     async def _status(_request):
         detected = _detect_comfyui_url()
--- a/browser_tests/unit/bundle-version.test.mjs
+++ b/browser_tests/unit/bundle-version.test.mjs
@@ -1,0 +1,206 @@
+/**
+ * Unit tests for #584/#611 — stale panel bundle detection + module-cache priming.
+ *
+ * The pack on disk can be current while a browser tab keeps running a CACHED
+ * older bundle (heuristic static-file caching survives restarts and plain
+ * reloads); the stale bundle then advertises old/unknown capabilities and the
+ * orchestrator's write fence refuses every mutation. These lock the verdict
+ * (never reload on an unreadable probe) and the cache priming crawl (visits the
+ * pack's whole static module graph with cache:"reload", stays inside the pack
+ * dir, bounded and cycle-safe).
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+
+import {
+  collectRelativeImportSpecifiers,
+  primeModuleCache,
+  resolveBundleStaleness,
+} from "../../web/js/lib/bundle-version.js";
+
+test("resolveBundleStaleness: equal versions are current", () => {
+  assert.equal(resolveBundleStaleness({ running: "0.11.39", installed: "0.11.39" }), "current");
+});
+
+test("resolveBundleStaleness: different versions are stale — upgrade AND downgrade", () => {
+  assert.equal(resolveBundleStaleness({ running: "0.11.34", installed: "0.11.39" }), "stale");
+  assert.equal(resolveBundleStaleness({ running: "0.11.39", installed: "0.11.34" }), "stale");
+});
+
+test("resolveBundleStaleness: a missing/malformed probe is UNKNOWN — never evidence for a reload", () => {
+  assert.equal(resolveBundleStaleness({ running: "0.11.39", installed: null }), "unknown");
+  assert.equal(resolveBundleStaleness({ running: "0.11.39", installed: undefined }), "unknown");
+  assert.equal(resolveBundleStaleness({ running: "0.11.39", installed: "" }), "unknown");
+  assert.equal(resolveBundleStaleness({ running: "0.11.39", installed: "   " }), "unknown");
+  assert.equal(resolveBundleStaleness({ running: "0.11.39", installed: 123 }), "unknown");
+  assert.equal(resolveBundleStaleness({ running: "", installed: "0.11.39" }), "unknown");
+  assert.equal(resolveBundleStaleness({}), "unknown");
+});
+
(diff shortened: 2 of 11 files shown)


artokun:
**Unblocked** — #621 merged as `7b1b27e`. The binding codepath (`graph-binding.js`, `subgraph-scope.js`, `getGraphCtx`/`resolveScope`, the mutation binding bar) is free.

Rebase on `main` before starting: #621 landed a reference-counted `acquireCanvasInteractionLock`/`releaseCanvasInteractionLock` pair with per-holder tokens, and a discriminated revert outcome (`restored`/`none`/`refused`/`failed`/`unknown`) in `lib/graph-revert.js`. If your cluster touches the canvas fence or renders a revert result, use those rather than reintroducing a bare boolean or a `null` that means two different things.

Two rules that came out of #621 the hard way and apply directly here:

artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 3 P0, 2 P1.** Bytes clean, no duplicate top-level declaration.

I pointed this gate at one question — **does every relaxation rest on a positive observation, or on the absence of a negative signal?** — and all three P0s are the second.

### P0-1 — `graph-binding.js:543`: `isModified !== true` is treated as proven clean

If the field is absent or unreadable while the tracker lags a dirty tab, `sealProvenRootBinding` stamps a root from stale state and **authorizes later dirty writes**. "Not observed to be modified" is not "observed to be unmodified". Require an explicit `isModified === false`.

### P0-2 — `comfyui-mcp-panel.js:4522`: exclusivity inferred from a list merely being an array

`sealProofExclusive = true` is set because `openWorkflows` **is an array** — never because that array was established as a complete, current inventory. It does not even require the list to contain `activeWorkflow`. So a transient or partial empty list, while an identical twin is open, permits sealing the ambiguous root.

**Absence of a twin in a list we never proved complete is not evidence of exclusivity.** This is the sharpest instance in the PR, because the whole seal rests on it.

### P0-3 — `comfyui-mcp-panel.js:24301`: the self-heal reloads after a PARTIAL prime

It destructures only `failed`, warns on failures, **ignores `truncated`, and navigates unconditionally.** A failed child fetch — or a graph exceeding `maxModules` — reloads a fresh entry module against stale dependencies, and the fresh `PANEL_VERSION` then advertises *current* over a mixed bundle.

That is a self-heal reporting success it did not observe, and it lands in the worst place: the version string is the very thing the fence trusts. Only reload when the crawl completes with **no failures and no truncation**; anything else is a heal that was attempted, not one that happened.

### P1-1 — the identical-twin remedy points at the wrong condition

`graph-binding.js:676`. A dirty unsealed root with an identical open twin returns `dirty-mutation-binding-unproven`, and the message points at **`multiple_active_tabs`** — a different condition. Reloading or reopening the active workflow leaves the twin present, so the seal stays ambiguous and the retry is refused again.

A dead-end remedy, in the cluster whose entire subject is dead-end remedies. It needs to name closing or disambiguating the identical workflow.

### P1-2 — the central relaxation is unverified

`bundle-version.test.mjs:24` tests only the pure staleness classifier. **Nothing proves the write fence still refuses a genuinely too-old panel**, and nothing proves a failed or truncated prime prevents the self-heal reload.

This is the test I asked for specifically when the gate was set up, and it is the one that matters most here: a loosening without a test that the tightening still holds is not verified, it is asserted. Add it before this lands.

### Note on merging
Verify on the merged tree — `main` has taken #621, #614, #643, #652, #653, #649, #633 and two releases, all touching these files. A green check is only green for the `main` it ran against; that staleness took `main` red earlier tonight.

Issue #612, opened 2026-08-04
panel_set_widget on a subgraph node reports a misleading "backend does not provide node type <uuid>" when the widget simply isn't promoted


artokun:
### What happened / root cause

Calling `panel_set_widget` on a **subgraph node** with a widget name that is NOT a promoted widget fails with an error that blames the node TYPE:

```
Cannot set widget on node 105 ("4c314f31-ecda-4b08-ae98-faaba1bf613f"): the ComfyUI
backend does not provide node type "4c314f31-ecda-4b08-ae98-faaba1bf613f" (not
installed, or its pack was removed) - refusing to write to a node the live backend no
longer defines (#458). Check the exact class_type via get_node_info.
```

That message is wrong and actively misleading. Nothing was uninstalled — the node is a perfectly healthy subgraph. The real situation is just "that widget isn't promoted on this subgraph node".

Root cause is in `web/js/lib/set-widget.js`, the promoted-target resolution around lines 96-131. Per its own comments:

* line ~103: `A truthy 'subgraph' field with NO promoted match is NOT a real promoted write`
* line ~127-130: `resolvedTargetNode = isResolvedPromotion ? promotedResolution.target.node : node`

So when the widget name doesn't match a promoted widget, `resolvedTargetNode` falls back to the subgraph node ITSELF, and `assertTypeAgainstFreshBackend` then authorizes that node's own `type` — which for a subgraph is a virtual UUID that is never in `/object_info`. The #458 fail-closed guard fires and reports it as a removed/uninstalled pack.

The guard behaviour is correct (fail closed); only the **diagnosis** is wrong. The subgraph-with-no-promoted-match case is a distinct, benign branch and should be reported as such.

### Steps to reproduce

1. Open a workflow containing a subgraph node (here: "Image to Video (MiniMax H3)", node 105).
2. Call `panel_set_widget(node_id=105, widget="control_after_generate", value="fixed")` — a widget that exists on an INNER node but is not promoted to the boundary.
3. Observe the "backend does not provide node type" error naming the subgraph UUID.

### Impact / why it matters

The message sent me diagnosing a phantom missing-custom-node problem. The real workaround (enter the subgraph, `panel_disconnect` the inner widget from its rail, set it, then reconnect afterwards) is several destructive-looking graph mutations to accomplish what should be one refused call with a clear reason. An agent that trusts the message would reasonably start reinstalling packs or restarting ComfyUI.

### Fix

[...]

Issue #583, opened 2026-08-03
panel_set_widget reports wrong previous value for promoted subgraph widgets


artokun:
### What happened / root cause
panel_set_widget successfully updates an exposed subgraph widget, but its response reports the promoted inner widget's previous value instead of the previous value visible on the requested outer subgraph widget. This makes the mutation result misleading and forces a follow-up panel_query_graph verification.

### Steps to reproduce
1. On a ComfyUI subgraph node, inspect an exposed numeric widget with panel_query_graph. In this case node 105 showed value_1: 15.
2. Call panel_set_widget({node_id:105, widget:"value_1", value:10}).
3. The response reports node_id 111, widget "value", previous:2, value:10, promoted_from.parent_widget_synced:true.
4. Query node 105 again; it correctly shows value_1:10.

### Exact output (scrubbed)
Before: outer node 105 value_1 = 15
Mutation response: {"node_id":"111","widget":"value","previous":2,"value":10,"promoted_from":{"subgraph_node_id":"105","inner_node_id":"111","parent_widget_synced":true}}
After: outer node 105 value_1 = 10

### Fix
applied locally: no
upstream-only: yes

Precise change needed: in the panel graph_set_widget subgraph-promotion path, capture the requested outer widget's previous value before mutating/syncing and return that as `previous`. If the promoted inner value is also useful, expose it separately as `inner_previous`; do not substitute it for the outer value the caller actually addressed.

### Environment
Windows 11
ComfyUI 0.30.0
NVIDIA GeForce RTX 5090 32 GB
comfyui-mcp 0.49.3
panel 0.11.38

artokun:
Fixed by PR #591 (merge b5d5acd): promoted writes report the OUTER rail's previous value with the inner value kept separately as inner_previous (web/js/lib/widget-write.js:1392-1396; test names #583 verbatim in widget-write.test.mjs). Shipped in panel 0.11.39.

Issue #512, opened 2026-08-01
panel_set_widget refuses promoted widgets on outer UUID subgraph nodes as 'unverifiable virtual-subgraph node' (#458 guard) though the write is correct


artokun:
Split out of #496, which is about the frontend-only-node allowlist and is being fixed in PR #511. This is a **different guard** and is not addressed there.

Reported twice on current versions (comfyui-mcp 0.48.32, panel 0.11.32, Windows 11, ComfyUI 0.29.0):

- Bundled `ltx-2.3-img2vid`, outer UUID subgraph node **320** (`2454ad83-157c-40dd-9f19-5daaf4041ce0`), promoted widget `value_4` — refused by the #458 outer-subgraph authorization even though `panel_enter_subgraph` can navigate the container. Reporter verified the intended write reaches inner node **301** and returns `parent_widget_synced: true`.
- Same node/graph, promoted `value` / `value_2` widgets — rejected as an "unverifiable virtual-subgraph node". After a frontend reload the intended local change updates inner node **319** and returns `parent_widget_synced: true`; wrapper values also reflect the update.

### Why this is its own bug

The refusal comes from `assertMutatedNodeAuthorized`'s **virtual-container branch**, not from the frontend-only-type allowlist that #496/#511 unify. The write is demonstrably correct — the reporter confirmed it lands on the inner node and syncs the parent rail — so the guard is failing closed on a legitimate mutation. That makes it a false refusal, the same shape as #496 and #507 but through a different oracle.

### What needs establishing

The open question needs the rig, not static reading: does ComfyUI's `SubgraphNode` carry backend provenance, or does it fail `isVirtualSubgraphContainer`? The guard's oracle is `/object_info`, and a virtual subgraph container is a frontend construct — so it may be unverifiable *by construction*, in which case the fix is to authorize it through a different signal (the container's navigability, or the resolved inner node's own authorization) rather than to weaken the container check.

### Constraint on the fix

Do **not** fix this by relaxing `assertMutatedNodeAuthorized` generally. #458's guard exists to stop writes landing on unresolved or wrong nodes — wrong-node writes are the corruption class this whole family of guards defends. The correct shape is to authorize the *resolved inner target*, which is verifiable, rather than to stop verifying the container.

Recommend verifying on the rig before implementing: enter the subgraph, confirm what provenance the container actually exposes, and confirm the inner node resolves and authorizes cleanly on its own.

artokun:
[...]

PR #625, opened 2026-08-04
panel_set_widget and the #458 guard: promoted subgraph widgets refused or mis-reported, and frontend-only node types failing closed

artokun:
## Issues in this cluster

- #612
- #583
- #512
- #496

## What changed

**#612 + the #512 recurrence — `web/js/lib/set-widget.js`.** A `panel_set_widget` call on a genuine subgraph node naming a widget that is **not promoted** no longer falls through to the #458 fresh-backend check on the node's own type (a subgraph UUID that is never in `/object_info`), which produced the misleading *"the ComfyUI backend does not provide node type \<uuid\> (not installed, or its pack was removed)"*. It now refuses up front with the honest diagnosis:

> Cannot set widget on subgraph node 78: "control_after_generate" is not a promoted widget on this subgraph (promoted: width, height, seed, …). Only promoted widgets are settable from outside a subgraph — panel_enter_subgraph and set it on the inner node directly, or promote it from inside the subgraph with panel_promote_widget first.

The refusal itself is unchanged (fail closed) — only the diagnosis. The ever-seen history gate is still consulted first (a container-shaped node whose type the backend reported earlier this session keeps the removed-type diagnosis), and a bare `subgraph:{}` marker that fails the container-shape check keeps the #458 fresh-auth path.

The #512 recurrence (krea2-txt2img-manual node 78: promoted `width` accepted, promoted `control_after_generate` refused with the UUID error) was root-caused against the shipped frontend (1.47.12 source maps): the pack's legacy `properties.proxyWidgets` entry `["75","control_after_generate"]` is **quarantined on load** by `proxyWidgetMigration.repairCreateSubgraphInput` because a canvas-only control widget has no connectable input slot — so the outer node genuinely has no such promotion. The refusal was correct; only its message was wrong. That is exactly this fix.

**#496 recurrence — `web/js/lib/node-resolve.js`.** KJNodes' frontend-only `SetNode`/`GetNode` (also reported with rgthree installed) added to `FRONTEND_ONLY_NODE_TYPES`, so the one shared predicate (`isAuthorizedFrontendOnlyType`) authorizes them for both `panel_add_node` and `panel_set_widget`. The guards are not weakened: live-registry membership, no backend provenance on class AND instance, and the non-forgeable ever-seen gate (applied first by every caller) still refuse squatters and mid-session removals — pinned by new negative tests.

**#583** was already fixed on main by #591 (`previous` now reports the requested outer rail widget's prior value, `inner_previous` exposes the inner value separately) and is covered by its regression test in `widget-write.test.mjs`; this PR closes the issue.

## Verification

- Full unit suite: **1782/1782 pass** (`node --test browser_tests/unit/*.test.mjs`).
- 8 new tests (5 in `set-widget-fresh-backend.test.mjs`, 3 in `node-resolve.test.mjs`).
- Delete-the-code checks: removing the new branch fails exactly the 3 tests about it; removing the allowlist entries fails exactly the 2 positive #496 tests; the negative pins stay green.
- Mutation-verified: dropping the `no-oracle` arm, dropping the container-shape check, dropping the verdict gate, and garbling the message are each caught (one also by the pre-existing #458 subgraph-shaped-bypass test).
- ES-module syntax checked as `.mjs`; committed blobs scanned for control bytes (0).
- Adversarial self-gate (codex, read-only): **PASS**, no findings.

## Deliberately not fixed here

- #612's two "related observations": (a) `panel_promote_widget` reporting `link-only` promotions as ordinary widgets (socket-only controls) — a different tool's reporting path; (b) `control_after_generate` surfaced under its current value as its key in `panel_query_graph` — a read-side/frontend-naming issue. Both are real but separate codepaths from this cluster's guard family.
- Live-browser confirmation that the krea2 node's `control_after_generate` is indeed quarantined (not visible on the boundary) is inferred from the frontend source + pack JSON, not observed on a running instance.

### DIFF
--- a/browser_tests/unit/node-resolve.test.mjs
+++ b/browser_tests/unit/node-resolve.test.mjs
@@ -1063,6 +1063,65 @@ test("#496 set_widget: MarkdownNote's text widget is writable end-to-end (the re
   assert.equal(node.widgets[0].value, "# README");
 });
 
+test("#496 recurrence: KJNodes' frontend-only SetNode/GetNode are ADDABLE on a healthy backend that never lists them", async () => {
+  // The reported recurrence (also seen with rgthree installed): SetNode/GetNode are
+  // registered purely by the pack's frontend JS — absent from /object_info BY DESIGN —
+  // and both panel_add_node and panel_set_widget refused them as a missing pack.
+  // They are on the shared allowlist now; the ever-seen gate + provenance checks
+  // below pin that this does not reopen the #458 hole.
+  const reg = loadedRegistry([], ["SetNode", "GetNode"]); // registered by pack JS, defless
+  const fresh = objectInfo(); // healthy backend — genuinely never lists these types
+  for (const t of ["SetNode", "GetNode"]) {
+    await assert.doesNotReject(
+      () => assertAddNodeResolvableRefreshing(() => reg, t, ADD_OPTS(fresh)),
+      `#496: "${t}" is frontend-only and must be addable`,
+    );
+  }
+});
+
+test("#496 recurrence: SetNode's widget is WRITABLE end-to-end via runSetWidget", async () => {
+  const reg = loadedRegistry([], ["SetNode"]);
+  const node = {
+    id: 32,
+    type: "SetNode",
+    widgets: [{ name: "Constant", type: "string", value: "" }],
+    constructor: reg["SetNode"],
+  };
+  const { set } = await runSetWidget(node, "Constant", "model", {
+    registry: reg,
+    getRegistry: () => reg,
+    getFreshObjectInfo: async () => objectInfo(), // healthy backend, no SetNode
+    wasTypeEverDefined: () => false,
+    ...HOOKS,
+  });
+  assert.equal(set.value, "model");
+  assert.equal(node.widgets[0].value, "model");
+});
+
+test("#496 recurrence: the SetNode/GetNode allowlist entries do NOT weaken the guards", async () => {
+  const fresh = objectInfo();
+  // (a) A class SQUATTING the allowlisted name but carrying backend provenance
+  //     (a real backend def, or a removed pack's unpurged class) is NOT frontend-only.
+  const squat = loadedRegistry(["SetNode"]); // registered WITH nodeData
+  await assert.rejects(
+    () => assertAddNodeResolvableRefreshing(() => squat, "SetNode", ADD_OPTS(fresh)),
+    /Unknown node type|does not provide/i,
+  );
+  // (b) The EVER-SEEN gate still wins: a "GetNode" the backend reported EARLIER this
+  //     session and no longer lists is a removed backend node — refused.
+  const reg = loadedRegistry([], ["GetNode"]);
+  await assert.rejects(
+    () => assertAddNodeResolvableRefreshing(() => reg, "GetNode", ADD_OPTS(fresh, (t) => t === "GetNode")),
+    /defined this node type earlier this session|removed/i,
+  );
+  // (c) An allowlisted name that is NOT registered in the live registry at all stays
+  //     refused (registry membership is also what createNode needs).
+  await assert.rejects(
+    () => assertAddNodeResolvableRefreshing(() => loadedRegistry(), "SetNode", ADD_OPTS(fresh)),
+    /Unknown node type|does not provide/i,
+  );
+});
+
 // The three guards whose oracle is /object_info. Each is reduced to a boolean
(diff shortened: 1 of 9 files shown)


artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **No P0. One P1 — and it needs fixing before merge rather than deferring, for the reason below.**

**The good result first:** both widenings of the #458 guard (#512's promoted outer-UUID widgets, #496's frontend-only allowlist) came back clean. Neither admits something genuinely unverifiable, and no path was found where a write reaches the wrong widget, node or graph, or where an applied write is reported as not-applied. That was the risk I was most worried about — two of these four issues *widen* a guard whose whole job is refusing writes to nodes it can't identify.

### P1 — `web/js/lib/set-widget.js:190`: `no-oracle` asserted as a definite verdict

The new branch treats **`no-oracle` as sufficient to assert "not a promoted widget"**. That contradicts `backendHistoryVerdict`'s own fail-closed contract: **no oracle establishes neither "never backend-defined" nor safe container identity.** It also bypasses the fresh `/object_info` result this call already fetched.

Concrete regression: a backend-defined, subgraph-capable node **added after the startup baseline**, with a direct own widget whose type *is* present in this call's `/object_info`, has `never-seen` history on first observation. The new branch refuses it as unpromoted **before** the existing fresh-type authorization can permit the valid direct write. Same false diagnosis for an unwired caller.

So the remedy handed to the user is actionable and **wrong** — it names a promotion problem that was never established.

This is the cluster's own defect class arriving in its fix: **absence of an oracle read as a positive finding.** #612 is the same shape ("the backend has no such node type" answering "is this widget promoted"), which is precisely what this PR set out to correct.

Use the fresh result you already have; where neither the oracle nor the fresh type settles it, say *undetermined* rather than *unpromoted*.

### Why this one doesn't get deferred

My standing policy is to merge on green and file non-data-loss P1s as follow-ups. I'm making the distinction explicit here: **a P1 that is a regression introduced by the PR gets fixed; a P1 that is a pre-existing gap gets filed.** Deferring the first means knowingly shipping a break in something that worked — a user who could set that widget yesterday cannot today, and the message tells them the wrong reason.

Mutation-verify it (a node absent from the startup baseline whose type is in the current `/object_info`, asserting the direct write is permitted), and assert the **reason**, not just the outcome — asserting the outcome alone passes through more than one path here.

artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 0 P0, 1 P1.** No P2, no duplicate top-level declaration.

**P0 came back clean, and that is the good news that matters here**: the loosened guard does not accept a write it cannot establish is safe, the `previous` value is not stale, and nothing reports a write it did not observe. Loosening a guard without opening a false-acceptance is the hard half of this cluster, and that half holds.

### P1 — the fix introduces a NEW false refusal (fix this rather than file it)

`web/js/lib/set-widget.js:190` treats **`no-oracle` as sufficient to assert the definite "not a promoted widget" diagnosis.** That contradicts `backendHistoryVerdict`'s own fail-closed contract: no oracle establishes *neither* "never backend-defined" *nor* safe container identity. It also bypasses the fresh result this call already fetched.

Concrete regression: a backend-defined virtual/subgraph-capable node added **after the startup baseline**, with a direct own widget whose type IS present in this call's `/object_info`, has `never-seen` history on that first observation. The new branch refuses it as unpromoted **before the existing fresh-type authorization can permit the valid direct write.** Same false diagnosis for an unwired caller.

So the user is refused a legitimate write, and handed a remedy that is actionable but **wrong** — which is worse than a bare refusal, because it sends them to fix something that is not broken.

This is the exact both-directions failure this cluster is about, reappearing inside its own fix: the issues were an over-strict check refusing something real, and the fix adds a new over-strict check refusing something real. **"Could not determine whether this is promoted" is not "determined it is not promoted."** `no-oracle` is the could-not-determine case by definition, so it must fall through to the fresh-type authorization rather than short-circuit to a definite negative.

I am asking for this to be fixed rather than filed because it is a regression **introduced by this PR**, not a pre-existing gap it declined to close. Add a test for the first-observation case specifically — a backend-defined node added after the baseline, direct own widget, type present in the current `/object_info` — and assert the write is **permitted, and permitted for the fresh-type reason**, not merely that it succeeded.

### On merging
Checks completed 19:4x and `main` has since taken #621, #614, #643 and the 0.11.39 release, all touching panel JS. **A green check is only green for the `main` it ran against** — that staleness took `main` red tonight. Verify on the actual merged tree.

Issue #620, opened 2026-08-04
panel_add_node falsely rejects nodes whose required inputs are link-only custom/MASK types ("Required custom widget X have not registered") — same nodes load fine via panel_load_workflow


artokun:
### What happened / root cause

While building a workflow node-by-node on the live canvas with `panel_add_node`, three node types were consistently refused, while `panel_load_workflow` with the SAME node types loaded them all onto the canvas perfectly. So the frontend widget registry was never actually broken — `graph_add_node`'s pre-creation guard is over-strict and its error text misleadingly tells the user to reload ComfyUI.

Refused:
- `LoadImage`, `LoadImageOutput` -> `Required custom widget "upload" did not initialize for "LoadImage". Reload ComfyUI so its node extension can register, then retry.`
- `SetLatentNoiseMask` -> `Required custom widget "MASK" have not registered. They may be custom widgets still loading; retry shortly.`
- `InpaintStitchImproved` (ComfyUI-Inpaint-CropAndStitch) -> `Required custom widget "STITCHER" have not registered. ...`

Succeeded in the same session: `CheckpointLoaderSimple`, `InpaintCropImproved` (optional MASK inputs, STITCHER as an *output*), `CLIPTextEncode`, `VAEEncode`, `KSampler`, `VAEDecode`. Pattern: only nodes with a REQUIRED link-only input of a type absent from a hardcoded allowlist, plus LoadImage's `upload` widget.

Two independent defects, both introduced by the #580/#593 guard (`42d7f76 fix(panel): materialize V3 custom widgets before add`):

**A. `unavailableRequiredCustomWidgetTypes` (`web/js/lib/node-widget-materialization.js`)** fails closed on any required input type not in the hardcoded `SAFE_SOCKET_TYPES` allowlist.
- `MASK` — a ComfyUI *core* datatype — is simply missing from that list. `SetLatentNoiseMask` (required `samples: LATENT`, `mask: MASK`) can therefore never be added.
- Worse, the design is unfixable-by-allowlist for third-party sockets: `STITCHER` is a link datatype from ComfyUI-Inpaint-CropAndStitch. No widget constructor will EVER register for it, so the 5s poll always times out and no retry / `panel_refresh_nodes` / `panel_reload` can ever clear it. Every custom pack that defines its own socket datatype and takes it as a *required* input is permanently un-addable via panel_add_node.

[...]

Issue #608, opened 2026-08-04
panel_add_node cannot add native SaveVideo: VIDEO widget not registered


artokun:
### What happened / root cause
On the live ComfyUI canvas, panel_add_node failed twice when adding the native SaveVideo node.

### Steps to reproduce
1. Open a workflow in the sidebar panel.
2. Call panel_add_node with class_type SaveVideo.
3. The operation fails before the node is added.

### Exact error (scrubbed)
Error: Required custom widget "VIDEO" have not registered. They may be custom widgets still loading; retry shortly.

### Fix / workaround
Applied locally: no. Upstream-only: likely panel node-registration/loading race or missing VIDEO widget metadata. Workaround: use CreateVideo plus the existing VHS_VideoCombine output, or load the official template.

### Environment
OS: Linux; ComfyUI 0.30.0; GPU: NVIDIA GeForce RTX 5090 31 GB; comfyui-mcp 0.49.4; panel 0.11.38.

Related (same root cause): #580

artokun:
Additional reproduction on the same current versions/environment: Linux, ComfyUI 0.30.0, NVIDIA RTX 5090 31 GB, comfyui-mcp 0.49.4, panel 0.11.38. `panel_add_node` fails for both `LoadImage` and `LoadImageOutput` before adding the node with:

`Required custom widget "upload" did not initialize for "LoadImage". Reload ComfyUI so its node extension can register, then retry.`

The suggested restart recovery was unavailable: `panel_restart_comfyui` refused because `COMFYUI_PATH` resolved to an install without `main.py` (`Resolved ComfyUI script does not exist on disk: main.py`), although the connected ComfyUI instance was running. Loading a UI workflow that already contains LoadImage nodes or restarting through the external launcher were the available workarounds.

artokun:
Fixed in PR #626 (merge 8958721): native SaveVideo's VIDEO widget resolves via the registry socket proof (registeredSocketTypes/inputDeclaredAsSocket in node-widget-materialization.js); test 'native VIDEO socket resolves via registry proof (#608 SaveVideo)'.

Issue #580, opened 2026-08-03
panel_add_node omits required custom widget values for V3 nodes


artokun:
### What happened / root cause
panel_add_node created a V3 custom node with UI-only custom inputs as bare required link slots and did not initialize/serialize their widget values. ComfyUI then rejected the graph with required_input_missing.

### Steps to reproduce
1. Install martin-rizzo/ComfyUI-ZImagePowerNodes v2.0.0.
2. Call panel_add_node with class_type "StylePromptEncoder2 //ZImagePowerNodes".
3. Set style/text, connect CLIP and CONDITIONING, then queue.
4. Validation fails for gallery and spacer.

### Exact error
Required input is missing — gallery
Required input is missing — spacer

panel_add_node returned inputs ZIPN_STYLE_GALLERY and ZIPN_SPACER with no widgets/default values. These are UI-only controls ignored by execute(**kwargs), not data connections.

### Fix
Applied locally: no. Upstream-only: yes.
panel_add_node should ensure frontend custom-widget conversion/default initialization has completed before returning and that the created node serializes required custom widget values. If the frontend extension cannot materialize a custom input widget, the tool should fail clearly instead of creating a graph that cannot validate.

### Environment
Windows 11; ComfyUI 0.30.1; RTX 5080 16GB; Python 3.13; comfyui-mcp 0.49.3; panel 0.11.38.

artokun:
Fixed in PR #626 (merge 8958721): node-widget-materialization.js fails closed when a required custom/V3 widget can't be registered, scanning the fresh /object_info def; 29 regression tests in browser_tests/unit/node-widget-materialization.test.mjs name #580/#608/#620.

PR #626, opened 2026-08-04
panel_add_node rejects nodes whose widgets it cannot register — link-only custom types, VIDEO, V3 required values

artokun:
## Issues in this cluster

- #620 — panel_add_node falsely rejects nodes whose required inputs are link-only custom/MASK types
- #608 — panel_add_node cannot add native SaveVideo: VIDEO widget not registered
- #580 — panel_add_node omits required custom widget values for V3 nodes (the guard this PR keeps intact)

## What changed

The #580 pre-creation guard failed closed on any required input type absent from a hardcoded allowlist — permanently, since no widget constructor ever registers for a link datatype. Both add-node guards now run off the **fresh /object_info entry** that graph_add_node already fetches for its resolvability assert, instead of the possibly-stale registered nodeData:

- **`unavailableRequiredCustomWidgetTypes`** admits a type when (a) a widget constructor is registered, (b) it is in the (extended) core allowlist — MASK, HOOKS, LATENT_OPERATION, MESH, VOXEL added, or (c) the **fresh backend defs prove it is a link datatype** (some current node declares it as an OUTPUT — STITCHER, VIDEO). A still-unknown type with no proof continues to fail closed, so #580's protection is intact.
- **`missingRequiredWidgetMaterializations`** scans the backend def, so frontend-injected inputs (LoadImage's `upload` button, serialize:false/canvasOnly:true) are never misread as a missing prompt value. A def present with no required inputs enforces nothing; a frontend-only type (no def) falls back to scanning the registered node data.
- **`driftedRequiredInputNames`** (new) refuses the add before createNode — with a reload remedy — when a pack upgraded mid-session added or **retyped** a required input on an already-registered class (shape signature: widget/socket type incl. combo choices + forceInput/force_input/defaultInput/widget:false; benign config like defaults deliberately not compared).
- `inputWidgetType` honours the raw snake_case `force_input` spelling from /object_info and treats `defaultInput` as a socket.

## Verification

- `node --test browser_tests/unit/*.test.mjs` → **1785 pass / 0 fail** (incl. 10 new tests in node-widget-materialization.test.mjs).
- Every clause mutation-verified (each semantic mutant fails exactly its matching test).
- `node --check` on the panel as .mjs passes; committed blobs control-byte-clean.
- 6 codex adversarial gate rounds → PASS.
- Pre-existing timer flake in manager-install.test.mjs (`waitForQueueDrain`, ~1 in 4 on this machine) is unrelated: the extracted function is byte-identical to base and that test file is untouched by this diff.

Live-browser confirmation still open: actually adding LoadImage / SetLatentNoiseMask / SaveVideo / InpaintStitchImproved on a live canvas.

### DIFF
--- a/browser_tests/unit/connect-match.test.mjs
+++ b/browser_tests/unit/connect-match.test.mjs
@@ -280,3 +280,105 @@ test("#406: a slot literally NAMED like a type still wins by name first", () => 
   const m = autoMatchSlots(origin, target, "IMAGE", "mask");
   assert.equal(m.outIdx, 0); // resolved by NAME to the MASK slot, not by type
 });
+
+// ---- #651: a WILDCARD socket accepts the type it is addressed by ------------
+//
+// Reported: `panel_connect(ImageCrop.IMAGE -> SetNode."IMAGE")` was refused with
+// "No input on node 225 accepts type IMAGE", while node 225's only input was a
+// wildcard `*` — a socket whose declaration is a POSITIVE statement that it accepts
+// anything, and which had carried an IMAGE from VHS_LoadVideo moments earlier.
+// The type-ref resolver compared type strings naively while the auto-matcher scored
+// with isTypeCompatible, so the two answered "does this slot accept IMAGE?"
+// differently. Reading `*` as "not IMAGE" is reading an unknown as a definite
+// negative — the mirror of the add path's over-permissive output-side inference,
+// and the reason both now go through ONE predicate.
+
+// Set/Get bus node: a single wildcard input, named "*" and typed "*".
+function setNodeWildcard() {
+  return { id: 225, type: "SetNode", inputs: [{ name: "*", type: "*", link: null }] };
+}
+function imageCrop() {
+  return { id: 324, type: "ImageCrop", outputs: [{ name: "IMAGE", type: "IMAGE" }] };
+}
+
+test("#651: a wildcard input is matched when addressed by a concrete TYPE name", () => {
+  assert.deepEqual(typeMatchedIndices(setNodeWildcard().inputs, "IMAGE"), [0]);
+  assert.deepEqual(typeMatchedIndices(setNodeWildcard().inputs, "LATENT"), [0]);
+});
+
+test("#651: the reported call now connects (IMAGE output -> existing SetNode wildcard)", () => {
+  const m = autoMatchSlots(imageCrop(), setNodeWildcard(), "IMAGE", "IMAGE");
+  assert.equal(m.outIdx, 0);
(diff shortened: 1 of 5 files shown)


artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 2 P0.** No P1, no P2, no duplicate top-level declaration.

Both are the risk that comes with loosening a refusal: **what now gets added isn't valid.** The old behaviour was a wrong refusal; the new behaviour is a node added in a state the backend will reject later, further from its cause.

### P0 — output-type compatibility is not proof that an input is link-only

`web/js/lib/node-widget-materialization.js:94` — `knownSocketTypes` waives registration for any input whose type appears as **some fresh definition's output**. That doesn't establish this particular input is socket-only: **ComfyUI's frontend explicitly supports converting widget inputs to links**, so widget-bearing inputs are link-compatible too.

Concrete: a V3 node requires `ACME_VALUE` as a custom widget, another node *outputs* `ACME_VALUE`. After the timeout this passes even though `comfyApp.widgets.ACME_VALUE` never registers — and the later materialization guard can't report it either, because it has no constructor. `panel_add_node` adds a node with **neither the required widget value nor a link**.

This is #620's own fold restated: "some node outputs this type" answering "does this input need a widget". The link-only decision has to come from an **input-level socket declaration**, not inferred from the output side.

### P0 — the node is constructed from stale `nodeData`, not `currentDef`

`node-widget-materialization.js:134` with `comfyui-mcp-panel.js:6875` — schema drift deliberately ignores `default`, `min` and `max`, and then `LG.createNode` builds from the **registered** `nodeData` rather than `currentDef`.

Concrete: a registered node's required `INT` moves from `{default: 1, min: 0, max: 10}` to `{default: 20, min: 20, max: 100}`. Both signatures are `INT`, so the add proceeds and creates **`1`** — out of range. The backend rejects it at queue time, which is exactly the "confusing failure far from its cause" outcome an add-time refusal exists to prevent.

And even when the stale value happens to remain valid, it is an **invented value in the user's graph** rather than the current definition's — which is #580's actual complaint.

### The shape
Both want the same correction: **decide from the current definition, and only waive a requirement on a positive statement that it is satisfiable.** Absence of a registered widget constructor is could-not-determine; a matching output type elsewhere in the graph is unrelated evidence.

Mutation-verify each — a required custom widget whose type is output by another node (assert the add still refuses or supplies the value), and a re-registered node whose `INT` range shifted (assert the created value comes from `currentDef`). Assert the **reason**, not just that an add succeeded or failed.

Issue #610, opened 2026-08-04
panel_get_errors keeps resolved models missing when object_info refresh exceeds 4 seconds


artokun:
### What happened / root cause
After two model downloads completed and were verified both on disk and in the connected ComfyUI model lists, panel_get_errors continued to report both loaders as missing. panel_refresh_nodes returned refreshed:true, but panel_get_errors still trusted the old Pinia candidates.

The forced /object_info + combo refresh takes about 14.5 seconds on this Windows install. GET_ERRORS_REFRESH_TIMEOUT_MS was 4000 ms, so graph_get_errors timed out its freshness verdict, set comboTrustedForQuery=false, and intentionally kept the stale missing-model candidates.

### Steps to reproduce
1. Open a workflow whose UNETLoader and LoraLoaderModelOnly reference absent files.
2. Download the files with download_model / download_civitai_model.
3. Confirm download_status says done and the connected ComfyUI model lists include both exact filenames.
4. Call panel_refresh_nodes, then panel_get_errors.
5. On an install where /object_info refresh exceeds 4 seconds, panel_get_errors still reports both files missing.

### Exact error (scrubbed)
panel_get_errors returned missing_models for krea2_turbo_fp8_scaled.safetensors and krea2_identity_edit_v1_2_r64.safetensors even though download_status reported verified placement and ComfyUI listed both files.

### Fix
applied locally: yes
upstream-only: no

```diff
-const GET_ERRORS_REFRESH_TIMEOUT_MS = 4000;
+const GET_ERRORS_REFRESH_TIMEOUT_MS = 20000;
```

After a frontend panel reload, panel_get_errors took about 13 seconds and correctly removed both missing_models entries. Only the genuinely missing input image and node types remained.

### Environment
Windows 11
ComfyUI 0.30.1
NVIDIA RTX 3060 12 GB
comfyui-mcp 0.49.4
comfyui-mcp-panel 0.11.38

Related (same root cause): #185, #569

artokun:
Fixed in PR #627 (merge 951d802): the /object_info refresh cap is now 15000ms (GET_ERRORS_REFRESH_CAP_MS, web/js/lib/get-errors-budget.js:44, citing the ~14.5s measurement), inside a shared 18s total budget; tests cover the old-4s-constant-gone and the 14.5s case.

artokun:
[...]

Issue #589, opened 2026-08-03
panel_get_errors times out on a live responsive workflow tab


artokun:
Calling panel_get_errors on the active workflow `LTX23_expression_12s_dance` timed out after 20 seconds with `Panel tab wf:workf did not reply to graph_get_errors`, while panel_graph_outline called at the same time completed successfully and returned the full 23-node graph. The tab was active and responsive, so the timeout appears specific to graph_get_errors rather than a disconnected/frozen panel.

This forced diagnosis to rely on the already-injected missing-asset block and graph outline instead of the error surface.

Environment: Windows 11, ComfyUI 0.29.0, comfyui-mcp 0.49.0, comfyui-mcp-panel 0.11.38.

artokun:
Fixed in PR #627 (merge 951d802): all elective waits inside one graph_get_errors call share a single 18s budget (web/js/lib/get-errors-budget.js), provably under the orchestrator's 20s ui-bridge read timeout; steps that find no budget fail closed. Note: the original 20s overrun was never reproducible from code — if it recurs, the cause is main-thread blockage, not budget math.

Issue #586, opened 2026-08-03
panel_get_errors keeps stale missing_media after LoadImage widget changes


artokun:
### What happened / root cause
panel_get_errors returned contradictory state for one live LoadImage node after the user changed its image widget. The live node summary and validation error correctly reported image=helloe.png / Invalid image file: helloe.png, but missing_media still reported the previous workflow value 20260707_130607992_iOS.jpg. graph_get_errors appears to merge a stale load-time missing-media cache with current graph and validation state without reconciling node_id + widget + current widget value.

### Steps to reproduce
1. Open a workflow with LoadImage node 2129 referencing a missing image 20260707_130607992_iOS.jpg.
2. Change the node image widget to helloe.png, which is also absent/invalid on the remote ComfyUI input server.
3. Queue once so validation records Invalid image file: helloe.png.
4. Call panel_get_errors.
5. Observe nodes[0].widgets.image=helloe.png and validation details for helloe.png, while nodes[0].reasons.missing_media and top-level missing_media still name 20260707_130607992_iOS.jpg.

### Exact error (scrubbed)
Live widget: helloe.png
Validation: image - Invalid image file: helloe.png
Stale missing_media: 20260707_130607992_iOS.jpg

### Fix
Applied locally: no
Upstream-only: yes
Precise change needed: when graph_get_errors composes missing_media, re-read the live widget value and discard or recompute any cached missing-media record whose file no longer equals that node's current widget value. Invalidate the node's missing-media/red-outline cache on image widget changes and successful uploads.

### Environment
OS Linux remote
ComfyUI 0.30.0
GPU NVIDIA GeForce RTX 4090 24 GB
comfyui-mcp 0.49.3
comfyui-mcp-panel commit 55535ce5b95c3f7519fded92ce47d644a1568604

artokun:
Fixed in PR #627 (merge 951d802): assetCandidateStillReferenced (web/js/lib/asset-staleness.js:166) cross-checks the live widget value; regression test mirrors the exact #586 repro (node 2129, iOS.jpg -> helloe.png).

PR #627, opened 2026-08-04
panel_get_errors: stale after a widget change, resolved models still missing, and a timeout on a live tab

artokun:
## Issues in this cluster

- #610
- #589
- #586

**Not yet started** — draft until an agent picks it up.

### DIFF
--- a/browser_tests/unit/asset-staleness.test.mjs
+++ b/browser_tests/unit/asset-staleness.test.mjs
@@ -229,6 +229,61 @@ test("assetCandidateStillReferenced fails OPEN when the node is gone", () => {
   assert.equal(assetCandidateStillReferenced(graphOf([]), 99, "x.safetensors"), true);
 });
 
+// ---- #586: stale missing_media after a LoadImage widget change ----
+// The missingMedia Pinia store is populated at workflow LOAD; after the user
+// repoints the node's image widget to a different file, the load-time candidate
+// for the OLD value must no longer be reported. The drop must come from the
+// WIDGET-REFERENCE cross-check — so pin it with trustCombo:false (the combo
+// clear path inert) and a node that provably still exists (the scope-drop path
+// inert). Mirroring the issue's repro: node 2129, load-time value
+// 20260707_130607992_iOS.jpg, widget changed to helloe.png.
+test("isStaleAssetCandidate drops a missing-MEDIA candidate whose LoadImage widget moved to a new file (#586)", () => {
+  const node = {
+    id: 2129,
+    type: "LoadImage",
+    widgets: [{ name: "image", value: "helloe.png", options: { values: ["helloe.png"] } }],
+  };
+  const candidate = {
+    nodeId: "2129",
+    name: "20260707_130607992_iOS.jpg",
+    widgetName: "image",
+    mediaType: "image",
+    isMissing: true,
+  };
+  // Assert the REASON, not just the verdict: with no combo trust and the node
+  // present, only the still-referenced check can drop it.
+  assert.equal(
+    assetCandidateStillReferenced(graphOf([node]), "2129", "20260707_130607992_iOS.jpg"),
+    false,
+    "no widget still holds the old file",
+  );
+  assert.equal(
+    isStaleAssetCandidate(graphOf([node]), candidate, { trustCombo: false }),
+    true,
+  );
+});
+
+test("isStaleAssetCandidate KEEPS the media candidate while the widget still holds the missing file (#586 control)", () => {
+  const node = {
+    id: 2129,
+    type: "LoadImage",
+    widgets: [{ name: "image", value: "helloe.png", options: { values: ["helloe.png"] } }],
+  };
+  const candidate = {
+    nodeId: "2129",
+    name: "helloe.png",
+    widgetName: "image",
+    mediaType: "image",
+    isMissing: true,
+  };
+  // trustCombo stays false: a genuinely missing current value must survive even
+  // though the combo lists it (an untrusted combo may be a stale snapshot).
+  assert.equal(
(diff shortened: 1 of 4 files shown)


artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 2 P0, 1 P1.**

Widget invalidation (#586) and routine healthy-workflow refreshes came back clean, and there are no P2s. But all three findings are the same thing: **the threshold moved; what happens at the threshold did not.**

### P0 — the refresh timeout still fabricates `missing_models` (#610 is not fixed)

`web/js/comfyui-mcp-panel.js:8421` — budget expiry makes `withRefreshTimeout` return `false`, and `collectMissingAssets(false)` then emits load-time candidates through the **definite** `missing_models` field.

`/object_info` plus combo refresh taking 15.1 s — or a 14.5 s fetch plus local registration work — reports a model that **is present** as missing, and the agent goes and re-downloads it.

That is #610 with a larger number. The defect was never the size of the window; it was that **a refresh which did not finish yields a positive finding.** At expiry the answer has to become *undetermined*, carrying what was actually confirmed, not a `missing` verdict for everything unreached.

### P0 — budget exhaustion fabricates `missing_media` without probing

`:8493` skips `filterServerConfirmedInputSubfolderMedia` when the shared budget is gone, and the new early returns at `:5204` **retain every raw candidate**.

So a slow model refresh consumes the budget, and a stale nested or `[output]` LoadImage candidate — pointing at a file that exists — is returned as definite `missing_media`. It cannot be cleared by the input combo and was never checked with `/view`. Nothing observed it; the verdict is asserted anyway.

Note this one is worse than #610 because the *skip itself* is the cause: the check that would have exonerated the file is the thing the budget dropped.

### P1 — #589 is a deadline increase, not timeout enforcement

The budget starts only once `graph_get_errors` is already executing (`:8417`) and permits 18 s of waits (`web/js/lib/get-errors-budget.js:36`) under a 20 s bridge deadline. It excludes bridge dispatch, timer delay and throttling, synchronous refresh work, result construction, and reply delivery.

**`Promise.race` cannot preempt synchronous work** — so a live tab stalled by a >2 s long task or a throttled timer still blows the bridge deadline. This is accounting, not enforcement, and it will fail on slower machines exactly where it matters.

That is fine *if it is stated*. What is not fine is the PR reading as though #589 is closed. Either budget from bridge dispatch rather than handler entry, or say plainly that the deadline was widened and the synchronous tail remains unbounded.

### The shape to fix
All three want the same change: **at budget expiry, the result becomes `undetermined` with whatever was confirmed — never a definite `missing`.** A caller can act on "I could not verify these four in time." A caller cannot recover from being told a present model is absent, because the remedy it will choose is a re-download.

Mutation-verify by driving the budget to expiry with a **present** model and a **present** media file, asserting neither is reported missing and the status says verification was incomplete. Assert on the reason, not the field's emptiness — an empty `missing_models` passes for more than one cause.

artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 3 P0, 2 P1.**

All three P0s are the same fold this cluster exists to fix, still present after the fix: **an incomplete observation reported as a definite negative.** The budget machinery is the right idea; what it currently does on expiry is assert absence it never observed.

### P0-1 — a timed-out `/object_info` refresh is emitted as a definite missing model
`comfyui-mcp-panel.js:8433`. When a load-time candidate is now installed but the forced refresh exceeds `refreshBudgetMs`, `withRefreshTimeout` returns `false`, `comboTrustedForQuery` stays false, and the raw candidate is returned as `missing_models`. We observed **an incomplete refresh**, not an absence — and the remedy handed to the user is a download for a model they already have. That is worse than silence: it is non-actionable *and* confidently wrong.

### P0-2 — a timed-out media verification is returned as definite `missing_media`
`:5199`. A stale nested/annotated candidate needs `/view` confirmation; the preceding refresh consumes most of the shared budget; `/system_stats` takes the rest; `probeBudget` hits zero and the helper returns `media` **without ever observing `/view`**. Same shape, different probe.

For both: expiry needs a third state. "Could not verify within the budget" is not "missing" — it needs its own verdict, its own wording, and a remedy that is actually actionable (raise the budget / retry), rather than borrowing the missing-file remedy.

### P0-3 — the widget-change race is still open
`:8495`. `assets.media` is collected **before** the await; the code then awaits server probing and afterwards verifies only workflow/root-graph identity. Change the LoadImage widget after collection but before the await completes, and the **old candidate is returned** on the same workflow and root graph — so the identity check passes and cannot catch it. That is the stale `missing_media` in the issue, unchanged, and the new static helper test does not cover it. The identity check answers "is this the same graph?", not "is this candidate still what the widget says?" — a value computed for one purpose read as the answer to a different question.

### P1-1 — `withRefreshTimeout` never clears its timer
`:8433`. No `clearTimeout` after the refresh settles, so a series of fast `graph_get_errors` calls accumulates one later-firing timer each. The callbacks are inert once `Promise.race` settles, so this is not a correctness bug today — but a guard's teardown must be unconditional, or the next change to what that callback does turns it into one.

### P1-2 — the wiring test is syntactic and does not discriminate
`browser_tests/unit/get-errors-budget.test.mjs:113`. Replace `withRefreshTimeout`'s implementation with a hard-coded 4-second timer — still accepting and still passed `refreshBudgetMs` — and **every assertion stays green.** It therefore does not protect the #610 behaviour it claims to. Apply the standing check to each new test here: if the code under test were deleted or neutered, would this test fail? Assert the reason, not the shape of the call.

### On merging
Checks completed 19:44 and `main` has since taken #621, #614, #643 and the 0.11.39 release, all touching panel JS. **A green check is only green for the `main` it ran against.** Verify on the actual merged tree.

Issue #609, opened 2026-08-04
Blind mode: storyboard note still instructs the agent to "review motion, sharpness, and temporal consistency" after images were withheld — agent confabulates a visual verdict


artokun:
### What happened / root cause

With the panel **Blind** toggle ON (issue #90), a finished video render delivered a storyboard notification to the agent that contained **no image data** — the gate worked — but whose text still ordered the agent to perform a visual review it could not possibly perform. The agent complied and invented a verdict.

The exact message received (text only, no attachments):

```
[panel event] 📽️ 20-frame storyboard (contact sheet) of the FINAL saved video
(file MiniMax_H3_00001_.mp4 — reference THIS filename) — frames run
top-left→bottom-right = start→end. Review motion, sharpness, and temporal consistency.
• MiniMax_H3_00001_.mp4 — path: video/MiniMax_H3_00001_.mp4 · 20-frame storyboard
· 1.6 MB · rendered in 5m 11s · finished 1:50:19 AM
```

The agent replied: *"The 10s clip rendered clean in 5m 11s — nice result for a first pass on that model."* — a claim about **motion and image quality of a video it never saw**. The user then had to ask "you can see that even though i put blind mode on?" to discover the assessment was fabricated. That is a trust-destroying failure mode: blind mode users are exactly the users who most need to know the agent is not looking.

**Root cause.** `panel-agent.ts` already guards this for the DEFAULT wording — it appends the "attached below" sentence only when images exist AND the backend has vision:

```ts
(imgs.length
  ? this.backend.capabilities.vision
    ? `The image(s) are attached below and already shown to the user in the panel. `
    : `You cannot view images on this provider, but they are already shown to the user in the panel. `
  : ``) +
```

with the comment: *"a text-only backend told 'attached below' would confabulate having viewed the render."* The concern is understood and handled — **for the default text**.

But a panel-supplied `note` **replaces** the default wording entirely (`panel-agent.ts:421-429`, "A custom `note` (e.g. the panel's video-storyboard summary) replaces the default image-acknowledgement wording"), and the panel's storyboard note hardcodes `Review motion, sharpness, and temporal consistency.` unconditionally. So the note bypasses the very guard that exists to prevent confabulation. The `imgs.length === 0` branch correctly suppresses "attached below" — and then the note asks for a visual review two sentences earlier.

### Steps to reproduce

1. Turn ON the panel's Blind toggle for a tab.
[...]

Issue #588, opened 2026-08-03
Repeated successful reconnects flood chat with waiting-for-agent messages


artokun:
The panel appends a durable chat transcript entry every time the bridge WebSocket reconnects successfully but the panel agent is not yet ready. During ComfyUI/orchestrator restarts this produces many repeated `Connected to ws://127.0.0.1:9181/ — waiting for the panel agent…` messages, even though connection recovery is working normally.

Root cause: `createBridgeClient` used `onLog(...)` in the WebSocket open handler, and `markConnected` reset `loggedWaiting`, allowing every reconnect cycle to append the same message again. The status pill already communicates this transient state.

Local fix: removed the `loggedWaiting` bookkeeping and the `onLog` transcript write, retaining the status-pill update only.

Verification: `npm.cmd run test:unit` completed with 1698 tests, 1695 passed, 0 failed, 3 skipped.

Environment: Windows 11, ComfyUI 0.29.0, comfyui-mcp 0.49.0, comfyui-mcp-panel 0.11.38.

artokun:
Fixed in PR #628 (merge c87f173): the transcript onLog write and loggedWaiting throttle were removed from the WS open handler; browser_tests/unit/bridge-reconnect-transcript.test.mjs. The status pill still shows transient connecting state.

PR #628, opened 2026-08-04
turn lifecycle told untruthfully: a stall reported as a user rejection, an unmatched result, reconnect flood, and blind mode asking for a verdict on withheld images

artokun:
## Issues in this cluster

- artokun/comfyui-mcp#886
- artokun/comfyui-mcp#885
- artokun/comfyui-mcp-panel#588
- artokun/comfyui-mcp-panel#609

## Outcome per issue

**#588 (reconnect flood) — FIXED here.** Every successful bridge reconnect reset the `loggedWaiting` throttle (`markConnected`/`start()`/`setUrl()`), so the transient `Connected to … — waiting for the panel agent…` line was appended to the **durable** chat transcript once per reconnect cycle. The transcript write and the throttle bookkeeping are removed; the status pill still shows the transient `connecting` state, and a genuinely wedged open socket keeps its distinct handshake-timeout warning (`⚠ Bridge open … no panel agent responded`).

**#609 (blind mode asked for a visual verdict) — FIXED here.** The video-storyboard note hardcoded `Review motion, sharpness, and temporal consistency.` even when Blind mode strips the storyboard pixels at the `sendFrame` gate. `buildVideoSegment` now returns both note variants and `composeRunCompletionFrame` picks **one per frame** via an injected `agentReceivesImages()` (wired at the call site to `() => !AGENT_BLIND`), evaluated after every segment resolves and immediately before send — so parallel video segments can never disagree with each other or with the gate when the toggle flips mid-composition. When withheld, the note says so affirmatively (`Blind mode is ON, so the storyboard was NOT sent to you … Do not comment on motion, sharpness, or visual quality`) and keeps the factual metadata. The storyboard is still built and painted for the user.

**#587 (stall reported to the agent as a USER REJECTION) — orchestrator-side; partially fixed upstream, NOT in this repo.** The stall watchdog and the agent-visible tool result live in comfyui-mcp (`src/orchestrator/panel-agent.ts`, backends). Upstream fix `8bc668a` (comfyui-mcp PR #782, released in **v0.49.4** — the issue was filed against 0.49.3) sends a Codex turn an explicit harness-stall notice (`turn/steer`: "The user did NOT reject or cancel this tool call …") before falling back to interrupt. **Residual gap:** `recoverStalledTurn` is implemented only by the Codex backend; a stalled **Claude** turn still goes straight to `interrupt()`, and the issue's exact repro was a Claude turn. Closing the Claude half requires an orchestrator change, which is out of this repo's scope — flagged here so it isn't lost.

**#590 (result matched no submitted turn) — orchestrator-side; working as designed upstream, NOT in this repo.** The quoted message is emitted by comfyui-mcp `claude-backend.ts` (`The result could not be matched to any submitted turn — reporting the turn as unverifiable rather than a success`, plus the `Nothing was lost — try again …` wrapper from `panel-agent.ts`). That fail-closed guard (the #740/#745 work) was already present in the reported version 0.49.3 — the message the user saw IS that guard firing. The issue's enhancement asks (turn registry surviving reconnects, reconciliation by stable turn id, surfacing "completed but unverified" with tool activity) are all orchestrator-state changes with no panel-side component. One upstream nit worth noting: `Nothing was lost` is questionable when the unmatched turn had in fact done real work.

## Verification

- `npm run test:unit` — 1783 passed, 0 failed (includes new `browser_tests/unit/bridge-reconnect-transcript.test.mjs` and +5 tests in `run-completion.test.mjs`).
- Both edited panel files pass `node --check` as `.mjs` (ES-module check).
- Mutation-verified: restoring the transcript write, snapshotting the blind decision at flush start, always picking the sighted note, and dropping the call-site `agentReceivesImages` dep each turn the matching test red.
- Committed blobs scanned for control bytes: 0. Post-state grepped on the committed blobs (old forms absent, new forms present).
- Rendered both note variants end to end against the issue's exact scenario text.
- Codex adversarial self-gate: **PASS** (round 1 found a real per-segment race — fixed and regression-tested; round 2 finding was shown to be a false positive with line-cited evidence and withdrawn by the gate itself in round 3).

- The transcript staying quiet across a real ComfyUI/orchestrator restart while the pill cycles (the unit tests pin the wiring by source inspection, per the repo's established pattern for the giant module).
- A real Blind-mode video completion delivering the prohibitive note variant to the agent.

### DIFF
--- a/browser_tests/unit/bridge-reconnect-transcript.test.mjs
+++ b/browser_tests/unit/bridge-reconnect-transcript.test.mjs
@@ -1,0 +1,81 @@
+// Regression guard for #588: every successful bridge reconnect used to append
+// "Connected to ws://… — waiting for the panel agent…" to the DURABLE chat
+// transcript (onLog → appendSystem), because markConnected()/start()/setUrl()
+// reset the `loggedWaiting` throttle. During a ComfyUI/orchestrator restart the
+// same transient line flooded the chat once per reconnect cycle, even though
+// connection recovery was working normally and the status pill already shows
+// the transient state.
+//
+// The load-bearing invariants locked here:
+//   * the socket-open handler never writes the transient "waiting for the
+//     panel agent" line to the transcript (the pill owns that state);
+//   * the pill still flips to "connecting" on open (the state is not hidden);
+//   * a genuinely WEDGED open socket (no agent handshake inside the window)
+//     still surfaces its distinct warning — the transient-line removal must
+//     not silence real wedge reporting;
+//   * the `loggedWaiting` throttle bookkeeping is gone entirely (a resettable
+//     throttle IS the flood — every reset re-arms exactly one transcript line).
+//
+// The callback depends on the real browser/ComfyUI environment, so inspect the
+// shipped handler source directly (the panel's established wiring test
+// pattern, cf. bridge-disconnect.test.mjs) rather than substitute a different
+// implementation.
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+
+const panelPath = fileURLToPath(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url));
+
+function panelSource() {
+  return readFileSync(panelPath, "utf8").replace(/\r\n/g, "\n");
+}
+
+function socketOpenHandler() {
+  const source = panelSource();
+  const start = source.indexOf('thisSock.addEventListener("open", () => {');
+  assert.notEqual(start, -1, "could not locate the bridge socket open handler");
+  const end = source.indexOf("\n    });", start);
+  assert.notEqual(end, -1, "could not locate the end of the open handler");
+  return source.slice(start, end);
+}
+
+test("#588: a socket (re)open never writes the transient waiting line to the transcript", () => {
+  const handler = socketOpenHandler();
+  assert.doesNotMatch(
+    handler,
+    /onLog\(\s*[`"'][\s\S]{0,300}?waiting for the panel agent/,
+    "the transient waiting state must not be appended to the durable chat transcript",
+  );
+  assert.doesNotMatch(
+    handler,
+    /appendSystem\([^)]*waiting for the panel agent/s,
+    "the transient waiting state must not reach the transcript by any other path",
+  );
+});
+
+test("#588: the status pill still communicates the transient connecting state", () => {
+  const handler = socketOpenHandler();
+  assert.match(
+    handler,
+    /emitStatus\("connecting"\)/,
+    "removing the transcript line must not hide the state from the status pill",
+  );
+});
+
+test("#588: a wedged open socket still warns (transient-line removal ≠ silence)", () => {
+  const handler = socketOpenHandler();
+  assert.match(
+    handler,
+    /no panel agent responded/,
+    "the handshake-timeout wedge warning is a REAL state and must survive",
+  );
+});
+
+test("#588: the resettable loggedWaiting throttle is gone (a reset re-arms the flood)", () => {
+  assert.doesNotMatch(
+    panelSource(),
+    /\bloggedWaiting\b/,
+    "the throttle that let every reconnect cycle re-log the line must be removed, not re-tuned",
+  );
+});
--- a/browser_tests/unit/run-completion.test.mjs
+++ b/browser_tests/unit/run-completion.test.mjs
@@ -12,6 +12,8 @@
  */
 import test from "node:test";
 import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
 
 import {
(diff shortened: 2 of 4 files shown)


artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 2 P0.**

**What passed, with specifics** — the reconnect suppression and the blind-storyboard work are clean. Successful opens now use only the status pill, the handshake timeout stays distinct (so the flood fix doesn't swallow a real wait), and the new per-frame blind decision genuinely prevents visual-review instructions when storyboard pixels are withheld. No ES-module redeclaration or import/export defect. No P1, no P2.

**Both P0s are outside the hunks you edited** — the issues named symptoms and the class runs further. Same thesis as the cluster: *never report a state that was not observed.*

### P0 — an unmatched result is reported as a successful turn
`web/js/lib/run-completion.js:536` with `web/js/comfyui-mcp-panel.js:19290`

An `executed` event with **no `prompt_id`** is buffered under `NO_PROMPT_KEY` and later emitted as a definite `kind:"executed"` completion. The frame omits `prompt_id` but the note still says **"Run finished."**

So a dropped or malformed lifecycle id turns an *unverifiable* result into an *apparent success* — which is artokun/comfyui-mcp#886's own defect still live in the path artokun/comfyui-mcp#886 is about. The honest outcome here is the unverifiable one you already have vocabulary for.

### P0 — an inferred verdict leaked as a definite one
`web/js/comfyui-mcp-panel.js:19158`

When history and queue reconciliation establish only that **the prompt is absent**, the panel sends `kind:"run_error"` and says it was *"likely cancelled or interrupted."*

A ComfyUI restart or history eviction **after a successful completion** produces exactly the same absence. So neither cancellation nor failure was observed — absence is the only fact, and it is consistent with success. This is artokun/comfyui-mcp#885's shape (a non-user event reported as a user action) reached by a different route: the agent is handed a lifecycle verdict nobody established, and will relay it to the user as fact.

Note the word "likely" does not rescue it. The frame's `kind` is `run_error`, and that is what downstream reads.

### Suggested shape
Both want the same thing: an **absent/unmatched** state that is neither success nor failure, carried as its own `kind` rather than folded into the nearest definite one. You already added that distinction for artokun/comfyui-mcp#886 — these two paths just don't reach it.

Mutation-verify each (an `executed` with no `prompt_id`; a reconciliation that finds the prompt absent after a restart), and assert the **reason**, not just the kind — asserting the kind alone passes via more than one path.

Everything else on this PR stands. Re-gated on green unless a P0 returns.

artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 1 P0, 1 P1.** No duplicate top-level declaration; bytes clean.

The three hard ones came back clean, and they were the ones worth worrying about: **the stall is no longer reported as a user rejection**, the turn-matching fix holds, and the reconnect path came back with no findings. The single-decision-point design for sighted/blind — one read after the slowest segment resolves, immediately before send, so segments cannot disagree with each other or with the `sendFrame` gate — is the right shape and the comment explaining why is worth keeping.

### P0 — the blind disclosure never fires for an image-only run

`web/js/lib/run-completion-frame.js:127` (and `:92-93`).

The stills segment pushes into `outImages`/`noteSections` at `:92-93` with `seg.note` only. The sighted/blind decision at `:135` then iterates **`videoSegs` alone**, and `noteWhenBlind` is by construction set only on the storyboard-success path — the code says so itself at `:287`.

So: **Blind ON, image-only run.** The `sendFrame` gate strips `frame.images`, and no blind note is ever produced. The agent receives text with no explanation for the missing render, and the most natural reading is that the render *failed or was unavailable* — not that it was deliberately withheld. That is the same harm artokun/comfyui-mcp-panel#609 is about, just reached through the commonest path rather than the storyboard one: an agent reasoning from an absence it was never told about.

It is also the specific thing I asked the gate to check — whether the fix covers **every** withheld-image path or only the one the issue named. The still-image path is the one most users will hit.

The fix wants to live where the decision already is: give the stills segment a `noteWhenBlind` too, and fold **all** segments — stills included — through the single `sighted` decision rather than folding stills in beforehand. That keeps the one-decision-point property that makes this design correct, instead of adding a second place where sighted/blind is decided.

### P1 — the new blind tests only exercise video/storyboard frames

`browser_tests/unit/run-completion.test.mjs:471`. Every blind-mode test uses a video or storyboard frame, so they all pass while the image-only path stays silent. Add an image-only blind completion that asserts the agent-facing note **explicitly says the render images were withheld** — assert the disclosure, not merely that a note exists, since a note exists today and is exactly the problem.

### Note on merging
These checks completed at 19:48, and `main` has since taken artokun/comfyui-mcp-panel#621, artokun/comfyui-mcp-panel#614, artokun/comfyui-mcp-panel#643 and the 0.11.39 release — all touching `comfyui-mcp-panel.js`. A green check is only green for the `main` it ran against; that staleness took `main` red earlier tonight. Verify on the actual merged tree before this lands.

Issue #605, opened 2026-08-04
Manager dialect cache stays legacy after backend restart to v4


artokun:
### What happened / root cause
After migrating a live ComfyUI tab from legacy ComfyUI-Manager 3.x to pip Manager v4.2.2 and restarting ComfyUI without reloading the browser tab, panel_list_nodes returned "ComfyUI-Manager not reachable". Direct GETs to /v2/manager/version, /v2/manager/queue/status, and /v2/customnode/installed all returned HTTP 200.

The panel keeps managerDialectCache="legacy" across the backend restart. nodes_list then routes to legacy endpoints, and its fallback repeats a legacy call rather than invalidating the dialect cache and probing the restarted backend.

### Steps to reproduce
1. Open the sidebar panel while ComfyUI runs legacy Manager 3.x.
2. Install comfyui-manager 4.2.2, disable the old custom_nodes/ComfyUI-Manager clone, and restart ComfyUI with --enable-manager while leaving the browser tab alive.
3. Call panel_list_nodes.
4. It reports Manager unreachable even though /v2 endpoints are healthy.
5. A full frontend reload clears the cache and makes the call work.

### Exact error (scrubbed)
Error: ComfyUI-Manager not reachable (is the built-in Manager enabled?)

### Fix
Applied locally: yes. Upstream-only: no.
Invalidate managerDialectCache after an unreachable routed call, re-probe once, then retain the existing legacy fallback.

--- web/js/comfyui-mcp-panel.js
+++ web/js/comfyui-mcp-panel.js
@@
       } catch (err) {
         if (!isManagerUnreachable(err)) throw err;
-        return { installed: await managerCall(route) };
+        managerDialectCache = null;
+        try {
+          return { installed: await managerGet(route) };
+        } catch (retryErr) {
+          if (!isManagerUnreachable(retryErr)) throw retryErr;
+          return { installed: await managerCall(route) };
+        }
       }

Verification:
- node --check web/js/comfyui-mcp-panel.js: exit 0
- frontend soft reload
- panel_list_nodes returned the complete installed-node map
- panel_node_queue_status returned the v4 queue payload

### Environment
Windows 11; ComfyUI 0.30.1; RTX 5070 Ti 16 GB; comfyui-mcp 0.49.4; panel 0.11.38; migrated Manager 3.x -> v4.2.2.

artokun:
[...]

Issue #485, opened 2026-08-01
panel_install_node reports Manager unreachable while panel_list_nodes works on Manager 3.x legacy


artokun:
### What happened / root cause
The live-panel panel_install_node tool could list installed Manager packs but failed to queue a repository install, returning "ComfyUI-Manager not reachable (is the built-in Manager enabled?)". ComfyUI-Manager is installed and enabled, and panel_list_nodes succeeds. This appears to be an incompatibility in the panel install route with Manager 3.x legacy.

### Steps to reproduce
1. On ComfyUI 0.29.2 with Manager 3.x legacy, call panel_list_nodes (succeeds).
2. Call panel_install_node with repository https://github.com/Fannovel16/ComfyUI-Frame-Interpolation.git and version nightly.
3. Observe the reachability error.

### Exact error (scrubbed)
Error: ComfyUI-Manager not reachable (is the built-in Manager enabled?)

### Fix
Applied locally: no. Upstream-only: yes.
Suggested change: detect legacy Manager API availability and either use its supported install endpoint or return a structured compatibility result with a supported fallback to comfyui-mcp install_custom_node rather than a generic reachability error.

### Environment
Windows 11; NVIDIA RTX 5070 Ti Laptop GPU 12 GB; ComfyUI 0.29.2; comfyui-mcp 0.48.28; panel 0.11.20; Manager 3.x legacy.

PR #629, opened 2026-08-04
Manager dialect cache stale after a backend upgrade, and sync no-oping on incoherent queue counts

artokun:
## Issues in this cluster

- artokun/comfyui-mcp-panel#605 — Manager dialect cache stays legacy after a backend restart to v4
- artokun/comfyui-mcp#887 — panel sync no-ops on incoherent Manager queue counts while reinstall succeeds

## What changed (#605)

The per-tab `managerDialectCache` was never invalidated: after a Manager 3.x → pip v4 migration and a ComfyUI restart with the browser tab alive, `panel_list_nodes` (and every other dialect-routed call) kept routing at the retired 3.x routes and reported a false "ComfyUI-Manager not reachable" until a full frontend reload.

The fix treats the cached dialect as a hypothesis that two signals falsify:

- **Lifecycle:** the `reconnected` listener (backend socket re-established = the restart signal) now drops the cache.
- **Per-call backstop:** an "unreachable" verdict from a dialect-routed call invalidates the cache, re-probes the live backend once (`reProbeManagerDialect`), and retries via the pure `dialectRetryTarget` ladder (fresh verdict wins when it changed; else the artokun/comfyui-mcp-panel#485 legacy absolute last resort; legacy-on-legacy surfaces the original error). Applied at `managerGet` (covers nodes_list / nodes_search / nodes_queue_status / update verification) and at the mutation paths (`nodes_install` submit, `graph_update_node` dispatch).

Mutation retries fire ONLY on a proven route-level rejection: the Manager transports now tag their HTTP 404 throw (`err.managerRouteMissing`, checked by `isManagerRouteMissing`) — an ambiguous no-response "not reachable" never re-sends an install/update (no double-fire), and landed-enqueue tracking (`enqueued` / `legacyEnqueued`) keeps a post-enqueue start failure from re-firing. The artokun/comfyui-mcp-panel#424 405→legacyUpdate fallback routes through the same ladder. The whole re-probe shares one budget (caller's signal combined via a feature-detected `anyAbortSignal` with the 15s Manager timeout), and an aborted probe never pins a partial verdict into the cache.

Codex adversarial gate: 5 finding rounds (r1 mutation-retry ambiguity P0, r2 405-fallback ladder bypass, r3 re-probe budget, r4 entry-detection budget + false sub-dialect pin, r5 `AbortSignal.any` baseline) — each fixed with mutation-verified tests; round 6 PASS. Full unit suite 1801/1801; committed blobs scanned for control bytes (0); panel verified as ES module (`node --check` as .mjs).

## What deliberately did NOT change (#598)

`install_panel(action="sync")` is implemented entirely in the orchestrator repo (`comfyui-mcp` `src/services/panel-installer.ts` / `panel-sync.ts`) — it talks to the Manager over HTTP directly; the panel is not in that codepath, so there is no panel-repo change possible here. The current orchestrator already detects the incoherent-counts signature (done > total) and deliberately refuses the git fallback with an actionable remedy (#724 codex gates). The issue's remaining ask — sync offering or automatically using the verified reinstall fallback when the Manager no-ops incoherently and no clean git fallback exists — is an orchestrator-side enhancement and stays open on artokun/comfyui-mcp#887.

## Live-browser confirmation

The stale-cache heal paths are covered by source-extraction unit tests against the real panel functions; the actual 3.x→v4 restart swap with a live tab (the issue's repro) can only be confirmed in a browser.

### DIFF
--- a/browser_tests/unit/manager-dialect.test.mjs
+++ b/browser_tests/unit/manager-dialect.test.mjs
@@ -14,6 +14,7 @@ import { fileURLToPath } from "node:url";
 import * as ManagerInstall from "../../web/js/lib/manager-install.js";
 const {
   dialectRetryTarget,
+  isManagerRouteMissing,
   isManagerUnreachable,
   isMethodNotAllowed,
   legacyUpdateBody,
@@ -23,6 +24,11 @@ const {
 
 const UNREACHABLE = "ComfyUI-Manager not reachable (is the built-in Manager enabled?)";
 const QUEUE_STATUS = { total_count: 0, done_count: 0, is_processing: false };
+
+/** The error managerCall/managerV2 throw for a PROVEN route-level rejection
+ *  (HTTP 404): the same message, tagged managerRouteMissing. Mutations may be
+ *  re-sent on this and ONLY this (codex P0). */
+const routeMissing = () => Object.assign(new Error(UNREACHABLE), { managerRouteMissing: true });
 
 function readPanelSource() {
   const panelPath = fileURLToPath(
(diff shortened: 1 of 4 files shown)


artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 1 P0.** No P1, no P2 beyond it, no duplicate top-level declaration. The dialect-cache invalidation itself came back clean.

### P0 — `comfyui-mcp-panel.js:10854`: refusing after the destructive step already happened

The raw start failure is rethrown **after the update task POST was observed to succeed** (`enqueued === true`). If `/manager/queue/task` accepts the destructive update and then `/manager/queue/start` loses its response, the caller gets only *"not reachable"* — no `ui_id`, and no statement that the task is already queued.

An agent reading "not reachable" does the reasonable thing and retries `graph_update_node`, **dispatching a second update**. And the new test locks this rejection shape in, so the misleading form is now the specified behaviour.

This is the repo's governing distinction: **refuse when the action has not happened; disclose when it already has.** Once `enqueued === true`, the update exists on the server, and the only honest outcome is a disclosure — "the update was queued as `<ui_id>`; the start signal could not be confirmed; do not retry" — with the `ui_id` so the caller can poll rather than re-dispatch. Rethrowing the transport error reports failure for work that succeeded and invites a destructive retry, which is exactly the harm the rule exists to prevent.

Fix the test alongside it: it should assert the **disclosure** (ui_id present, retry explicitly discouraged), not the raw rejection.

### On artokun/comfyui-mcp#887
The gate notes that **no panel-side sync/reinstall codepath is changed in this diff**, so it cannot itself fix an orchestrator-side `install_panel` sync verdict. That's a scoping call worth making explicit rather than leaving implied — if artokun/comfyui-mcp#887's fix belongs on the orchestrator side, say so in the PR body and leave the issue open with that note, rather than closing it against a panel-only change. A partial fix reported as complete is the outcome to avoid here.

Issue #599, opened 2026-08-03
CivitAI browser model search returns `Failed to fetch` while backend CivitAI API succeeds


artokun:
### What happened / root cause
The docked CivitAI browser opens, but every checkpoint search returns an empty grid with `error: { status: null, message: "Failed to fetch" }`. The same public CivitAI searches succeed through comfyui-mcp's `search_civitai_models` and direct `https://civitai.com/api/v1/models` requests on the same machine, so this appears isolated to the panel/frontend fetch path (likely cross-origin/proxy handling), not CivitAI availability.

### Steps to reproduce
1. Call `panel_open_civitai` with tab `checkpoints`, query `Wan2.2 TI2V 5B GGUF Q8_0`, browsingLevels `[1,2]`.
2. Call `panel_civitai_results`.
3. Result is empty with `Failed to fetch`.
4. Call `panel_civitai_search` with query `WAN 2.2 5B`; `dispatched:true`.
5. Call `panel_civitai_results` again; same failure.
6. Backend `search_civitai_models` and direct CivitAI REST API requests return matching records immediately.

### Exact error (scrubbed)
`{"items":[],"total":0,"loading":false,"error":{"status":null,"message":"Failed to fetch"},"authenticated":false}`

### Fix
Applied locally: no. Upstream-only: yes. The panel's CivitAI browser should route public model searches through the server-side proxy/backend client (or otherwise avoid browser cross-origin fetch failure), and return an HTTP status/message that distinguishes network/CORS from a true empty result.

### Environment
Windows 11; ComfyUI 0.30.0; NVIDIA RTX 5060 Ti 16 GB; comfyui-mcp 0.49.3; panel 0.11.35.

artokun:
Two changes have landed for this, but **leaving this open** — the root cause is still not established.

- `040f4b7` (PR #632, merge `521f651`): one bounded retry for idempotent specs, plus a classified `kind:"transport"` error so `panel_civitai_results` can distinguish a transport failure from an upstream CivitAI error from a genuinely empty result.
- `f50efde` (PR #669): fixed the two P0s that #632's own gate raised and that were merged unaddressed — the error no longer claims the browser→ComfyUI hop was the cause or that "this is not a CivitAI error", and the retry no longer destroys the first attempt's evidence.

What is fixed for certain: the **reporting**. `Failed to fetch` is no longer surfaced bare, is never mistaken for an empty result, and no longer names a hop it cannot identify.

[...]

PR #632, opened 2026-08-04
CivitAI browser search fails in the frontend while the backend API succeeds

artokun:
## Issues in this cluster

- #599

**Not yet started** — draft until an agent picks it up.

### DIFF
--- a/browser_tests/unit/civitai-drive-results.test.mjs
+++ b/browser_tests/unit/civitai-drive-results.test.mjs
@@ -7,7 +7,7 @@
 // returns METADATA + URLs ONLY, never image bytes, and clamps `limit`.
 import { test } from "node:test";
 import assert from "node:assert/strict";
-import { serializeCivitaiResults, CIVITAI_PROMPT_CAP } from "../../web/js/cmcp-civitai-ui.js";
+import { serializeCivitaiResults, CIVITAI_PROMPT_CAP, civitaiErrorState } from "../../web/js/cmcp-civitai-ui.js";
 
 // Media rows as produced by CivitaiClient._fromRest / _fromMeili.
 const mediaRows = [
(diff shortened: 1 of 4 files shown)


artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **No P0. One P1 — a regression this PR introduces, so it wants fixing before merge rather than filing.**

**Clean results worth stating:** no credential-bearing request routed to an unverified origin, no API key reaching a log, URL or frontend-visible surface, no false-empty result path (the error is still recorded non-null, so a failure can't read to the user as "CivitAI has nothing"), and no duplicate top-level declaration.

### P1 — `web/js/cmcp-civitai.js:347-356`: a diagnosis that wasn't established

Every statusless `fetchApi` rejection is turned into a definite **`"browser→ComfyUI hop"`** failure, stated as **"not a CivitAI error."**

`Failed to fetch` establishes exactly one thing: **the browser did not obtain a response.** It does not identify which hop failed, and it does not exclude an upstream CivitAI failure whose response was blocked or discarded on the way back. A service worker, a browser policy, or a response-path failure *after* the request reached ComfyUI produces the identical rejection.

So the agent is sent to investigate ComfyUI reachability when the cause is undetermined — and told, incorrectly, that CivitAI has been ruled out.

This is the bug's own shape reappearing in its fix. #599 exists because `Failed to fetch` refuses to distinguish causes; the fix responds by **picking one and asserting it**. The browser withholds that information deliberately, and no amount of framing recovers it.

**Keep everything you observed, drop the inference.** `status: null`, the original message, and the retry step are all real and useful. Report the cause as **undetermined**, list the candidates rather than naming one, and give the check step. A remedy that names the wrong layer costs more than "could not determine why" — the user rules out a healthy component and stops looking.

Mutation-verify with a rejection occurring *after* the request reached ComfyUI, asserting the diagnosis is undetermined and CivitAI is not excluded. Assert on the **reason**, not just that an error was produced — asserting the outcome alone passes through either branch.

### Note on policy
Merging on green with non-data-loss P1s filed as follow-ups is the norm, with one exception I'm applying here: **a P1 that is a regression introduced by the PR gets fixed; a pre-existing gap gets filed.** Before this change the user got an unhelpful-but-honest `Failed to fetch`; after it they get a confident wrong answer, which is worse.

artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 2 P0.** No P1, no P2. No secret reaches the browser, a URL, or a log — that check came back clean, which was the one I was most concerned about on this path.

### P0 — the fix narrates the opaque bucket as a cause, which is the defect it exists to remove

`web/js/cmcp-civitai.js:345`. The transport error says:

> "This is a connection-level failure — **not a CivitAI error**, and not an empty result."

A rejected `fetchApi` establishes exactly one thing: **the browser received no HTTP response.** It does not establish which hop failed. Concretely: the panel's request reaches ComfyUI, ComfyUI's own upstream CivitAI request fails or its response connection drops before a response is returned, `fetchApi` rejects — and the panel tells the user CivitAI was definitively not involved, when it may have been the entire cause.

`Failed to fetch` is the browser's single opaque bucket covering network failure, CORS rejection, a blocked request, DNS, offline, and abort. **Replacing one bucket-narrated-as-cause with a more confident bucket-narrated-as-cause is not the fix** — it is the same defect with better prose, and it will misdirect diagnosis of the next #599. State what was observed (no HTTP response came back at the browser→ComfyUI hop), list the candidates, and assert none. The one thing that *is* safe to say is that no HTTP status was received.

### P0 — the retry discards the first failure's evidence

`:334-349`. On an idempotent request the first transport rejection is dropped by `continue`; if the retry also fails, only attempt 2's message survives. So attempt 1 fails with an actionable browser/proxy error, attempt 2 fails for an unrelated reason, and the user is shown only the second. **Never destroy evidence before it can be reported** — this repo has had three bugs that were a rescue path deleting the thing it was rescuing, and a retry that eats the original cause is that shape. #599 is an intermittent failure; losing the first error is losing the diagnosis.

Carry both: report the retry's error with the first attached (`cause`, or named in the message), so an intermittent failure leaves a full trail.

### On merging
Checks completed 19:46 and `main` has since taken #621, #614, #643 and the 0.11.39 release, all touching panel JS. **A green check is only green for the `main` it ran against** — that staleness took `main` red tonight. Verify on the actual merged tree.

PR #672, opened 2026-08-05
graph_disconnect silently deletes adjacent nodes instead of only removing the input link

### LINKED ISSUE #668: panel_disconnect (graph_disconnect) silently DELETES adjacent nodes instead of only removing the input link
### What happened

Three consecutive `panel_disconnect` calls on a **subgraph node**'s inputs removed **two unrelated NODES** from the graph, not just the links. Each call returned a normal success payload with no hint that anything was deleted.

Graph before (from `panel_graph_outline`, 10 nodes):

```
115 ResolutionSelector
114 LoadImage            -> 121.image
121 ImageCrop            <- 114.IMAGE   -> 105.first_frame, 119.image
119 ImageScaleToTotalPixels <- 121.IMAGE -> 120.image
120 GetImageSize         <- 119.IMAGE   -> 105.width, 105.height
105 <subgraph 4c314f31-ecda-4b08-ae98-faaba1bf613f> "Image to Video (MiniMax H3)"
                          <- 121.IMAGE, 120.width, 120.height  -> 92.video
92  SaveVideo [OUTPUT]   <- 105.VIDEO
(+ 3 MarkdownNote)
```

Calls made (nothing else mutating in between):

```
panel_disconnect(node_id=105, input="first_frame")  -> {"disconnected":{"node_id":"105","input":"first_frame"}}
panel_disconnect(node_id=105, input="width")        -> {"disconnected":{"node_id":"105","input":"width"}}
panel_disconnect(node_id=105, input="height")       -> {"disconnected":{"node_id":"105","input":"height"}}
```

Graph after (8 nodes) — **`114 LoadImage` and `92 SaveVideo` are gone**:

```
115 ResolutionSelector
121 ImageCrop            (image input now EMPTY)  -> 119.image
119 ImageScaleToTotalPixels -> 120.image
120 GetImageSize
105 <subgraph>           (no inputs, VIDEO output links:0)
(+ 3 MarkdownNote)
```

The two deleted nodes were **not** the disconnect target and were not even both upstream:

- `114 LoadImage` was **two hops upstream** — it fed `121 ImageCrop`, never node 105 directly.
- `92 SaveVideo` was **DOWNSTREAM**, a consumer of node 105's `VIDEO` output — not an input at all.

Meanwhile `121`, `119`, `120` (the nodes actually on the disconnected input path) all survived, which makes the selection of what got deleted look arbitrary rather than a deliberate orphan-prune.

### Why this is bad

Silent destruction of the user's graph. The three tool results all reported plain success, so nothing signalled the loss. I only caught it because a later `panel_query_graph` showed node 105's `VIDEO` output with `links:0` and `total:8`. Had I trusted the success payloads and queued the render, it would have run to completion (a ~45-70 minute video generation on this machine) and **saved nothing**, because `SaveVideo` no longer existed.

### Steps to reproduce

1. Open a workflow containing a **subgraph node** with multiple linked inputs (repro'd on ComfyUI's `video_minimax_h3_i2v` MiniMax H3 template; node 105 is subgraph type `4c314f31-ecda-4b08-ae98-faaba1bf613f`).
2. Ensure the subgraph node has an upstream chain at least 2 nodes deep feeding one input, and a downstream output node (e.g. `SaveVideo`) consuming its output.
3. Call `panel_disconnect` on each of its linked inputs in turn.
4. Re-read with `panel_graph_outline` and compare node counts.

### Suspected area

`web/js/comfyui-mcp-panel.js`, `graph_disconnect` (~line 7144):

```js
graph_disconnect({ node_id, input }) {
  const { graph } = getGraphCtx();
  const node = resolveNode(graph, node_id);
  const inIdx = resolveSlot(node.inputs, input ?? 0, "input");
  graph.beforeChange();
  try {
    node.disconnectInput(inIdx);
  } finally {
    graph.afterChange();
  }
  graph.setDirtyCanvas(true, true);
  return {
    disconnected: { node_id: node.id, input: node.inputs?.[inIdx]?.name ?? inIdx },
  };
}
```

Hypothesis (NOT confirmed — I was interrupted before I could run an isolated repro, so please treat this as a lead only): on a **SubgraphNode** the frontend's `disconnectInput()` may do more than drop the link — e.g. remove the boundary IO slot from the subgraph definition, which then cascades into the parent graph. That would also explain why slot indices/identities appear to shift between the successive calls. It does not by itself explain the deletion of the downstream `SaveVideo`.

### Suggested fix

Two independent hardening steps, both cheap:

1. **Verify, don't assume.** Mirror the honesty pattern already used in `graph_connect` (the `isLinkPersisted` / `#397 CONNECT HONESTY` block): snapshot `graph._nodes.length` (and ideally the node id set) before `disconnectInput`, re-check after, and if any node other than the intended link disappeared, **restore it / fail loudly** rather than returning a bare success.
2. **Report what actually changed.** Include the removed link's former source (`{node_id, output}`) in the result — `graph_connect` already returns `replaced_link`, so `graph_disconnect` returning the symmetric info would have made this obvious immediately.

A narrower alternative for the subgraph case: when `node.isSubgraphNode?.()`, drop only the link (`graph.removeLink(inputSlot.link)` / null out `inputSlot.link`) instead of calling the node's own `disconnectInput`, so the boundary slot definition is left untouched.

### Fix status

applied locally: **no**  ·  upstream-only: **no**

I deliberately did not patch. The root cause isn't confirmed (see hypothesis above) and a blind edit to `graph_disconnect` risks breaking the ordinary non-subgraph path, which works correctly. The isolated repro that would have pinned it was interrupted. Recovery in the session was manual: re-added `SaveVideo` + `LoadImage` and rewired.

### Workaround for other agents

After any `panel_disconnect`, re-read the graph (`panel_graph_outline`) and compare the node count before trusting the success payload. Where possible prefer `panel_connect` to *replace* a link (it reports `replaced_link`) over disconnect-then-reconnect.

### Environment

- OS: Windows 11 Enterprise 10.0.26200
- ComfyUI: 0.30.0 (local-portable), frontend 1.47.12, Manager 3.41
- GPU: NVIDIA RTX 5070 Ti, 16 GB VRAM · torch 2.13.0+cu130 · Python 3.13.14
- comfyui-mcp: **0.49.8**
- comfyui-mcp-panel: **0.11.38**

artokun:
Severe mutation-integrity defect: `graph_disconnect` (web/js/comfyui-mcp-panel.js:8601) is a bare `node.disconnectInput(inIdx)` with no subgraph-boundary handling, no pre/post node-set snapshot, and no verification.

closes #668

### DIFF
--- a/browser_tests/unit/disconnect-verify.test.mjs
+++ b/browser_tests/unit/disconnect-verify.test.mjs
@@ -1,0 +1,240 @@
+/**
+ * Unit tests for the graph_disconnect pre/post verification (#668) —
+ * web/js/lib/disconnect-verify.js. Run with `node --test`.
+ *
+ * Bug: three panel_disconnect calls on a SUBGRAPH node's inputs silently DELETED
+ * two unrelated nodes (a LoadImage two hops upstream, a downstream SaveVideo)
+ * while every call returned a plain success payload. The panel assumed
+ * node.disconnectInput() only drops one wire and never checked the post-state.
+ * These tests pin the property: after a disconnect, the intended link is gone
+ * and NOTHING else changed — otherwise the verdict is not ok and the caller
+ * must disclose, never report a bare success.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import {
+  snapshotGraphState,
+  describeInputLink,
+  verifyDisconnect,
+} from "../../web/js/lib/disconnect-verify.js";
+
+/** Minimal mock graph: nodes by id, links as a legacy record store. */
+function mockGraph(nodes, links) {
+  return {
+    _nodes: nodes,
+    links,
+    getNodeById: (id) => nodes.find((n) => n.id === id) ?? null,
+  };
+}
+
+function node(id, inputs = [], outputs = []) {
+  return { id, inputs, outputs };
+}
+
+// The #668 repro shape (reduced): 121 ImageCrop ← 114 LoadImage; subgraph node
+// 105 ← 121 (first_frame) and 105.VIDEO → 92 SaveVideo. Link ids 1, 2, 3.
+function reproGraph() {
+  const n114 = node(114, [], [{ name: "IMAGE", links: [1] }]);
+  const n121 = node(121, [{ name: "image", link: 1 }], [{ name: "IMAGE", links: [2] }]);
+  const n105 = node(
+    105,
+    [{ name: "first_frame", link: 2 }],
+    [{ name: "VIDEO", links: [3] }],
+  );
+  const n92 = node(92, [{ name: "video", link: 3 }], []);
+  return mockGraph([n114, n121, n105, n92], {
+    1: { id: 1, origin_id: 114, origin_slot: 0, target_id: 121, target_slot: 0 },
+    2: { id: 2, origin_id: 121, origin_slot: 0, target_id: 105, target_slot: 0 },
+    3: { id: 3, origin_id: 105, origin_slot: 0, target_id: 92, target_slot: 0 },
+  });
+}
+
+test("snapshotGraphState: node ids string-normalized, floating links excluded", () => {
+  const g = mockGraph([node(1), node("sub")], {
+    5: { id: 5, origin_id: 1, origin_slot: 0, target_id: "sub", target_slot: 0 },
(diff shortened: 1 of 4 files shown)


artokun:

PR #673, opened 2026-08-05
panel_install_node times out while the canvas remains responsive

### LINKED ISSUE #671: panel_install_node times out while the canvas remains responsive
Attempting to install `ComfyUI-MelBandRoFormer` from the sidebar timed out after 30 seconds. The same panel tab successfully answered `panel_graph_outline` immediately beforehand, so the install command/reply path appears wedged or absent.

### Steps to reproduce
1. Open a live workflow containing missing custom nodes.
2. Call `panel_search_nodes` with `MelBandRoFormer`.
3. Call `panel_install_node` with id `ComfyUI-MelBandRoFormer`.

### Exact error (scrubbed)
`Error: Panel tab wf:b7468 did not reply to "nodes_install" within 30000 ms — the ComfyUI tab may be backgrounded or frozen`

### Fix
Applied locally: no. Upstream-only: yes. The panel should acknowledge or forward `nodes_install` reliably, or return a structured Manager job id while the install is running.

### Environment
Windows 11; NVIDIA RTX 5070 12 GB; CUDA 13.0; torch 2.10.0; Python 3.13; ComfyUI 0.30.2; comfyui-mcp 0.49.3; panel 0.11.41.

artokun:
Install path times out despite a live, responsive canvas tab.

closes #671

### DIFF
--- a/browser_tests/unit/manager-dialect.test.mjs
+++ b/browser_tests/unit/manager-dialect.test.mjs
@@ -714,7 +714,7 @@ function nodesInstallDeps(overrides) {
     isStallError: (err) =>
       err?.name === "AbortError" ||
       err?.name === "TimeoutError" ||
-      /aborted mid-probe/.test(String(err?.message ?? "")),
+      String(err?.message ?? "").startsWith("ComfyUI-Manager dialect detection was aborted mid-probe"),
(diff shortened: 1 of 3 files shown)


artokun:

PR #674, opened 2026-08-05
run targeting: to_node_id drift guard false-positives on COMFY_AUTOGROW_V3 dynamic-input nodes

### LINKED ISSUE #659: panel_run to_node_id always refused with "graph CHANGED after the run was queued" (#556 drift guard false-positive) on a graph containing a COMFY_AUTOGROW_V3 dynamic-input node

`panel_run` with `to_node_id` is **100% unusable** on one of my workflows: every attempt is refused by the #556 run-scope guard with `reason: "graph_changed"`, on a completely settled canvas with **no user edits and no agent edits between calls**. Reproduced **5 consecutive times** (including 3 back-to-back calls with only a read-only `panel_query_graph` in between).

Since the graph genuinely did not change, `promptContentHash(await app.graphToPrompt())` computed pre-dispatch in `dispatchScopedRun` must be differing from `promptContentHashFromBody(POST /prompt body)`. i.e. **two serializations of the same unmodified graph do not produce the same content hash** on this workflow.

**Hypothesis (not verified — I could not instrument the browser):** the graph contains `MiniMaxH3ReferenceToVideo`, whose `ref_images` / `ref_videos` / `ref_audios` inputs are **`COMFY_AUTOGROW_V3`** dynamic/autogrow slots (it also has `COMFY_DYNAMICCOMBO_V3`-family siblings in core, e.g. `ClaudeNode`). Autogrow nodes mutate their own slot list at serialization time (appending the next empty slot), so `graphToPrompt()` is likely **not idempotent** for them — the pre-dispatch hash and the deferred post-time serialization legitimately differ, and the drift guard reads that as a user edit. `collectVolatileInputs` only excludes `beforeQueued` widget hooks, so it cannot cover autogrow slot churn.

### Steps to reproduce

1. Open a workflow containing a node with `COMFY_AUTOGROW_V3` inputs (here: ComfyUI core template `video_minimax_h3_r2v`, node `MiniMaxH3ReferenceToVideo`).
2. Add any output node (I used `PreviewAny`) on a small independent branch.
3. `panel_run({ to_node_id: <that output node id> })`.
4. Refused every time. Retrying does not help; waiting/settling does not help.

### What I ruled out

- **Not the seed / `control_after_generate` hook.** The graph has `RandomNoise` with `control_after_generate=randomize`. The error text explicitly suggests this cause, so I set it to `fixed` and re-ran — **still refused, identical error**. So the #572 `linkedWidgets` exclusion path is not what's failing here.
- **Not a concurrent edit.** No canvas interaction occurred; consecutive calls with only read-only tools in between still fail.
- **Not a transient race.** 5/5 failures, deterministic.
- **Not node-list churn from my own edits.** Failures continued after the graph was fully settled and re-read.

### Exact error (scrubbed)

```
ComfyUI refused to queue the workflow: run-to-node scope for node 143 was NOT
applied: the workflow graph CHANGED after the run was queued — the deferred
dispatch would render a modified workflow, not the one that was scoped. Nothing
was queued — refusing to fall through to a full-graph execution (#556).
```

### Impact / workaround

Run-to-node is the documented way to debug a wrong render and to preview one branch cheaply — it is completely unavailable on this graph. My workaround was to rebuild the branch as a standalone API-format graph and `enqueue_workflow` it headlessly, which works but **runs outside the user's canvas** (no live result on their graph, no undo, and it duplicates the wiring by hand). Full (unscoped) `panel_run` is unaffected.

### Fix

**applied locally: no** · **upstream-only: yes**

I deliberately did not patch. The guard is intentionally fail-closed and loosening `promptContentHash` to make this pass would re-open the #556 hole (a real mid-flight edit would stop being detected), so a blind local relaxation would be worse than the bug.

Precise change needed, in `web/js/lib/run-scope-guard.js`:

1. **Confirm the non-idempotency first** — call `app.graphToPrompt()` twice back to back on this workflow and diff `promptContentHash` of the two results. If they differ with zero edits, that is the bug and it is independent of the whole queue-mark/observation machinery.
2. If autogrow is the cause, either:
   - **(a)** normalize autogrow/dynamic slots out of the canonical form in `promptContentHash` — drop trailing dynamic inputs whose value is `null`/absent (the empty tail slot an autogrow node appends), so a grown-by-one serialization hashes identically; or
   - **(b)** extend the volatile-input mechanism: add a `collectDynamicSlotInputs(rootGraph)` companion to `collectVolatileInputs` that emits `execId inputName` pairs for `COMFY_AUTOGROW_V3` / dynamic-slot inputs, and union it into `volatileInputs`. This reuses the existing per-node-pair exclusion and the existing `drift_coverage` reporting, so the gap stays visible to the caller rather than becoming silent.
   
   (b) is the smaller, more consistent change and keeps the honesty guarantee (`volatileInputs` is already surfaced in the run result).
3. **Regardless of cause — make the refusal diagnosable.** Today `graph_changed` tells the caller nothing about *what* changed, which is why ruling causes out took 5 runs and a widget experiment. Please include, on the `graph_changed` verdict, the set of `execId inputName` pairs (or at least node ids) whose canonical entries differed between the pre-dispatch hash input and the post body. `promptContentHash` already builds the canonical array — retaining it and diffing on mismatch is cheap and would have made this self-diagnosing.

### Environment

- OS: Windows 11 Pro for Workstations 10.0.26200
- GPU: NVIDIA GeForce RTX 5090 (32 GB VRAM), 127 GB RAM, CUDA 13.0
- torch 2.10.0, python 3.12
- ComfyUI 0.30.0 (local)
- **comfyui-mcp 0.49.7**
- **comfyui-mcp-panel 0.11.38**
- Workflow: ComfyUI core template `video_minimax_h3_r2v` (+ 4 nodes I added: `OllamaConnectivityV2`, `OllamaGenerateV2`, `PreviewAny`, `MarkdownNote`)
- Note: 4 model files referenced by the graph were still downloading at the time (nodes red). Partial execution to `PreviewAny` does not touch those nodes, and the refusal happens client-side before any POST, so this should be unrelated — flagging it only for completeness.

Related (same root cause): #572

artokun:
`panel_run to_node_id` always refused with "graph CHANGED after the run was queued" (#556 drift guard false positive) when the graph contains a COMFY_AUTOGROW_V3 dynamic-input node: `promptContentHash`/`collectVolatileInputs` (web/js/lib/run-scope-guard.js:186,251) exclude only beforeQueued-hook inputs; non-idempotent `graphToPrompt()` on autogrow nodes trips the guard.

closes #659

### DIFF
--- a/browser_tests/unit/run-scope-guard.test.mjs
+++ b/browser_tests/unit/run-scope-guard.test.mjs
@@ -25,6 +25,8 @@ import {
   repairScopeInBody,
   promptContentHash,
   promptContentHashFromBody,
+  canonicalizePrompt,
+  diffPromptCanons,
   collectVolatileInputs,
   verifyScopedPromptBody,
   scopeDroppedError,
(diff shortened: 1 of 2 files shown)


artokun:

Issue #635, opened 2026-08-04
panel_restart_comfyui reports dispatched:true when the target port is wrong; panel_refresh_nodes stuck at refreshed:false


artokun:
### What happened / root cause

Two related defects, both hit while recovering a workflow whose custom-node pack needed an update + ComfyUI restart.

**(1) `panel_restart_comfyui` false-positive dispatch.**

The user's ComfyUI had lost its bind on 8188 at boot and fell back to **8189**:

```
[ERROR] Failed to initialize database. ... Could not acquire lock on database
'~/ComfyUI-Installs/ComfyUI/ComfyUI/user/comfyui.db'. Another ComfyUI process
may already be using it.
[INFO] Starting server
[INFO] To see the GUI go to: http://127.0.0.1:8189
```

`panel_restart_comfyui` returned:

```json
{"rebooting": true, "ready": false, "confirmed_cycle": false, "dispatched": true,
 "note": "ComfyUI restart was dispatched and accepted; it is restarting out-of-band..."}
```

Nothing restarted. The process start time never changed and the log gained no reboot entries. `dispatched: true` plus "dispatched and accepted" reads as success, so I told the user it was restarting and waited on a server that was never going to cycle. The reboot only actually happened when I bypassed the tool and POSTed `/api/v2/manager/reboot` directly against 8189 — after which the updated node signature loaded correctly.

Meanwhile every headless comfyui MCP tool failed with a bare `fetch failed` / `{"error":"CONNECTION_ERROR","message":"ComfyUI unreachable: fetch failed"}` because it was still aimed at 8188. Neither surface said "nothing is listening on the port I am configured for" — which is the one fact that would have explained all of it immediately.

**(2) `panel_refresh_nodes` permanently returns `refreshed: false`.**

`{"ok": true, "refreshed": false}` on every call — including immediately after a confirmed successful ComfyUI reboot where `/object_info/Krea2EditModelPatch` demonstrably served the NEW signature. `ok:true` with `refreshed:false` gives no reason and no remedy, so it is indistinguishable from a no-op. I had to verify the node definition by curling `/object_info` directly.

### Steps to reproduce

1. Start ComfyUI so it falls back off its default port (e.g. leave a stale process holding 8188; it binds 8189).
2. Call `panel_restart_comfyui` → returns `dispatched: true`, nothing cycles.
3. Call any headless comfyui MCP tool → `fetch failed`, with no mention of which host:port was tried.
[...]

PR #677, opened 2026-08-05
reconnect recovery: proactive rebind, tmp-key invalidation, subgraph scope receipts, refresh_nodes honesty

### LINKED ISSUE #617: Frontend reconnect can loop on root-shape-mismatch for an unsaved workflow

### LINKED ISSUE #619: Intermittent root-shape-mismatch immediately after panel_enter_subgraph

### LINKED ISSUE #646: Graph mutation races with panel reconnect after ComfyUI restart
After panel_restart_comfyui completed a confirmed server cycle, panel_set_workflow_target({mode:"current"}) and one panel_query_graph call succeeded. The immediately following panel_set_widget disconnected mid-command with outcome unknown; the next panel_set_widget and graph query reported no connected tab. The restart result had already warned panel_tab_reconnected:false, but graph calls briefly succeeded before the socket vanished. This appears to be a reconnect race where graph tools become callable before the workflow tab connection is stable.

### Steps to reproduce
1. Use panel_restart_comfyui and accept the restart.
2. Wait for the tool to report server_ready:true, panel_tab_reconnected:false.
3. Call panel_set_workflow_target({mode:"current"}).
4. Call panel_query_graph successfully.
5. Immediately call panel_set_widget twice.

### Exact error (scrubbed)
First edit: panel tab wf:workf disconnected mid-command ("graph_set_widget") - OUTCOME UNKNOWN.
Second edit: graph_set_widget could not be dispatched; no connected tab. Connected: none.

### Fix
Keep graph mutation tools unavailable until the workflow tab socket has remained connected/stable, or make the rebind handshake await a durable connection. A mutation that loses the socket should also preserve enough idempotency state for automatic verification after reconnect.

### Environment
Windows 11; ComfyUI 0.30.0 LOCAL; NVIDIA RTX 5090 32 GB; comfyui-mcp 0.49.5; panel 0.11.38.

### LINKED ISSUE #654: Panel does not reconnect after panel_restart_comfyui
After `panel_restart_comfyui` completed an observed down/up cycle and reported `server_ready:true`, the sidebar panel tab did not reconnect. `panel_set_workflow_target({mode:"current"})` remained deferred and `panel_graph_outline` failed with `no connected tab`, forcing a manual browser refresh. The backend recovery succeeds, but the frontend socket/canvas registration is not automatically restored.

### Steps to reproduce
1. Start with an active workflow and connected Agent panel.
2. Call `panel_restart_comfyui` and approve the confirmation card.
3. Wait until the tool confirms the ComfyUI server is healthy.
4. Call `panel_set_workflow_target({mode:"current"})`, then `panel_graph_outline`.

### Exact error (scrubbed)
`panel_restart_comfyui`: `server_ready:true`, `graph_tools_ready:false`, `panel_tab_reconnected:false`, `ready:false` after a confirmed restart cycle.
`panel_set_workflow_target`: `deferred:true`, no panel tab connected.
`panel_graph_outline`: `could not reach the ComfyUI panel ... no connected tab`.

### Fix
Precise change needed: after a ComfyUI restart, the frontend Agent panel should automatically reconnect its socket and re-register the active workflow canvas/routing key; the restart recovery path should wait for or trigger that registration so graph tools become usable without requiring a browser-tab refresh.

### Environment
Windows 11; ComfyUI 0.30.2 (local Desktop); NVIDIA RTX 5070 Ti 16 GB; comfyui-mcp 0.49.6; panel 0.11.39.

### LINKED ISSUE #663: graph_* commands keep failing "out of sync with the active workflow" for a long window after a ComfyUI restart, even across panel_reload retries

During a long session, ComfyUI was restarted several times by another agent (installing custom nodes). After one restart/reconnect, every `graph_*` bridge command (graph_outline, graph_query, graph_set_widget) started throwing:

```
The live graph is out of sync with the active workflow: the workflow reports 37 node(s), but the canvas is bound to a different graph. A load, tab switch, or reconnect left this command pointed at the wrong canvas, so it was NOT applied. Re-open the active workflow tab (panel_open_workflow) or reload the panel to rebind the graph, then retry.
```

This error is thrown by `assertGraphBoundToActiveWorkflow()` in `web/js/comfyui-mcp-panel.js` (~line 4083), called from the graph-command dispatcher (~line 4118) via `getGraphCtx()` (~line 4048). `getGraphCtx()` already has self-heal logic: `resolveScope(app)` detects a stale canvas/root binding and calls `app.canvas.setGraph(scope.rootGraph)` to reconcile (comment at ~4054: "keeps read + edit in lockstep"). In this session that self-heal did not take effect for several minutes despite multiple retries.

I initially assumed the fix was switching `panel_set_workflow_target` from `mode:"current"` to `mode:"pinned"` (with the exact active path) — that call did immediately precede success. But reading the source shows `assertGraphBoundToActiveWorkflow` is called unconditionally for every `graph_*` command (~line 11320-11326) *before* the pinned-target guard block (~line 11327 onward) even runs, and it does not consult `workflow_path`/pin mode at all — it only compares live `rootGraph`/`app.graph` node identity against the `activeWorkflowRef()` service object. So pinning cannot have been the actual mechanism that fixed it; it just happened to be the last thing I tried after several other calls (`panel_set_workflow_target(mode:"current")`, `panel_open_workflow`, `panel_reload`) had already elapsed real wall-clock time and additional frontend activity, which is the more likely reason `resolveScope` finally reconciled.

Net effect: this is very likely a **timing/race window** in the post-reconnect canvas-rebuild path, not a `mode:"current"` vs `mode:"pinned"` behavior difference — but the tool's own error message and `panel_set_workflow_target` docs point agents at `mode:"current"` as *the* recovery signal, and in this session that call reported success (`{"mode":"current","note":"Following the user's current workflow tab."}`) without the underlying desync actually being resolved yet, which is misleading (a success response for a state change that hadn't taken effect).

### Steps to reproduce

Hard to reproduce deterministically (timing-dependent), but the shape that triggered it:
1. Have an active, previously-working panel session against an open workflow tab.
2. Trigger a ComfyUI backend restart/reconnect (e.g. via a concurrent custom-node install) while the panel session keeps issuing `panel_*` graph tool calls.
3. Immediately after reconnect, call any `graph_*`-backed tool (`panel_graph_outline`, `panel_query_graph`, `panel_set_widget`) against the SAME tab that was active before and after the restart (confirmed via `panel_list_workflows` → `active_confirmed:true`, correct filename, correct node count).
4. Observe the "out of sync" error. Retry immediately — still fails.
5. Try `panel_set_workflow_target(mode:"current")` — returns success but the very next `graph_*` call still fails identically.
6. Try `panel_open_workflow(path:<same path>)` — itself errors: "workflow_open could not prove that the active canvas was rebound to the requested workflow."
7. Try `panel_reload()` — returns "soft reload (orchestrator) scheduled" but an immediate follow-up `graph_*` call still fails.
8. `panel_free_vram()` (non-graph tool) succeeds throughout — confirms base panel/session connectivity is fine; only the graph-binding path is affected.
9. After several minutes and multiple intervening tool calls, `panel_set_workflow_target(mode:"pinned", path:<same path>)` succeeds, and the immediately-following `panel_graph_outline` also succeeds with the correct 37-node graph and all prior edits intact.

### Exact error (scrubbed)

```
Error: The live graph is out of sync with the active workflow: the workflow reports 37 node(s), but the canvas is bound to a different graph. A load, tab switch, or reconnect left this command pointed at the wrong canvas, so it was NOT applied. Re-open the active workflow tab (panel_open_workflow) or reload the panel to rebind the graph, then retry.
```
(from `web/js/comfyui-mcp-panel.js` ~line 4118, thrown by `assertGraphBoundToActiveWorkflow`)

### Fix

applied locally: no
upstream-only: yes

I did not attempt a blind patch here — this is a timing-sensitive frontend state issue (LiteGraph canvas/root-graph identity vs. the Vue/Pinia workflow-service object) that I can't reliably reproduce or verify a fix for from outside the browser session, and a wrong patch to `resolveScope`/`assertGraphBoundToActiveWorkflow` risks masking real desyncs (the whole point of artokun/comfyui-mcp-panel#220/#308/#587's fail-loud design). Precise things worth investigating upstream:

1. Why `resolveScope(app)`'s staleness detection (and the `app.canvas.setGraph(scope.rootGraph)` self-heal in `getGraphCtx()`, ~line 4056-4063) can fail to reconcile for an extended period (multiple retries + a `panel_reload()`) after a backend reconnect, rather than resolving on the first or second retry.
2. Whether `panel_set_workflow_target(mode:"current")` should actively verify/force a rebind (e.g. by also calling the same reconciliation `getGraphCtx()` performs) rather than only updating the in-memory target and reporting success — right now a `mode:"current"` call can report success while the underlying canvas binding is still broken, which is misleading feedback for an agent using it as the documented recovery path.
3. Consider giving `assertGraphBoundToActiveWorkflow`'s error message an explicit hint that pinning is NOT expected to fix this class of desync (since it doesn't touch the assert at all), so agents don't waste retries on a workaround that isn't actually the mechanism, or conversely, add a genuine forced-rebind path if pinning is intended to help.

### Environment
OS Windows 11 Pro 10.0.26200 · ComfyUI 0.29.0 · GPU NVIDIA GeForce RTX 5090 (32GB VRAM) · comfyui-mcp 0.49.7 · panel 0.11.35 (Manager 3.x legacy)

artokun:
The remaining halves of the restart/reconnect family after the #621/#622/#623/#624/#655 wave:
- no proactive canvas/tab rebind on 'reconnected'; unsettled restore hard-refuses until manual recovery
- stale tmp: routing keys not invalidated on frontend reconnect
- graph_enter_subgraph: bare canvas.openSubgraph, no scope/binding receipt, no post-nav reconcile
- mutations not gated until the post-restart socket is proven stable
- tab that never reconnects still requires a manual browser refresh
- panel-side half of #635: refresh_nodes stuck at refreshed:false with no reason/remedy (comfyui-mcp-panel.js:6934)

closes #617
closes #619
closes #646
closes #654
closes #663
refs #635 (panel half only — the restart dispatched:true/wrong-port half is orchestrator-side, tracked in comfyui-mcp)

### DIFF
--- a/browser_tests/unit/canvas-navigation.test.mjs
+++ b/browser_tests/unit/canvas-navigation.test.mjs
@@ -416,10 +416,125 @@ test("#619: the shipping enter does not claim settled when the canvas moved afte
   assert.match(reply.note, /binding check passed, but the view has already moved elsewhere/);
 });
 
-test("#619: the shipping exit carries the same receipt wiring", () => {
-  // Source-level pin: exit must run the same confirmCanvasNavigation receipt
+test("#619: the shipping exit carries the same receipt wiring", () => {  // Source-level pin: exit must run the same confirmCanvasNavigation receipt
   // (delete it and this fails) and must keep the immediate-parent resolution.
   const body = extractMethod("async graph_exit_subgraph()");
   assert.match(body, /await confirmCanvasNavigation\(\{/, "exit runs the receipt");
   assert.match(body, /findSubgraphOwner\(rootGraph, graph\)\?\.parentGraph/, "exit keeps the immediate-parent walk (#412)");
+});
+
+// ---------------------------------------------------------------------------
+// r9: the terminal observation is three-state — an UNREADABLE canvas read is
+// neither "still on target" nor "moved elsewhere"
+// ---------------------------------------------------------------------------
+
+function buildExitSubgraph(doubles) {
+  const body = extractMethod("async graph_exit_subgraph()").replace(
+    /^async graph_exit_subgraph\(/,
+    "async function graph_exit_subgraph(",
+  );
+  const factory = new Function(
+    "getGraphCtx",
+    "findSubgraphOwner",
+    "confirmCanvasNavigation",
+    "describeActiveGraph",
+    "assertGraphBoundToActiveWorkflow",
+    "coerceMessageText",
+    `return (${body});`,
+  );
+  return factory(
+    doubles.getGraphCtx,
+    doubles.findSubgraphOwner,
+    doubles.confirmCanvasNavigation ?? confirmCanvasNavigation,
+    doubles.describeActiveGraph,
+    doubles.assertGraphBoundToActiveWorkflow,
+    doubles.coerceMessageText ?? ((v) => String(v)),
+  );
+}
+
+test("#619: an UNREADABLE terminal canvas read discloses uncertainty, not displacement (codex r9, enter)", async () => {
+  const rootGraph = { _nodes: [], name: "root" };
+  const sub = { _nodes: [], name: "sub", rootGraph };
+  const node = { id: 105, type: "SubgraphNode", subgraph: sub };
+  let reads = 0;
+  const canvas = {
+    _g: rootGraph,
+    // Reads 1-4 complete the receipt (landing, assert's ctx read, survival
+    // re-read); the reply's own reads throw.
+    get graph() {
+      reads += 1;
+      if (reads >= 5) throw new Error("canvas getter exploded");
+      return this._g;
+    },
+    openSubgraph(s) {
+      this._g = s;
+    },
+    setDirty() {},
+  };
+  const app = { graph: rootGraph, canvas };
+  const enter = buildEnterSubgraph({
+    getGraphCtx: () => ({ app, graph: canvas.graph, rootGraph, canvas }),
+    resolveNode: () => node,
+    describeActiveGraph: (g) => ({ scope: g === rootGraph ? "root" : "subgraph" }),
+    assertGraphBoundToActiveWorkflow: () => {},
+  });
+  const reply = await enter({ node_id: 105 });
+  assert.equal(reply.settled, false);
+  assert.match(reply.note, /could not determine which/, "uncertainty is disclosed");
+  assert.doesNotMatch(reply.note, /navigated away|moved elsewhere/, "no displacement claim without an observation");
+  assert.match(reply.note, /panel_graph_outline/, "the remedy is a scope read");
+});
+
+test("#619: the shipping exit settles at root on a clean navigation (driven, not just source-pinned)", async () => {
+  const rootGraph = { _nodes: [], name: "root" };
+  const sub = { _nodes: [], name: "sub", rootGraph };
+  const canvas = {
+    _g: sub,
+    get graph() {
+      return this._g;
+    },
+    setGraph(g) {
+      this._g = g;
+    },
+    setDirty() {},
+  };
+  const app = { graph: rootGraph, canvas };
+  const exit = buildExitSubgraph({
+    getGraphCtx: () => ({ app, graph: canvas.graph, rootGraph, canvas }),
+    findSubgraphOwner: () => ({ id: 105, parentGraph: rootGraph }),
+    describeActiveGraph: (g) => ({ scope: g === rootGraph ? "root" : "subgraph" }),
+    assertGraphBoundToActiveWorkflow: () => {},
+  });
+  const reply = await exit();
+  assert.equal(reply.settled, true);
+  assert.equal(reply.viewing.scope, "root");
+});
+
+test("#619: an UNREADABLE terminal canvas read discloses uncertainty, not displacement (codex r9, exit)", async () => {
+  const rootGraph = { _nodes: [], name: "root" };
+  const sub = { _nodes: [], name: "sub", rootGraph };
+  let reads = 0;
+  const canvas = {
+    _g: sub,
+    get graph() {
+      reads += 1;
+      if (reads >= 5) throw new Error("canvas getter exploded");
+      return this._g;
+    },
+    setGraph(g) {
+      this._g = g;
+    },
+    setDirty() {},
+  };
+  const app = { graph: rootGraph, canvas };
+  const exit = buildExitSubgraph({
+    getGraphCtx: () => ({ app, graph: canvas.graph, rootGraph, canvas }),
+    findSubgraphOwner: () => ({ id: 105, parentGraph: rootGraph }),
+    describeActiveGraph: (g) => ({ scope: g === rootGraph ? "root" : "subgraph" }),
+    assertGraphBoundToActiveWorkflow: () => {},
+  });
+  const reply = await exit();
+  assert.equal(reply.settled, false);
+  assert.match(reply.note, /could not determine which/);
+  assert.doesNotMatch(reply.note, /navigated away|moved elsewhere/);
 });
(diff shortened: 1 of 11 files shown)


artokun:

Issue #350, opened 2026-07-30
[Feature] Integrate 'record a skill' into the panel


artokun:
Requested by **seanmcmagic** in Discord #help.

**Ask:** add a 'record a skill' flow in the panel — capture a sequence of agent/panel actions and save it as a reusable skill (aligning with the existing skills system the agent already loads).

**Notes / feasibility:** the agent already has a skills mechanism (plugin/skills + read_skill/list_skills). A panel 'record' affordance would let a user demonstrate a workflow once and persist it as a named skill for replay. Product-direction call for @artokun to prioritize.

(Inspiration reel Sean shared: instagram.com/reel/DbbELp8uf4s — not fetched/acted on; linked for context only.)

artokun:
Parked: feature requests are on hold while the bug backlog is worked (PR #678 remains a draft placeholder).

artokun:
Claimed by grok-swarm 2026-08-19T09:09:57Z — driving this in worktree `wt-350` (branch `feat/350-record-a-skill`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp-panel/pull/1398 (worktree C:/Users/Artokun/code/comfyui-mcp-panel/.worktrees/wt-350).

artokun:
Shipped — closing, since the PR merged and the issue was left open.

`/record-skill` saves the open graph as a reusable skill: panel PR #1398, merged 2026-08-19T09:31:48Z.

Closing manually because a same-repo owner-qualified reference does not auto-close, so this sat open for six hours after its own fix landed. Worth noting for the next release audit rather than as a defect in the work.

PR #678, opened 2026-08-05
feature: record-a-skill capture flow in the panel

artokun:
**Status: placeholder branch — no implementation.** `cluster/record-a-skill` carries one empty
commit (`e29ca85d`, subject prefixed `placeholder:`) plus merges of `main` taken purely as
upkeep. It contributes **zero changed files**: `git diff origin/main...cluster/record-a-skill`
lists nothing. This PR tracks the request; it does not deliver it.

**Re-measured 2026-08-14 against `main` at `afb8db64` (v0.14.41).** `main` is now merged in
rather than merely test-merged, and the branch tip's tree is byte-identical to `main`'s:
both are `ca1b58006485ec65ff3f7efbbd37e52f2afe6800`. (An earlier revision of this body cited
tree `1a468145`; that was a correct measurement against an older `main` and is superseded —
the number moves every time `main` does, so treat the *equality* as the claim, not the hash.)
`npm run test:unit` on that tip: **4409 tests, 4408 pass, 0 fail, 1 todo** — which is exactly
`main`'s own result, necessarily, since the trees are identical.

**Parked under the feature freeze (bug/stability only).** Not a merge candidate. Kept open
and in draft so the request is not dropped.

Scope: integrate a "record a skill" capture flow into the panel. Refs #350.

**State of main (re-verified 2026-08-14):** no capture flow exists. A search of `web/` for a
record/capture entry point (`record.a.skill`, `recordSkill`, `skill.capture`, `captureSkill`)
returns nothing. Skills are surfaced only passively, as SDK slash commands pushed by the
orchestrator into the completion menu — `web/js/comfyui-mcp-panel.js` describes them at the
two sites that handle them as "SDK slash commands (built-ins like `/compact`, plus any loaded
skills)". Reading and running a skill is all that exists; there is no way to author one from
the panel.


artokun:
Parked: feature requests on hold while the bug backlog is worked. This draft stays as the placeholder; do not pick up until unparked.

Issue #428, opened 2026-07-31
[Feature] Video-gen app: LTX 2.3/Director + Wan 2.2 + Bernini + Hunyuan Video backend, panel agent as UI driver


artokun:
Requested by **seanmcmagic** in Discord #help.

**Ask:** a dedicated app/product built on top of a video-gen node stack — LTX 2.3, LTX Director, Wan 2.2, Bernini, Hunyuan Video, and the Easy-Use Media nodes as the ComfyUI backend — with a new panel agent as the UI driver that controls all generation parameters. The pitch: bring "Seedance 2.5"-style capabilities to an open-source stack, agent-operated end to end (not just workflow execution — the agent reasons about and sets the params).

**Context from the thread:** Sean separately noted the existing `Dispatch` tool already resembles part of what this would need (background/parallel task execution), just wired differently than what he's picturing here — i.e. this may be less "new engine" and more "new panel surface + agent skill set" over stack we largely already have pieces of (agent tool-use, dispatch, node/workflow control).

**Notes / feasibility:** this is a big architectural direction (new app surface, not an incremental panel feature) — needs @artokun's call on scope before any design/build work starts. Flagging for product-direction discussion rather than immediate implementation.

Reference for Sean's earlier related asks in the same thread: #350 (record a skill), #351 (cowork mode).

artokun:
Parked: feature requests are on hold while the bug backlog is worked (PR #679 remains a draft placeholder).

artokun:
Claimed by grok-swarm 2026-08-19T09:09:57Z — driving this in worktree `wt-428` (branch `feat/428-video-gen-app`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp-panel/pull/1397 (worktree C:/Users/Artokun/code/comfyui-mcp-panel/.worktrees/wt-428).

artokun:
Shipped — closing, since the PR merged and the issue was left open.

Converting a video-gen workflow to an app: panel PR #1397, merged 2026-08-19T09:39:42Z.

Closing manually because a same-repo owner-qualified reference does not auto-close, so this sat open for six hours after its own fix landed. Worth noting for the next release audit rather than as a defect in the work.

PR #679, opened 2026-08-05
feature: video-gen app (LTX 2.3/Director + Wan 2.2 + Bernini + Hunyuan), panel agent as UI driver

artokun:
**Status: placeholder branch — no implementation.** `cluster/video-gen-app` carries one empty
commit (`ed7c651d`, subject prefixed `placeholder:`) plus merges of `main` taken purely as
upkeep. It contributes **zero changed files**: `git diff origin/main...cluster/video-gen-app`
lists nothing. This PR tracks the request; it does not deliver it.

**Re-measured 2026-08-14 against `main` at `afb8db64` (v0.14.41).** `main` is now merged in
rather than merely test-merged, and the branch tip's tree is byte-identical to `main`'s:
both are `ca1b58006485ec65ff3f7efbbd37e52f2afe6800`. (An earlier revision of this body cited
tree `1a468145`; that was a correct measurement against an older `main` and is superseded —
the number moves every time `main` does, so treat the *equality* as the claim, not the hash.)
`npm run test:unit` on that tip: **4409 tests, 4408 pass, 0 fail, 1 todo** — which is exactly
`main`'s own result, necessarily, since the trees are identical.

**Parked under the feature freeze (bug/stability only).** Not a merge candidate. Kept open
and in draft so the request is not dropped.

Scope: video-gen app surface, with the panel agent as UI driver. Refs #428.

**State of main (re-verified 2026-08-14):** only targeted LTXDirector widget-write support
exists — `web/js/lib/ltx-director.js`, written for #314. There is no video-gen app surface.

Stated precisely, because a naive grep contradicts it: the names *do* occur in `web/`, but
never as support for generating with these models. `Hunyuan` appears once as
`"EmptyHunyuanLatentVideo"` inside `cmcp-apps.js`'s generic node-type allowlist (alongside
`EmptyLTXVLatentVideo` and `EmptyMochiLatentVideo`) used only to classify widgets into app
inputs, and `Wan` / `Hunyuan` appear in `cmcp-civitai.js` as CivitAI **base-model name
strings** for model browsing and disambiguation. `Bernini` occurs nowhere. None of that is a
generation path.

**Model-naming caveat.** The model list in the title dates from when #428 was filed and
should be re-checked before any implementation rather than treated as a spec. "Wan 2.2" is
two generations behind and drifting further: `main`'s CivitAI base-model list now carries
`"Wan Video 2.7"` and `"Wan Image 2.7"` as well as `"Wan Video 2.5 T2V"` and the older
`"Wan Video 2.2 TI2V-5B"` (`web/js/cmcp-civitai.js`) — at the last pass 2.5 was the newest
entry there, so this list moved again inside a week. LTX 2.3, Hunyuan Video and Bernini
could not be confirmed superseded from this repo — verify current versions at implementation
time rather than inheriting these names.


artokun:
Parked: feature requests on hold while the bug backlog is worked. This draft stays as the placeholder; do not pick up until unparked.

artokun:
will focus on a separate h3 director, closing

Issue #710, opened 2026-08-07
panel_show_media renders audio files (.mp3/.wav/.flac) as broken <img> instead of an audio player


artokun:
### What happened / root cause
`panel_show_media` accepts any ComfyUI output ref, but the panel only renders **image** and **video** media cards. Passing an AUDIO output (`.mp3`) returns `ok:true, painted:N, unconfirmed:0` — a full success — while the chat actually shows a **broken image icon** with the caption underneath. The user sees nothing playable and reasonably replies "??????????????????".

This is a silent-wrong-output bug on two levels:
1. The card renders as `<img>` for an audio MIME type instead of an `<audio controls>` element (or at minimum a download link).
2. The tool reports `painted` success, so the agent believes the user can hear the file and continues the workflow on that false premise. There is no signal that the kind was unsupported.

Note `upload_image action:"stage"` already has a `kind: image|video|audio` concept, so audio is a first-class media kind elsewhere in the suite — the panel card renderer just doesn't cover it.

### Steps to reproduce
1. Generate or place any audio in ComfyUI's output dir (e.g. via TTS-Audio-Suite → `SaveAudioMP3`, prefix `synlara/vo_sophie` → `output/synlara/vo_sophie_00001.mp3`).
2. Call:
   `panel_show_media({items:[{source:{filename:"vo_sophie_00001.mp3", subfolder:"synlara", type:"output"}, caption:"..."}]})`
3. Tool returns `{ok:true, count:3, painted:3, unconfirmed:0}`.
4. Panel chat shows broken-image placeholders (screenshot confirmed by the user), no playback control.

### Exact error (scrubbed)
No error is raised — that is the defect. Tool result:
```json
{"ok":true,"count":3,"painted":3,"unconfirmed":0,"previews":[],
 "note":"3 items were displayed to the USER in the panel chat."}
```
Rendered result: three broken `<img>` icons.

### Fix
applied locally: no · upstream-only: no (panel-side renderer change; I did not patch it this session to avoid a second local divergence on top of the #706 patch already applied to this file)

[...]

Issue #706, opened 2026-08-06
Legacy Manager 3.x security-gated install surfaces as "ComfyUI-Manager not reachable" (404 body ignored in managerV2/managerCall)


artokun:
### What happened / root cause
`panel_install_node` with a git URL against a LEGACY ComfyUI-Manager **V3.41** failed with `Error: ComfyUI-Manager not reachable (is the built-in Manager enabled?)` — while the Manager was provably up (`GET /manager/version` → `V3.41`, `GET /manager/queue/status` → 200 at the same moment).

Root cause: legacy Manager 3.x answers a **security-gated** operation (git-URL install of a pack outside its registry, default security_level) with **HTTP 404 + body "A security error has occurred. Please check the terminal logs"**. In `web/js/comfyui-mcp-panel.js` both `managerV2()` and `managerCall()` treat EVERY 404 as "not reachable" and never read the body, so a refusal is mislabeled as unreachability, and the 404 also gets tagged `managerRouteMissing = true` (the #605 re-send authorization) even though a handler DID run and refuse.

Same-session side effect: an earlier queued git-URL install (Comfy-Org/Nvidia_RTX_Nodes_ComfyUI) was security-blocked the same way — Manager marked the queue task done, pack never appeared (the documented queue-done≠installed trap, but with zero hint that the security gate was the reason).

### Steps to reproduce
1. ComfyUI with legacy ComfyUI-Manager 3.x (V3.41), default security level, network_mode public.
2. `panel_install_node` with `repository: <any git URL not in the legacy registry>` (e.g. https://github.com/WhatDreamsCost/WhatDreamsCost-ComfyUI).
3. Panel throws "ComfyUI-Manager not reachable (is the built-in Manager enabled?)" although the Manager responds on /manager/version.

### Exact error (scrubbed)
```
Error: ComfyUI-Manager not reachable (is the built-in Manager enabled?)
```
while direct probe at the same time:
```
GET /manager/version → V3.41 ; GET /manager/queue/status → 200
POST /manager/queue/install → 404, body: "A security error has occurred. Please check the terminal logs"
```

### Fix
applied locally: yes (runtime verification pending a frontend hard-refresh — the patched JS loads on next tab reload) · upstream-only: no

[...]

Issue #705, opened 2026-08-06
CivitAI search errors drop the upstream response body — user sees a bare "503: Service Unavailable" with no cause and no retry hint


artokun:
### What happened / root cause

User searched `min` on the CivitAI **LoRA** tab and got `CivitAI error: CivitAI API 503: Service Unavailable`, repeatedly. Because the message carries only the HTTP status, it reads as query-specific ("my search for 'min' is broken") and the user went looking for a bug in our code.

It is actually a **CivitAI-side outage of their model-search backend**, and their API says so in the response body. Verified from this machine:

```
GET https://civitai.com/api/v1/models?limit=5&types=LORA&query=min    -> 503 {"error":"Model search is temporarily overloaded — please retry."}
GET .../models?limit=3&types=LORA&query=flux                          -> 503 (same)
GET .../models?limit=3&types=LORA&query=anime                         -> 503 (same)
GET .../models?limit=3&types=LORA          (NO query param)           -> 200 OK, items returned
```

So: every `query=` search 503s right now; browsing without a query works. The actionable string "Model search is temporarily overloaded — please retry." IS present in the upstream body, and `py/civitai_proxy.py` already forwards that body verbatim with the upstream status (`web.Response(body=..., status=resp.status, content_type="application/json")`, ~line 285). The panel client then **throws it away**: `_request()` in `web/js/cmcp-civitai.js` builds its Error from `res.status`/`res.statusText` only and never reads the body. All the diagnostic value the proxy carefully forwarded is discarded at the last hop.

Secondary observation (not fixed here): the proxy retries 3x on 429/503 and then surfaces a hard failure. For a known-flaky endpoint it would be worth telling the user "CivitAI's search is down — browsing without a search term still works", since that fallback demonstrably works during the outage.

### Steps to reproduce

1. Open the docked CivitAI browser, LoRA tab.
2. Type any search term (during a CivitAI model-search outage).
3. Error card reads `CivitAI API 503: Service Unavailable` — no cause, no hint that it is upstream-wide, no hint that unfiltered browsing still works.

### Exact error (scrubbed)

`CivitAI error: CivitAI API 503: Service Unavailable`

Upstream body that was dropped: `{"error":"Model search is temporarily overloaded — please retry."}`

### Fix

applied locally: **yes**  ·  upstream-only: **no**

Read the forwarded body (best-effort, bounded, never allowed to mask the original error) and append it to the message; also expose it as `err.detail` so the UI can render it separately.

[...]

Issue #703, opened 2026-08-06
Connection popover "Prompts" button 404s — /prompts route does not exist on the console server


artokun:
### What happened / root cause

The **Prompts** button in the connection popover (`● connected` → `Prompts`) opens a new tab that renders:

```json
{"ok":false,"error":"not_found"}
```

**Root cause: the panel opens a route the orchestrator does not serve.**

- Panel builds the URL at `web/js/comfyui-mcp-panel.js:15617`:
  ```js
  window.open(`${cmcpConsoleUrl}/prompts?token=[REDACTED] "_blank", "noopener");
  ```
- The console HTTP server only registers these GET routes (`dist/orchestrator/panel-console-http.js`):
  - `/api/secrets` (370)
  - `/api/status` (415)
  - `/api/lora-preview` (431)
  - `/credentials` (435)
  - `/` and `/console` (445)

There is **no `/prompts` page route and no `/api/prompts` data route** anywhere in `dist/` — grep for `api/prompts` returns nothing. A nearby panel comment at line 15604 implies both were expected to exist:

> `// page is same-origin to the console server, so its /api/prompts fetches need no ...`

So the feature appears unimplemented server-side rather than mis-wired.

**Not a version-skew issue:** `npm view comfyui-mcp version` → `0.49.8`, which is the running build. The latest published mcp has no `/prompts` route, so upgrading will not fix it.

### Steps to reproduce
1. Open the ComfyUI Agent Panel.
2. Click the `● connected` status pill.
3. Click **Prompts**.
4. A new tab opens at `http://127.0.0.1:9183/prompts?token=[REDACTED] showing `{"ok":false,"error":"not_found"}`.

### Exact error (scrubbed)
```
GET http://127.0.0.1:9183/prompts?token=[REDACTED]  ->  404
{"ok":false,"error":"not_found"}
```
Probed alternates on :9183, all 404: `/prompt`, `/system-prompts`, `/settings`, `/config`, `/api/prompts`.
Other ports: `:9180/prompts` → 426 (expects websocket upgrade), `:9181/prompts` → 400.
`:9183/` itself serves fine (title: "ComfyUI MCP Console").

### Fix
applied locally: **no**   upstream-only: **yes**

Not patched locally: the missing piece is an entire prompts-editor page plus its `/api/prompts` GET/PUT handlers, which is a feature implementation rather than a minimal reversible patch.

Precise change needed — one of:
[...]

PR #722, opened 2026-08-07
0.11.43 — ten fixes from a day of live panel reports

artokun:
Version bump so today's fixes actually reach users. **Merging this publishes to the Comfy Registry** (the publish fires on the `pyproject.toml` change).

Every fix below came from a **via-panel report filed in the last day**, and none has shipped — 0.11.42 carried the 0.50.0 tool-surface migration, and all nine landed after it.

| issue | fix |
|---|---|
| **#711** | `ask_user` / `request_secret` fenced to the conversation whose turn asked. An abandoned turn could paint a **secure-input card into whatever conversation the tab was showing**, and the token typed there returned as that turn's result |
| **#695** | socket-vs-widget classified structurally, so comma-joined unions (`"IMAGE,MASK"`) stop failing closed — **17 more node classes addable, 0 newly refused** across all 2887 |
| **#700** | the backend node def is snapshotted, not read by reference, so a refresh can't mutate the "backend truth" the guard compares against — `LoadImage` had become permanently unaddable |
| **#706** | a Manager **security refusal** no longer reports as "not reachable", and no longer carries the flag that authorizes a mutation **re-send** |
| **#710** | `panel_show_media` **plays** audio instead of painting a broken image, and a kind it cannot present is no longer counted as `painted` |
| **#705** | a CivitAI outage surfaces the upstream reason instead of a bare `503`, so it stops reading as a broken search |
| **#696 #701 #702 #663** | a content difference is wrong-canvas evidence only when **structural** — the guard compared full `serialize()` against a tracker that only snapshots on *user* input, so any node-driven widget write looked like a different graph |
| **#698** | a widget that structurally cannot hold a value says so, instead of a retryable-sounding "did not retain the requested value" |
| **#703** | the **Prompts** button no longer opens a console page this build doesn't serve; it disables itself on a definitive 404 and re-enables when the route ships |

Two themes run through them, and they're the same classes the 0.50.0 work kept hitting: **false refusals** (valid operations rejected) and **dishonest results** (a tool reporting success, or a wrong cause). Both mislead the agent rather than merely inconveniencing the user.

## Verification

- full suite **2665/2665**
- panel **parses and links** as an ES module
- `PANEL_VERSION` and `pyproject` bumped together — they must match or the publish is wrong
- control-byte scan 0 on both changed files

Each underlying PR was independently mutation-verified before merge, and `main` was re-verified composed afterwards — each was CI-green against a *different* `main`, which is exactly the stale-green trap that produced a regression earlier in this cycle.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "comfyui-agent-panel"
 description = "Autonomous AI agent in the ComfyUI sidebar - local-first, bring any LLM. Run it on your Claude or ChatGPT subscription (no API keys, no extra LLM costs), on a Kimi K3 / GLM / OpenRouter key, or fully offline on free local models via Ollama, LM Studio, or llama.cpp. The agent drives your live canvas at full parity: add, wire, move and tweak nodes with Ctrl+Z undo, load whole workflows and installer packs in one shot, generate images, video and audio, browse CivitAI, and run your workflow - asking before it spends paid API credits. The sidebar UI for comfyui-mcp, the local-first, agent-native control plane for ComfyUI."
-version = "0.11.42"
+version = "0.11.43"
 license = { file = "LICENSE" }
 requires-python = ">=3.9"
 # UI-only pack: no Python deps. The frontend floor guarantees
--- a/web/js/comfyui-mcp-panel.js
+++ b/web/js/comfyui-mcp-panel.js
@@ -843,7 +843,7 @@ const DOCS_URL = "https://comfyui-mcp.artokun.io/docs";
 // Panel version — surfaced in the "Need help?" diagnostics blob. Bump via
 // `node scripts/set-version.mjs <v>` (updates this AND pyproject together); CI
 // and the publish gate FAIL if the two ever drift, so this can't go stale.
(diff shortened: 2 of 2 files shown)


artokun:
Refreshed — now **ten** fixes. Added since the last update:

- **#703** — the **Prompts** button no longer opens a console page this build doesn't serve. It disables itself on a definitive 404 (and only a 404 — a flaky console leaves it enabled) and re-enables automatically the day the route ships.
- **#712** *(P1)* — an empty CivitAI model grid is now **attributable**. Every search was returning zero results with `error: null`, indistinguishable from "nothing matched", because `browsingLevels` defaults to PG-only and a model whose covers all sit outside the mask is dropped entirely. The filter and the default are **unchanged** — what it removes is a content-policy decision — but the silence is gone: the result now reports how many the level hid. Counted *before* the keyword filter, since a keyword miss is "no match", not "hidden by your level".

Rebased onto current `main`, re-verified: full suite **2670/2670**, panel parses and links as an ES module, `pyproject` and `PANEL_VERSION` both at 0.11.43.

Still one word from shipping — merging publishes to the Comfy Registry.

Issue #688, opened 2026-08-06
panel_* graph reads permanently wedged with "workflow instance mismatch" after reconnect — neither panel_open_workflow nor panel_set_workflow_target({mode:'current'}) clears it


artokun:
### What happened / root cause
Every `panel_*` graph call fails with `workflow instance mismatch` after the panel session reconnected (ComfyUI was restarted mid-task) and the user switched the open workflow tab. The panel prepends the standard "user switched the open workflow" notice, so the orchestrator/panel clearly knows about the switch — yet the instance guard still rejects every command against the now-active tab.

Neither documented recovery clears it:
1. `panel_set_workflow_target({mode:"current"})` returns `{"mode":"current","note":"Following the user's current workflow tab."}` — the very next `panel_graph_outline` still errors with the same mismatch.
2. `panel_open_workflow("rife")` SUCCEEDS and returns a fresh routing key/uuid (`routing_key: "wf:workflows/rife.json"`, `workflow_uuid: 7e96d101-…`, `open_seq: 3`) — the next `panel_graph_outline` still errors with the same mismatch.
3. Pinning explicitly (`panel_set_workflow_target({mode:"pinned", path:"workflows/rife.json"})`) also succeeds, and reads still fail.

So the session is permanently wedged: the guard's cached instance identity is never refreshed by any of the three tools whose documented job is to refresh it. Likely the guard compares against an identity captured at socket/session establishment (pre-reconnect) rather than the one returned by the most recent `panel_open_workflow` / `panel_set_workflow_target`.

Two secondary problems:
- **Read-only, non-graph tools are gated by the same guard.** `panel_list_workflows` (which just lists open tabs and should be instance-independent) fails with the identical mismatch — so the agent cannot even enumerate tabs to recover. That turns a recoverable state into a dead end.
- **The error text is misleading.** It tells the agent to run exactly the two recoveries that do not work, so an agent will loop on them.

### Steps to reproduce
1. Have a panel session working on workflow A.
2. ComfyUI restarts / the panel connection drops and the session resumes.
3. User switches the canvas to a different workflow (here: `rife`).
4. Call any `panel_*` read (`panel_graph_outline`, `panel_find_nodes`, `panel_list_workflows`) -> mismatch error.
5. Run `panel_set_workflow_target({mode:"current"})` (succeeds) -> retry read -> same error.
6. Run `panel_open_workflow("rife")` (succeeds, returns new uuid/open_seq) -> retry read -> same error.
[...]

PR #759, opened 2026-08-07
fix(fence): let the recovery probe through both target guards

artokun:
## The circularity

`workflow_list` was refused by both canvas-target guards — the workflow-instance uuid
fence (#570) and the pinned-target guard (#349/#186). It is also the **only** probe the
recovery path has: `rebindWorkflowFence()` re-derives a session's target from the active
record `workflow_list` returns.

So repairing a stale target required the target to already be correct:

```
stale stamp
  → panel_set_workflow_target({mode:"current"})
    → rebindWorkflowFence()
      → ctx.call({cmd:"workflow_list"})
        → REFUSED: "workflow instance mismatch"     ← by the very fence being repaired
  → "Could NOT read the live canvas identity … still fenced to the previous
     workflow instance, which the panel will keep refusing."
```

That is what artokun/comfyui-mcp#932, #607 and #688 describe as *"the documented recovery
paths do not clear the guard"*. The advertised remedy was a no-op and a browser reload was
the only exit.

## Reproduced live before fixing

On released builds — orchestrator `mcp=0.50.14`, panel `panel=0.11.44` — a plain
`panel_new_workflow` → `panel_query_graph` → `panel_add_node` → `panel_query_graph`
sequence produced one success and three hard `workflow instance mismatch` refusals, with
the failed rebind quoting the refusal as its own underlying cause. `node_count: 0`
afterwards confirms the refusals were real: the node never landed.

## Why the exemption is safe

Both guards exist to stop an operation landing on the **wrong** workflow. `workflow_list`
lands on nothing:

- it mutates no graph;
- it selects no target — it enumerates *every* open tab rather than naming one;
- its reply is tab metadata (path, title, routing key, active/modified flags), never
  graph content;
- its handler reads the live binding at execution time (`liveWorkflowListActive()`), so a
  stale stamp cannot make it answer for a stale tab — the reply describes whatever is
  genuinely active, which is precisely the fact the caller needs;
- unlike a graph read, the reply **carries its own identity**, so a caller holding a stale
  expectation can see the difference instead of misattributing the result;
- the orchestrator still corroborates the active record against the open list
  (`corroborateActiveForFence`) before adopting a uuid from it.

## Both guards, not one

The new predicate `commandIsCanvasTargetless` feeds the fence **and** the pin guard.
Exempting the fence alone would have left a pinned session wedged the same way one guard
over — with the pin refusal advertising the very `mode:"current"` re-target that runs
through `workflow_list`.

It stays **out** of `CANVAS_INDEPENDENT_COMMANDS`: `workflow_list` does observe the canvas,
so claiming otherwise would falsify that set's contract. Canvas-observing, not
canvas-targeting.

## Tests

50 in the fence file, 2831 in the panel suite, plus typecheck and the vocabulary gate.
Mutation-verified on both classes:

| mutation | result |
|---|---|
| delete the exemption from the classifier | 6 tests fail |
| revert the pin-guard call site to the narrow predicate | wiring test fails |

Marker counts were asserted before and after each mutation — these files are CRLF, where a
naive substitution silently no-ops and reports a false clean sweep.

Two existing tests asserted the old behaviour on the premise that the reply "describes the
active workflow, so a stale-stamped call would answer for the wrong tab". Both are updated
**in place** with why that premise does not hold, rather than deleted.

## Not claimed

This does not touch the second failure mode seen in the same session — an orchestrator-side
`no connected tab can be chosen … issued from multiple workflows at once`, which fires when
`panel_new_workflow` moves the canvas mid-turn and is self-clearing on the next
single-origin message. That is a separate defect and is filed separately.

Does not close the reports on its own: they close when someone confirms on a released build.

### DIFF
--- a/browser_tests/unit/workflow-chat-identity.test.mjs
+++ b/browser_tests/unit/workflow-chat-identity.test.mjs
@@ -7,7 +7,9 @@ import { dirname, join } from 'node:path'
 import {
   activeWorkflowFenceApplies,
   commandIsCanvasIndependent,
+  commandIsCanvasTargetless,
   commandWorkflowMismatch,
+  commandTargetsActiveWorkflow,
   hasEmbeddedUuidSuccessionEvidence,
   selectorSearchIncludesListed,
   selectorTargetsNonActiveWorkflow,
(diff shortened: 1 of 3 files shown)


artokun:
## Verified end to end on a live rig

All three fixes applied together, then the exact reported sequence run over the loopback panel MCP — **no agent involved**, so nothing could paper over a refusal.

Setup: orchestrator built from this branch (`mcp=0.50.15` + this fix), panel carrying #759 and #761, ComfyUI 0.30.2, a **fresh browser document** (verified by `performance.timeOrigin`, after every patch — a stale module is the standing false-negative on this rig). All three fixes confirmed present in the *running* module, not merely on disk:

```
fix759_predicate        true      fix759_listExempt      true
fix759_pinGuard         true      fix761_unsavedBranch   true
fix759_graphStillFenced true   ← the fence still refuses graph work
```

The tab under test was an **unsaved** workflow — the #761 scenario exactly.

| step | before | after |
|---|---|---|
| `panel_new_workflow` | `created:true`, fence left stale | `created:true`, key `tmp:c2512bcc…` |
| `panel_query_graph` | `workflow instance mismatch` | **success** — 0 nodes |
| `panel_add_node` | `workflow instance mismatch` | **success** — node id `1` |

And confirmed against the live canvas rather than the tool's own reply (the #710 lesson — a handler must be checked against observed effect, not its own report):

```js
activeTitle: "Unsaved Workflow (2)"   // the workflow new_workflow created
nodeCount:   1                        // the CheckpointLoaderSimple genuinely landed
```

Before these fixes the same sequence produced one success and three hard refusals, with `node_count: 0` proving nothing landed.

### One observation, not a claim

The 4th call failed with `no connected tab with id "tmp:10cc"` — my harness pinned a tab id by prefix, and that id migrated to `tmp:c251` when the new workflow was created. Calls 2 and 3 routed fine, so this is my test harness holding a stale pin across a workflow switch, not a reproduction of the reported bug. Whether the loopback MCP's path resolution should consult the same migration aliases `resolveTarget` does is worth a look on its own, and I have not investigated it.

This does not close the reports — they close when someone confirms on a released build.

Issue #716, opened 2026-08-07
Add bulk widget writes to avoid repeated full object_info fetches


artokun:
### What happened / root cause
A prompt-editing task required 29 panel_set_widget calls on one live workflow. Each successful widget write performs a fresh full ComfyUI /object_info fetch before mutation. On an installation with many custom nodes this makes single writes very slow, and the latency scales badly for multi-widget edits.

The production path in comfyui-mcp-panel web/js/comfyui-mcp-panel.js, GRAPH_TOOL_EXECUTORS.graph_set_widget, awaits awaitObjectInfoHistorySeed() and then runSetWidget(...). Its getFreshObjectInfo callback calls api.getNodeDefs() for every individual write. The security/correctness fence is valid, but the single-widget API has no way to amortize the full backend metadata fetch across a deliberate batch.

### Steps to reproduce
1. Open a normal saved workflow in the sidebar panel.
2. Call panel_set_widget repeatedly for 20-30 distinct text/combo widgets on the same active workflow.
3. Sequential calls take minutes; a 29-call sequence had to be terminated after about 3 minutes and had only partially completed.
4. Retrying independent writes concurrently completed correctly, but batches of 4 and 7 writes each still took roughly 40-45 seconds before acknowledgements returned.
5. panel_query_graph confirmed the writes had applied; there was no workflow error.

### Exact error
No thrown error. This is a severe latency/awkward-workaround defect: successful mutations are delayed by repeated full fresh /object_info authorization, and terminating the long batch leaves an unknown partial-application state.

### Fix
applied locally: no
upstream-only: yes

Add a bulk panel_set_widgets / graph_set_widgets operation that:
- accepts multiple {node_id, widget, value} edits for one stamped workflow;
- fetches and records fresh /object_info once;
- verifies the workflow stamp again immediately before the first write and between synchronous writes if needed;
- authorizes every concrete target against that same fresh snapshot;
- applies the batch in one undo envelope and returns per-edit previous/new values;
- fails atomically during preflight before mutation when any edit is invalid.

Alternatively, safely coalesce concurrent getNodeDefs() requests and reuse the same fresh snapshot for commands already in flight on the same stamped workflow. A short blind cache would weaken the removed-node guard, so a deliberate bulk command is preferable.

### Environment
Windows 11
ComfyUI 0.30.0 local
NVIDIA RTX 5090 32 GB
comfyui-mcp 0.50.1
comfyui-mcp-panel 0.11.42

Issue #640, opened 2026-08-04
Isolate saved-workflow bridge routes per browser tab


artokun:
### What happened / root cause
After a ComfyUI/frontend reconnect, `panel_list_workflows` showed the intended saved workflow as active and `panel_set_workflow_target({mode:"current"})` returned success, but the next `panel_graph_outline` was rejected with a workflow-instance mismatch. A recovery `panel_open_workflow` also timed out.

The saved-workflow routing key in the panel was path-only (`wf:<path>`). Two browser tabs/sessions that open the same saved workflow therefore register on the same bridge route. Commands or replies can reach the other live tab, whose workflow UUID correctly fails the instance fence. Process evidence showed multiple concurrent agent sessions using the identical saved-workflow panel URL.

### Steps to reproduce
1. Open the same saved workflow file in two ComfyUI browser tabs/sessions.
2. Connect the Agent panel in both.
3. Reconnect/reload one session and call `panel_list_workflows`, then a graph read such as `panel_graph_outline`.
4. The graph read can be routed to the other tab and fail its workflow UUID fence.

### Exact error (scrubbed)
`workflow instance mismatch: this command targets a different workflow than the active canvas — the workflow was switched or replaced after it was issued. Re-select the intended workflow (panel_open_workflow) or re-target with panel_set_workflow_target({mode:"current"}), then retry.`

`Panel tab wf:workf did not reply to "workflow_open" within 15000 ms — the ComfyUI tab may be backgrounded or frozen`

### Fix
Applied locally: yes. Upstream-only: no.

Use the existing per-browser-tab routing helper for saved workflows, matching the regression test already present in the installed panel:

```diff
+import { savedWorkflowTabId } from "./lib/workflow-tab-id.js";

 function workflowTabId(wf = activeWorkflowRef()) {
   if (!wf) return getTabId();
   const saved = savedWorkflowPath(wf);
-  if (saved) return "wf:" + wf.path;
+  if (saved) return savedWorkflowTabId(getTabId(), saved);
```

This changes saved workflow routes from `wf:<path>` to `wf:<browser-tab-id>:<path>`, keeping same-file sessions isolated while remaining stable inside one browser tab.

Verification:
- `node --check web/js/comfyui-mcp-panel.js` passed.
- `python3 -m unittest ... test_workflow_tab_id.py -v` passed after failing before the patch.
- The broader installed-panel test directory still has three unrelated pre-existing failures; they were not modified.

### Environment
[...]

PR #761, opened 2026-08-07
fix(fence): let an UNSAVED canvas publish its established identity

### LINKED ISSUE #760: An UNSAVED canvas can never rebind the workflow fence — the last cause of the mismatch wedge
Reproduced live and traced end to end today. This is the **remaining** cause of the mismatch wedge after artokun/comfyui-mcp-panel#759 and artokun/comfyui-mcp#1047, and unlike those two it is a deliberate design rule rather than an oversight — so it needs a decision, not just a patch.

## The chain

`panel_new_workflow` creates an **unsaved** workflow and makes it the live canvas. From there:

1. `establishedWorkflowReplyIdentity(wf)` returns `null` for it, because of:

   ```js
   const savedPath = savedWorkflowPath(wf);
   if (!savedPath) return null;
   ```

   with the stated reason: *"A tmp: routing handle is intentionally not durable enough to publish as a command-fence refresh source."*

2. `liveWorkflowListActive()` therefore yields `activeIdentity = null`, so `workflow_list`'s **top-level** `active` record is built with `activeRoutingKey = null`:

   ```js
   key: activeDiskPath ? active.key : activeRoutingKey,   // → null
   routing_key: activeRoutingKey,                          // → null
   ```

3. But each **list entry** derives its identity independently, via `workflowTabId(w)`, which resolves fine for an unsaved tab. Measured on a live rig, both records describing the *same* just-created workflow:

   ```
   top-level `active`  : path None, filename None, title 'Unsaved Workflow', key None,             routing_key None
   flagged list entry  : path None, filename None, title 'Unsaved Workflow', key 'tmp:68b3d2dd…',  routing_key 'tmp:68b3d2dd…'
   ```

4. The orchestrator's `corroborateActiveForFence()` compares them, `identityVerdict` returns `undefined`, and it fails closed:

   > the active record and the open-workflow list share no comparable identity field, so nothing corroborates it

5. Nothing is adopted, the session stays fenced to the previous instance, and every following `panel_*` graph call returns `workflow instance mismatch`.

## Why this is the load-bearing one

The fence can only be repaired from a record the panel publishes an identity for. An unsaved canvas never publishes one. So **a session whose active workflow is unsaved cannot rebind its fence at all** — the recovery is not failing, it is unavailable. That is precisely what artokun/comfyui-mcp#932 / #607 / #688 describe as permanent, and why the advertised `panel_set_workflow_target({mode:"current"})` returns a hollow non-recovery.

And `panel_new_workflow` is the single most common way to land on an unsaved canvas.

## Verified on

Orchestrator `mcp=0.50.14`, panel `panel=0.11.44`, ComfyUI 0.30.2, driven over the loopback panel MCP so no agent was involved. `panel_new_workflow` → `created: true`; the very next `panel_add_node` → `workflow instance mismatch`.

## The decision to make

The rule exists for a real reason, stated at the top of the function: *"A reply is an observation, never an identity-initialization point"* — `workflowTabId()` / `workflowStableUuid()` **mint** for an unfamiliar object, and minting during a reply would manufacture a plausible stamp for something never established as a real workflow (#716).

That reasoning applies to *minting*. It does not obviously apply to an unsaved workflow whose per-instance uuid is **already established** in the live-object WeakMap (`_tempWorkflowInstanceIds`) — that value is per-instance by construction, a copy or import never shares the live object, and `resolveUnsavedInstanceUuid` already treats it as the trustworthy source precisely because of that (#570).

So the candidate fix is to accept an unsaved workflow's identity **only when it is already established** (a WeakMap hit, never a mint), keeping `workflow_uuid` gated exactly as #716 requires.

Two things make this delicate enough that I did not want to land it unreviewed at the end of a session:

- the `tmp:` routing-key spelling must match `workflowTabId()`'s **exactly**, or every record's `active` flag flips false and the agent is told no tab is active — the divergence the #640 comment warns about in this same function;
- #186's misroute protection leans on unsaved tabs being hard to address, so widening what they publish deserves an adversarial look.

A cheaper, narrower alternative worth weighing: leave `establishedWorkflowReplyIdentity` alone and make the **top-level `active` record reuse the identity the flagged list entry already computed** for the same object. That removes the self-inconsistency (which is a bug on its own terms — one reply describing one workflow two different ways) without changing what counts as an established identity. It fixes corroboration; whether it also yields a usable `workflow_uuid` needs checking, since #716 gates that separately.

I lean toward doing the narrow one first, since a reply contradicting itself is indefensible regardless of how the broader rule is settled.

artokun:
Closes #760.

> **This reverses a deliberate, tested design decision.** The test asserting the old
> behaviour is updated in place with the reasoning rather than deleted. It deserves a
> considered review, not a rubber stamp.

## What was happening

`establishedWorkflowReplyIdentity()` returned `null` for any workflow without a saved path,
so an **unsaved canvas could never publish an identity in a reply**. Measured live on
`mcp=0.50.14` / `panel=0.11.44`, both records describing the *same* just-created workflow:

```
top-level `active`   : key None,             routing_key None
flagged list entry   : key 'tmp:68b3d2dd…',  routing_key 'tmp:68b3d2dd…'
```

The list entry derives identity independently via `workflowTabId()` and resolves fine. The
top-level record goes through `activeIdentity` and comes back empty. The orchestrator's
`corroborateActiveForFence()` then reports:

> the active record and the open-workflow list share no comparable identity field, so
> nothing corroborates it

…refuses to adopt, and the session stays fenced to the previous instance. Every following
graph command returns `workflow instance mismatch`.

Since `panel_new_workflow` **creates** an unsaved workflow, recovery for those sessions was
not failing — it was *unavailable*. That is the permanent half of artokun/comfyui-mcp#932,
#607 and #688, and why the advertised `panel_set_workflow_target({mode:"current"})` returns
a hollow non-recovery.

## Why the old premise does not hold

The rule read: *"a tmp: routing handle is not durable enough to publish as a command-fence
refresh source."*

1. **It conflates two values.** What is published is `workflowObjectUuid` — the canonical
   per-instance uuid from the **live-object** map. The tmp handle is only the routing key.
2. **The handle is not ephemeral.** `_tempWorkflowInstanceIds` is keyed on the live object
   precisely so that one unsaved workflow keeps **one** tmp: id for its lifetime — a
   churning id would cause the re-hello storm that comment warns about.
3. **A fence refresh is not a durable-resume record.** It re-stamps commands for the canvas
   that is live *now* and is discarded when the tab changes. Durability is the wrong axis;
   per-instance correctness is the right one, and the live-object uuid is exactly that
   (#570 — a copy, import or in-place replace never shares it).
4. **Decisively:** the panel's own fence comparison uses `workflowStableUuid()`, which for
   an unsaved workflow returns *this same objectUuid*. Publishing it makes the
   orchestrator's stamp match the panel's active uuid exactly — which is the entire purpose
   of the refresh.

## What #716 still forbids, unchanged

> A reply is an observation, never an identity-initialization point.

`workflowObjectUuid()` is a pure WeakMap read — it never mints. An object the panel has not
established still publishes nothing, and the function bails **before** it would even ask for
a routing handle. Both facts are pinned:

- an unsaved canvas with **no** established uuid → `null`, and `workflowTabId` is called
  **zero** times;
- a **non-canonical** uuid on an unsaved canvas → still refused (the shape gate is not
  bypassed by the new branch).

The handle comes from `workflowTabId()` — the same function `brief()` uses for every list
record — so the two spellings agree by construction. The #640 divergence hazard flagged in
that function is *reduced* here, not introduced.

## Tests

2826 in the panel suite, typecheck and the vocabulary gate green.
Mutation-verified: restoring the early `return null` fails the #760 test.

## Sequencing

Third of three in this cluster, and the last one standing:

| | | |
|---|---|---|
| #759 | the recovery probe was fenced by the fence it repairs | panel |
| artokun/comfyui-mcp#1047 | a save's `tmp:`→`wf:` rename counted as two workflows | orchestrator |
| **this** | an unsaved canvas can never publish a fence identity | panel |

Each was hiding the next: with #759 in place the failure moved from "the panel did not
answer" to "the panel answered, but its reply could not be corroborated", which is what
exposed this one.

Does not close the reports on its own: they close when someone confirms on a released build.

### DIFF
--- a/browser_tests/unit/open-outcome.test.mjs
+++ b/browser_tests/unit/open-outcome.test.mjs
@@ -1031,7 +1031,30 @@ test("#716 P1: a malformed truthy active binding cannot mint reply identity", ()
   assert.equal(workflowUuids.has(malformedActive), false, "must not mint an ephemeral workflow UUID");
 });
 
-test("#716 P1: a temporary workflow UUID is never published as a durable refresh source", () => {
+// #760 — this test used to assert the OPPOSITE ("a temporary workflow UUID is
+// never published as a durable refresh source"), on the reading that "its tmp
+// routing identity is ephemeral, so it cannot refresh the MCP's durable command
+// fence". Two things in that premise do not hold, and the cost of acting on it
+// was that an unsaved canvas could never rebind its fence at all:
+//
+//   • it conflates two different values. What gets published as the fence source
+//     is `workflowObjectUuid` — the canonical per-instance uuid from the LIVE-
+//     OBJECT map — NOT the tmp handle. The tmp handle is only the routing key.
+//   • the tmp handle is not ephemeral either: `_tempWorkflowInstanceIds` is keyed
+//     on the live object exactly so "the same unsaved workflow must keep ONE tmp:
+//     id for its lifetime" (a churning id would cause a re-hello storm).
+//
+// And a fence REFRESH is not a durable-resume record: it re-stamps commands for
+// the canvas that is live right now and is discarded when the tab changes.
+// Decisively, the panel's own fence comparison uses workflowStableUuid(), which
+// for an unsaved workflow returns this same objectUuid — so publishing it makes
+// the orchestrator's stamp match the panel's active uuid exactly, which is the
+// whole point of the refresh.
+//
+// The #716 rule that DOES still hold — a reply must never INITIALIZE an identity
+// — is unchanged and separately pinned by the test below: workflowObjectUuid is a
+// pure read, so an object with no established uuid still publishes nothing.
+test("#760: an unsaved canvas with an ESTABLISHED uuid can refresh the fence", () => {
   const src = readFileSync(PANEL_JS, "utf8");
   const pathSource = namedFunctionSource(src, "savedWorkflowPath");
   const canonicalUuidSource = namedFunctionSource(src, "isCanonicalWorkflowInstanceUuid");
@@ -1039,10 +1062,12 @@ test("#716 P1: a temporary workflow UUID is never published as a durable refresh
   const helperSource = namedFunctionSource(src, "activeWorkflowUuidForOpenReply");
   const temporaryActive = { isPersisted: false, isTemporary: true };
   const workflowUuids = new WeakMap([[temporaryActive, "33333333-3333-4333-8333-333333333333"]]);
+  const tempIds = new WeakMap([[temporaryActive, "tmp:33333333-3333-4333-8333-333333333333"]]);
   const replyUuid = new Function(
     "activeWorkflowRef",
     "workflowObjectUuid",
     "savedWorkflowHandle",
+    "workflowTabId",
     `${pathSource}; ${canonicalUuidSource}; ${identitySource}; ${helperSource}; return activeWorkflowUuidForOpenReply;`,
   )(
     () => temporaryActive,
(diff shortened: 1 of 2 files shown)


artokun:
## Verified end to end on a live rig

All three fixes applied together, then the exact reported sequence run over the loopback panel MCP — **no agent involved**, so nothing could paper over a refusal.

Setup: orchestrator built from this branch (`mcp=0.50.15` + this fix), panel carrying #759 and #761, ComfyUI 0.30.2, a **fresh browser document** (verified by `performance.timeOrigin`, after every patch — a stale module is the standing false-negative on this rig). All three fixes confirmed present in the *running* module, not merely on disk:

```
fix759_predicate        true      fix759_listExempt      true
fix759_pinGuard         true      fix761_unsavedBranch   true
fix759_graphStillFenced true   ← the fence still refuses graph work
```

The tab under test was an **unsaved** workflow — the #761 scenario exactly.

| step | before | after |
|---|---|---|
| `panel_new_workflow` | `created:true`, fence left stale | `created:true`, key `tmp:c2512bcc…` |
| `panel_query_graph` | `workflow instance mismatch` | **success** — 0 nodes |
| `panel_add_node` | `workflow instance mismatch` | **success** — node id `1` |

And confirmed against the live canvas rather than the tool's own reply (the #710 lesson — a handler must be checked against observed effect, not its own report):

```js
activeTitle: "Unsaved Workflow (2)"   // the workflow new_workflow created
nodeCount:   1                        // the CheckpointLoaderSimple genuinely landed
```

Before these fixes the same sequence produced one success and three hard refusals, with `node_count: 0` proving nothing landed.

### One observation, not a claim

The 4th call failed with `no connected tab with id "tmp:10cc"` — my harness pinned a tab id by prefix, and that id migrated to `tmp:c251` when the new workflow was created. Calls 2 and 3 routed fine, so this is my test harness holding a stale pin across a workflow switch, not a reproduction of the reported bug. Whether the loopback MCP's path resolution should consult the same migration aliases `resolveTarget` does is worth a look on its own, and I have not investigated it.

This does not close the reports — they close when someone confirms on a released build.

Issue #779, opened 2026-08-08
Agent panel renders completely blank on a fresh 0.11.44 install — tab registers, content never paints, no error attributed to us


artokun:
Reported by **bunnlilah** in Discord `#help` — a brand-new user, completely blocked on first install.

## Symptom

The **Agent** sidebar tab registers and is selectable, but its content area renders **entirely empty** — no header, no provider chip, no Connect button, no "not connected" state. Just black.

![blank panel](https://cdn.discordapp.com/attachments/1535573950124531752/1535573950631772220/Screenshot_2026-08-08_022140.png)

## Environment

```
panel: 0.11.44
mcp:   0.50.35
comfyui: 0.30.0 (LOCAL), triton installed, sage installed
page:  http://127.0.0.1:8188
ua:    Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:140.0) Firefox/140.0  [Waterfox]
```

## What has been eliminated

The reporter did all of this before I had to ask twice:

- **Not the install** — uninstalled, reinstalled, restarted ComfyUI. Unchanged.
- **Not the browser** — blank in Waterfox *and* Chrome.
- **Not a stale cache** — hard refresh (Ctrl+Shift+R) in both.
- **Not the orchestrator being down** — reproduced with `npx -y comfyui-mcp@latest connect` running in an open Command Prompt.
- **Not a syntax/parse error in the shipped bundle** — I parse-checked **all 99** `web/js/**/*.js` files at `origin/main` (= 0.11.44) as ES modules. All clean. So this is *not* the duplicate-top-level-declaration class that `node --check` misses.
- **No error from our code in the console.** Nothing red mentions `comfyui-mcp-panel.js` on its surface line.

## Console

```
GET http://127.0.0.1:8188/user.css                                   404
GET http://127.0.0.1:8188/api/userdata/user.css                      404
[ComfyUI] The legacy queue/history menu is deprecated...             (warn)
ComfyApp graph accessed before initialization                        ERROR  settingStore-*.js:223
GET http://127.0.0.1:8188/api/userdata?dir=subgraphs&recurse=true... 404
GET http://127.0.0.1:8188/api/userdata/comfy.templates.json          404
```

## Current hypothesis (unconfirmed — expanded stack requested)

[...]

PR #804, opened 2026-08-08
fix(panel): the #779 silence detector — a selected tab that never paints must say so, verified live on real 1.50.3

artokun:
## What this PR is, and what it is not

The blank-panel-on-1.50.x **cause** was already fixed on `main` before this branch existed: #784 (the sidebar guard must not destroy on an unknown active tab), #785 (a throwing `render()` paints a failure shell instead of nothing), #786 (the second reader of the moved tab-button marker). The sidebar-tab **contract never moved** — 1.50.3 calls `render()`; our own guard deleted the content the instant it was attached.

This PR is the two halves that were still missing:

1. **The silence detector** — the diagnostic that would have made #779 a one-line support answer, and the only thing that will speak the next time a frontend change breaks us in a way nobody predicted.
2. **Live verification on a real 1.50.3** — which the #784 thread explicitly recorded as still owed ("I would still rather see the panel render on a real 1.50.3 before calling it confirmed").

Refs artokun/comfyui-mcp-panel#779. That issue is **already closed** — it auto-closed when #784 merged, and this branch carries the remaining halves discussed in its thread.

The reference is deliberately owner-qualified, which does **not** auto-close. This line previously used a same-repo closing keyword, which was inert only for as long as the issue stayed closed: had anyone reopened #779, merging this PR would have slammed it shut again underneath them. An owner-qualified ref links the two without claiming the authority to close anything.

## 1. The watchdog (`web/js/lib/sidebar-render-watchdog.js`)

#785 covers a `render()` that **throws**. A `render()` that is never **called** — which is exactly what a real sidebar-tab contract change in a future frontend would produce — still failed in perfect silence. **One** evidence-only check closes that:

- **Starvation**: our tab is *provably* the selected one (dual-generation marker read via the #784 helper — `data-testid` first, class fallback) and neither `.cmcp-root` nor the #785 failure shell exists, continuously for 3s, re-verified at expiry. One `console.error`, ever, naming the panel version, the frontend version, that it is **not** a connection problem, that reinstalling cannot help, where to report, and a working `--front-end-version` pin.

### The check that was REMOVED, and why

An earlier draft of this branch carried a second check — **appearance**: "the rail exists but our tab button never joined it within 10s", read as a frontend that accepts `registerSidebarTab()` and silently drops the legacy spec. It has been **removed rather than narrowed**, because the DOM cannot support that conclusion:

ComfyUI ships a supported view that renders a deliberately **filtered** rail —
`src/views/LinearView.vue:120` mounts `<SideToolbar :visible-tab-ids="['assets', 'apps']" force-connected …>`, confirmed in the frontend repo at **v1.50.3** (the exact version this branch live-verified against) **and v1.51.3**, and reachable by any user who turns on linear mode, since `GraphView.vue` renders `<LinearView v-if="linearMode" />`. There, `.side-tool-bar-container` is present and our button is absent **while registration succeeded exactly as designed** — the frontend's own registry (`sidebarTabStore.sidebarTabs`) still lists us; `visibleTabIds` filters only at RENDER time.

On that view the check fired a compatibility-fault line naming both versions and advising a relaunch pinned to a different frontend — a **false statement**, aimed at the same reader who lost an hour to reinstalls on a guess. A filtered rail and a genuine contract break are indistinguishable from the DOM, so under this module's own rule (the #784 lesson: "I cannot tell" must never be reported as "it is broken") the honest answer is silence. Narrowing was considered and rejected — a registry check cannot discriminate either, since a frontend that registers a tab and then declines to render it looks identical to a filtered rail.

Removing the phase also closed the second review finding, that the 60s give-up bound could cancel the 10s appearance window before it was owed. A structural test now fails if a second report — or any reasoning about our rail button's presence — comes back.

**The false-positive bar was treated as the primary requirement**, because this line will be read as "something is broken":

- an **unreadable tab marker disarms rather than counts** — the #784 lesson applied to diagnostics ("I cannot tell" must never be reported as "it is broken");
- **no rail recognized = permanent silence**, watchdog-wide (hardened in codex round 1);
- keep-alive detaches (user switches tabs mid-window) disarm; a slow first build is absorbed by the deadline **re-check** — it fires on fresh evidence, never a stale sample; a stalled main thread cannot fire early, because the deadline is itself a timer;
- **a glimpse of paint retires nothing** — only a paint that *survives* 1.5s while our tab is active retires the watchdog. This is not theoretical: the first draft trusted the glimpse, and a live drill reproducing the *actual* #779 shape (root attached, then removed in the same mutation flush) put the watchdog to sleep instead of firing it. Caught live on the rig, fixed, re-drilled on both frontends, and pinned by unit tests.

### The remedy line, corrected against reality

The pin names **`Comfy-Org/ComfyUI_frontend`** — *not* `comfyanonymous/ComfyUI`, the spelling the issue thread circulated. The flag fetches GitHub releases from the repo it names; the frontend's 1.x tags exist only in the frontend repo, and the wrong spelling silently falls back to the installed default package (watched happen in a live boot log; confirmed against 0.30.2's `frontend_management.py` and the GitHub API). The pin also derives from `VERIFIED_FRONTENDS` (#787's registry) and prefers the newest verified frontend **that differs from the one failing** — never a hardcoded relic, never the version the user is already stuck on.

### The wording

> `[comfyui-mcp-panel] the Agent tab has been selected for ~3s but no panel content exists (no .cmcp-root in the document). The tab registered and was selected, yet the panel was either never asked to render or its content was removed as soon as it was attached. That is a compatibility fault between panel <ver> and ComfyUI frontend <ver>. This is NOT a connection problem, and reinstalling the pack or ComfyUI cannot change it. Please report it at <issues> and include both version numbers from this message. Until it is fixed, relaunching ComfyUI with --front-end-version Comfy-Org/ComfyUI_frontend@<pin> restores the panel (frontends 1.47.12 and 1.50.3 are verified to render this panel version).`

It reports what was **observed** (naming both shapes it cannot distinguish), never a guessed cause — the same rule as #785, learned from the hour this reporter spent reinstalling on a guess.

## 1b. Review finding: the watchdog could not RUN (P1, reproduced then repaired)

A pre-merge review caught the failure shape this project pays for most often — a mechanism that is thoroughly tested and cannot execute in production. Green tests prove the code matches the tests; they never prove the code is reachable.

**The defect.** The watchdog subscribed to the FIRST `.side-tool-bar-container` element it found, and the poll then stood down permanently ("once the observer is watching the rail, the poll has nothing left to discover"). ComfyUI does not keep one rail element for the life of the page.

**Reachability, verified against shipped frontends — not assumed.** The rail is `v-if`-gated, so it is destroyed and recreated rather than hidden. Checked at **1.47.12, 1.48.7, 1.50.3 and 1.51.5** (two from sourcemapped bundles on this machine, two from the tagged source):

```vue
<!-- src/components/graph/GraphCanvas.vue -->
<SideToolbar v-if="showUI && !isBuilderMode && !linearMode" />
```
```ts
const showUI = computed(() => !workspaceStore.focusMode && betaMenuEnabled.value)
```

Linear mode mounts a **second, separate instance** of its own (`src/views/LinearView.vue`). The compiled 1.50.3 bundle confirms it is a `v-if` and not a `v-show` — the slot compiles to a ternary yielding either a block or `createCommentVNode`, with no `vShow`/`style.display` on the node. So **focus mode, linear mode and builder mode each replace the `<nav>` with a brand-new element.**

After any such toggle the subscription pointed at a node that had left the document, the poll was long gone, and nothing was left to call `sample()`. Selections on the **replacement** rail were never sampled — restoring the exact silence this module exists to break, on a code path whose tests were entirely green.

**Reproduced before repairing.** Three new tests drive the real installer, let it attach, then replace the rail element as a remount does. All three **fail on the parent commit** (`914843a7`); one fails on `notStrictEqual` with the observer's target being the rail object itself, which is the defect stated as an assertion.

The old observer stub is why this shipped green: it ignored `observe()`'s target and `mutate()` fired every observer unconditionally, which made a rail-bound subscription and a document-bound one **indistinguishable**. The stub now models both halves of a real `MutationObserver` — *where* it is registered (a detached node mutates no more) and *what* it subscribed to (`{attributes:true}` delivers no childList records).

**The repair.** The rail becomes a **gate, never a handle**. Its existence still proves "this page has a sidebar we understand" — no rail, no evidence, no claim — and nothing more. The subscription that drives sampling watches the **document**, the one anchor on the page that cannot be replaced. `sample()` already read the selected button at document scope, so it never cared which rail it was looking at; only the subscription did.

This removes the failure rather than detecting it: nothing about a rail is retained, so no reference can go stale, and there is **no re-binding step to get wrong**. No timer was added — the discovery poll keeps its existing give-up bound, so nothing is permanently armed in the render lifecycle. `childList` joins the class filter because the two halves of a remount are different mutations: toggling tabs on the new rail is a class change, but a rail rebuilt *while our tab is active* comes back with its button **already selected** — no attribute ever transitions on it, and the only observable is that the element appeared.

**Still one line, not one per remount.** The once-ever latch is untouched, and a test drives eight consecutive remounts against a permanently starved panel to pin it. That bar is deliberate: the re-arming-notice flood tracked in artokun/comfyui-mcp#1489 is a live bug in this panel, and a detector that re-binds must not become another one.

**Nothing outlives teardown.** Asserted, not assumed: after the check resolves either way — and after an explicit `stop()` — zero live subscriptions and zero pending timers remain, and no subscription is ever left pointing at a departed rail.

**Mutation-tested against a verified-green baseline.** Four mutations, each proven applied by counting the mutated text in the source (a mutation that fails to apply is a false PASS, not a pass), each killed:

| mutation | tests killed |
|---|---|
| observe the rail element again (the original defect) | 3 |
| `childList` dropped from the subscription | 1 |
| `stop()` no longer disconnects the subscription | 3 |
| once-ever latch removed (report per remount) | 5 |

## 2. Verification — all live, real browsers, real frontends

Isolated second ComfyUI instance (CPU, scratch `--base-directory`, this branch via junction, port 8388) — the owner's rig on 8188 was never restarted or repointed.

| check | frontend | result |
|---|---|---|
| panel renders, `.cmcp-root` connected + visible | **1.50.3** (real bundle, `data-testid`-only buttons) | **PASS** — the verification the #784 thread recorded as owed |
| keep-alive: detach on switch-away, re-attach on return | **1.50.3** | PASS |
| watchdog silent through open → dwell → switch → return | **1.50.3** | PASS (0 lines) |
| drill: root removed on attach (the historical failure shape) | **1.50.3** | PASS — exactly one line, names `frontend 1.50.3`, pin backs off to `@1.47.12` |
| panel renders + keep-alive + watchdog silent | **1.47.12** (this branch) | PASS |
| drill: root removed on attach | **1.47.12** | PASS — exactly one line, names `frontend 1.47.12` |
| `main`'s #784–#786 render + keep-alive, no panel console errors | **1.47.12**, owner's live rig, read-only | PASS |

Static cross-check of the extracted 1.50.3 bundle confirmed every marker this branch depends on: `"data-testid":\`${e.id}-tab-button\``, `side-bar-button-selected`, `side-tool-bar-container`, and `mountCustomExtension=(e,t)=>{e.render(t)}` (render IS invoked; no handler).

Unit: **35 watchdog tests** (state machine, the starvation wording, installer against a hand-cranked clock and fake rail in both marker generations, the live-drill regression, the rail-remount suite from 1b, integration greps) — including two behavioural specs that pin the **filtered-rail (LinearView) shape as silent**, and a structural spec asserting exactly one report survives and that the module never reads `findSidebarTabButton`. Latest full suite on this branch (`npm run test:unit`, which also runs the i18n, tool-vocabulary and panel-scope gates): **4443 pass / 0 fail** (4444 with 1 todo). The six added tests are the rail-remount ones described in 1b. The #787 DOM-reach gate passes — the watchdog reads only already-declared selectors, whose registry entries now name it as a reader.

Note on the live-verification table above: it was recorded **before** the appearance check was removed. Every row in it exercises panel rendering, keep-alive, watchdog silence, or the starvation drill — the half that shipped. No row depended on the appearance check firing, so the removal does not invalidate any of them.

## Out of scope, deliberately

- **No version bump** — releasing belongs to the release owner. (This line originally read "`main` is unreleased at 0.11.44"; `main` has since released well past that and now sits at **0.14.41**, so the release urgency that framed this PR no longer applies — the #784–#786 render fixes are long since in users' hands. What remains here is the diagnostic for the *next* time.)
- The `TODO(v2): defineSidebarTab` migration stays a TODO: the contract demonstrably did not move, and migrating under a P0 label would be churn, not fix.
- Playwright browser tests were not run (they require `--workers=1` against the live ComfyUI and are not in CI); the checks above were performed directly in a real browser instead.


artokun:
Superseding the earlier body: this was never stranded work — the worktree was live and mid-verification when it was surfaced (the same file-heuristic trap as the documented dont-kill-agents-on-file-heuristics incident). The branch is now COMPLETE and fully verified: live on a real 1.50.3 (isolated second instance, correct Comfy-Org/ComfyUI_frontend pin via local dist cache) and on 1.47.12, healthy paths silent, drills fire exactly one correctly-worded line on both frontends, codex gate round 2 PASS, full suite 3045/3045. See the updated PR body for the verification table. Not merging — that call belongs to the merge/release owner.

artokun:
## Both review findings addressed — by retiring the appearance check, not narrowing it

Merged `origin/main` (clean) and pushed `919fadc5`.

### Finding 1 (the filtered toolbar) is correct, and it is worse than "a supported view"

I verified it against the frontend source rather than taking it on description:

```
v1.50.3:src/views/LinearView.vue:120:  :visible-tab-ids="['assets', 'apps']"
v1.51.3:src/views/LinearView.vue:120:  :visible-tab-ids="['assets', 'apps']"
v1.50.3:src/components/sidebar/SideToolbar.vue:142-143:
    return visibleTabIds ? all.filter((tab) => visibleTabIds.includes(tab.id)) : all
```

Two things make this decisive rather than theoretical:

1. **It is present in 1.50.3 — the exact frontend this PR live-verified against.** The verification table above is real, but it exercised the default rail; it never entered this view.
2. **It is reachable by an ordinary user**, not a dev flag: `GraphView.vue` renders `<LinearView v-if="linearMode" />`.

In that view `.side-tool-bar-container` is present, our button is absent, and **registration succeeded exactly as designed** — `visibleTabIds` filters at RENDER time only, and the frontend's own `sidebarTabStore.sidebarTabs` still lists our tab. So the check would have printed a compatibility-fault line naming both versions and telling the user to relaunch ComfyUI pinned to a different frontend — to a user whose panel is completely healthy. That is the expensive kind of false statement, aimed at exactly the reader who already lost an hour to reinstalls that could not help.

### Why removed rather than restricted

The suggested alternative was "restrict this to a provably unfiltered main toolbar". I could not find a way to prove that which is not itself a guess:

- **Counting buttons** is a heuristic (the filtered rail renders two *legitimate* tabs), not evidence.
- **Detecting LinearView** couples the diagnostic to one version's internals — the precise failure mode this whole issue is about.
- **Checking the registry** (`app.extensionManager.sidebarTab.sidebarTabs`) looked promising and is readable in both 1.47.x and 1.50.3 — but it does not discriminate either. `registerSidebarTab` pushes unconditionally, so a frontend that registers our tab and then declines to render it is byte-identical to a filtered rail. The registry can only detect a frontend that *refuses* registration, which is a different failure, is observable immediately without a 10s DOM window, and already surfaces loudly if the call throws.

Filtering and a contract break are genuinely indistinguishable from the DOM. Under this module's own rule — the #784 lesson that "I cannot tell" must never be reported as "it is broken" — the honest answer is silence, so the appearance half is gone.

### Finding 2 resolves with it

The 60s give-up bound could cancel the 10s appearance window before it was owed. Correct, and now moot: with no appearance phase the poll only waits for the rail and stands down once the observer is watching it.

### What is unchanged

Starvation — the half this PR is titled for, and the half that actually answers #779. On the "can it false-positive on a slow paint" question specifically: it cannot, and the bound is not a guess. `render()` attaches `.cmcp-root` (or the #785 shell) **synchronously**, in the same task ComfyUI invokes it from, so there is no async gap for a slow machine to widen; a `buildPanel()` slow enough to matter blocks the main thread and delays the deadline timer *with* it rather than firing it early; and the deadline **re-reads the document** instead of deciding on the stale sample that armed it, so a paint landing anywhere inside the window disarms the check.

### Tests

`npm run test:unit` green in full (i18n, tool-vocabulary and panel-scope gates chained): **4428 tests**. One unrelated failure, `manager-install.test.mjs` "#671 verifyInstalled … inside the reply window", is a wall-clock-budget assertion that blew under whole-suite load — 122/122 green running that file alone.

The 29 watchdog tests now pin the *absence* of the claim:

- two behavioural specs for the filtered-rail shape (silent forever; still silent when the button turns up late), and
- a structural spec that fails if a second report — or any reasoning about our rail button — comes back.

Mutation-checked rather than assumed: restoring the appearance report **and** the old poll-continuation kills all three (`pass 26 / fail 3`). An earlier, sloppier mutation that only restored the report killed just the structural one, which is what prompted me to make the mutation faithful.

Not merged, still yours to land. Refs artokun/comfyui-mcp-panel#779.

One thing worth a maintainer eye: the PR body still opens with "Closes #779". That issue is already closed, so the keyword is inert here, but it will close it again on merge if it is ever reopened.

Issue #775, opened 2026-08-08
LTX pack loading and remote restart leave the panel workflow out of sync


artokun:
While preparing the bundled local LTX-2.3 workflow on a remote ComfyUI/RunPod target, several panel failures occurred. 1) panel_load_workflow(pack:"ltx23-distill-3stage") returned "Error: workflow is in API/prompt format; provide the UI workflow JSON (the pack workflow.json is UI format)" even though the pack advertises a UI workflow. 2) panel_strip_workflow(pack:"ltx23-distill-3stage") returned "Cannot read properties of undefined (reading 'map')". 3) panel_flatten_workflow(pack:"ltx23-distill-3stage",apply:true) reported success but loaded a 0-node graph. 4) panel_restart_comfyui refused with "could not confirm that this panel's ComfyUI is the local instance", although the generic ComfyUI target is a configured remote instance. 5) After reconnect/rebinding, panel_graph_outline/panel_get_errors repeatedly reported root-workflow-uuid-mismatch until the workflow was explicitly reopened. This forced recovery through list/open workflow and left the graph showing stale missing-node diagnostics after custom-node installation. Environment: remote RunPod H100 PCIe, ComfyUI 0.28.0, torch 2.9.1, CUDA 12.8, comfyui-mcp 0.50.23 per environment line, panel 0.11.44. No secrets included.

Related (same root cause): #769, #702, #607

artokun:
Five findings, and they split cleanly: **three share one root cause I can confirm, and two are already fixed.**

## The root cause for (1), (2) and (3)

`packs/ltx23-distill-3stage/workflow.json` is in **API/prompt format**, not UI format. Measured on `main`:

```
top-level keys: ['1','10','102','107','110','114','122','15','1956', …]   (59 numeric)
has nodes[]: False        has links[]: False
sample entry keys: ['_meta', 'class_type', 'inputs']
```

That is the API shape — a map of node id → `{class_type, inputs}` — with no `nodes[]`/`links[]` and no layout. So the pack does not ship what it advertises, and every one of your three failures follows from feeding that to a tool expecting UI format.

What makes it worth reporting as three bugs rather than one is that the **three tools disagree about what to do with it**:

| tool | behaviour | verdict |
|---|---|---|
| `panel_load_workflow` | refused: *"is in API/prompt format; provide the UI workflow JSON"* | **correct** — `assertUiWorkflow` checks for a top-level `nodes` array and says so |
[...]

Issue #695, opened 2026-08-06
panel_add_node rejects every node with a required MASK input — MASK (and other core socket types) missing from SAFE_SOCKET_TYPES


artokun:
### What happened / root cause
panel_add_node fails for ANY node whose required inputs include a `MASK` socket (core `InvertMask`, `SetLatentNoiseMask`, KJNodes `GrowMaskWithBlur`, `MaskToImage`, …) with:

> Required custom widget "MASK" have not registered. They may be custom widgets still loading; retry shortly.

Root cause: `web/js/lib/node-widget-materialization.js` → `SAFE_SOCKET_TYPES` (the whitelist of built-in connection datatypes that are safe to leave as sockets when no widget constructor exists) is missing `MASK`, which is a stock ComfyUI core datatype. `unavailableRequiredCustomWidgetTypes()` therefore classifies the MASK input as a possibly-still-registering custom V3 widget, `awaitRequiredCustomWidgetRegistration()` polls for 5s, and the add is rejected. Also missing from the list: `MESH`, `VOXEL`, `PHOTOMAKER` (core load3d/photomaker types) — same failure class.

Impact: the agent cannot build any masking/inpaint chain on the live canvas (every mask-consuming node is unaddable). Hit in production while adding a SetLatentNoiseMask-based product-protection refine.

### Steps to reproduce
1. Open any workflow in the panel.
2. Call `panel_add_node { class_type: "InvertMask" }` (or `SetLatentNoiseMask`, `GrowMaskWithBlur`).
3. After a ~5s poll it errors: `Required custom widget "MASK" have not registered…` — 100% reproducible; retrying never helps because MASK is a socket type, not a widget, and will never register a constructor.

### Exact error (scrubbed)
```
Error: Required custom widget "MASK" have not registered. They may be custom widgets still loading; retry shortly.
```

### Fix
applied locally: yes  ·  upstream-only: no

Add the missing core connection datatypes to `SAFE_SOCKET_TYPES`:

```diff
diff --git a/web/js/lib/node-widget-materialization.js b/web/js/lib/node-widget-materialization.js
index 1fdb801..f707a68 100644
--- a/web/js/lib/node-widget-materialization.js
+++ b/web/js/lib/node-widget-materialization.js
@@ -42,8 +42,8 @@ export function requiredWidgetInputTypes(nodeOrNodeData) {
 const SAFE_SOCKET_TYPES = new Set([
[...]

PR #822, opened 2026-08-09
fix(add_node): the socket proof must not be read off a single-class payload (#821)

### LINKED ISSUE #821: panel_add_node widget-registry guard blocks a node whose input types ARE produced by nodes already added; panel_refresh_nodes reports refreshed:true but does not clear it

After installing `seedvr2_videoupscaler` (via `panel_install_node`) and restarting ComfyUI with `restart_comfyui`, I added the pack's two loader nodes successfully:

- `SeedVR2LoadDiTModel` -> added, id 11, outputs `SEEDVR2_DIT`
- `SeedVR2LoadVAEModel` -> added, id 12, outputs `SEEDVR2_VAE`

Both are on the canvas and their outputs are `SEEDVR2_DIT` / `SEEDVR2_VAE`. Adding the pack's main node then fails:

```
Cannot add "SeedVR2VideoUpscaler": 2 required input types had no widget after 5.0s waiting for node extensions to register.
  - input "dit" (declared type "SEEDVR2_DIT"): no installed node outputs "SEEDVR2_DIT" and no frontend widget is registered for it.
  - input "vae" (declared type "SEEDVR2_VAE"): no installed node outputs "SEEDVR2_VAE" and no frontend widget is registered for it.
```

The guard's stated premise ("no installed node outputs SEEDVR2_DIT") is demonstrably false at that moment: two nodes that output exactly those types were just added to this same graph by this same tool, and the backend `/object_info` (fetched via `create_workflow action:node_info`) lists all three node classes with those output types. The ComfyUI server log confirms the pack imported cleanly (`3.3 seconds: ...\custom_nodes\seedvr2_videoupscaler`).

The guard appears to consult the browser tab's cached frontend node registry, which was populated at page load (before the pack was installed), rather than the live backend `/object_info` or the node types already present in the graph.

Second, related defect: `panel_refresh_nodes` — whose own description says to use it "after installing a node pack" so that "panel_set_widget / panel_add_node will accept the new value" — returned `{ ok: true, refreshed: true }` but the very next `panel_add_node` failed with the identical error. So it reports an authoritative refresh it did not actually achieve for this code path.

Net effect: the agent is blocked on a graph edit that is valid, and must ask the user to manually F5 the browser tab. Two `retry_of` retries and one `panel_refresh_nodes` in between all failed identically.

### Steps to reproduce

1. Have a ComfyUI tab open in the browser.
2. `panel_install_node({ id: "seedvr2_videoupscaler" })` -> installed + verified.
3. `restart_comfyui({ action: "restart" })` -> ComfyUI healthy, pack imported (per log).
4. Do NOT reload the browser tab (the panel reconnects on its own).
5. `panel_add_node({ class_type: "SeedVR2LoadDiTModel" })` -> SUCCEEDS.
6. `panel_add_node({ class_type: "SeedVR2LoadVAEModel" })` -> SUCCEEDS.
7. `panel_add_node({ class_type: "SeedVR2VideoUpscaler" })` -> FAILS with the error above.
8. `panel_refresh_nodes()` -> `{ ok: true, refreshed: true }`.
9. Retry step 7 -> FAILS identically.

### Exact error (scrubbed)

See the fenced error block above; no secrets involved.

### Fix

I did not patch: the guard lives in the orchestrator process that serves the `panel_*` tools, which per the panel docs only reloads on a full process restart (not `panel_reload`, not Disconnect/Connect), so I could not verify a patch from here.

Precise change needed, in the `graph_add_node` pre-flight widget/socket check:

1. Before declaring "no installed node outputs <TYPE>", resolve the type against the LIVE backend `/object_info` (the same snapshot `create_workflow action:node_info` reads), not only the browser tab's cached frontend node registry. If any backend node class declares `<TYPE>` in its `output`, the input is a socket and the node should be added immediately — same fast path the check already grants to `MASK`/`IMAGE`/`LATENT`.
2. Additionally (cheap and sufficient for this case): if any node ALREADY PRESENT IN THE OPEN GRAPH has an output of `<TYPE>`, treat it as a socket and skip the wait. Here nodes 11 and 12 were sitting on the canvas with exactly those outputs.
3. Make `panel_refresh_nodes` either actually re-register frontend node defs for newly installed packs, or report honestly when it cannot (`refreshed: false` + a hint to hard-refresh the tab). Returning `refreshed: true` when the very next `panel_add_node` fails on stale registry state is misleading and cost two extra retries.

### Environment

- OS: Windows 11 Home 10.0.26200
- ComfyUI: 0.27.0 (Desktop, base-directory C:\ComfyUIapp)
- ComfyUI FRONTEND: 1.45.20
- GPU: NVIDIA GeForce RTX 5090 Laptop GPU, 24 GB VRAM (23.9 GB reported)
- Python 3.12.11 / PyTorch 2.10.0+cu130 / CUDA 13.0
- comfyui-mcp: 0.50.45
- comfyui-mcp-panel: 0.11.45 (updated from 0.7.3 earlier in this same session)
- Offending pack: seedvr2_videoupscaler 2.5.24

artokun:
Fixes #821.

## Root cause — a confirmed regression from #780, shipped in 0.11.45

`graph_add_node` uses `freshDefs` for **two different jobs**:

1. *"does the live backend still provide `class_type`?"* — #458's authority test, which needs **one** entry; and
2. `registeredSocketTypes(freshDefs)` — *"which datatypes does some installed node declare as an OUTPUT?"*, which needs the **whole schema**.

#780 (`7ed9291`) made job 1 cheap by fetching `/object_info/<class_type>` when the type is already registered. Its module docstring claims:

> THIS CANNOT CHANGE A VERDICT, ONLY THE COST OF REACHING ONE.

That is false for job 2. The same single-class map is handed to `registeredSocketTypes`, so the socket-proof set collapses from *every output type of every installed node* to *the output types of this one class*.

Every custom link datatype produced by a **sibling** node then reads as unproven, `linkProven` goes false, the input waits 5 s for a widget constructor that will never exist, and the refusal asserts something demonstrably untrue:

```
- input "dit" (declared type "SEEDVR2_DIT"): no installed node outputs "SEEDVR2_DIT" ...
```

`SeedVR2LoadDiTModel` outputs exactly that, and was sitting on the canvas.

Timeline confirms the reporter's build is affected:
- `7ed9291` #780 fast path — 2026-08-08 03:24
- `d2c1aa3` 0.11.45 release — 2026-08-08 11:43
- reporter: panel 0.11.45

## Shape of the fix

Keep #780's win (the whole point of the fast path is that the common add never pays for the full schema), and restore correctness where the partial payload is not sufficient evidence: when the guard's report is **non-empty** and the socket proof came from a partial payload, widen it once against the full `/object_info` and re-check — *before* spending the 5 s wait, since a link-proven type has nothing to wait for.

Same idiom as #775's `readImportFailures`: consulted only on the path that is about to refuse, so a healthy add pays nothing.

## Deliberately NOT doing

The report's suggestion 2 — *"treat a type as a socket if a node already in the open graph outputs it"* — is not implemented. Nodes on the canvas are the same class of stale evidence as the LiteGraph registry, which `registeredSocketTypes`' own docstring rejects for this exact reason (a removed/schema-changed pack keeps stale `nodeData.output` positives). The full-schema widen is strictly sound and supersedes it.

---

### DIFF
--- a/browser_tests/unit/add-node-socket-proof-scope.test.mjs
+++ b/browser_tests/unit/add-node-socket-proof-scope.test.mjs
@@ -1,16 +1,459 @@
+// panel#821 — `panel_add_node` refused a node whose input types ARE produced by nodes
+// that were already on the canvas:
+//
+//   Cannot add "SeedVR2VideoUpscaler": 2 required input types had no widget after 5.0s ...
+//     - input "dit" (declared type "SEEDVR2_DIT"): no installed node outputs "SEEDVR2_DIT"
+//     - input "vae" (declared type "SEEDVR2_VAE"): no installed node outputs "SEEDVR2_VAE"
+//
+// The premise was demonstrably false: `SeedVR2LoadDiTModel` (id 11) and
+// `SeedVR2LoadVAEModel` (id 12) had just been added BY THIS SAME TOOL and output exactly
+// those types, and `/object_info` listed all three classes.
+//
+// WHY IT HAPPENED — a regression from #780 (`7ed9291`), shipped in 0.11.45 (`d2c1aa3`).
+//
+// `graph_add_node` reuses one `freshDefs` map for two questions of different arity:
+//
+//   1. #458's authority test — "does the live backend still provide `class_type`?"
+//      One entry is enough; `hasOwnProperty` reads the same either way.
+//   2. the socket proof — `registeredSocketTypes(freshDefs)`, "which datatypes does SOME
+//      installed node declare as an OUTPUT?"  This quantifies over the WHOLE install.
+//
+// #780 made (1) cheap by fetching `/object_info/<class_type>` when the class is already
+// registered — 3 KB instead of 5.4 MB. The same one-entry map then flows into (2), whose
+// answer collapses to "the output types of this one class". Every custom link datatype
+// produced by a SIBLING node becomes unproven, `linkProven` goes false, and the input
+// waits 5 s for a widget constructor that no one will ever register for a link datatype.
+//
+// It is a SIBLING bug, which is why the reporter's own two loaders adding fine was not a
+// contradiction: their required inputs are combos, so they never consulted the proof.
+//
+// THE HARNESS: these tests run the SHIPPED `graph_add_node` body, extracted from the panel
+// source and given injected collaborators — the same technique as
+// add-node-upload-widget.test.mjs. A helper-only test could not have caught this: both
+// `registeredSocketTypes` and `fetchSingleNodeDef` are individually correct. The defect is
+// entirely in which payload the CALL SITE hands to which question.
 import test from "node:test";
 import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
 
-import { registeredSocketTypes } from "../../web/js/lib/node-widget-materialization.js";
+import {
+  applyCurrentDefWidgetValues,
+  driftedRequiredInputNames,
+  missingRequiredWidgetMaterializations,
+  registeredSocketTypes,
+  unavailableRequiredWidgetMessage,
+  unavailableRequiredWidgetReport,
+} from "../../web/js/lib/node-widget-materialization.js";
+import {
+  assertAddNodeResolvableRefreshing,
+  isRegisteredNodeType,
+} from "../../web/js/lib/node-resolve.js";
+import { fetchSingleNodeDef } from "../../web/js/lib/single-node-def.js";
+import {
+  describeUnmaterializedRequiredWidgets,
+  snapshotBackendDef,
+} from "../../web/js/lib/add-node-widget-guard.js";
 
-// panel#821 — placeholder; the real assertions land with the fix.
-test("registeredSocketTypes over a single-class payload cannot see a sibling's output type", () => {
-  const whole = {
-    SeedVR2LoadDiTModel: { output: ["SEEDVR2_DIT"] },
-    SeedVR2LoadVAEModel: { output: ["SEEDVR2_VAE"] },
-    SeedVR2VideoUpscaler: { output: ["IMAGE"] },
+const panelPath = fileURLToPath(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url));
+const panelSrc = readFileSync(panelPath, "utf8");
+
+const addNodeMatch = panelSrc.match(
+  /\n {2}async graph_add_node\(\{ class_type, pos, title \}\) \{[\s\S]*?\n {2}\},/,
+);
+assert.ok(addNodeMatch, "could not locate graph_add_node in panel source");
+
+const awaitWidgetsMatch = panelSrc.match(
+  /\nasync function awaitRequiredCustomWidgetRegistration\([\s\S]*?\n\}/,
+);
+assert.ok(awaitWidgetsMatch, "could not locate awaitRequiredCustomWidgetRegistration");
+
+const placementMatch = panelSrc.match(/\nfunction placementFor\(graph, pos\) \{[\s\S]*?\n\}/);
+assert.ok(placementMatch, "could not locate placementFor");
+
+// ---------------------------------------------------------------------------
+// The reporter's install: seedvr2_videoupscaler, freshly installed, ComfyUI restarted,
+// browser tab NOT reloaded (the panel reconnects on its own and registers the new defs).
+// ---------------------------------------------------------------------------
+
+/** The live whole-schema /object_info. A fresh object each call, exactly like a fetch. */
+function backendObjectInfo() {
+  return {
+    // The two loaders. Their own required inputs are plain combos, which is why they
+    // added successfully and never consulted the socket proof.
+    SeedVR2LoadDiTModel: {
+      name: "SeedVR2LoadDiTModel",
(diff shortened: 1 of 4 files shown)


artokun:
## Browser verification — the reported failure reproduced and fixed, in real Chromium

The panel's Playwright e2e suite needs a live ComfyUI on `BASE_URL`, and this rig's ComfyUI cannot start right now (its `standalone-env` is missing PyTorch — `ModuleNotFoundError: No module named 'torch'`, which is also why Comfy Desktop's Electron shell is up with no backend listening). That is a machine-state problem unrelated to this change.

So the verification was done the way that needs no backend: serve the repo over HTTP, import the panel's lib modules as **real ES modules** in Chromium, evaluate the **shipped** `graph_add_node` body, and drive the reporter's steps 5 → 6 → 7 with the real 5 s registration window.

**Against unfixed `origin/main`:**

| step | result | elapsed |
|---|---|---|
| `SeedVR2LoadDiTModel` | added (id 11) | 1 ms |
| `SeedVR2LoadVAEModel` | added (id 12) | 0 ms |
| `SeedVR2VideoUpscaler` | **refused** | **5025 ms** |

```
Cannot add "SeedVR2VideoUpscaler": 2 required input types had no widget after 5.0s waiting for node extensions to register.
  - input "dit" (declared type "SEEDVR2_DIT"): no installed node outputs "SEEDVR2_DIT" and no frontend widget is registered for it.
  - input "vae" (declared type "SEEDVR2_VAE"): no installed node outputs "SEEDVR2_VAE" and no frontend widget is registered for it.
```

That is the reported error verbatim, including the node ids, on the reported sequence.

**Against this branch:**

| step | result | elapsed |
|---|---|---|
| `SeedVR2LoadDiTModel` | added (id 11) | 1 ms |
| `SeedVR2LoadVAEModel` | added (id 12) | 0 ms |
| `SeedVR2VideoUpscaler` | **added (id 13)** | **0 ms** |

`dit`, `vae` and `images` land as sockets; `seed` is the only widget. Three nodes on the canvas. Whole-schema fetches: **1** — the widen ran only for the add that needed it, and the two loaders never triggered it, so #780's saving is intact. Zero page errors, so the changed modules load cleanly as ES modules in a browser.

## Test evidence

- 3039 panel unit tests pass.
- New suite `browser_tests/unit/add-node-socket-proof-scope.test.mjs` (8 tests) drives the shipped `graph_add_node` body, not the helpers — a helper-only test could not catch this, since `registeredSocketTypes` and `fetchSingleNodeDef` are each individually correct.
- Five mutations, each confirmed applied before its result was trusted, each killed by at least one test:

| mutation | killed by |
|---|---|
| unwire the widen at the call site | reported-add + the #695 source-wiring assertion |
| never mark the payload partial | reported-add |
| widen unconditionally | "already fetched whole" — no double fetch |
| widen after the wait instead of before | the elapsed-time assertion |
| let a failed widen propagate | best-effort fail-closed |

The third mutation initially **survived**; the widen is invoked lazily, so arming it unconditionally costs nothing on the healthy path. That exposed a genuinely untested direction of the flag — don't re-fetch a schema you already hold whole — and a test for it was added.

artokun:
## Codex adversarial review — 1 real finding, fixed in `ac053c5`

Reviewed for six specific hazards. Five came back clean with evidence:

| # | hazard | verdict |
|---|---|---|
| 1 | does the widen introduce a #580 false accept? | safe — the waiver still needs output-proof **and** an input-level socket declaration, so a widget-valued input stays in the report |
| 2 | TOCTOU from the extra await | safe — the widen is entirely before `LG.createNode`, and `revalidateGraphMutationContext` still runs with no await after it |
| 3 | `recordObjectInfoTypes` side effect on the #458 history | safe — `recordTypes` only adds keys; recording a type the backend currently returns makes a later absence correctly read as "removed" |
| 4 | is `currentDef` left stale? | safe by design — it is the per-class snapshot; the widen only recomputes the cross-schema output set |
| 5 | is widen-before-poll the right order? | safe — the poll never refetches, so waiting cannot turn an unproven type into a link-proven one |

**Finding (6) was real and is fixed.** `registeredSocketTypes` maps a null/empty payload to an **empty Set**, which passes the `.has` check and so *replaced* the single-class proof with something strictly weaker.

It is not a false accept — the node still fails closed. What broke is the refusal **message**: the single-class payload does prove the class's own output types, and after the empty replacement the guard would name those too, asserting that no installed node outputs a datatype the panel had proven moments earlier *in the same call*. That is the false-cause refusal #695/#700 were filed about, and it contradicted the best-effort claim made for this widen.

`widenSocketProof` now returns `null` on any doubt — `getNodeDefs` absent, returning `null`, `{}`, or a non-object — and the existing proof stands.

Pinned by five new cases built on the only node shape that reaches the interesting state: two custom required socket types, one produced by the node **itself** (already proven by the single-class payload) and one produced only by a **sibling** (so the report is non-empty and the widen runs). Reverting the guard fails all five on `the proof already in hand must survive a widen that found nothing out`.

3044 unit tests pass. Re-verified in Chromium after the change: the reported sequence still adds cleanly, one whole-schema fetch, no page errors.

*Note: Codex's sandbox denied local shell access, so it reviewed the PR head via the GitHub API rather than the worktree. The finding was independently reproduced and mutation-tested locally before being accepted.*

Issue #251, opened 2026-07-30
panel_search_nodes fails outright when legacy ComfyUI-Manager endpoint is unavailable


artokun:
### What happened / root cause
`panel_search_nodes` returned an immediate error when the connected local ComfyUI used Manager 3.x LEGACY and its built-in Manager endpoint was not reachable. This blocks the normal install-discovery flow even though the agent can often continue using core registered nodes. The tool should detect legacy/partial Manager support and provide a graceful fallback or a more actionable capability result.

### Steps to reproduce
1. Run ComfyUI 0.28.3 with ComfyUI-Manager 3.x LEGACY (partial feature set).
2. Call `panel_search_nodes({query:"RIFE frame interpolation"})`.
3. Observe the call fails instead of returning results or a supported fallback.

### Exact error
`Error: ComfyUI-Manager not reachable (is the built-in Manager enabled?)`

### Fix
Applied locally: no. Upstream-only: yes. Suggested change: capability-detect Manager v3/legacy endpoints and either query the compatible legacy route, fall back to the comfyui-mcp node registry search, or return structured `{supported:false, managerVersion, fallback}` data instead of throwing.

### Environment
Windows 11; NVIDIA RTX 5070 Ti Laptop GPU 12 GB; ComfyUI 0.28.3; comfyui-mcp 0.48.8; sidebar panel version not exposed in the startup environment; Manager 3.x LEGACY.

PR #826, opened 2026-08-09
fix(manager): an EMPTY catalogue is not 'no matches' (#808)

### LINKED ISSUE #808: An unreachable Manager catalogue renders as an empty node list, not as 'could not reach it'
Split out of artokun/comfyui-mcp#1136 — the orchestrator half is fixed in artokun/comfyui-mcp#1139; this is the half that lives here.

## The symptom

A Chinese-speaking user was told to update the panel through ComfyUI Manager and replied:

> 我这边搜不到任何内容 — *"I can't find anything when I search"*

Read at the time as a stale Manager cache. A blocked registry fits the evidence better. Three consecutive pieces of advice were spent sending them at a door that could not open, and neither of us could see that from the symptom — because **empty and unreachable look identical**.

Empty reads as "nothing matched", so the user concludes their query was wrong, or that the thing does not exist, and goes on trying variations of an action that can never succeed.

## Ask

Anywhere the panel renders a list that came from an external host — the Manager catalogue in particular, since that is the route we tell everyone to use for installing and updating — a fetch that failed (DNS, timeout, connection refused, TLS) must render **distinctly from an empty result**, and must **name the host** so a user behind a filter recognises their own situation immediately.

## The pattern already exists here

`panel_civitai_results` gets this right and its tool description says so explicitly: a `total:0` result is **not** automatically "no matches", because an `error` field (e.g. a 503) may be present, and callers are told to check it before concluding anything from an empty grid.

That is the rule. It has simply never been applied to the Manager-catalogue surface.

## For reference: what the orchestrator side now does

`search_custom_nodes` (`api.comfy.org`) and `download_model action:"search"` (`huggingface.co`) now say, on a transport failure:

- **the host** that could not be reached;
- that **nothing was queried**, so nothing follows about whether the thing exists;
- that corporate/campus/national filtering is a likely cause, since a user cannot be expected to infer that from `ENOTFOUND`;
- and for HuggingFace, the `HF_ENDPOINT` mirror escape hatch.

A **timeout is called a timeout** and gets "retry", not "your network is filtered" — the host may be reachable and merely slow, and sending someone to a VPN for a problem a retry would have solved is the same class of wrong answer pointing the other way. Worth mirroring that distinction here.

## Not asked for

Not CDN mirrors, and not a China-specific mode. Just make unreachable say *unreachable*.

artokun:
Fixes #808. Panel half of the split; the orchestrator half shipped in artokun/comfyui-mcp#1139.

## What the panel can honestly say

The panel does not perform the outbound registry fetch — ComfyUI-Manager does — so it cannot observe DNS/timeout/TLS directly and must not claim to have diagnosed the transport.

But it *can* distinguish the two states the report says look identical, and the discriminator is already local:

| observed | today | should be |
|---|---|---|
| Manager unreachable on every route | structured `{supported:false}` capability result | unchanged, already correct |
| Manager reachable, catalogue has packs, none match the query | `count: 0` | unchanged — a genuine no-match |
| **Manager reachable, catalogue has ZERO packs** | **`count: 0` — indistinguishable from a no-match** | **distinct: nothing was searched** |

`searchNodesVia` requests `customnode/getmappings?mode=cache`, so Manager serves from its LOCAL cache. A cache that was never populated — because Manager itself could not reach the registry, which is exactly the blocked-registry case in the report — answers HTTP 200 with `{}`. `parseNodeMappings` then filters zero entries and returns `count: 0`, which reads as "nothing matched", so the user concludes their query was wrong and tries variations of an action that can never succeed.

That is the reported failure, and a zero-entry catalogue is sound evidence for it: a working Manager on any real install has thousands of packs.

## Plan

1. Expose the catalogue size (entries **before** query filtering) from `parseNodeMappings`, so "empty" and "missed" stop collapsing.
2. When the catalogue is structurally empty, return a distinct result saying **nothing was searched**, so nothing follows about whether the pack exists — naming the hosts Manager depends on so a filtered user recognises their own situation, and pointing at Manager's cache-update as the remedy.
3. Keep a genuine no-match exactly as it is today.

Following `panel_civitai_results`, which the report correctly cites as the existing right answer: a `total:0` is not automatically "no matches" when an error field may be present.

## Deliberately not doing

Not claiming a cause the panel did not observe. The panel saw "Manager answered; its catalogue is empty" — it did not see a DNS failure, so it will not say one occurred. Per the report's own timeout-vs-filtering point, sending someone to a VPN for a problem a cache refresh would have solved is the same class of wrong answer pointing the other way.

Not touching `nodes_list`: an empty INSTALLED list is legitimate.

### DIFF
--- a/browser_tests/unit/manager-catalogue-empty.test.mjs
+++ b/browser_tests/unit/manager-catalogue-empty.test.mjs
@@ -1,14 +1,200 @@
-// panel#808 — an unreachable Manager catalogue renders as an empty node list, not as
-// "could not reach it". Placeholder pinning the CURRENT conflation; the fix follows.
+// panel#808 — an unreachable Manager catalogue rendered as an empty node list rather
+// than as "could not reach it".
+//
+// THE REPORTED FAILURE. A Chinese-speaking user was told to update the panel through
+// ComfyUI Manager and replied "我这边搜不到任何内容" — I can't find anything when I search.
+// That was read as a stale cache; a blocked registry fits the evidence better. Three
+// rounds of advice were spent sending them at a door that could not open, and neither
+// side could see it from the symptom, because EMPTY AND UNREACHABLE LOOK IDENTICAL.
+//
+// THE MECHANISM. `searchNodesVia` requests `customnode/getmappings?mode=cache`, so
+// Manager serves from its own local cache. A cache Manager never managed to populate —
+// because Manager itself could not reach the registry — answers HTTP 200 with `{}`.
+// `parseNodeMappings` filters zero entries and returns `count: 0`, which is exactly what
+// a healthy catalogue returns when the query matches nothing. The reader takes the first
+// for the second, concludes the pack does not exist, and keeps trying variations.
+//
+// THE DISCRIMINATOR is local and needs no new signal from Manager: how many packs the
+// payload CONTAINED, before the query filter. Zero is sound evidence — a working install
+// has thousands.
+//
+// WHAT IS DELIBERATELY NOT CLAIMED. The panel does not make the registry request, so it
+// never observed DNS/timeout/TLS and must not report one. These tests pin that the
+// message states only what was observed.
 import test from "node:test";
 import assert from "node:assert/strict";
-import { parseNodeMappings } from "../../web/js/lib/manager-install.js";
 
-test("#808 today: an EMPTY catalogue and a populated one that misses are indistinguishable", () => {
-  const populated = { "https://github.com/a/b": [[], { title: "Impact Pack" }] };
-  const empty = {};
-  // Both collapse to the same "nothing here" answer, so a caller cannot tell whether
-  // the query missed or the catalogue was never populated.
-  assert.deepEqual(parseNodeMappings(empty, "seedvr2", 15), { count: 0, results: [] });
-  assert.deepEqual(parseNodeMappings(populated, "seedvr2", 15), { count: 0, results: [] });
+import {
+  catalogueSize,
+  emptyCatalogueResult,
+  parseNodeMappings,
+  searchNodesVia,
+} from "../../web/js/lib/manager-install.js";
+
+/** A catalogue in the documented MAP shape: repo URL -> [ [classes…], meta ]. */
+const POPULATED = {
+  "https://github.com/ltdrdata/ComfyUI-Impact-Pack": [
+    [],
+    { title: "Impact Pack", description: "detailer and background tools" },
+  ],
+  "https://github.com/rgthree/rgthree-comfy": [[], { title: "rgthree-comfy" }],
+};
+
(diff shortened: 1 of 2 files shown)


artokun:
## Review — codex wedged, disclosed Claude adversarial review substituted

The codex review task backgrounded and its runner could not poll its own job, so it returned a task handle and, on a follow-up, an honest "no findings yet — I have not fabricated anything" rather than inventing a verdict. Per the established fallback for a wedged codex, an adversarial review was run here instead and is disclosed as such.

**It found a real defect in my own first cut, and it is the same defect #808 is about.**

The original message told the user an empty catalogue meant the registry hosts were blocked. That was an *assumption*. ComfyUI-Manager's own source says otherwise — `glob/manager_core.py`, `get_data_by_mode`:

```python
except Exception as e:
    print("[ComfyUI-Manager] Due to a network error, switching to local mode.")
    uri = os.path.join(manager_util.comfyui_manager_path, filename)
    json_obj = await manager_util.get_data(uri)
```

A network failure does **not** empty the catalogue — it falls back to the bundled `extension-node-map.json` (2.2 MB and populated on this machine's stock install). `{}` comes from the `network_mode == 'offline'` path with neither cache nor local file, or an empty/unreadable data file.

Also wrong: the message named `api.comfy.org`. `DEFAULT_CHANNEL` is `https://raw.githubusercontent.com/ltdrdata/ComfyUI-Manager/main` (manager_core.py:51); api.comfy.org serves pack installs, not this mapping. Naming the wrong host sends a filtered user to check something irrelevant — failing the report's ask as surely as naming no host.

Both corrected in `994b440`, and both pinned by mutation.

**The branch survives the review and is better justified.** Zero packs means Manager assembled a catalogue from *none* of channel, cache, or bundled copy — genuinely anomalous, so it cannot fire on a healthy install.

### Other hazards checked

| hazard | verdict |
|---|---|
| response shape — is `getmappings` a wrapper? | bare map, confirmed at `manager_server.py:778`; `catalogueSize` counting top-level keys agrees exactly with what `parseNodeMappings` iterates |
| does the new branch shadow #251/#255 (Manager unreachable)? | no — pinned by test |
| does it shadow #426 (installed-node `/object_info` fallback)? | no — pinned by test |
| result-shape compatibility | the orchestrator relays `nodes_search` verbatim via `ctx.call` with no destructuring, so added fields are safe |
| false positive on a populated catalogue that misses | pinned — still an ordinary no-match, and mutating the discriminator to `count` kills that test |

### Known limitation — stated, not papered over

Because Manager degrades to the bundled copy, a genuinely blocked channel surfaces as a **stale** catalogue rather than an empty one, and the panel cannot yet tell stale from current. Closing that needs a signal Manager does not expose today. Inventing a staleness claim here would repeat the exact fault this issue reports, so it is documented in the code instead.

3058 unit tests pass. Five mutations each killed.

Issue #356, opened 2026-07-30
panel_get_errors reports stale missing_model after panel_set_widget fixes it; panel_run completion notification never delivered


artokun:
Two separate panel defects hit in the same session. Both are recoverable but both actively mislead the agent.

---

### What happened

A Z-Image Turbo workflow loaded with node 63 `VAELoader` set to `vae_name = "ae.safetensors"`, which this install doesn't have. Correctly flagged red at load time.

I fixed it with `panel_set_widget(63, "vae_name", "z-image-ae.safetensors")` — the file that *is* installed. The call returned `{previous: "ae.safetensors", value: "z-image-ae.safetensors"}`, and the workflow then **ran successfully** (`Prompt executed in 86.03 seconds`, VAE decoded normally, image saved).

But `panel_get_errors` afterwards still reports:

```json
{
  "id": 63,
  "type": "VAELoader",
  "widgets": { "vae_name": "z-image-ae.safetensors" },
  "red_outline": true,
  "reasons": [{
    "kind": "missing_model",
    "file": "ae.safetensors",
    "directory": "vae",
    "widget": "vae_name",
    "download_url": "https://huggingface.co/Comfy-Org/z_image_turbo/resolve/main/split_files/vae/ae.safetensors"
  }]
}
```

Note the self-contradiction **inside one response object**: `widgets.vae_name` shows the new correct value while `reasons[].file` still names the old missing one. The `⚠️ MISSING ASSETS` turn-start banner also re-fired with the stale entry, telling me the canvas had red nodes "RIGHT NOW" when it didn't.

### Root cause (suspected)

The load-time missing-asset scan appears to cache per-node results keyed by node id and is never invalidated when `panel_set_widget` writes to a widget that a `missing_model` reason points at. The reason survives even though the graph state it was derived from is gone — and even though a successful execution has since proven the asset resolves.

### Suggested fix

On `panel_set_widget`, drop any cached `missing_model` / `missing_media` reason whose `widget` matches the widget being written on that node id (cheap, targeted). Optionally also clear a node's missing-asset reasons when a run completes without error. At minimum, `panel_get_errors` should not emit a `reasons[].file` that disagrees with the node's own current `widgets` value in the same payload — that combination is always a bug.

### Impact

I nearly re-downloaded a 300 MB VAE that was already present and correctly wired, and had to tell the user to ignore a red node on their canvas. An agent that trusts the banner will either download redundant weights or refuse to run a workflow that is actually fine.

[...]

PR #831, opened 2026-08-09
fix(run-completion): a run that produces no media must still tell the agent it finished (#356)

artokun:
Addresses the second half of #356 — `panel_run` promises a completion notification that never arrives. (Bug 1 in that issue was fixed in 0.11.27.)

## Root cause

The reporter's most recent recurrence (comfyui-mcp 0.50.36 / panel 0.11.44) is the clean one: a cache-hit run, `total_duration_ms: 3`, `status_str: success`, **three text outputs**, no completion delivered.

The completion path is gated on media in **two** places, and text outputs are neither images nor videos:

**1. The live path** — `web/js/lib/run-completion-frame.js:150`

```js
if (!outImages.length && !noteSections.length) return null;
```

and the call site treats that as success:

```js
// frame===null ⇒ empty batch (nothing to deliver) ⇒ delivered.
if (frame == null || framePushed) runCompletion.markDelivered(promptId);
```

So a media-less run composes nothing, is marked **delivered**, and no frame is ever sent.

**2. The recovery path** — `web/js/lib/run-completion.js:471`

```js
const hasBatch = parsed.images.length > 0 || parsed.videos.length > 0;
if (hasBatch) { ...onFlush({...}) }
return { promptId, status: "success", delivered: hasBatch };
```

The `/history` reconcile — the safety net built for #370/#468 precisely to recover missed completions — also refuses to flush without media. So the one mechanism that should have caught this cannot.

Net effect is what the reporter called corrosive, and they were right about why: `panel_run` tells the agent *"you will be notified automatically — do NOT poll — just end your turn now and wait"*. For a media-less run that notification can never arrive, so the agent waits forever, the user thinks it is stuck, and the only escape is prompting again. The instruction is what converts a dropped event into an indefinite stall.

## Intended fix

- Deliver a completion for a run that finished with no image/video outputs, saying so plainly and reporting the duration, rather than composing nothing.
- Let the reconcile path flush a media-less terminal run too, so the `/history` safety net covers it.

## Scoping — deliberately narrow

Only runs the **panel queued** get the media-less completion. `onQueued` is called from the panel's own queue path, so the pending ledger already distinguishes an agent-initiated run — the one that carried the "end your turn and wait" instruction — from a run the user started in the ComfyUI UI. A user's own media-less run stays silent, so this adds no new wakeups beyond the promise that was actually made.

## Not in scope

The first recurrence on this issue (`to_node_id: 146`, queued twice) may be a different mechanism; this PR targets the reproducible media-less path and will say so rather than claiming the whole issue.

### DIFF
--- a/browser_tests/unit/run-completion-no-media.test.mjs
+++ b/browser_tests/unit/run-completion-no-media.test.mjs
@@ -3,11 +3,20 @@ import test from "node:test";
 import assert from "node:assert/strict";
 import { composeRunCompletionFrame } from "../../web/js/lib/run-completion-frame.js";
 
+const deps = (sent) => ({
+  sendFrame: (f) => (sent.push(f), true),
+  coerceMessageText: (s) => String(s ?? ""),
+  formatDuration: (ms) => `${ms}ms`,
+  formatClock: () => "12:00:00",
+  agentReceivesImages: () => true,
+  warn: () => {},
+});
+
 test("#356 today: a run with no images and no videos composes NO frame at all", async () => {
   const sent = [];
   const frame = await composeRunCompletionFrame(
     { promptId: "p1", images: [], videos: [], durationMs: 3 },
-    { sendFrame: (f) => (sent.push(f), true), agentReceivesImages: () => true, warn: () => {} },
+    deps(sent),
   );
   // The agent was told to end its turn and wait. Nothing is ever sent.
   assert.equal(frame, null);
--- a/browser_tests/unit/run-completion-no-media.test.mjs
+++ b/browser_tests/unit/run-completion-no-media.test.mjs
@@ -1,24 +1,208 @@
-// panel#356 Bug 2 — placeholder pinning the CURRENT behaviour; the fix follows.
+// panel#356 Bug 2 — `panel_run` promises a completion notification that never arrives.
+//
+// THE REPORTED FAILURE, in the reporter's own words: panel_run returns
+//
+//   [IMPORTANT] You will be notified automatically with the output image(s)/video when
+//   the render finishes — do NOT poll get_queue, get_history, or list_output_images.
+//   Just end your turn now and wait for the result to be delivered to you.
+//
+// …and then nothing arrives. They called it corrosive *because* the tool tells the agent
+// to trust it and not poll: a dropped notification becomes an indefinite silent stall,
+// the agent believes it is still rendering, and the only escape is the user prompting
+// again. That is exactly right, and it is why silence here is a defect and not merely an
+// omission — the PROMISE is what converts a missing event into a wedged agent.
+//
+// THE MECHANISM, for the clean 0.11.44 recurrence (cache hit, total_duration_ms: 3,
+// status success, THREE TEXT OUTPUTS, nothing delivered). The completion path was gated
+// on media in four places, and a text output is neither an image nor a video:
+//
+//   1. onExecuted            — `if (!images.length && !videos.length) return;`
+//                              so a media-less `executed` never creates a buffer;
+//   2. flush()               — `if (!buf) return;` then a second empty-batch guard;
+//   3. reconcileOne()        — `if (hasBatch)` gates the flush, and it reports
+//                              `delivered: hasBatch`, so the /history safety net built
+//                              for #370/#468 could not recover this case either;
+//   4. composeRunCompletionFrame — `return null`, which the call site reads as
+//                              "empty batch ⇒ nothing to deliver ⇒ delivered".
+//
+// So the run was marked delivered and no frame was ever sent.
+//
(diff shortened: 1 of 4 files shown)


artokun:
## Verification

**Browser (real Chromium, modules loaded over HTTP as the panel loads them):**

| scenario | flushes | noMedia | notes |
|---|---|---|---|
| the reporter's cache hit — panel-queued, text-only, 3 ms | **1** | `true` | duration reported as 3 ms |
| user's own canvas run, no media | **0** | — | stays silent; no new wakeups |
| panel-queued **with** media | **1** | `false` | 1 image, unchanged |
| composed frame | sent **1** | — | note carries "This IS the completion you were told to wait for" |

Zero page errors.

**Unit:** 3110 pass.

**Mutations — six, each confirmed applied before its result was trusted, each killed:**

| mutation | killed by |
|---|---|
| drop the media-less flush entirely | the panel-queued cache-hit test |
| widen the scope from `panelQueued` to `pending` | "a user canvas run stays silent" |
| fire even for runs that DID buffer media | "no extra flush alongside the real one" |
| composer ignores the `noMedia` flag | "composes and SENDS a frame" |
| composer sends for EVERY empty batch | "an empty compose without the flag still returns null" |
| read `panelQueued` AFTER `markDelivered` | the /history reconcile test |

## Two things the tests caught that I had wrong

**1. `pending` is not the agent-initiated set.** I first scoped the media-less completion to `pending`, assuming it held panel-queued runs. It does not — `onExecutionStart` and `onExecuted` both call `trackPending`, so it holds every run seen on the socket, including the user's own canvas runs. The existing test `empty completion (no buffered output) never emits an event` failed and was pinning a genuine invariant, not stale harness. Hence the dedicated `panelQueued` set, populated only by `onQueued`; that existing test still passes untouched.

**2. The reconcile path read the flag after retiring it.** `reconcileKey` calls `markDelivered(k)` — which clears both `pending` and `panelQueued` — about thirty lines *before* the media-less branch consulted `panelQueued.has(k)`. So `/history` recovery for a media-less run could never fire, while reading as though it did. Fixed by capturing before the retire, matching what the live path already does. Reverting it fails the reconcile test.

## Scope

Second half of #356 only (Bug 1 shipped in 0.11.27). The earlier `to_node_id: 146` recurrence on that issue may be a different mechanism and is **not** claimed by this PR.

artokun:
## Codex adversarial review — 1 finding accepted and fixed, 1 already fixed, 1 assessed

Codex reviewed an earlier head, so its **Finding 1** (reconcile reads `panelQueued` after `markDelivered` clears it) had already been caught by the reconcile test I added and fixed in `b6c7161`.

But Finding 1 carried a **second half I had not fixed**, and it was real:

> `markUndelivered` re-pends the run but does not restore `panelQueued`, so a later recovery attempt also stays silent.

The tracker's internal `markDelivered` is **optimistic** — flush, `execution_success` and reconcile all call it before the caller has confirmed the frame reached the agent — and `markUndelivered` re-pends when `sendFrame` reports the bridge down. Retiring the panel-queued mark there made that re-pend unrecoverable for a media-less run: on the way back it looked like an ordinary canvas run and stayed silent forever. **The fix for this stall would have reintroduced the stall one layer down.**

Fixed in `39ee097`: the mark now survives the optimistic retire and is cleared on *confirmed* delivery (the public `markDelivered`, where the #585 delivery hold is also released) and on eviction. Pinned by a bridge-down test — live success delivers, `markUndelivered` re-pends, `/history` reconcile redelivers. Restoring the optimistic delete fails it.

**Finding 2** (a media-less run whose `execution_success` is dropped while the socket stays up gets no recovery from `onExecutionStart`'s sequential flush, since it has no buffer to flush) is accurate as written, but its consequence is smaller than stated: the run stays in `pending`, and the panel arms a periodic `/history` reconcile sweep whenever runs are pending — added under this very issue. With Finding 1 fixed, that sweep can now actually deliver a media-less completion, which it could not before. So the path is covered; closing the gap *at* the sequential flush would be a narrowing, not a fix for a stall, and is not attempted here.

Codex's clean verdicts on double delivery, `onExecutionSuccess` ordering, premature/cancelled completion, leaks, and frame propagation match what I checked independently.

## Also in this push

- **Vocabulary gate failure fixed.** The reporter's verbatim `panel_run` text quoted in the header named two tools retired in 0.49.0/0.50.0, and the gate correctly refuses a dead tool name even inside a quotation. Paraphrased, with a note explaining why.
- **One test relabelled rather than kept.** "A confirmed delivery retires the panel-queued mark" **survived** its mutation — the no-duplicate guarantee comes from the `terminal` fence, not that line. The retire is memory-bounding and its absence is not observable through the public API, so it is now explicitly documented as unpinned instead of being guarded by a test that would pass either way.

3112 unit tests pass. Eight mutations run; seven killed, and the one that survived is disclosed above rather than quietly dropped.

Issue #132, opened 2026-07-23
Long Codex turns appear to end silently after 120s while the turn is still running


JusticeWay:
## Environment

- ComfyUI Agent Panel: current `main` inspected at `add992082301ca03f0052a8658cdfb1253ba70cc`
- ComfyUI Desktop `1.0.29`
- `comfyui-mcp 0.46.0`
- Codex app-server backend, `gpt-5.6-sol`, effort `high`
- Windows 11

## Symptom

During a long Codex turn, the working/thinking indicator disappears after about two minutes. The composer then looks idle even though the orchestrator still owns the turn and newly submitted messages remain queued. If the agent is executing tools without emitting user-visible text, the conversation can appear to have ended on the user's last message with no reply.

In the live reproduction, a passive bridge observer continued receiving normal `stream` frames, an `action` frame (`panel.panel_query_graph`), and eventually a committed `say` frame. The orchestrator logged the turn as `completed`. The backend and bridge were therefore alive; the panel had hidden the only visible sign of progress.

## Root causes

1. `THINKING_SAFETY_MS` is 120 seconds and `armSafety()` unconditionally calls `hideThinking()` when it fires. It does not check the authoritative `agentWorking` state, so a healthy long-running turn is visually cleared.
2. Desktop Panel ignores bridge frames of `{ type: "action", name }`. Long tool phases therefore look completely silent even when the orchestrator is reporting the current operation.
3. `resetFeed()` removes `.cmcp-thinking` from the DOM but can leave `thinkingEl` / `thinkingLabel` pointing at the detached element. A later `showThinking()` then becomes a no-op. This is reachable when an async history/thread repaint runs during a live turn.
4. A transient bridge disconnect also hides the indicator even if the turn is still authoritative and will be rebound/resumed.

The first two issues are present on current `main`; the third was exposed while testing the IndexedDB history branch but is a general stale-DOM-reference problem in the panel lifecycle.

## Expected behavior

- A live turn remains visibly active until an authoritative terminal frame (`turn:done`, terminal error, or explicit local interruption).
- A safety timer may repair/recreate the indicator, but must not make a still-running turn look finished.
- `action` frames should update the activity label, e.g. `panel.panel_query_graph` → `Using query graph…`.
- Feed/history repaint must clear detached indicator references and recreate the indicator when `agentWorking` is still true.
- A transient disconnect should show reconnecting state without falsely presenting the turn as complete.

## Focused PoC

I implemented a local focused patch that:

[...]

PR #837, opened 2026-08-09
fix(a2ui): a same-thread repaint must not silently kill a live card (#832)

### LINKED ISSUE #832: panel_ui_update loses freshly rendered card during same-turn repaint
A card created by panel_ui_render returned a valid card_id, but an immediate panel_ui_update in the same agent turn failed because the card was no longer present in the panel's liveA2uiCards registry. No user click, dismissal, or workflow/tab switch occurred between the two calls.

Inspection of comfyui-agent-panel/web/js/comfyui-mcp-panel.js shows onUiUpdate requires liveA2uiCards.get(card_id), while resetFeed() unconditionally clears liveA2uiCards and paintA2UIRecord() replays persisted cards inert. A same-thread history repaint between tool calls therefore appears able to invalidate a freshly returned card_id.

### Steps to reproduce
1. Call panel_ui_render with a valid A2UI card.
2. Receive card_id (observed: a2ui-mslhgcem-8).
3. Immediately call panel_ui_update with that card_id and a valid replacement spec.
4. Observe the error below without any intervening user action or view switch.

### Exact error (scrubbed)
Error: no live card "a2ui-mslhgcem-8" (already resolved, dismissed, or from a previous view)

### Fix
Applied locally: no
Upstream-only: yes for this session, because the running frontend would require a hard refresh and changing it would interrupt the user's active planning interaction.

Precise change needed: preserve or re-register unresolved A2UI records as live during a same-thread repaint, while retaining the existing thread/view ownership guard. One option is for paintA2UIRecord() to render unresolved records with live handlers and reinsert their handle into liveA2uiCards; resolved records remain inert. Alternatively, avoid resetFeed() clearing live card handles when repainting the authoritative active thread during an active turn. Add a regression test covering render -> same-thread repaint -> update.

Temporary workaround: render a fresh corrected card instead of updating the lost card.

### Environment
Windows 11
ComfyUI 0.31.0
ComfyUI frontend 1.48.7
NVIDIA RTX 4080 SUPER 16 GB
comfyui-mcp 0.50.46
comfyui-mcp-panel 0.11.43

Related (same root cause): #132

artokun:
Fixes #832. Same root cause as #132.

## Root cause — confirmed, and one layer deeper than reported

The report is right about the mechanism:

- `appendA2UICard` renders a **live** card and registers it: `liveA2uiCards.set(handle.cardId, { handle, rec })`
- `resetFeed()` unconditionally does `liveA2uiCards.clear()`
- `paintA2UIRecord()` replays every persisted record **inert** via `renderA2UIInert(...)` and re-registers nothing
- `onUiUpdate` requires `liveA2uiCards.get(card_id)`

So a same-thread repaint between two tool calls turns an unresolved card inert and drops the handle, and `panel_ui_update` fails with `no live card "…"` — with no user action, dismissal, or view switch, exactly as reported.

**The part the report could not see from the outside:** even re-rendering the record live on repaint would not fix it, because **the record never stores its `card_id`**:

```js
const rec = { role: "card", kind: "a2ui", spec, resolved: false, choice: null };
```

The id lives only on the transient `handle`. And `renderA2UICard` mints a fresh one every call:

```js
export function renderA2UICard(spec, { onAction, onDismiss } = {}) {
  const cardId = `a2ui-${Date.now().toString(36)}-${++_cardSeq}`;
```

So a repaint would re-register under a **new** id while the agent still holds the old one — the update would keep failing, just for a different reason. Restoring liveness requires restoring *identity*, which needs both a persisted id and a way to render with it.

## Shape of the fix

1. `renderA2UICard` accepts an optional `cardId` and mints one only when none is supplied.
2. `appendA2UICard` persists `rec.cardId`, so the id the agent holds survives into history.
3. `paintA2UIRecord` renders an **unresolved** record live — same action/dismiss handlers — and re-registers it under its original `cardId`. A **resolved** record stays inert exactly as today.

The thread/view ownership guard is untouched: repaint still only replays the records of the thread being painted, so a card cannot come back to life in a view it does not belong to.

## Regression coverage

The report asks for render -> same-thread repaint -> update, which is the test that would have caught this. That plus: a resolved card stays inert after repaint; a dismissed card is not resurrected; and the id is stable across the repaint rather than re-minted.

### DIFF
--- a/browser_tests/unit/a2ui-repaint-live-card.test.mjs
+++ b/browser_tests/unit/a2ui-repaint-live-card.test.mjs
@@ -121,3 +121,15 @@ test("#832 first paint and repaint mount through the SAME function", () => {
 test("#832 the live registry is written on both paths, via that shared function", () => {
   assert.match(fn(panelSrc, "mountLiveA2UICard"), /liveA2uiCards\.set\(handle\.cardId, \{ handle, rec \}\)/);
 });
+
+test("#832 the card is RECORDED before it is mounted — record() can repaint", () => {
+  // record() can reach detachInvalidCurrentThread(), which calls resetFeed() (clearing
+  // liveA2uiCards) and repaints. Mounting first would place the card into a feed that
+  // the very next statement can wipe, and the card is not yet in any thread, so the
+  // repaint would not bring it back. This is the ordering the pre-refactor code had.
+  const append = fn(panelSrc, "appendA2UICard");
+  assert.ok(
+    append.indexOf("record(rec)") < append.indexOf("mountLiveA2UICard(rec)"),
+    "record(rec) must precede the mount",
+  );
+});
--- a/browser_tests/unit/a2ui-repaint-live-card.test.mjs
+++ b/browser_tests/unit/a2ui-repaint-live-card.test.mjs
@@ -136,3 +136,20 @@ test("#832 the card is RECORDED before it is mounted — record() can repaint", 
     "record(rec) must precede the mount",
   );
 });
+
+test("#832 (codex): the PRE-HYDRATION restore paint replays cards inert", () => {
+  // The synchronous restore pass declares itself "paint-only": settings are not hydrated,
+  // so the thread it paints comes from a tab pointer and may not be the authoritative one.
+  // Mounting a card LIVE there would register one belonging to a thread about to be
+  // replaced, and an update could land on it — a correctness bug traded for a convenience
+  // one. Liveness waits until the real thread is chosen.
+  assert.match(panelSrc, /let a2uiPaintProvisional = false;/);
+  assert.match(
+    fn(panelSrc, "paintA2UIRecord"),
+    /m\.resolved !== true && !a2uiPaintProvisional/,
+    "the live branch is suppressed during the provisional paint",
+  );
+  const restore = panelSrc.match(/\(function restoreLastThread\(\) \{[\s\S]*?\n  \}\)\(\);/)[0];
+  assert.match(restore, /a2uiPaintProvisional = true;/);
+  assert.match(restore, /finally \{\s*a2uiPaintProvisional = false;/, "reset even if paint throws");
+});
--- a/browser_tests/unit/a2ui-repaint-live-card.test.mjs
+++ b/browser_tests/unit/a2ui-repaint-live-card.test.mjs
@@ -1,0 +1,26 @@
+// panel#832 — placeholder pinning the CURRENT behaviour; the fix follows.
+import test from "node:test";
+import assert from "node:assert/strict";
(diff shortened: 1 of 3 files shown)


artokun:
## Codex review — 2 real findings. One fixed; **one blocks merge.**

First, a correction to my own process: I called codex wedged and started a substitute review. It was not wedged, just slow (~6 min), and it caught two things my own pass missed. The substitute review is not a replacement here.

### Fixed — id collision (`cmcp-a2ui.js:537-556`)

`_cardSeq` restarts at 0 on every module load, and an id **reused** from a persisted record never advances it. So a freshly minted id can equal one a repaint has already put back into `liveA2uiCards`, silently overwriting that entry — an update would then land on the wrong card. Minted ids now carry a random nonce, which removes the class rather than narrowing it. My own review had dismissed this by reasoning about `Date.now()` differing; that argument does not cover a reused id failing to advance the counter.

### BLOCKS MERGE — cross-thread resurrection (`comfyui-mcp-panel.js:27072-27087` vs `:27103-27140`)

> The synchronous reload-restore path calls `paintThread(pointed)` off `CURRENT_THREAD_KEY` *before* scope/settings hydration selects the authoritative thread, so a prior thread's unresolved cards can mount live in a window before the real thread is chosen.

This is the one thing this change must not do. My own review checked that `paintA2UIRecord` is only called within the painted thread's `msgs` loop and concluded "no cross-thread resurrection" — but that only establishes it is scoped to *whatever thread `paintThread` was handed*, and the reload path can hand it the wrong one. Before this PR that window was harmless, because every replayed card was inert; with live replay it can mount a live card belonging to a thread that is about to be replaced.

I am not merging on that. The fix is to defer live mounting on the pre-hydration restore path until the authoritative thread is selected (or render inert there specifically), and it needs to be done deliberately rather than bolted on at the end of a long session.

### Acknowledged, not fixed — test coverage (finding 7)

Codex is right that the suite is source-text assertions and would catch none of findings 1–6. I disclosed that limitation in the test file and the PR description rather than implying otherwise, but disclosure is not coverage. A DOM harness exercising `resetFeed` → repaint → `onUiUpdate` is what this actually needs, and finding 2 is the proof: a structural assertion cannot see which thread `paintThread` was handed.

### Clean per codex

Duplicate registration/leaks (`resetFeed` removes every log child first), `setChatSurfaceForCards`/`scrollLog` ordering, and other `liveA2uiCards` readers. Stale "inert" comments were flagged and one is corrected in `4a1…`; no executable path relied on them.

**Status: staying in review.** The bug is real and correctly diagnosed, and the identity half of the fix is sound — but shipping it with a known live-card-in-the-wrong-thread window would trade a recoverable annoyance for a correctness bug.

PR #839, opened 2026-08-09
fix(a2ui): a same-thread repaint must not lose a card the same turn just created (#832)

artokun:
> A card created by `panel_ui_render` returned a valid `card_id`, but an immediate `panel_ui_update` in the same agent turn failed because the card was no longer present in the panel's `liveA2uiCards` registry. No user click, dismissal, or workflow/tab switch occurred between the two calls.

The reporter's analysis is right, and there is a second half not visible from the source they had.

## Two code paths for one card

Creating a card and repainting it were **different code**. `appendA2UICard` rendered live, wired the handlers and registered the handle; `paintA2UIRecord` rendered `renderA2UIInert` — no handlers, no registration. So any `paintThread` between the two tool calls left the card visible and the registry empty.

Worse, **the id could not have survived even if it had re-registered**: `renderA2UICard` minted a fresh `cardId` on *every* call, and the record never persisted one. The agent's only handle on a card was per-mount state.

## The fix

One mount path. `mountLiveA2UICard(rec)` wires the handlers, registers the handle and appends the element, and both paths go through it. `renderA2UICard` accepts an existing `cardId` and reuses it — a blank or non-string one is ignored rather than adopted, since an empty id would collide every such card in the registry — and the record carries its id from first mount, persisted, so it keeps naming the same card across repaints and reloads.

`appendA2UICard` mounts **before** it records, so the stored record already carries its id; a record persisted without one cannot be re-registered on replay.

## What stays refused

**Resolved** records still repaint inert — they were answered or dismissed, and nothing should answer them twice. And `resetFeed()` still clears the whole registry, so only records belonging to the thread now being painted come back. A `ui_update` naming another thread's card still finds nothing and still refuses. The guard's stated reasons hold exactly as before: no detached element is repainted, and no background thread's record is mutated. It simply no longer refuses for a card sitting in front of the user.

## A quieter defect this also fixes

An unresolved card replayed through `renderA2UIInert` is resolved with a null choice — so a user who reloaded mid-question was left looking at a card that **rendered as already answered and could not be clicked**. A pending question is answerable again after a reload, which is what it always looked like it should be.

`setChatSurfaceForCards()` also now runs after the replay: `resetFeed()` runs it against an empty registry, so a repainted unresolved `surface:"wide"` card would otherwise come back live in a narrowed sidebar.

## Review

Codex's sandbox denied it shell access on the first two attempts, so it returned its four safety questions unanswered rather than a grounded review. They were the right questions; I answered them from the code and **pinned all three testable ones** so the answers stay true — cross-thread leakage (impossible by construction: `paintThread`'s first statement is `thread = t`), double registration (exactly one caller, after `resetFeed` cleared the map), and a post-reload click when the agent is gone (`sendCardReply` already surfaces "agent disconnected" rather than dropping the reply).

Re-run with the source inlined, it raised one finding, which I am **declining**: that a `ui_update` for a card in a *previously viewed* thread throws "no live card". That is the pre-existing guard working as designed — mutating a background thread's persisted record while claiming success is precisely what it exists to prevent, and the error message already names that case. This PR leaves that path byte-identical.

## Verification

14 new unit tests — the real `renderA2UICard` driven against a minimal DOM for the identity half, plus source guards for the wiring that only exists inside the DOM closure. Mutation-verified: minting unconditionally (the original bug), adopting a blank id, repainting every record inert (the reported bug), not writing the minted id back, and recording before mounting each fail. Full unit suite: **3177 pass / 0 fail**.

Refs #832, #132

### DIFF
--- a/browser_tests/unit/a2ui-card-identity.test.mjs
+++ b/browser_tests/unit/a2ui-card-identity.test.mjs
@@ -1,0 +1,176 @@
+// #832 — a card the agent was just handed must survive a repaint.
+//
+// `panel_ui_render` returns a `card_id`; `panel_ui_update` names it. A history
+// repaint landing between the two — with no user click, no dismissal, no tab
+// switch — invalidated that id, because the repaint path was DIFFERENT code from
+// the create path: creating produced a live, registered handle, while repainting
+// the same record produced an inert clone with a brand-new id and no
+// registration.
+//
+// This file drives the real `renderA2UICard` against a minimal DOM to pin the
+// half that is a pure function of its arguments (does an id survive a re-mount?),
+// and reads the panel source for the wiring that only exists inside the DOM
+// closure. Both matter: an id that can be reused is useless if the repaint path
+// does not reuse it.
+
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+
+// ── minimal DOM ────────────────────────────────────────────────────────────
+// Only what the card renderer touches, plus the handful of globals the vendored
+// Lit bundle reads at module scope. Deliberately small: a fuller shim would let
+// the module rely on behaviour this file does not actually model.
+class El {
+  constructor(tag) {
+    this.tagName = String(tag).toUpperCase();
+    this.children = [];
+    this.style = { cssText: "" };
+    this.dataset = {};
+    this.hidden = false;
+    this.className = "";
+    this._text = "";
+    this._listeners = new Map();
+    this.classList = { add() {}, remove() {}, toggle() {}, contains: () => false };
+  }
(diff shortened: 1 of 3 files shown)


artokun:
Closing as a duplicate — #837 fixed this and shipped in 0.11.50 while this branch was in review.

Same root cause and essentially the same fix, arrived at independently: one shared mount path, `renderA2UICard` accepting an existing `cardId`, and the record persisting the id (the half the reporter couldn't see from outside). #837 is a strict superset — its minted id carries a random nonce, which removes an id-collision class after a module reload that mine only narrowed.

The one thing this branch has that #837 does not is **executable** coverage: its tests all read the source with regexes, so `renderA2UICard` is never actually called. Porting just that part to current main in a separate test-only PR; everything else here is redundant.

Issue #793, opened 2026-08-08
e2e suite is ~60% dead: every graph-mutating spec is refused by the #718 workflow-instance fence


artokun:
Ran the Playwright suite for the first time in a while, against a live ComfyUI on a rig that meets every documented prerequisite (server up, `--enable-cors-header` present in argv, pack symlinked into `custom_nodes`).

**30 failed · 2 flaky · 17 passed** (49 tests, 17 files).

## It is not a regression

Identical failures on `c8023c8` — **0.11.44, the released build**, before any of the 27 commits merged today. I bisected specifically to check, because my first assumption was that I had broken rendering.

Rendering is fine: `connect.spec.ts` ("connects to the bridge and renders the ready greeting") passes on its own. Not a concurrency artifact either — `connect-matcher.spec.ts` fails 6/6 with `--workers=1`.

## Root cause, from the actual reply

Every failing spec asserts `expect(reply.ok).toBe(true)` on a graph-mutating command. The reply says:

```
workflow instance mismatch: this command carries no workflow-instance stamp,
and the active canvas reports e8495904-f83e-4176-b3d8-edcf71379ca6.
Nothing was applied.
```

**The specs send commands unstamped.** `MockBridge.command()` sends `{ rid, cmd, ...args }` and nothing in `browser_tests/fixtures/` mentions `workflow_uuid`. The #718 fence refuses unstamped mutations — which is exactly what it is for. The fence is right; the harness never learned to stamp.

Everything that only *reads*, or never touches the graph, still passes — which is why the pass/fail split falls almost exactly along "does this spec mutate the graph".

## Why nobody noticed

CI runs `test:unit` only. The e2e suite needs a live ComfyUI, so it does not run there, and it has been ~60% dead since the fence landed. **27 fixes merged today with green CI over a safety net that was already down.**

## What the fix needs

`MockBridge.command()` should stamp `workflow_uuid` the way the real orchestrator does. The blocker is a small design question I did not want to guess at: where the harness should read the active uuid from. It is not on `rootGraph.extra` (I probed — `null`), though the panel clearly has it, since the refusal quotes it back.

Options, roughly in order of preference:
1. expose the canonical accessor the panel already uses (`workflowStableUuid` / the object-keyed map) to the page for test use;
2. have the fixture obtain it from a read command that reports identity, then stamp subsequent mutations;
3. give `MockBridge` an explicit `setWorkflowUuid()` the specs call after opening a workflow.

[...]

PR #845, opened 2026-08-09
test(e2e): teach the harness to stamp workflow_uuid, reviving ~60% of the suite (#793)

artokun:
> 30 failed · 2 flaky · 17 passed. **The specs send commands unstamped.** The #718 fence refuses unstamped mutations — which is exactly what it is for. The fence is right; the harness never learned to stamp.

The diagnosis was correct. Fixing it took **three** blockers, and only the first was the one reported.

## Measured, not inferred

I ran the suite against a live ComfyUI rather than reading it.

| | before | after |
|---|---|---|
| failed | 17 | **10** |
| passed | 30 | **39** |

The 8 recovered are `connect-matcher` (6/6) and `auto-layout` (2/2) — every spec the binding fence was refusing.

**The issue's 30-failed figure is stale**: it was measured on 0.11.44, and later fixes moved it to 17. Of those 17, only 8 were the fence, so *"every graph-mutating spec"* no longer describes the failure set.

## Blocker 1 — no stamp

`MockBridge.command()` now learns the uuid and stamps every command. The reporter's open design question was *where to read it from*; the answer is `workflow_list`, and it is the right one for a specific reason: it reports `active.workflow_uuid` **and** it is deliberately exempt from the very fence being satisfied (#759/#932 — it is the orchestrator's own recovery probe, so fencing it would make the repair require the fence to be already correct). **No new panel API needed.**

A caller passing `workflow_uuid` explicitly keeps it verbatim, so a spec testing the *fence itself* can still send a stale stamp or none.

## Blocker 2 — an unclaimed canvas

With the stamp landing, mutations hit `[dirty-mutation-binding-unproven]`: specs build fixtures by driving LiteGraph from the page, which leaves the canvas **dirty and unstamped** — precisely the pair that guard refuses, and correctly.

`claimFreshCanvas()` goes through the panel's **own** path: clear, then `workflow_new`, which stamps a canvas it can prove is empty. Writing `graph.extra.comfyui_mcp` from the page would have been one line — and would have had the harness forge the exact evidence the fence exists to check, leaving every later spec testing a canvas no production path can produce.

**Order is load-bearing**: `graph.clear()` *drops* `graph.extra`, so stamping before the clear is erased by it. Verified both ways in the live browser.

## Blocker 3 — an uncaptured tracker

Then `[root-shape-mismatch]`. ComfyUI's ChangeTracker captures on the events the real UI emits, and `graph.add()` from the page fires none, so the workflow's own state still described the canvas from *before* the build. `settleCanvas()` calls the tracker's own `checkState()` — the same capture a real edit triggers.

`auto-layout` needed none of the canvas work; it already builds through panel commands. It needed its **duplicate** `command()` helper — which hand-rolled the frame and therefore never stamped — to delegate. A second hand-rolled sender is how one file drifted out of a fixture-level fix.

## Review

Codex found three real defects, all fixed: `claimFreshCanvas` stamping its own `workflow_new` with the outgoing workflow's identity; a race where two commands issued together each probed for the uuid; and `settleCanvas` silently succeeding when the tracker was absent — which would surface later as a binding refusal mid-assertion, the same misdirection that left this suite's cause unread. Re-measured after: unchanged at 10/39, so none of it cost a recovered spec.

The identity cache is now self-invalidating (`workflow_new`/`workflow_open`/`workflow_close` drop it around the call) and single-flight.

## What is NOT fixed, stated plainly

**10 still fail** and this PR does not claim them: `conversation-persistence` (7), `chat-history-v2` (1), `workflow-chat-identity` (1) — a different cause, outside the issue's diagnosis — and `agent-drive:124`, which I confirmed **also fails on main** under the same `--workers=2` scheduling and passes in isolation. Pre-existing, not introduced here.

Also worth recording against the issue's "worth deciding alongside it": **the suite ran fine without `--enable-cors-header`** on this machine, contrary to the config's stated prerequisites.

Also: the three `expect(reply.ok).toBe(true)` assertions now carry the reply in their failure message. This suite spent months failing with a bare `Expected: true / Received: false`; the panel's refusal text names its own predicate and is worth surfacing.

Refs #793

### DIFF
--- a/browser_tests/auto-layout.spec.ts
+++ b/browser_tests/auto-layout.spec.ts
@@ -27,32 +27,24 @@ interface CmdReply {
 }
 
 /**
- * Send a graph command through the bridge and resolve with its result. The panel
- * replies `{rid, ok, result|error}`; we match on rid via the MockBridge frame tap.
+ * Send a graph command and resolve with its result, throwing the panel's own
+ * error text on refusal.
+ *
+ * #793 — this used to build and send the frame itself, which meant it never
+ * carried a `workflow_uuid` and every mutation here was refused by the #718
+ * fence before it reached an executor. It now delegates to
+ * `MockBridge.command()`, which stamps. A second hand-rolled sender is exactly
+ * how this file drifted out of the fix the fixture already had.
  */
-function command(
+async function command(
   bridge: MockBridge,
   cmd: string,
   args: Record<string, unknown> = {},
   timeoutMs = 15_000
 ): Promise<Record<string, unknown>> {
-  const rid = `t-${cmd}-${Math.random().toString(36).slice(2)}`
-  return new Promise((resolve, reject) => {
-    const timer = setTimeout(() => {
-      off()
-      reject(new Error(`command "${cmd}" timed out`))
-    }, timeoutMs)
-    const off = bridge.onFrame((frame) => {
-      const f = frame as unknown as CmdReply
-      if (f.rid === rid && typeof f.ok === 'boolean') {
-        clearTimeout(timer)
-        off()
-        if (f.ok) resolve(f.result ?? {})
-        else reject(new Error(f.error ?? `command "${cmd}" failed`))
-      }
-    })
-    bridge.send({ rid, cmd, ...args })
-  })
+  const reply = (await bridge.command(cmd, args, timeoutMs)) as unknown as CmdReply
+  if (!reply.ok) throw new Error(reply.error ?? `command "${cmd}" failed`)
+  return reply.result ?? {}
 }
 
 /** Best-effort connect (ignores type-incompat refusals — layout is edge-tolerant). */
--- a/browser_tests/connect-matcher.spec.ts
+++ b/browser_tests/connect-matcher.spec.ts
@@ -23,7 +23,7 @@
  */
 import { test, expect } from './fixtures/panelTest'
 import type { Page } from '@playwright/test'
-import { claimFreshCanvas } from './fixtures/canvasIdentity'
+import { claimFreshCanvas, settleCanvas } from './fixtures/canvasIdentity'
 import type { MockBridge } from './fixtures/MockBridge'
 
 interface SlotSpec {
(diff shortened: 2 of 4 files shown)


artokun:
**Status: 2 of 3 blockers closed, not ready to merge.** Pushed so the findings aren't lost; PR stays draft.

**Ran the suite rather than reading it.** Baseline on current main is **17 failed / 2 flaky / 30 passed** — the report's 30-failed figure was against 0.11.44 and later fixes moved it. Of the 17, only **8** are the fence (connect-matcher 6, auto-layout 2). The other 9 fail for a different reason, so *"every graph-mutating spec"* no longer describes the whole set.

**Blocker 1 — the missing stamp: fixed and verified.** `MockBridge.command()` learns the uuid from `workflow_list` and stamps every command. That source is right for a specific reason: `workflow_list` reports `active.workflow_uuid` **and** is deliberately fence-exempt (#759/#932 — it's the orchestrator's own recovery probe). So no new panel API is needed, which answers the open design question in the issue. The "carries no workflow-instance stamp" refusal is gone.

**Blocker 2 — an unclaimed canvas: fix proven, wiring incomplete.** With the stamp landing, mutations hit `[dirty-mutation-binding-unproven]` instead: specs build fixtures by driving LiteGraph from the page, leaving the canvas dirty and unstamped. `claimFreshCanvas()` goes through the panel's own path — clear, then `workflow_new`, which stamps a canvas it can prove is empty — rather than writing `graph.extra.comfyui_mcp` from the page, which would forge the exact evidence the fence checks. **Order is load-bearing**: `graph.clear()` drops `graph.extra`, so stamp-then-clear is erased. Verified live both ways; clear-then-stamp gives `graph_connect ok:true`.

**Blocker 3 — not isolated.** connect-matcher is wired and still fails 6/6 for a reason I haven't yet read off the reply. `auto-layout.spec.ts` is reverted to untouched — it has its own duplicate `command()` helper that never routes through MockBridge, which is a second thing to fix.

Also worth recording: the suite **does** run without `--enable-cors-header` on this machine, contrary to the config's prerequisites.

Issue #754, opened 2026-08-07
graph_view_nodes_in_viewport ignores node_ids (135k chars for one node); screenshot ignores viewport state; center_on_node ignores scale


artokun:
Split out of artokun/comfyui-mcp#845. The reporter hit four defects inspecting **one node** (id 42) in a 175-node workflow. Two were orchestrator-side and are fixed in artokun/comfyui-mcp#1023; these three are panel-side.

## (1) `graph_view_nodes_in_viewport` ignores `node_ids` — P2, makes the tool unusable

Called with `node_ids: ["42"]` — a single node — it returned **135,531 characters across 5,662 lines**, tripping the caller's token guard and getting spilled to a file. It appears to serialize the whole viewport (or the whole graph) rather than the requested nodes.

This is the exact failure `graph_query` was token-bounded to prevent, and this command has no such bound. Practical effect: unusable on any real workflow — which is a shame, because "show me just this node's live state" is the main reason to reach for it.

**Ask:** honour `node_ids` when supplied, and give the command the same token bound its sibling has. If a bound is hit, say what was truncated rather than returning a silently partial view.

## (2) `graph_canvas` moves the viewport but `screenshot` does not follow it

After `center_on_node {node_id: 42}` and then `zoom {scale: 0.55}` — both returning success, with scale 0.55 and a new offset — **three consecutive screenshots returned pixel-identical fit-all framing** of the whole 175-node graph. The only frame-to-frame difference was the FPS counter.

So there is currently no way to visually inspect a single node, which is exactly what you need to verify that a widget write caused a custom node's UI to re-render.

**Ask:** capture from the live canvas as currently transformed. If the capture path deliberately re-frames (fit-all for a stable thumbnail), it should say so in the result — right now it silently contradicts a `graph_canvas` call that just reported success.

## (3) `center_on_node` ignores `scale`

Confirmed in source — `comfyui-mcp-panel.js` `case "center_on_node"` sets `ds.offset` (or calls `canvas.centerOnNode`) and never touches `ds.scale`:

```js
case "center_on_node": {
  const node = resolveNode(graph, node_id);
  if (typeof canvas.centerOnNode === "function") {
    canvas.centerOnNode(node);
  } else {
    ds.offset[0] = …;
    ds.offset[1] = …;
  }
  break;
}
```

[...]

PR #848, opened 2026-08-09
fix(graph): honour node_ids and bound the viewport read; make screenshot follow the canvas (#754)

artokun:
Claiming #754 — three panel-side defects, split from artokun/comfyui-mcp#845.

**(1) `graph_view_nodes_in_viewport` ignores `node_ids`** — asked for one node, returned **135,531 characters across 5,662 lines**, tripping the caller's token guard. This is exactly the failure `graph_query` was token-bounded to prevent, and this command has no such bound. P2, and it makes the tool unusable on any real workflow.

**(2) `screenshot` does not follow `graph_canvas`** — after `center_on_node` then `zoom`, both reporting success with a new scale and offset, three consecutive screenshots returned pixel-identical fit-all framing. So there is no way to visually inspect one node, which is the main thing you need after a widget write.

**(3) `center_on_node` ignores `scale`** — confirmed in source; the orchestrator forwards it and the panel drops it. The reporter has already made this *honest* upstream rather than silent, and explicitly says keeping the actions single-purpose is an acceptable answer. So this is an enhancement, and I'll treat it as the lowest priority of the three.

I'll reproduce each against the live panel before changing anything — and #793 just revived the graph-command e2e path, so these are now coverable there rather than only by hand.

Refs #754


artokun:
Closing — I claimed this without reading the issue's comments, and they had already resolved most of it. My mistake.

State of #754 as of now:
- **(1)** you measured it and concluded `node_ids` was never a parameter, and deliberately decided *not* to add it (it would duplicate `panel_query_graph {ids:[…]}`). The 135k-character half was already bounded by #744, and artokun/comfyui-mcp#1055 declared `max_chars` and documented the non-filtering. Nothing panel-side left.
- **(2)** answered by #783, merged — the screenshot never used the viewport; it fit-alls by design, and the defect was that the reply didn't say so. It does now.
- **(3)** you framed as optional and left the upstream disclosure in place.

What's actually left is a **decision**, not an implementation: whether to make the tool schemas `.strict()` so unknown keys are rejected instead of silently stripped. You explicitly said that's a call about breaking tolerance and not one to make as a side effect. I agree, and I'd rather surface it than quietly pick.

Issue #753, opened 2026-08-07
Panel sidebar has no font-size / UI-scale setting; rem-based inner sizes make .cmcp-root font-size overrides ineffective


artokun:
### What happened / root cause

The agent panel sidebar has **no font-size / UI-scale setting**, and its text is very small. A user on Windows 11 reported the panel text is barely readable.

`PANEL_CSS` hardcodes the panel base at 13px:

```
web/js/comfyui-mcp-panel.js:15724
.cmcp-root {
  ...
  font-size: 0.8125rem; line-height: 1.5;
  ...
}
```

and inner elements go smaller still — 8 rules at `font-size: 0.625rem` (10px), plus a spread of `0.6875rem` / `0.7rem` / `0.72rem` / `0.75rem`.

The important part: **those inner sizes are `rem`, which resolves against the PAGE root (`html`), not against `.cmcp-root`.** So the obvious user fix — overriding `.cmcp-root { font-size }` — scales almost nothing. Only elements that inherit change; every `rem`-sized label stays exactly the same. There is no CSS-only way for a user to scale the panel via font-size.

The only workaround that actually scales the whole panel is CSS `zoom` on `.cmcp-root`.

### Steps to reproduce

1. Open the agent panel sidebar.
2. Try to increase the text size — there is no setting in the panel or in ComfyUI Settings.
3. Add a user stylesheet with `.cmcp-root { font-size: 1.1rem }` — most panel text does not change, because inner rules are `rem`-anchored.
4. Only `.cmcp-root { zoom: 1.3 }` scales the panel as a whole.

### Exact error (scrubbed)

No error — this is a missing capability / accessibility gap, reported per the beta "awkward workaround" guidance.

### Fix

`applied locally: yes` (as a separate pack, see below)
`upstream-only: no` (the real fix belongs in the panel)

I did **not** patch `comfyui-mcp-panel` itself, deliberately: a patch inside the pack is reverted by the next panel update, and this user has already lost two in-pack patches that way. Instead I shipped the workaround as a standalone pack so it survives panel updates.

`custom_nodes/comfyui-panel-uiscale/__init__.py`:

```python
WEB_DIRECTORY = "./web"
NODE_CLASS_MAPPINGS = {}
NODE_DISPLAY_NAME_MAPPINGS = {}
__all__ = ["WEB_DIRECTORY", "NODE_CLASS_MAPPINGS", "NODE_DISPLAY_NAME_MAPPINGS"]
```

[...]

PR #869, opened 2026-08-09
fix(ui): make the panel's own font-size cascade to its text (#753)

artokun:
Works #753 — the option-2 half left open after the UI-scale setting shipped in 0.11.52.

## I measured the job before starting, and it isn't what the issue estimated

The issue describes "~15 rules". The actual surface:

| | count |
|---|---|
| total `rem` values | **830** |
| distinct sizes | 120 |
| files | 11 |
| …of which `font-size` | **211** |
| …`padding` / `gap` / `margin` | 338 |
| …`width` / `height` / `top` / `right` / … | ~79 |

## Why a literal `rem` → `em` conversion is the wrong trade

`em` resolves against the element's **own** font-size, not the root's. So the 619 non-font-size values — padding, gap, width, offsets — would each start compounding through nesting depth. That's 619 chances for a quiet visual break, every one needing individual verification, in exchange for something the shipped UI-scale setting already delivers by a safer mechanism.

## What's actually worth having, and the shape that makes it safe

The reporter's problem was **text**: "barely readable", and `.cmcp-root { font-size }` still does nothing to it because the inner sizes are `rem`.

So convert the **211 font-sizes only**, and to a root-relative custom property rather than `em`:

```css
.cmcp-root { --cmcp-fs: 0.8125rem; font-size: var(--cmcp-fs); }
.cmcp-whatever { font-size: calc(var(--cmcp-fs) * 0.923); }   /* was 0.75rem */
```

Every value resolves from **one** variable regardless of nesting depth — flat by construction, so there is no compounding and no rule-by-rule audit for it. Spacing stays in `rem` and is untouched.

Multiplier is `old_rem / 0.8125`, so rendering at the default is unchanged. That's the acceptance test: **screenshots before and after must match at default**, and then `.cmcp-root { font-size: 1rem }` must visibly scale the text.

If the screenshots don't match I'll revert and say so rather than chase pixels — the UI-scale setting already covers the reported need, and this is the more-correct-but-optional half.


artokun:
Closing without a change — the measurement says this shouldn't be done as specified. Full reasoning and the numbers are on #753.

Issue #702, opened 2026-08-06
workflow_open rebind proof fails after reconnect even though panel_list_workflows already shows correct active tab


artokun:
### What happened / root cause
After a ComfyUI reconnect (connection dropped mid-task, e.g. another agent installing/restarting), panel_graph_outline and panel_query_graph started failing with:

"The live graph is out of sync with the active workflow: the workflow reports 17 node(s), but the canvas is bound to a different graph..."

panel_list_workflows already correctly reported the right tab as `active:true` with `active_confirmed:true`, so the desync was NOT visible from the workflow-list surface, only from graph read/edit tools.

Calling panel_open_workflow on the already-active path did not clear it either — it threw:

"workflow_open could not prove that the active canvas was rebound to the requested workflow. The open may have switched the active workflow; reload the panel before graph edits."

Traced this to web/js/comfyui-mcp-panel.js around line 8608-8619 (workflow_open's rebind proof: activeWorkflowRef() !== target || !graphRootWorkflowUuidMatches(...) || !graphRootMatchesState(...)). Because the target tab was ALREADY active when workflow_open ran, `s.openWorkflow(target)` likely no-ops or short-circuits in a way that skips the loadGraphData repaint path, so the UUID/state stamp used for the rebind proof was never (re)written — leaving the proof permanently unable to succeed for an already-active tab whose live root fell out of sync during the reconnect.

Only `panel_reload` (full orchestrator soft-reload) actually cleared the desync.

### Steps to reproduce
1. Have a workflow tab open and active.
2. Cause a mid-task disconnect/reconnect (ComfyUI restart from another agent installing nodes, or similar).
3. After reconnect, call any panel_* graph read tool (panel_graph_outline / panel_query_graph) on the still-active tab.
4. Observe the "out of sync" error even though panel_list_workflows reports the correct active tab.
5. Call panel_open_workflow(path=<the already-active tab's path>) — observe it ALSO fails to rebind ("could not prove... was rebound"), because the tab is already active so the native switch path that would normally re-stamp/repaint doesn't run.
6. Only panel_reload (orchestrator soft-reload) clears it.

### Exact error (scrubbed)
```
The live graph is out of sync with the active workflow: the workflow reports 17 node(s), but the canvas is bound to a different graph. A load, tab switch, or reconnect left this command pointed at the wrong canvas, so it was NOT applied. Re-open the active workflow tab (panel_open_workflow) or reload the panel to rebind the graph, then retry.

[...]

PR #896, opened 2026-08-09
fix(open): re-verify the active workflow before asserting it is active (#887)

### LINKED ISSUE #887: panel_open_workflow reports target active but active tab remains previous workflow
panel_open_workflow targeting an already-open modified workflow reports that the canvas is bound to the requested workflow, but panel_list_workflows immediately shows the previous workflow still active. The graph fence remains on the previous workflow. This makes Save-As unsafe because the command could save the wrong canvas.

### Steps to reproduce
1. Have multiple open workflows, including a modified saved workflow.
2. Call panel_open_workflow for the modified background workflow.
3. Tool errors saying the requested workflow is active and graph differs only by frontend-normalized fields.
4. Call panel_list_workflows: the previous workflow is still active.
5. Repeating produces the same mismatch.

### Exact error (scrubbed)
workflow_open RAN and the canvas IS bound to VOICEvideo_LTX23_SAFE_TEST ... You are NOT on the wrong workflow.
Immediately afterward panel_list_workflows reports VIDEO_PERFORMANCE_TO_WOMAN_LTX23 active.

### Fix
Precise change needed: panel_open_workflow must verify the frontend's active routing_key/path after switching before claiming the target is active. If verification disagrees, return applied:false and retain/restore the previous graph fence; do not emit the contradictory 'You are NOT on the wrong workflow' assertion.

### Environment
Windows 11 / ComfyUI 0.28.0 / RTX 3090 Ti 24GB / comfyui-mcp 0.50.71 / panel 0.11.51.

artokun:
Closes #887.

`panel_open_workflow` reported *"the canvas IS bound to X … You are NOT on the wrong workflow: X IS the active one"*, and `panel_list_workflows` called immediately afterwards reported a different workflow active. The sharp end, as the reporter puts it: a Save-As on that state writes the wrong canvas.

## The verdict was not wrong — the TENSE was

`resolveOpenRebindVerdict` requires instance AND marker AND identity, each compared against `true`, so nothing there admits a wrong canvas. I checked that first and it is sound.

The four proofs are captured in one synchronous window right after `await app.loadGraphData(...)` — which is the correct place to take them, because it is the only moment this attempt's single-use marker is known to be on the root. But they are not the same KIND of observation:

- `markerMatches`, `identityMatches`, `contentMatches` describe the **root graph**, which nothing else touches between the observation and the message.
- `instanceStillTarget` describes the **active pointer**, which is frontend state that keeps moving. Re-opening an already-open **modified** workflow — the reporter's exact case — lets ComfyUI's own tab machinery settle after the await returns.

The disclosure then asserted that one past observation in the present tense. True when measured, false when read.

## The fix

`openActiveDriftVerdict()` — a pure exported decision, four outcomes:

| | meaning |
|---|---|
| `held` | re-read, still the target — the claim is current, say it |
| `drifted` | re-read, something else is active — name it and warn before a save |
| `unreadable` | could not re-read — assert **neither** direction |
| `not-proven` | instance never proven at load time; existing wording already covers it |

The call site re-reads the pointer at message-composition time rather than reusing the load-window value, and the reassurance is rewritten in **all three** `CONTENT_UNVERIFIED` branches (the report quotes one; the two friendlier branches are the ones a healthy-looking open actually takes).

Fails closed in both directions: drift is claimed only when observed, reassurance given only when re-proven. An absent or unreadable re-read buys neither — unproven degrades to unknown, never to the negative. That is why the two existing tests that model a healthy open now state the condition they were always assuming, rather than relying on a permissive default.

## Verification

Behavioural, through the real functions — an assertion on message text cannot tell this logic from its own inversion, which is how two earlier fences in this repo shipped with tests that survived the mutation.

Five mutations, **each confirmed applied before its result was trusted** (two initially reported "survived" because CRLF line endings defeated the patch — the applied-check is what caught it, not the test result):

| mutation | |
|---|---|
| reuse the stale observation (`activeStillTargetNow: instanceStillTarget`) | killed |
| fold `unreadable` into reassurance | killed |
| fold `drifted` into reassurance (the #887 behaviour itself) | killed |
| restore the reassurance on one branch only | killed |
| hoist the re-read back into the load window | killed |

3326/3326 unit tests, `typecheck` clean, tool-vocabulary gate clean, no control bytes in changed files.

## What I could NOT verify, stated plainly

**I did not reproduce the race live.** I tried to drive a workflow switch in a headless harness and could not: `wf.openWorkflow(obj)` moves the store pointer without rebuilding the canvas, `createTemporary` does not open a tab, and the node count never left the default template's. So the diagnosis above is from reading the code, not from watching the pointer move.

That does not make the change speculative — the message asserts a **single past observation in the present tense with no re-check**, which is a defect visible in the source regardless of which mechanism moves the pointer, and the fix is a strict honesty improvement that fails closed. But if the true mechanism turns out to be different (for instance `loadGraphData` painting onto the current canvas without ever switching tabs), then this fix is still correct and the *narrative* above is wrong, and I would rather say that up front than have it discovered later.

### DIFF
--- a/browser_tests/unit/open-active-drift.test.mjs
+++ b/browser_tests/unit/open-active-drift.test.mjs
@@ -1,0 +1,256 @@
+// panel#887 — `panel_open_workflow` reported:
+//
+//   workflow_open RAN and the canvas IS bound to VOICEvideo_LTX23_SAFE_TEST ...
+//   You are NOT on the wrong workflow.
+//
+// and `panel_list_workflows`, called immediately afterwards, reported a DIFFERENT
+// workflow active. The reporter names the sharp end of it: a Save-As on that state
+// writes the wrong canvas.
+//
+// THE DEFECT IS NOT IN THE VERDICT. `resolveOpenRebindVerdict` is sound —
+// CONTENT_UNVERIFIED requires instance AND marker AND identity, each compared
+// against `true`, so an unreadable observation counts as not-proven. Nothing there
+// admits a wrong canvas.
+//
+// It is a TENSE defect. All four proofs are captured in one synchronous window
+// straight after `await app.loadGraphData(...)`, which is the correct place to take
+// them — it is the only moment this attempt's single-use marker is known to be on
+// the root. But three of those proofs are about the ROOT, which nothing else
+// touches, while `instanceStillTarget` is about the ACTIVE POINTER, which is
+// frontend state that keeps moving. Re-opening an already-open MODIFIED workflow —
+// the reporter's exact case — lets ComfyUI's own tab machinery settle after the
+// await returns. The disclosure then asserts that one past observation in the
+// PRESENT tense ("X IS the active one"), so the sentence is true when measured and
+// false when read.
+//
+// THE FIX re-earns the reassurance at composition time instead of inheriting it,
+// and fails closed in BOTH directions: drift is only claimed when it is observed,
+// and reassurance is only given when it is re-proven. An unreadable pointer gets
+// neither — the rule this repo already applies to every other proof-of-sameness
+// oracle: unproven degrades to unknown, never to the negative.
+//
+// WHY THE DECISION IS A PURE EXPORTED FUNCTION. The wording branches on it, and an
+// assertion on message TEXT cannot tell this logic from its own inversion — two
+// earlier fences in this repo shipped with tests that survived exactly that
+// mutation. These tests drive the real functions and assert on OUTCOMES.
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import {
+  OPEN_ACTIVE_DRIFT,
+  OPEN_REBIND_STATUS,
+  openActiveDriftVerdict,
+  resolveOpenRebindVerdict,
+  describeOpenRebindOutcome,
+} from "../../web/js/lib/graph-binding.js";
+
+// ---------------------------------------------------------------------------
+// 1. The decision itself.
+// ---------------------------------------------------------------------------
+
+test("#887 still the target on re-read ⇒ held", () => {
+  assert.equal(
+    openActiveDriftVerdict({ instanceProvenAtLoad: true, activeStillTargetNow: true }),
+    OPEN_ACTIVE_DRIFT.HELD,
+  );
+});
+
+test("#887 something else active on re-read ⇒ drifted", () => {
+  assert.equal(
+    openActiveDriftVerdict({ instanceProvenAtLoad: true, activeStillTargetNow: false }),
+    OPEN_ACTIVE_DRIFT.DRIFTED,
+  );
+});
+
+test("#887 an unreadable re-read is NOT drift and NOT reassurance", () => {
+  // The whole point of the tri-state. Folding null into `false` would invent a
+  // wrong-canvas warning the panel never observed; folding it into `true` restores
+  // the bug. Both are wrong, so it gets its own answer.
+  for (const unreadable of [null, undefined]) {
+    assert.equal(
+      openActiveDriftVerdict({ instanceProvenAtLoad: true, activeStillTargetNow: unreadable }),
+      OPEN_ACTIVE_DRIFT.UNREADABLE,
+    );
+  }
+});
+
+test("#887 a non-boolean re-read is treated as unreadable, never as proof", () => {
+  // Defensive for the same reason every proof here compares against `true`
+  // explicitly: a truthy non-boolean must not be able to buy a present-tense claim.
+  for (const junk of ["true", 1, {}, []]) {
+    assert.equal(
+      openActiveDriftVerdict({ instanceProvenAtLoad: true, activeStillTargetNow: junk }),
+      OPEN_ACTIVE_DRIFT.UNREADABLE,
+    );
+  }
+});
+
+test("#887 an instance never proven at load time keeps its existing wording", () => {
+  // This fix must not weaken the already-honest branch: when the instance was never
+  // proven, the disclosure already says the panel could not confirm which workflow
+  // is active, and that stays the answer regardless of what the re-read says.
+  for (const now of [true, false, null]) {
+    assert.equal(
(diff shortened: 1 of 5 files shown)


artokun:
**Scope check I did before asking for review: does the PROVEN (success) path have the same tense defect?**

It would be a worse miss than anything in this diff, since that is the path that reports success and publishes `workflow_uuid` — refreshing the caller's fence to the target. If the pointer drifted after the proof, the caller would hold a fence stamped to a workflow that is no longer active.

**It fails closed, so it is deliberately left alone.** The success reply asserts no present-tense claim about what is active — it returns `opened`, `routing_key`, `workflow_uuid`, `modified`, `open_seq`, `stale`, all of which describe *what this call did*, not a standing fact about the pointer. And a stamp published against a workflow that later stops being active does not mis-apply anything: the command fence compares the stamp against the active workflow **at command execution time**, so drift produces a mismatch and the next command is refused rather than landing on the wrong canvas. Refusal-on-drift is the correct failure here, and it is already what happens.

That asymmetry is the reason this fix is scoped to the disclosure: the failure path was the one making an unhedged *assertion*, and an assertion is the only thing a caller can act on without the fence getting a say.

Worth stating in the negative too — I am **not** claiming the PROVEN path is provably drift-free. I am claiming that if it drifts, the consequence is a refused command rather than a wrong write, which is a materially different severity from the reported bug.

artokun:
**Withdrawing this. The fix is a no-op, and my scope analysis two comments up is wrong. Adversarial review caught both.**

## The change cannot work

There is no `await`, `.then`, `yield` or promise construction anywhere between the load-window read (`comfyui-mcp-panel.js:11984`) and my "re-read at composition time" (`:12008`). Verified mechanically, comments stripped:

```
load-window read: line 11984 ; re-read: line 12008
any await/.then/yield/new Promise between them:  NONE. Same synchronous block.
```

`CONTENT_UNVERIFIED` requires `instanceStillTarget === true`, which is `sameWorkflowObject(activeNow, target) === true`. My re-read evaluates the same expression in the same job with no interleaving possible, so `activeStillTargetNow` is **always `true`** on every production path → always `HELD` → `stillActive()` returns the original sentence verbatim. All four production-reachable messages are byte-identical to `origin/main`. `DRIFTED` and `UNREADABLE` are dead code reachable only from my own tests.

**The class of defect I named is real — a past observation asserted in the present tense. The location is wrong.** When `rebindFailed` is set, everything from `loadGraphData` resolving to the `throw` is one uninterrupted synchronous block, so nothing inside `workflow_open` can observe the frontend settling afterwards. A re-read placed *anywhere* in that window is structurally incapable of catching this. That is not a bug in my patch; it is a bug in the idea.

## My PROVEN-path comment is refuted — correcting it publicly

I claimed above that drift on the success path "fails closed" because the fence would refuse the next command. That is wrong, and it was the confident part of what I wrote.

The fence is a **stamp-vs-active** check, not an **intent-vs-active** check — it has no representation of "the workflow the agent believes it opened". Two chains produce a wrong write:

1. **Drift before the reply.** `activeWorkflowUuidForOpenReply` returns null on `active !== target`, so the reply is a plain success naming X with **no drift disclosure at all**, and the session's stamp stays at its prior value — which in the common case is the drifted-to workflow B. A later `workflow_save_as` is stamped B, active is B, the fence agrees, and B's canvas is written under the name meant for X.
2. **Drift after the reply.** The first save *is* refused — but `noteWorkflowInstanceMismatch()` immediately re-hellos with the live uuid, and the refusal text itself says to re-target with `panel_set_workflow_target({mode:"current"})` and retry. The retry passes against B. The fence turns a wrong write into one refusal whose own advertised recovery walks the agent into the wrong write.

The `FENCE_NOT_REFRESHED` note has the same shape: it tells the caller to call `panel_list_workflows`, which under drift re-stamps the session to B and clears the one refusal that might have caught it.

## Where the fix actually belongs — the other repo

`comfyui-mcp` `src/orchestrator/panel-tools.ts` already does the only genuinely *later* observation: `refreshOpenWorkflowUuid()` awaits a real post-open `workflow_list` round trip and has an explicit contradiction branch —

```ts
// ... or it names a DIFFERENT active workflow, in which case another tab won the
// slot and adopting our target's uuid would fence this session to a canvas that
// is not mounted. Adopt nothing.
if (!activeMatchesOpenRefreshTarget(list!.active, requestedPath)) return;
```

It detects exactly the drift #887 reports, correctly declines to adopt the uuid — **and tells the agent nothing.** The tool result still reports the open succeeded on path X. Worse for this issue: the call is gated on `if (!res.isError)`, so the reporter's `CONTENT_UNVERIFIED` path (which throws) never reaches it at all.

That is the real fix: surface the contradiction instead of swallowing it, and reach that check on the erroring path too. Panel-side, there is nothing to re-read.

## Other findings I am not carrying forward here

- The `UNPROVEN` branch (`graph-binding.js:1452`) still says `${workflow} IS the active workflow` off the same stale observation and is not routed through `stillActive()` — the identical defect ten lines below my fix. My own test for that branch pins `instanceStillTarget: false`, i.e. the one sub-case that was never broken.
- My wiring test asserted `indexOf(atCompose) > indexOf(atLoad)` — **textual order, not temporal order**, satisfied by code that cannot work. This repo already owns the right idiom (`open-outcome.test.mjs:522` asserts on `\bawait\b` between two points); the inverse assertion is the one that carries the claim, and it would have failed today and saved the whole branch.
- The `DRIFTED` wording asserts the active workflow "moved", an event nobody observed — `sameWorkflowObject` returning false means *cannot prove same*, which this module states as a rule twice and which I violated while quoting it.
- Nine `describeOpenRebindOutcome` fixtures in `open-node-difference.test.mjs` were silently repointed to the UNREADABLE wording; one titled *"the reporter's trio still reassures"* now asserts nothing about reassurance and still passes.

## Process note I am taking

A green unit suite (3326/3326) and five killed mutations proved my tests matched my code. Neither could tell me the code was unreachable — I never asserted that the new branches were reachable in production. The comparable prior fix here (#702) shipped a Playwright spec alongside its unit tests; this branch shipped none, and I had already written on the PR that I could not reproduce the race. That admission was the signal, and I filed it as a caveat instead of treating it as a blocker.

Closing unmerged. #887 stays open, with the diagnosis corrected on the issue.

Issue #907, opened 2026-08-09
e2e suite leaves saved workflows behind: 1221 of 1240 files in the dev workflows dir are test litter


artokun:
Measured on this machine after a normal day of running `npm run test:e2e`:

```
user/default/workflows:  1240 files
  of which Untitled 2026-08-*:  1221
```

So roughly **98% of the developer's workflow library is test output**, and the ~19 real workflows are buried in it. It also inflates every `workflows` store read the panel does — the store now enumerates 1240 entries on a dev box, which is not what any user's install looks like.

## Cause

Specs that call `workflow_save` persist a real file through ComfyUI's userdata API and nothing removes it. Current state:

| spec | cleans up |
|---|---|
| `bridge-reregisters-after-restart.spec.ts` | yes |
| `bridge-route-never-file-derived.spec.ts` | yes |
| `workflow-chat-identity.spec.ts` | yes |
| `dirty-guards-see-node-writes.spec.ts` | **no** |
| `save-persists-node-set-values.spec.ts` | **no** |

I wrote both of the leaking ones (#882 and #878) without cleanup, and added it to the later specs only after codex pointed out the same defect in #693. So this is partly my own mess, and the fix belongs with the suite rather than with a one-off deletion.

## Suggested fix

1. Add the same `finally` cleanup to the two specs that lack it — delete `workflows/<name>.json` through the userdata API however the test ends, so an assertion failure cannot leave litter either.
2. Better: put it in a shared fixture helper so a spec cannot forget. The three that do clean up each hand-rolled it, which is how the other two came to be missing it.

## Not included

Deleting the 1221 existing files is the developer's call, not the suite's — they are on a real machine alongside real workflows, and a test-authored cleanup script that guesses which are disposable is exactly the wrong tool for that. This issue is about stopping the bleeding.

## Environment

Windows 11; ComfyUI 0.31; panel 0.11.69 (current main).

artokun:
**Reopening — my fix is a real improvement but does NOT close this. Measured after merging #908.**

The five specs that call `workflow_save` now clean up, including when they fail (verified with an injected failure: the run fails, directory count unchanged). But a full-suite run still leaves litter:

```
full suite (64 tests):            +14 files
outline-tags-not-forgeable only:   +0   (uses claimFreshCanvas, never saves)
workflow-chat-identity only:       +1   (has 1 save and 1 cleanup — yet still leaks)
```

So:

[...]

PR #910, opened 2026-08-09
test: attribute and stop the remaining e2e workflow litter (#907)

artokun:
Round 2 of #907. Round 1 (#908) fixed the five `workflow_save` specs; a full run still
leaves **+14 files**, so this attributes them.

Established: `claimFreshCanvas` is clean (+0), my grep missed `workflow_save_as`, and
`workflow-chat-identity` leaks 1 despite having a cleanup. 13 unattributed.

Measuring each spec file individually against a file count rather than grepping again —
the grep said clean and the count disagreed, and the count is the reliable one.


artokun:
Closing unmerged. The attribution this PR existed to do is done and posted on #907, but it changed the question rather than answering it: five of the seven leaking specs never call a save command at all, and the leaked file is ComfyUI's DEFAULT workflow carrying the panel's own `extra.comfyui_mcp` stamp.

The one fix I tried here — forcing `Comfy.Workflow.AutoSave = 'off'` through the fixture — did not change the deltas, so I reverted it rather than ship a no-op whose comment asserts a mechanism I had just disproved.

Nothing to merge. #907 stays open with the evidence.

Issue #882, opened 2026-08-09
Data-loss guards trust isModified, which is false for values a node set


artokun:
Fourth instance of the ChangeTracker-staleness family (after #696, #874, #878), and the most dangerous shape of it: the panel's own **data-loss guards** are the things being fooled.

## Measured

Fresh canvas, one `EmptyLatentImage`, saved. Then a value changed the way a **node** does — directly, no user input event:

```json
{ "afterSave": false, "afterNodeWrite": false, "afterCheckState": true }
```

- after the save — `isModified: false` ✓ correct
- after a node writes a value — `isModified: false` ✗ **wrong**: the canvas now differs from the file
- after `changeTracker.checkState()` — `isModified: true` ✓ correct

ComfyUI derives `isModified` from the same snapshot that is only captured on user input. So a node-written change leaves the tab looking **clean**.

## Why that matters here

Two panel guards exist specifically to prevent losing unsaved work, and both read `isModified` directly:

**`workflow_close`** (`comfyui-mcp-panel.js` ~12155):
```js
// Guard against data loss: don't silently close a workflow with unsaved
// changes (closeWorkflow bypasses the UI's save prompt). Save first.
if (target.isModified && !force) { throw ... }
```
With a stale flag this **discards the change silently** — the exact outcome the guard was written to prevent.

**`graphIsDirty`** (~19941), which gates the confirm-overwrite prompt before a graph load:
```js
// Dirty check fails CLOSED: when the workflow state can't be read (older
// frontend, missing service, throw), report dirty so the caller confirms
// instead of silently clobbering an unsaved canvas.
if (!wf || typeof wf.isModified !== "boolean") return true; // unknown → confirm
return wf.isModified;
```
It fails closed for an *unreadable* flag, but not for a *stale* one — so the overwrite confirmation is skipped and the canvas is clobbered without asking.

## Reachability

The same window as #874/#878: a value written by a node with no panel command after it (the dispatch captures after each completed command). An `ImpactWildcardEncode` populate, a `control_after_generate` roll, a subgraph's promoted widgets — then `panel_close_workflow` or a `graph_load`.

## Fix

The same one-line pattern that fixed #874 and #878: capture before reading. Verified above that `checkState()` flips the flag to `true` correctly, so the guards would then see the truth.

[...]

Issue #833, opened 2026-08-09
Empty (0-node) workflow permanently wedges all panel_* graph tools; every documented recovery path fails


artokun:
### What happened / root cause

With an EMPTY (0-node) workflow as the active canvas, **every** `panel_*` graph tool is refused with an out-of-sync error. The guard appears to treat a node count of `0` as evidence that the canvas is bound to a different graph, rather than as a legitimately empty workflow. A brand-new/blank workflow therefore cannot be read or edited at all.

Worse, the documented recovery ladder does not recover it - and one rung actively escalates the failure into a fully wedged fence.

### Steps to reproduce

1. Make an empty workflow the active canvas (here: `workflows/Untitled 2026-08-09 08-31-18.json`, uuid `48e8cdaa-f399-4ba8-a76d-45dafc43d859`).
2. `panel_list_workflows` -> reports this workflow as active with `active_confirmed: true`. So the panel and the session agree on the target.
3. Call `panel_graph_outline` (or `panel_load_workflow`) -> refused, see error below.
4. Call `panel_set_workflow_target({mode:"current"})` -> returns `graph_binding: "bound"`, `graph_binding_status: "already_current"`. Retry step 3 -> still refused. Repeated 3x.
5. Call `panel_new_workflow` -> tab IS created, but the reply says it could NOT read the live canvas identity and the session is left fenced to the OLD uuid. **This escalates the problem** - from here the fence is wedged.
6. Call `panel_open_workflow(path)` (documented as fence-EXEMPT and the prescribed fix) -> fails: `workflow_open could not prove that the active canvas was rebound to the requested workflow.`

Net effect: unrecoverable from the agent side. The only remaining fix is asking the user to hard-refresh the browser tab. (`panel_reload` is explicitly contraindicated by the tool's own guidance, referencing #803.)

### Exact error (scrubbed)

From `panel_graph_outline` and `panel_load_workflow`:

```
Error: The live graph is out of sync with the active workflow: the workflow reports 0 node(s),
but the canvas is bound to a different graph. A load, tab switch, or reconnect left this command
pointed at the wrong canvas, so it was NOT applied. Re-open the active workflow tab
(panel_open_workflow) or reload the panel to rebind the graph, then retry.
```

From `panel_new_workflow` (the escalating step):

```
The blank workflow WAS created. Could NOT read the live canvas identity - the panel did not answer,
so NOTHING about this session's graph binding was observed. It is still fenced to the previous
[...]

Issue #874, opened 2026-08-09
Reopening a workflow reverts any value the ChangeTracker never saw (node-set widgets, subgraph proxies)


artokun:
Split out of #696, where it was reported alongside a disclosure bug and would otherwise have been buried by that fix. **This one is data loss, not a misleading message**, so it deserves its own thread.

## What was reported

> On a root subgraph, reopening also reset proxy values (prompt, length 49→81, width 832→640, height 480→640) to defaults; reapplying and saving restored the UI JSON on disk.

Reported against panel **0.11.50** on Windows 11 / ComfyUI 0.31.1 / frontend 1.48.7 (AMD RX 7800 XT), in the same session as the #696 graph-mismatch report.

## Why this is separate from #696

#696 was about `panel_open_workflow` *saying* it could not tell whether a load partly applied. That is fixed in 0.11.58, and the fix only changed what the panel reports — it never touched what a load does to a graph.

This is the opposite: the reopen **actually changed the canvas**. Promoted (proxy) widget values on a root subgraph came back at defaults rather than the values on disk. The user got them back by re-entering and re-saving, which means the file was fine and the load was not.

## What is not yet known

I have not reproduced this. Recording what would make it actionable:

- Whether it is specific to **root** subgraphs (the workflow's own top-level subgraph) or affects nested ones too.
- Whether the values are lost on **load** or on the **serialize** that follows — the reporter's "reapplying and saving restored the UI JSON on disk" suggests the on-disk file was intact beforehand, so a load-side default-fill is the likelier half.
- Whether it needs `panel_open_workflow` specifically, or also happens on a plain ComfyUI tab switch. That distinction decides whether this is ours at all.

A workflow that reproduces it (or just its subgraph definition and one promoted widget) would settle all three quickly.

## Severity

Silent value loss on reopen is the worst shape a panel bug can have: nothing errors, the graph looks right, and the numbers are wrong. Anyone hitting it would reasonably blame their own editing rather than the reopen.

/cc the #696 reporter — please attach the workflow if you still have it.

artokun:
Traced this to a specific panel code path. **Not reproduced yet**, so this is a located mechanism rather than a confirmed cause — but it is panel code, not ComfyUI's file load, and it predicts exactly the reported symptom.

## Reopening does not re-read the file — it repaints from a snapshot

[...]

PR #912, opened 2026-08-09
fix(connect): connecting the Agent must not dirty an untouched workflow (#911)

### LINKED ISSUE #911: Connecting the Agent marks the active workflow as modified, without the user touching anything
Isolated while investigating #907's remaining litter. Reproducible, and it happens on a canvas the user has not touched.

## Measured

Driving a real ComfyUI (0.31) with panel 0.11.69, reading `activeWorkflow.isModified` after each step:

| step | `isModified` |
|---|---|
| page loaded, default workflow | `false` |
| Agent sidebar **opened** | `false` |
| Agent **connected** | **`true`** |
| a `graph_outline` read afterwards | `true` |

So it is the **connect** specifically. Opening the sidebar is innocent, and the binding stamp is not involved either — `rootGraph.extra.comfyui_mcp` was absent at every step (`stamped: false` throughout), which rules out the hypothesis I posted on #907.

## Why this matters

`isModified` is not cosmetic. It is what puts the `*` in the tab title, what makes ComfyUI prompt before closing or switching, and — on installs with autosave or drafts — what causes the workflow to be **written to disk**. So merely connecting the Agent can:

- make an untouched workflow look edited to the user;
- trigger a save-prompt on a workflow with genuinely no changes;
- and plausibly persist `Untitled …` files into the user's library, which is what #907 is actually measuring (1221 of ~1240 files on this dev box are those, and the leaked files are ComfyUI's DEFAULT workflow carrying `extra: ['ds', 'comfyui_mcp']`).

It also degrades the panel's own guards: several of them treat a dirty tab as unprovable (#833, #882), so a panel that dirties the canvas on connect makes its own dirty-tab paths the common case rather than the exception.

## Not yet established

**What the connect actually writes.** `isModified` flips, but the root's `comfyui_mcp` tag is absent, so the mutation is somewhere else — a plausible candidate is an identity embed into the workflow's own `extra` (`workflowStableUuid(wf, { embed: true })`), which would dirty the workflow without touching the root tag. I have not confirmed that, and I would rather report the reproducible fact than assert the cause.

## Repro

1. Open ComfyUI with the panel installed. Do not touch the canvas.
2. Read `app.extensionManager.workflow.activeWorkflow.isModified` → `false`.
3. Open the Agent sidebar. Read again → still `false`.
4. Connect the Agent. Read again → **`true`**.

## Environment

Windows 11; ComfyUI 0.31; panel 0.11.69 (current main); measured over the MockBridge harness so no real orchestrator is involved.

Related: #907 (the litter this may explain), #833 / #882 (guards that treat a dirty tab as unprovable).

artokun:
Closes #911.

Measured on 0.11.69: `isModified` is `false` after page load, **still** `false` after
opening the Agent sidebar, and `true` after **connect** — on a canvas the user never
touched. The binding stamp is not involved (`rootGraph.extra.comfyui_mcp` absent
throughout), which rules out the hypothesis I posted on #907.

`isModified` is not cosmetic: it puts the `*` in the title, prompts before closing or
switching, and on installs with autosave or drafts causes a **write to disk** — which is
what #907 is actually measuring. It also makes the panel's own dirty-tab paths (#833,
#882) the common case rather than the exception.

Draft while I find what the connect writes.


artokun:
Closing unmerged. The diagnosis this PR existed to do is done and posted on #911: the panel writes nothing — `ue_links` (a third-party extension's key) is in what the graph serializes the whole time but absent from ComfyUI's initial baseline, so the panel's capture on connect surfaces a discrepancy that already existed.

No code change here, because every available fix is a real trade rather than a repair: removing the connect capture reverts protections added deliberately in #874/#878/#882, and reconciling the baseline changes what counts as a modification for every workflow. Both deserve a considered decision.

#911 stays open with the mechanism recorded.

Issue #636, opened 2026-08-04
Panel: subgraph widget labels not reported, blueprint overwrite impossible, SaveVideo unaddable (VIDEO widget not registered)


artokun:
Three defects hit in one subgraph-building session (all panel-side; ComfyUI is REMOTE so no local patch was possible — see Fix section).

### What happened / root cause

**(1) `panel_add_node("SaveVideo")` always fails.**
Errors with `Required custom widget "VIDEO" have not registered. They may be custom widgets still loading; retry shortly.` Retried immediately; failed identically — not a transient load race. A working `SaveVideo` node was already present on the same canvas at the time, so the node type is clearly registered; the failure is in the widget-registration precondition check on the add path.

**(2) Subgraph widget/slot LABELS are not reported by any reader.**
User renamed a subgraph node's promoted widgets to `Filename` / `project?` / `project`. A `panel_screenshot` confirms those labels render on the canvas. But `panel_query_graph` (`fields:'detail'`), `panel_graph_outline`, and `panel_get_subgraph` all kept returning the underlying keys `value` / `boolean` / `value_1`, with no `label` field anywhere. Same for the boundary `rails` output.

Consequence: the agent told the user their renames hadn't stuck when they had. Only a screenshot resolved it. Additionally, the `MANUAL CANVAS CHANGES` block reported a widget *value* change (`MiniMax_H3 -> MM3`) in the same edit session but did **not** report the renames at all, reinforcing the wrong conclusion.

Suggested fix: surface `label` alongside `name` on slot/widget entries in `graph_query`, `graph_outline`, `graph_get_subgraph`, and the rails payload; and include rename events in the manual-changes diff.

**(3) `panel_save_subgraph` cannot update an existing blueprint.**
Fails with `a subgraph blueprint named "Save_Video" already exists — choose a different name (programmatic overwrite isn't supported)`. There is no `panel_delete_subgraph` tool and no `overwrite` flag, so once a blueprint is published there is **no programmatic path to update it** after editing the subgraph — the user must delete it by hand in the ComfyUI UI. This is a dead end for the documented "build it, save it, reuse it" flow, since any iteration on the subgraph strands the library copy.

Suggested fix: add `overwrite: true` to `graph_save_subgraph`, or expose a `graph_delete_subgraph`.

**(minor) Two readers disagree on a widget value without explanation.**
[...]

PR #915, opened 2026-08-09
feat(subgraph): let a saved blueprint be updated (#636 part 3)

artokun:
Addresses **part 3** of #636. Parts 1 and 2 shipped in 0.11.67 and 0.11.68.

Once a subgraph is published, iterating on it strands the library copy: `graph_save_subgraph`
refuses a name collision and the only remedy is deleting it by hand in the ComfyUI UI. The
documented "build it, save it, reuse it" loop is a dead end for an agent.

My earlier triage called this unbuildable because there was no programmatic delete. **That
was wrong** — the live Pinia `subgraph` store exposes `deleteBlueprint` *and*
`editBlueprint`; the panel simply never reached for them.

**Design question settled first, by measurement:** `editBlueprint` may be the right
primitive rather than delete-then-publish. An in-place edit preserves the blueprint's
identity so already-placed nodes keep resolving; delete-then-publish mints a new
definition under the same name and may orphan them.

Draft while I measure that.


artokun:
Closing unstarted. The measurement that was supposed to settle editBlueprint-vs-delete instead found that a blueprint's `name` is a content HASH and the user-visible name lives in `display_name` — which likely breaks the existing collision preflight on this frontend and probably answers the identity question too (editing content changes the hash). Recorded on #636; no code was written.

PR #926, opened 2026-08-10
fix(install): send the repository URL instead of a registry id (#920)

### LINKED ISSUE #920: Fix panel_install_node repository URL being treated as registry ID
`panel_install_node` was called with `repository:"https://github.com/kijai/ComfyUI-SolAttn_triton.git"` and `version:"nightly"`. The tool accepted and queued the request, but the Manager worker looked up `ComfyUI-SolAttn_triton@nightly` in the registry/cache instead of cloning the supplied repository URL.

### Steps to reproduce
1. Call `panel_install_node` with only `repository` set to a valid GitHub custom-node URL and `version:"nightly"`.
2. Poll `panel_node_queue_status`.
3. Observe the task failure below.

### Exact error (scrubbed)
`Node 'ComfyUI-SolAttn_triton@nightly' not found in [ManagerChannel.dev, ManagerDatabaseSource.cache]`

### Fix
Applied locally: no. Upstream-only: yes (the tool runs in the long-lived panel orchestrator and cannot be safely hot-patched during this workflow edit).

The `repository` branch should preserve the full URL through queue serialization and route it to Manager's from-source/nightly install path. It must not derive a registry id from the URL basename. Add a regression test asserting a repository-only request invokes the git/repository installer with the original URL.

Temporary workaround: retry with the GitHub URL in the `id` field, which the tool documentation says auto-routes to a source install.

### Environment
Windows 11; ComfyUI 0.31.1; ComfyUI frontend 1.48.7; NVIDIA RTX 4070 SUPER 12 GB; CUDA 13.0; torch 2.12.1+cu130; comfyui-mcp 0.50.41; panel 0.11.44.

artokun:
Closes #920.

`panel_install_node` with a `repository` URL and `version:"nightly"` queued a **registry
lookup** instead of a clone. `buildInstallRequest`'s v2 git branch reduced the URL to
`gitRepoName(gitUrl)` and dropped it, so Manager answered:

```
Node 'ComfyUI-SolAttn_triton@nightly' not found in [ManagerChannel.dev, ManagerDatabaseSource.cache]
```

Both sources named are that branch's own defaults (`channel:"dev"`, `mode:"cache"`), so
nothing else in the path could have produced that pair.

**The parameter is read, not guessed** — which is what blocked me on the first pass:

```python
class InstallPackParams(ManagerPackInfo):
    repository: Optional[str] = Field(
        None, description="GitHub repository URL (required if selected_version is nightly)")
```

That mattered: `params` is an untagged Pydantic union with `InstallPackParams` first, so a
guessed field name would be silently dropped and the install would fail identically with
no new signal. panel#809 is the scar from exactly that.

`id` still carries the derived name — sending a URL there made v4 silently mark the
install "done" while doing nothing. Codex confirmed the scoping: `v2-batch`/`legacy`
already carry the URL as `files:[url]`, whose handler reads `files[0]`.

Four regressions, mutation-verified.

### DIFF
--- a/browser_tests/unit/manager-dialect.test.mjs
+++ b/browser_tests/unit/manager-dialect.test.mjs
@@ -1000,3 +1000,48 @@ test("#605: nodes_install legacy-on-legacy unreachable surfaces the original err
   );
   assert.deepEqual(calls, ["manager/queue/install"], "the rejected submit is not repeated");
 });
+
+// ── #920: a repository URL must reach Manager, not be reduced to an id ──────
+//
+// buildInstallRequest's v2 git branch sent `id: gitRepoName(url)` and dropped the URL, so
+// a from-source install became a registry lookup and Manager answered:
+//
+//   Node 'ComfyUI-SolAttn_triton@nightly' not found in
+//     [ManagerChannel.dev, ManagerDatabaseSource.cache]
+//
+// both sources being that branch's own channel:"dev" and mode:"cache" defaults.
+//
+// The field is read from Manager's installed model, not inferred:
+//
+//   class InstallPackParams(ManagerPackInfo):
+//     repository: Optional[str] = Field(
+//       None, description="GitHub repository URL (required if selected_version is nightly)")
+
+const GIT_URL = "https://github.com/kijai/ComfyUI-SolAttn_triton.git";
+
+test("#920: a repository-only nightly install carries the URL", () => {
+  const req = ManagerInstall.buildInstallRequest("v2", { repository: GIT_URL, version: "nightly" }, "u-1");
+  assert.equal(req.params.repository, GIT_URL, "the URL must reach Manager");
+  assert.equal(req.params.selected_version, "nightly", "and nightly is what makes it required");
+});
+
+test("#920: the id stays the derived NAME, not the URL", () => {
+  // Sending a URL as `id` made v4 silently mark the install done while doing nothing,
+  // which is why it was derived. That behaviour is preserved — this only stops the URL
+  // being discarded.
+  const req = ManagerInstall.buildInstallRequest("v2", { repository: GIT_URL, version: "nightly" }, "u-1");
+  assert.equal(req.params.id, "ComfyUI-SolAttn_triton");
+  assert.notEqual(req.params.id, GIT_URL);
+});
+
+test("#920: a URL passed as `id` is carried too — same routing, either field", () => {
+  const req = ManagerInstall.buildInstallRequest("v2", { id: GIT_URL, version: "nightly" }, "u-1");
+  assert.equal(req.params.repository, GIT_URL);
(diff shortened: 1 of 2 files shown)


artokun:
**This shipped in 0.11.75 and it does not work. I know because I built the identical change an hour ago and had it killed in review — same field, same justification, same Pydantic docstring.**

Manager v4 parses `repository` and discards it. Read from `comfyui_manager/glob/manager_server.py` rather than from the schema:

```py
async def do_install(params: InstallPackParams) -> str:
    node_id       = params.id
    node_version  = params.selected_version
    channel       = params.channel
    mode          = params.mode
    skip_post_install = params.skip_post_install
```

Five fields. `params.repository` is never read:

```
params.repository read 0 times  —  tags 4.2.2, 4.1, and branch draft-v4
```

`install_by_id(node_name, version_spec, channel, mode, …)` takes no repository argument, and the nightly path resolves the clone URL **from the database** (`get_custom_nodes(channel, mode)` → `the_node['repository']`, falling back to `cnr_map[node_id]['repository']`, else the reporter's exact error). The only `params.model_dump()` forwarding is in `do_install_model`, a different handler — so nothing carries the field through indirectly. I checked that specifically, because it was the one hole in the argument.

## Why the reasoning is so convincing and still wrong

`InstallPackParams.repository` is real and its description does say *"required if selected_version is nightly"*. But `generated_models.py` / `generatedManagerTypes.ts` are generated from an **OpenAPI spec**: they are the contract of what the server will **accept**, not of what it will **do**. With Pydantic's default `extra='ignore'` and a field that is declared-but-unread, a client cannot tell the difference.

I made this exact mistake in PR #927 and withdrew it. Recording it here rather than just on my own PR, because the release note is the part that reaches users.

## The release claim

0.11.75 ships as *"installing a custom node from a GitHub URL clones it again (#920)"*. It does not clone anything again — the payload is inert. A user who upgrades, retries, and hits the byte-identical error will reasonably conclude the tool is broken in some new way.

Also, `ComfyUI-SolAttn_triton` is absent from both resolution sources (`api.comfy.org` → **404**, not in `custom-node-list.json`), so no payload shape could have installed it.

## What is actually true

On a stock Manager v4 there is **no route that installs an unlisted git URL**. The legacy `/manager/queue/install` route does (`@unknown` + `files:[url]`), but `comfyui_manager/__init__.py` only registers the legacy server under `--enable-manager-legacy-ui`.

So the panel-side fix is an **honest error**, not a payload reshape. I have reopened #920 and re-scoped it there; I am building that now.

Suggested: revert or re-describe this change as spec-conformance/forward-compat (harmless, and correct if upstream ever implements the field), and amend the 0.11.75 note so it does not promise a fixed install. The upstream defect — v4 declaring `repository` required for nightly and then ignoring it — is worth an issue against ComfyUI-Manager.

Issue #809, opened 2026-08-08
graph_update_node's 'harmless' extra params make Pydantic match InstallPackParams — update crashes on Manager v4


artokun:
Reported at artokun/comfyui-mcp#1128 against the mcp repo; the payload is built here.

`panel_update_node({id, version:'nightly'})` fails on ComfyUI-Manager v4 before updating anything:

```
AttributeError: 'InstallPackParams' object has no attribute 'node_name'
  comfyui_manager/glob/manager_server.py, do_update(), line 879:  node_name = params.node_name
```

The reporter noted the queued task was serialized as `InstallPackParams(id='ComfyUI-Ollama-Describer', version='nightly', …)` — for a task whose `kind` was `update`.

## Root cause

`web/js/comfyui-mcp-panel.js`, `graph_update_node`, v2 branch:

```js
const params = {
  // The Manager's update task keys off node_name; include id too for
  // correlation parity with install (harmless if the server ignores it).
  node_name: id,
  id,
  selected_version: sel,
  version: sel,
  mode: mode || "remote",
  channel: channel || "default",
};
```

Those extras are not harmless. In Manager v4.2.2, `QueueTaskItem.params` is an **untagged Pydantic union**, and `InstallPackParams` is **first**:

```python
class QueueTaskItem(BaseModel):
    ui_id: str
    client_id: str
    kind: OperationType
    params: Union[
        InstallPackParams,      # <-- first match wins
        UpdatePackParams,
        UpdateAllPacksParams,
        ...
    ]
```

`InstallPackParams` requires exactly `selected_version` + `mode` + `channel` (plus `id` from `ManagerPackInfo`) — every one of which we send. So it validates first, Pydantic drops the unknown `node_name`, and `kind` is never consulted for the choice. `do_update` then reads `params.node_name` and raises.

`UpdatePackParams` is `{ node_name: str, node_ver?: str }` and nothing else. Sending only those two makes it the sole match.

The fields added "for correlation parity with install" are precisely what break it.

## Fix

Send `UpdatePackParams` alone for `kind: "update"`:

```js
const params = { node_name: id };   // + node_ver when a current version is known
```

If the id is genuinely needed for correlation, `ui_id` already carries it at the envelope level, outside the union.

## Worth checking while in there

Manager guards two of its sibling handlers against exactly this collision:

```python
is_unknown = getattr(params, 'is_unknown', False)  # guard: pydantic Union may match UpdatePackParams
```

[...]

PR #927, opened 2026-08-10
fix(install): send the repository URL Manager v4 requires for a nightly install (#920)

### LINKED ISSUE #920: Fix panel_install_node repository URL being treated as registry ID

artokun:
Fixes #920.

`panel_install_node({repository:"https://github.com/kijai/ComfyUI-SolAttn_triton.git", version:"nightly"})` queued a **registry lookup** instead of a source install, and failed with:

```
Node 'ComfyUI-SolAttn_triton@nightly' not found in [ManagerChannel.dev, ManagerDatabaseSource.cache]
```

`buildInstallRequest`'s v2 branch reduces the URL to its basename via `gitRepoName()` and never sends it. Manager receives `id:"ComfyUI-SolAttn_triton"`, `channel:"dev"`, `mode:"cache"` — which is exactly the pair the error names.

## The blocking unknown is now answered

I previously left this unfixed because I did not know Manager v4's source-install parameter, and guessing at `QueueTaskItem.params` — an **untagged** Pydantic union — is how #809 happened (extra fields silently dropped, wrong model selected).

The authoritative shape is generated from the Manager v4 API in `Comfy-Org/ComfyUI_frontend`, `src/workbench/extensions/manager/types/generatedManagerTypes.ts`:

```ts
InstallPackParams: ManagerPackInfo & {
  selected_version: string | 'latest' | 'nightly'
  /** GitHub repository URL (required if selected_version is nightly) */
  repository?: string
  pip?: string[]
  mode: ManagerDatabaseSource
  channel: ManagerChannel
  skip_post_install?: boolean
}

ManagerPackInfo: {
  /** Either github-author/github-repo or name of pack from the registry */
  id: string
  version: string
  ui_id?: string
}
```

Two things follow, and the current code gets **both** wrong:

1. **`repository` is an accepted field and is REQUIRED when `selected_version` is `nightly`.** Dropping it does not degrade the request — it guarantees failure for exactly the call in this report.
2. **`id` should be `github-author/github-repo`**, per its own description — not the bare basename the panel derives. `kijai/ComfyUI-SolAttn_triton`, not `ComfyUI-SolAttn_triton`.

This also means adding `repository` makes the payload match `InstallPackParams` *more* specifically rather than less, which is the safe direction across that untagged union.

## Status

Draft — implementation and the regression test the reporter asked for to follow. What this must prove: a repository-only request sends the URL through, `id` carries the owner, and the registry path for a plain pack id is unchanged.

### DIFF
--- a/browser_tests/unit/install-repository-url.test.mjs
+++ b/browser_tests/unit/install-repository-url.test.mjs
@@ -1,0 +1,103 @@
+// panel#920 — `panel_install_node({repository:"https://github.com/kijai/ComfyUI-SolAttn_triton.git",
+// version:"nightly"})` queued a REGISTRY LOOKUP instead of a source install:
+//
+//   Node 'ComfyUI-SolAttn_triton@nightly' not found in
+//   [ManagerChannel.dev, ManagerDatabaseSource.cache]
+//
+// `buildInstallRequest`'s v2 branch reduced the URL to its basename via
+// gitRepoName() and never sent it. Manager got id:"ComfyUI-SolAttn_triton",
+// channel:"dev", mode:"cache" — which is exactly the pair the error names.
+//
+// WHY IT WAS WRITTEN THAT WAY, and why that reasoning does not extend here: a
+// full URL sent as `id` made Manager v4 silently mark the install "done" while
+// doing nothing, and made 3.x fail late. So the panel derives a name instead.
+// That is right for `id`. It is wrong when the caller supplied `repository`,
+// because then the URL IS the request and no registry entry exists to find.
+//
+// THE CONTRACT (not a guess — this is why the fix waited). Generated from the
+// Manager v4 API in Comfy-Org/ComfyUI_frontend,
+// src/workbench/extensions/manager/types/generatedManagerTypes.ts:
+//
+//   InstallPackParams: ManagerPackInfo & {
+//     selected_version: string | 'latest' | 'nightly'
+//     /** GitHub repository URL (required if selected_version is nightly) */
+//     repository?: string
+//     mode: ManagerDatabaseSource
+//     channel: ManagerChannel
+//     ...
+//   }
+//   ManagerPackInfo: {
+//     /** Either github-author/github-repo or name of pack from the registry */
+//     id: string
+//     version: string
+//   }
+//
+// The consequence this fixes: `repository` is REQUIRED for a nightly install and
+// was being dropped. `id` is left alone — see the note in the source.
+//
+// `params` is an UNTAGGED Pydantic union with InstallPackParams first, so a
+// guessed field would be dropped SILENTLY and the wrong model could win — that
+// is #809. Adding `repository` makes this payload match InstallPackParams more
+// specifically, which is the safe direction across that union.
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { buildInstallRequest } from "../../web/js/lib/manager-install.js";
+
+const URL_GIT = "https://github.com/kijai/ComfyUI-SolAttn_triton.git";
+
+// ---------------------------------------------------------------------------
+// 1. The report, end to end.
+// ---------------------------------------------------------------------------
+
+test("#920 a repository-only nightly install SENDS the url", () => {
+  const req = buildInstallRequest("v2", { repository: URL_GIT, version: "nightly" });
+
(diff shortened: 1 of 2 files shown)


artokun:
**Withdrawing. This does not fix #920 — Manager v4 parses `repository` and throws it away.** Adversarial review found it; I verified before accepting.

## Measured against Manager's actual Python, which is what I should have done first

`do_install` reads exactly five fields:

```
params.id
params.selected_version
params.channel
params.mode
params.skip_post_install
```

`params.repository` is never accessed. Across every 4.x tag and the in-development branch:

```
4.2.2     params.repository read 0 times
4.1       params.repository read 0 times
draft-v4  params.repository read 0 times
```

`install_by_id(node_name, version_spec, channel, mode, …)` has no repository parameter at all, and the nightly branch resolves the clone URL **from the database** — `get_custom_nodes(channel, mode)` → `the_node['repository']`, falling back to `cnr_map[node_id]['repository']`, else the reporter's exact error string.

And the target is genuinely unresolvable either way: `api.comfy.org/nodes/ComfyUI-SolAttn_triton` → **404**.

So the payload change is behaviourally inert. The reporter would re-run it and get the byte-identical error.

## Where my reasoning broke, precisely

I verified the **schema** and stopped. `generatedManagerTypes.ts` is generated from an OpenAPI spec — it is the contract of the **wire format**, not of the **behaviour**. My source comment said *"generated FROM the v4 API, so this is the contract rather than a guess"*, and that sentence is exactly the error: it is a contract about what the server will *accept*, not about what it will *do*.

The sharper part: I cited **#809** as the cautionary precedent in both the code comment and the commit message. #809 was diagnosed by reading `manager_server.py`'s `do_update()`. I invoked its discipline while not applying its method — I cited only TypeScript. A schema field that no handler reads is indistinguishable, from the client side, from one that works.

## What is actually true about #920

On a stock Manager v4 there is **no route that installs an unlisted git URL**. The legacy `/manager/queue/install` route does support it (`@unknown` + `files:[url]`), but `comfyui_manager/__init__.py` only registers the legacy server under `--enable-manager-legacy-ui`.

That reframes the panel-side fix entirely: the useful change is an **honest error** — telling the user the pack is not in the registry and that v4 cannot install an arbitrary URL — instead of a misleading registry-lookup failure that reads like a lookup bug. And the underlying defect is upstream: v4's own schema declares `repository` required for a nightly install and the handler discards it.

Closing unmerged. #920 stays open, re-scoped on the issue.

## One thing that did hold up

The union-mis-selection risk (#809's actual failure mode) was checked and is genuinely absent: every competing member is excluded on required fields, `UpdateAllPacksParams` is all-optional but sits after `InstallPackParams`, and there is no `ConfigDict`/`extra` anywhere, so Pydantic's default `extra='ignore'` applies. The payload already matched `InstallPackParams` uniquely before this change. That analysis was sound — it just protected a change that does nothing.

Issue #697, opened 2026-08-06
panel_set_widget transiently loses active graph node until current-workflow rebind


artokun:
### What happened / root cause
panel_set_widget transiently returned "No node with id 105 in the current graph" immediately after panel_graph_outline and panel_query_graph had both confirmed node 105 on the active workflow. The active workflow and target mode remained correct. Rebinding with panel_set_workflow_target(mode:"current"), rereading the graph, and retrying the identical edit succeeded. This suggests stale workflow routing/session resolution between consecutive panel calls.

### Steps to reproduce
1. Keep workflows/video_minimax_h3_i2v.json active with panel target mode "current".
2. Confirm subgraph node 105 using panel_graph_outline and panel_query_graph.
3. Call panel_set_widget(node_id:105, widget:"prompt", value:<long prompt>).
4. Observe transient error: "No node with id 105 in the current graph".
5. Call panel_set_workflow_target(mode:"current"), reread the graph, and retry; the edit succeeds and promotes to inner node 104.

### Exact error (scrubbed)
Error: No node with id 105 in the current graph

### Fix
applied locally: no source patch
upstream-only: yes
Temporary verified workaround: explicitly rebind to the current workflow, reread the graph, then retry the identical panel_set_widget call.
Precise change needed: keep panel graph mutations bound to the same active workflow UUID/routing key as immediately preceding reads, or transparently refresh/re-resolve the current workflow once before returning "node not found" when the target is in current mode.

### Environment
Windows 11; ComfyUI 0.30.0; NVIDIA GeForce RTX 5090 32 GB; comfyui-mcp 0.49.8; panel 0.11.41.

Issue #302, opened 2026-07-30
panel_move_node cannot move subgraph rail IDs returned by panel_query_graph


artokun:
### What happened / root cause
Inside a subgraph, `panel_query_graph` reports the input and output rails as `rail_node_id` values `-10` and `-20`, but `panel_move_node` rejects those same IDs. This prevents programmatic subgraph layout even though the rail positions and IDs are exposed.

### Steps to reproduce
1. Enter a subgraph with `panel_enter_subgraph`.
2. Call `panel_query_graph` and note `rails.input.rail_node_id = -10`.
3. Call `panel_move_node({node_id:-10,pos:[160,2500]})`.

### Exact error
`Error: No node with id -10 in the current graph`

### Fix
Applied locally: no
Upstream-only: yes
Precise change needed: make `panel_move_node` recognize the special input/output rail IDs reported by `panel_query_graph` while viewing a subgraph, or add a dedicated rail-position tool. The write path should update the subgraph rail object's canvas position and preserve Ctrl+Z undo.

### Environment
Windows 11; ComfyUI 0.29.0; NVIDIA RTX PRO 4000 Blackwell 24 GB; comfyui-mcp 0.48.12; sidebar panel version paired with comfyui-mcp 0.48.12.

PR #930, opened 2026-08-10
fix(subgraph): a rail id is not a missing node (comfyui-mcp#1294)

artokun:
Addresses the diagnosis half of artokun/comfyui-mcp#1294. **Deliberately not a closing reference** — the other half of that issue (unexpose/remove-boundary tools) is unbuilt and stays open.

`panel_query_graph` reports boundary rails as `rails.output.rail_node_id: "-20"`. Fed back to a write, the panel answered that the id may be from a different workflow or that the node was removed, and told the caller to re-read `panel_graph_outline` — which hands back the same id. Every clause after the first was false, and the remedy was a loop.

This is #697's mistake in a new place: there the missing axis was SCOPE, here it is KIND.

The failure path now asks what the id *is* before answering that it is not there, reusing the `resolveRailNode` that `move_node` has taken since #302. Its real-node collision guard is what makes that safe — a rail is only recognised after `getNodeById` has already declined the id — and a test pins the inversion, since it would misreport ordinary missing-node failures as rails.

**Scope is the diagnosis only.** The message does not imply a removal path, because there isn't one; it names the invasive workaround that does work today rather than inventing a tool name.

3353 unit tests pass, typecheck clean. Verified by mutation: deleting the *call site* fails the wiring test.

### DIFF
--- a/browser_tests/unit/node-scope-locator.test.mjs
+++ b/browser_tests/unit/node-scope-locator.test.mjs
@@ -132,7 +132,7 @@ test("WIRING: resolveNode builds its error through describeMissingNode", async (
   // at source. Without it every one of them reverts to the bare message.
   const { readFile } = await import("node:fs/promises");
   const src = await readFile(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url), "utf8");
-  assert.match(src, /import \{ describeMissingNode \} from "\.\/lib\/node-scope-locator\.js";/);
+  assert.match(src, /import \{ describeMissingNode(?:, describeRailNodeTarget)? \} from "\.\/lib\/node-scope-locator\.js";/);
   const fn = src.slice(src.indexOf("function resolveNode(graph, nodeId) {"));
   const body = fn.slice(0, fn.indexOf("function normalizeLegacyNodeId"));
   assert.ok(body.includes("describeMissingNode(nodeId, rootGraph, viewingRoot)"),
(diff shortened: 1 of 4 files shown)


artokun:
## Codex adversarial review — NO-SHIP, and it caught me reintroducing the defect

**P2 — the message claimed something it cannot prove. Fixed.** It said the id is "not a stale or foreign id". Node ids are arbitrary integers, so an ordinary node may once have held this id and been deleted; the same id then reaches the rail check and matches. That caller *is* holding a stale id and was about to be told they weren't — the exact defect this PR exists to remove, reintroduced one clause later.

The fact established is "this resolves to a rail in the graph you are viewing". "Therefore your id is fine" does not follow. The message now reports what resolved and names the one case it cannot rule out, so `panel_graph_outline` stays available to the caller it's right for without being prescribed for a `rail_node_id` — which is the loop the reporter ran.

**And a second error, found while checking the review's other claim rather than assuming it.** Codex noted it couldn't verify the tool-surface claims (its shell was denied). I could, and two of them were wrong:

| tool | how it actually takes a rail |
|---|---|
| `panel_move_node` | the rail **id** — but `pos` only; the handler already refuses other fields |
| `panel_move_rail` | `rail: "input"｜"output"` — **never** an id |

The draft said "rail ids ARE accepted by panel_move_node and move_rail", which would send a caller to pass `-20` to a schema with no such parameter: a remedy that fails on contact. Checked against the registrations, not memory.

The claims codex could not verify, verified: `expose_subgraph_input`, `expose_subgraph_output` and `graph_move_rail` are the whole boundary surface — no unexpose exists, so the message promises none.

Its other findings I agree with as stated: `resolveRailNode` throwing is caught and preserves the original diagnosis, a null graph fails earlier at `getNodeById` (baseline, not new), and branch order is correct — the last of which is pinned by an assertion, since it can regress silently.

3354 unit tests pass, typecheck clean. Wiring mutation-verified: deleting the call site fails the wiring test.

PR #934, opened 2026-08-10
fix(#584): make the tab measure its own module cache instead of asking for DevTools

artokun:
Draft — claiming #584.

## Why another round at a different layer

Five fixes have shipped against this issue and it keeps recurring. Measurements I took against the live rig today move the ground under the current plan of record:

**1. It is 112 modules, not 12.** `/api/extensions` lists every `.js` recursively — 12 top-level, 96 `lib/`, 4 `vendor/`. The last analysis on this thread assumed `lib/*` arrived as nested imports beneath a dozen entry points; they are independent auto-imported fetches.

**2. The scan filters by suffix, and it is live.** A `.mjs` dropped next to a `.js` is not listed, and `/api/extensions` reflects the change with no restart. So "collapse to one entry point" is achievable — by taking modules out of the `.js` scan — but that is a 100-file rename and not this PR.

**3. The heal already reaches all 12 top-level modules.** I expected `primeModuleCache` to miss the siblings ComfyUI imports directly; computed over the real import graph, 108 of 112 are reachable and every top-level module is. That hypothesis is dead.

**4. All 112 are served `no-store` — and that is ComfyUI's doing, not ours.** `middleware/cache_middleware.py` sets `Cache-Control: no-store` on any path ending `.js`, introduced by ComfyUI e0982a71 (*"use no-store cache headers to prevent stale frontend chunks"*, #12911) — whose own commit message describes this issue's mechanism, down to ETags from mtime+size returning 304 with stale content. It predates every reporter's ComfyUI version.

So on a direct connection the HTTP cache **cannot** serve a stale module — yet a hard refresh is what fixes it, which means on affected machines the response is not arriving with that header. I cannot reproduce that here; every request on this rig is a real 200 with `no-store`.

## What this PR does

Stops asking users to open DevTools. The page already knows: `performance.getEntriesByType("resource")` reports, per module, whether it came from the network or from cache (`transferSize`), and what status it got. The panel will summarise its own module fetches and surface it where a wedged session actually looks.

That yields the decisive datum automatically on the next recurrence, and distinguishes the two live branches — version skew across modules vs. a header-stripping intermediary — which no amount of source reading can settle.

Not claiming a fix for the staleness itself. Instrumenting it first, because five attempts have now been built on an unmeasured hypothesis.

### DIFF
--- a/browser_tests/unit/module-cache-report.test.mjs
+++ b/browser_tests/unit/module-cache-report.test.mjs
@@ -49,7 +49,10 @@ test("a healthy page reads all-network — and must not warn", () => {
   assert.equal(s.verdict, "all-network");
   assert.equal(s.total, 2);
   assert.equal(s.cached, 0);
-  assert.match(describeModuleCache(s), /staleness here is NOT the HTTP cache/);
+  // The claim is bounded to what the buffer can support (codex review): "every
+  // module STILL IN THE BUFFER", never "every module".
+  assert.match(describeModuleCache(s), /STILL IN THE BUFFER was fetched/);
+  assert.ok(!/Every module was fetched/.test(describeModuleCache(s)), "an unbounded claim");
 });
 
 test("MIXED is the finding this issue is missing — some modules stale, version healthy", () => {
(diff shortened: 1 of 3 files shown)


artokun:
## Codex adversarial review — NO-SHIP, four findings, all four real

Every one was a way this could report the **false all-clear** it exists to prevent — which is the only failure mode that matters, since a false all-clear sends the sixth attempt back down the path this closes.

**H1 — a 304 was classified as `network`. The worst possible miss here.** A revalidation transfers headers, so `transferSize` is nonzero while the body comes from cache. A page built entirely of stale 304s would have reported *"staleness here is NOT the HTTP cache"* — clearing the prime suspect. And it is not a hypothetical suspect: ComfyUI's own e0982a71 describes this issue's mechanism as an aiohttp ETag from mtime+size matching, 304, stale content served. Now its own state, counted with cache.

**H2 — `all-network` was an unsupported completeness claim.** The buffer's floor is 250 entries and this page loads 112 panel modules on top of ComfyUI's own. Evicted cached entries plus surviving fetched ones reads as "everything was fetched". The summary now records `bufferMaybeFull` and no verdict claims completeness — *"every panel module **still in the buffer** was fetched"*.

**M1 — cache provenance is not content equality.** *"This page is running a MIXTURE of versions"* asserted the conclusion the evidence only makes possible; a cached module can be byte-identical to the served one. It now reports the observation, says skew is possible and not proven, and keeps the part that matters: skew is invisible to the single-version check that just reported healthy.

**M2 — a cached stylesheet could manufacture a finding.** Resource Timing reports a stylesheet and a `link rel=modulepreload` identically as "link", so the initiatorType filter admitted CSS. The URL decides now (`.js`, query strings tolerated).

Its read on compression I agree with and did not change: `decodedBodySize > 0` survives gzip/Brotli, and zero decoded size stays conservatively `unknown` for opaque cross-origin entries.

Six tests added, one per finding plus the neighbour that keeps each honest — a lone 304 among fetches must be `mixed`, a small buffer must carry no caveat. 3372 unit tests pass, typecheck clean, and both wiring sites remain mutation-verified.

Issue #847, opened 2026-08-09
e2e: 2 specs fail — chat-history-v2 row count, workflow-chat-identity first record


artokun:
Split out of #793, which fixed the 8 specs the workflow-instance fence was refusing. These 9 have a **different cause** and #793 deliberately did not claim them.

Measured against `main` at f226d10, live ComfyUI, `--workers=2`: **9 failed / 40 passed** (was 17/30 before #793).

## The failures

```
browser_tests/conversation-persistence.spec.ts:216  restores the latest panel-owned conversation after sessionStorage is lost
browser_tests/conversation-persistence.spec.ts:261  restores from IndexedDB after the localStorage shadow is lost
browser_tests/conversation-persistence.spec.ts:293  panel delete remains final when a stale tab republishes the removed thread
browser_tests/conversation-persistence.spec.ts:345  clear all removes durable transcripts, preserves workflow identity, and fences a stale tab
browser_tests/conversation-persistence.spec.ts:444  reload keeps the pointed conversation and live tab session during durable hydration
browser_tests/conversation-persistence.spec.ts:502  strict workflow storage sync detaches transcript todos and session before the next record
browser_tests/conversation-persistence.spec.ts:587  workflow rename publishes alias tombstones and a stale tab cannot echo them back
browser_tests/chat-history-v2.spec.ts:66            keeps multiple chats, supports search and restores from IndexedDB without localStorage
browser_tests/workflow-chat-identity.spec.ts:33     opening a workflow does not dirty it and first record embeds silently
```

Seven of nine are one file, and all nine are about **conversation persistence / workflow-chat identity** rather than graph mutation — so this is plausibly one root cause, not nine.

## Separately: a parallelism-sensitive failure

`agent-drive.spec.ts:124` ("append-on-scroll re-applies the highlight to a card from a LATER page") fails under `--workers=2` and passes in isolation. I confirmed it fails **on main** independently of #793, so it is pre-existing. Either a real race in the modal's append-on-scroll handling or a test that assumes exclusive use of something shared.

## Why this is worth a ticket rather than a note

CI runs `test:unit` only, so nothing here is enforced anywhere. #793's summary of that problem still stands: "17 files of e2e coverage" reads as more assurance than it provides, and the part that is still down is the part covering **whether a user's conversations survive a reload** — which is not a small thing to be unable to detect a regression in.

## Prerequisites note

[...]

Issue #557, opened 2026-08-02
panel_save_workflow desyncs the graph binding: all panel_* commands fail until a frontend reload


artokun:
### What happened / root cause

After a successful `panel_save_workflow`, every subsequent `panel_*` graph command fails — reads *and* writes — with:

> The live graph is out of sync with the active workflow: the workflow reports 205 node(s), but the canvas is bound to a different graph.

`panel_open_workflow` cannot recover it either (`workflow_open could not prove that the active canvas was rebound to the requested workflow`). Only `panel_reload(scope:"frontend")` restores the binding, which ends the agent's turn.

**Root cause** (traced in the installed panel):

`assertGraphBoundToActiveWorkflow()` in `web/js/comfyui-mcp-panel.js` (~line 4109) throws when `graphRootWorkflowUuidMismatches({rootGraph, activeWorkflowUuid})` is true — `web/js/lib/graph-binding.js:235`:

```js
export function graphRootWorkflowUuidMismatches({ rootGraph, activeWorkflowUuid } = {}) {
  if (typeof activeWorkflowUuid !== "string" || !activeWorkflowUuid) return false;
  const rootUuid = rootGraph?.extra?.comfyui_mcp?.workflow_uuid;
  return typeof rootUuid === "string" && rootUuid && rootUuid !== activeWorkflowUuid;
}
```

The save re-registers/replaces the active workflow object. `_workflowObjectUuids` is a `WeakMap` keyed on the workflow object, so the **new** object has no cached entry; `workflowStableUuid()` therefore mints a **fresh** UUID. Meanwhile `rootGraph.extra.comfyui_mcp.workflow_uuid` still carries the **pre-save** UUID. Both sides are "established", they differ, and the guard fails closed on every command.

The `loadGraphData` wrapper has careful re-stamp/invalidate logic for this exact class of identity change (the `#570 P0b` block), but the **save** path has no equivalent, so the root tag and the object cache drift apart.

### Steps to reproduce

1. Open a saved workflow (repro was 205 nodes / 31 groups; size may not matter).
2. Make graph edits via panel tools (e.g. `panel_set_node_color`).
3. Call `panel_save_workflow` with no `name` (save in place) — returns `{saved: true}`.
4. Call any `panel_*` graph tool. It throws the desync error.
5. `panel_open_workflow` fails to rebind; only `panel_reload({scope:"frontend"})` recovers.

[...]

PR #936, opened 2026-08-10
fix(e2e): the remaining conversation-persistence spec failures (#847)

artokun:
Draft — claiming #847.

Measuring the current failure set first: the issue body lists 9 specs, the title says 2 remain. Neither is trustworthy without a run.

Refs #847


artokun:
Closing unmerged — this carries no code, only a measurement, and that now lives on #847 where it belongs.

What it established: 8 of the 9 specs #847 opened with now pass; the one that remains is a genuine product bug, not a flaky test. Two conversations in one workflow get stamped with two different storage uuids because the uuid is re-minted at the unsaved→named boundary, so the earlier chat drops out of "Current workflow only" — the first chat about a workflow vanishes when you filter to that workflow.

I stopped rather than land a partial fix: the change belongs in `workflowStableUuid`'s identity resolution, where #557, #558 and #570 all converge, and I did not have room to implement and verify it properly.

PR #940, opened 2026-08-10
fix(e2e): the suite cleans up the workflows it saves, and proves it

artokun:
Draft — claiming #907.

Re-measured on this machine before starting: **1269 of 1286** files in `user/default/workflows` are `Untitled 2026-08-*` test output. The issue reported 1221 of 1240, so it is still leaking.

Two things follow that shape the fix:

1. Per-spec cleanup is what this repo already tried. Three specs clean up, two do not, and the two that do not were written before the review that added it to the others. A convention nobody can forget is not a convention — so the guarantee belongs to the suite.
2. **A leak nobody notices is the actual defect.** 1269 files accumulated silently. Whatever cleanup lands has to fail loudly when it stops working, or this is back in a month.

Also worth testing while here: #847's two remaining failures are a chat-history **row count** and a workflow-chat-identity **first record**. A workflows store holding 1286 entries on a dev box is exactly the kind of thing that perturbs those, and no user install looks like that. I will measure whether the litter causes them before assuming the two issues are independent.

I will **not** mass-delete the existing 1269 files — that is the user's workflow directory and the deletion is theirs to authorise. A dry-run-by-default cleanup script comes with this instead.

### DIFF
--- a/browser_tests/bridge-reregisters-after-restart.spec.ts
+++ b/browser_tests/bridge-reregisters-after-restart.spec.ts
@@ -44,7 +44,11 @@ test('the tab re-registers after the bridge dies and respawns', async ({
   await panel.openSidebar()
   await panel.connect()
 
-  const saved = await mockBridge.command('workflow_save', {})
+  // #907 — an unnamed save gets ComfyUI's `Untitled <date> <time>`, the SAME
+  // name it gives the developer's own unnamed saves, so nothing downstream can
+  // tell them apart and the cleanup must not try. Name it ours.
+  const e2eSaveName = `cmcp-e2e-${Date.now().toString(36)}-${Math.random().toString(36).slice(2, 8)}`
+  const saved = await mockBridge.command('workflow_save_as', { name: e2eSaveName })
   expect(saved.ok, 'the workflow must save so the route is the composed wf: form').toBe(true)
   const savedName = String(saved.result?.workflow || '')
   expect(savedName).toBeTruthy()
--- a/browser_tests/bridge-route-never-file-derived.spec.ts
+++ b/browser_tests/bridge-route-never-file-derived.spec.ts
@@ -81,7 +81,11 @@ test('two tabs on one saved workflow never share a bridge route', async ({
 
     // A SAVED workflow specifically. An unsaved tab routes on a per-object `tmp:<uuid>`
     // and could never collide, so using one would pass for the wrong reason.
-    const saved = await mockBridge.command('workflow_save', {})
+    // #907 — an unnamed save gets ComfyUI's `Untitled <date> <time>`, the SAME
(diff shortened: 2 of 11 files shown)


artokun:
## Codex adversarial review — NO-SHIP, two P0s, and it was right to refuse

Both P0s were on code that deletes from the developer's workflow library, so the strictest reading is the correct one.

**P0 — a developer's own unnamed save could be deleted.** The cleanup matched ComfyUI's `Untitled <date> <time>`, because that is what the 1272 leaked files are called. It is *also* what ComfyUI calls the user's own unnamed saves. Someone who hits save while the suite runs creates a file that is both new since baseline and matching — my "BOTH conditions" rule, which I had written up as *the* safety property, is no protection when one condition is a name anyone can produce by accident. My test even covered a manual mid-run save, but chose a name that did not match, so it agreed with me instead of checking me.

Fixed upstream of the deletion: the specs now save under `cmcp-e2e-<nonce>`, and the automated cleanup deletes nothing else. Ownership is **proven by a name only this suite produces**, not inferred.

**P0 — the manual script had no baseline at all.** `--apply` would have deleted every Untitled it found. It now requires `--include-untitled` on top of `--apply` for that family and writes the full candidate list to a file first.

**P1 — concurrent runs shared one ownership record** at a fixed temp path, so run A's teardown could delete run B's in-flight files against a baseline that was never A's. Now per-run.

**P1 — the teardown failed runs it should not.** A developer's mid-run save was reported as a leak and threw, while the message itself said it was safe to ignore; and a failed final listing was read as "every delete failed". Now it throws only when a file we recognise as ours could not be deleted, warns on the unfamiliar, and says "unverified" when the check could not be taken.

**P2 — requests bounded at 10s.** Its read on DELETE URL encoding I agree with and did not change.

## What this fixes, and what it does not — measured per spec

| spec | left behind |
|---|---|
| `dirty-guards-see-node-writes` | **0** (was leaking) |
| `save-persists-node-set-values` | **0** (was leaking) |
| `workflow-chat-identity` | 2 Untitled |
| `bridge-reregisters-after-restart` | 1 Untitled |
| `bridge-route-never-file-derived` | 1 Untitled |

The two specs #907 names are fixed. The other three still produce unnamed saves from a path I could not identify: not `workflow_save` (converted), not `workflow_new` (it opens a temporary tab), and ComfyUI's autosave is off on this install. The teardown deliberately leaves them alone rather than guess.

**That is the guard working, not a gap in it.** It names them every run — which is how I found them, and how the remaining source gets found. So this is `Refs #907`, not `Fixes`: the issue stays open with the unaccounted-for path written down.

Verified in a real browser throughout: 9 tests across 5 specs, exit 0, and the two converted specs measured at delta 0. 3392 unit tests pass.

Issue #817, opened 2026-08-09
panel_graph_outline rejects active workflow after tab switch with root-workflow-uuid-mismatch


artokun:
### What happened / root cause
Immediately after the panel announced a switch to an active workflow, panel_graph_outline rejected the graph because the canvas identity tag was stale. Re-opening the already-active workflow fixed the binding. The panel should rebind the root workflow UUID atomically on tab switch, or automatically retry/rebind a read-only outline operation.

### Steps to reproduce
1. Switch the open workflow tab.
2. Immediately call panel_graph_outline.
3. Observe root-workflow-uuid-mismatch despite panel_list_workflows reporting the same workflow as active.
4. Call panel_open_workflow on that same active path.
5. Retry panel_graph_outline; it succeeds.

### Exact error (scrubbed)
[root-workflow-uuid-mismatch] The live canvas carries a different workflow's identity tag than the active workflow.

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: synchronize/rebind the canvas workflow identity tag as part of the tab-switch event before graph tools are admitted, or make read-only graph calls perform a safe one-time rebind/retry when active path and workflow UUID are otherwise consistent.

### Environment
Windows 11; ComfyUI 0.30.0; frontend version not exposed in the session environment line; RTX 4060 Ti 8GB; comfyui-mcp 0.50.42; panel 0.11.45.

artokun:
Claimed — draft PR open, branch `fix/817-uuid-mismatch-after-tab-switch`. Will reproduce against the real binding predicates before changing anything.

artokun:
Shipped in **0.11.51** (PR #843, released via #844 — published to the Comfy Registry).

## What was happening

Reproduced against the real predicates before changing anything; your error text came back verbatim.

ComfyUI reuses **one** canvas object across tabs and its clear/configure does not reset the canvas metadata. So switching A → B leaves A's identity tag sitting on a canvas that now holds B's graph — and the tag guard refuses every graph tool.

The sharp part is why nothing recovered it. The panel *could* already heal a canvas carrying **no** tag: it proves the canvas is the active workflow's by comparing content, then stamps it. It could **not** heal one carrying the **wrong** tag, because that stamp is deliberately never overwritten. So a wrong tag was stickier than no tag — a byte-identical canvas was allowed in one case and refused in the other, and re-opening the already-open workflow was the only way out. Which is exactly the workaround you found.

## The fix

[...]

PR #944, opened 2026-08-10
fix(#833): name the canvas swap point, from the canvas-rebuild side

artokun:
Draft — claiming #833's remaining half.

The last pass on this issue eliminated `changeTracker` as the identity signal (it holds only serialized state, no live graph handle) and named the real blocker:

> Every remaining route needs the same missing thing: **a moment where the canvas-to-workflow binding is provable**, to seal from. That is not a gap in the proof predicates — it is that nothing in the panel can currently observe *when the canvas is swapped*.

It also said where to look next: **the canvas-rebuild side**, after four failed attempts from the store-API side in #817.

That is what this PR is. I can now drive the real frontend under Playwright, so instead of reading the store's surface again I am going to instrument what actually happens when a canvas is replaced — `loadGraphData`, `LGraph#configure`, the `app.graph` reference itself, and the workflow store's activation — and record the ORDER, plus whether any moment exists where the new graph and its workflow are both observable and agree.

If such a moment exists, it unblocks both #833's write half and #817, which are blocked on the identical unknown.

**I may not find one.** If the answer is that no such moment is observable from a custom node's position, that is the deliverable — written down with the probe that shows it, so the fifth attempt does not start from zero. The write-half refusal stays in place either way; the e2e spec asserts it, and relaxing it is an explicit act.

### DIFF
--- a/browser_tests/canvas-swap-point.spec.ts
+++ b/browser_tests/canvas-swap-point.spec.ts
@@ -1,166 +1,254 @@
-// #833 / #817 — NAMING THE CANVAS SWAP POINT.
+// #833 / #817 — WHAT IS ACTUALLY OBSERVABLE WHEN THE CANVAS IS SWAPPED.
 //
-// Both issues are blocked on one missing thing, stated on #833:
+// Both issues are blocked on the same missing thing, stated on #833: a moment
+// where the canvas-to-workflow binding is PROVABLE, to seal from. Four attempts
+// from the store-API side failed to find one, so this comes at it from the
+// canvas-rebuild side — instrumenting the live frontend rather than reading its
+// surface.
 //
-//   > Every remaining route needs the same missing thing: a moment where the
-//   > canvas-to-workflow binding is PROVABLE, to seal from. That is not a gap in
-//   > the proof predicates — it is that nothing in the panel can currently
-//   > observe when the canvas is swapped.
+// THIS SPEC IS EVIDENCE, AND IT IS DELIBERATELY NARROW. A first version claimed
+// "the binding is provable at loadGraphData:after" and codex refused it: the
+// claim was broader than the measurement in three ways, and two issues were
+// about to be built on it.
 //
-// Four attempts from the store-API side failed, so this comes at it from the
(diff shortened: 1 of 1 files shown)


artokun:
## Codex review — NO-SHIP, three P0s, all aimed at the conclusion rather than the code

Correct on every count, and the right call: two issues were about to be built on a claim broader than its evidence.

**P0 — pairing was not per-swap.** `indexOf`/`lastIndexOf` over a flat label array cannot establish that one call is *nested in* another. Calls now carry ids and every event records the stack of `loadGraphData` invocations in progress, so nesting is measured rather than inferred from adjacency.

**P0 — "the binding is provable at the exit" was not measured.** I asserted it from "a path is truthy and a graph exists". Each activation is now tied to its specific bracketing call, and the assertion is that *that* call's exit names *that* activation's workflow.

**P0 — one swap route cannot support an all-routes conclusion.** The spec now counts rebuilds observed **outside** any bracket and reports it — 0 of 3 here — and states plainly that a fix sealing at the bracket must still fail closed on an activation it never bracketed. That is a limit of the measurement, not a proof of coverage.

Also fixed: patch installation is asserted (otherwise every finding is vacuous rather than negative), the prototype patch is restored, and the fixed sleep is gone.

**And my first correction was worse than the sleep it replaced** — worth recording. Waiting for "the log grew" returned on the first `load:enter` and read mid-swap, so the run concluded *no activation was observable*: a pure harness artifact that looked like a real negative finding. Then waiting for a bracket to close matched ComfyUI's own startup restore, still in flight when the instrument installs. It waits for quiescence first now, and that alone is the difference between measuring nothing and measuring this:

```
load:enter   path=<old>   nodes=<old>
  configure  path=<old>   nodes=<NEW>   graph rebuilt, store still names the old one
  open:exit  path=<NEW>   nodes=<NEW>   store activates — nested inside the call
load:exit    path=<NEW>   nodes=<NEW>   both agree
```

3 of 3 swaps, and `app.graph` is the same object throughout.

3403 unit tests pass, typecheck clean, and the spec passed 3 consecutive runs.

Issue #945, opened 2026-08-10
embeddedWorkflowUuid(wf, {allowGraph:false}) is always null — two fork guards arbitrate a value they never receive


artokun:
Found while fixing #847. Filing separately rather than folding it in, because the fix is not obvious and the blast radius is larger than that issue.

## The measurement

On ComfyUI 0.31.0, instrumented live for both an unsaved and a freshly-grounded workflow:

```
wf.extra           : undefined
wf.workflow.extra  : undefined
wf.data.extra      : undefined
wf.activeState.extra : { comfyui_mcp: { workflow_uuid: "ff7890d8-…" } }   <- the uuid IS here
app.graph.extra      : { comfyui_mcp: { workflow_uuid: "ff7890d8-…" } }
```

`workflowOwnedExtra()` reads `wf.extra || wf.workflow.extra || wf.data.extra`. All three are undefined, so it returns null, so **`embeddedWorkflowUuid(wf, { allowGraph: false })` returns null unconditionally** on this frontend.

## Why that matters

`workflowStableUuid` resolves a saved workflow as:

```js
let id = objectUuid || embedded || pathAlias || crypto.randomUUID();
```

`embedded` is a permanent null, so that layer is gone. Downstream, `shouldForkEmbeddedWorkflowUuid` and `shouldForkEmbeddedUuidForLiveOwner` exist to decide whether an embedded uuid belongs to a copy and must be forked — they are handed `embedded`, so they are arbitrating a value that never arrives. Whatever they were written to prevent, they are not currently preventing it; they are dead code that reads as a live guard.

The embed WRITE has the same problem: `record()` comments that "workflow mode does [embed] on first record", and the write goes through `workflowOwnedExtra(wf)` — null — so it never lands on the workflow object either.

This is one of three carriers that all failed together at the grounding boundary in #847. The other two (object WeakMap, path alias) fail for structural reasons; this one fails because it is looking in the wrong place.

## What I tried and reverted

Adding `wf.activeState?.extra` to the end of the candidate chain, so it only speaks where the current carriers say nothing. Two reasons I backed it out:

1. **It did not fix #847.** `activeState.extra` holds the original uuid right after the save and is **null** by the time the history filter runs, so the lineage was already gone.
[...]

PR #947, opened 2026-08-10
fix(#945): answer whether the embedded-uuid carrier was ever real

artokun:
Draft — claiming #945.

The issue ends on the question that decides what the fix even is:

> Whether `wf.extra`/`wf.workflow.extra`/`wf.data.extra` were ever populated by any ComfyUI version this pack supports, or whether the chain has been aspirational from the start. If the latter, every guarantee written in terms of "the embedded uuid" needs re-reading.

That is answerable from evidence — the frontend's own workflow object, and this pack's history — and it separates two very different situations:

- **a regression**: the fields existed and went away, so the chain should be repaired;
- **aspirational**: they never existed, so two guards have been arbitrating a value that has never once arrived, and every claim written in terms of "the embedded uuid" is describing something that does not happen.

I am deliberately **not** changing where the embed lands. The previous pass tried adding `activeState.extra` and backed it out for good reasons — it moved identity persistence off `app.graph.extra` to fix a filter bug, which is not a trade worth making silently. Under the current freeze this should end as evidence plus honest naming, not a change to what carries identity.

Verification note: the dev rig's ComfyUI is currently serving another session's branch, so specs that need the panel to MOUNT cannot run. #945's surface is pure functions, so `node --test` covers it — which is part of why I picked it.

### DIFF
--- a/browser_tests/unit/embedded-uuid-carrier.test.mjs
+++ b/browser_tests/unit/embedded-uuid-carrier.test.mjs
@@ -5,7 +5,10 @@
 // `wf.extra` / `wf.workflow.extra` / `wf.data.extra` EVER populated by a ComfyUI
 // this pack supports, or has the chain been aspirational from the start?
 //
-// ANSWER: aspirational, on all the evidence available here.
+// ANSWER: no supported path reaches it. Three pieces of evidence, and one limit
+// on how far they generalise (codex: they cannot establish a historical
+// universal across every release this pack claims to support — nobody has an
+// old frontend in hand to check).
 //
 //  1. THE CLASS DOES NOT HAVE THOSE FIELDS. ComfyUI 0.31.1 / frontend 1.44.19,
 //     read off the shipped bundle:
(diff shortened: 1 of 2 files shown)


artokun:
## Codex adversarial review — NO-SHIP, four findings, and one of them was a false claim

**P1 — "neither guarantee holds" was wrong, and it was the load-bearing sentence.** I wrote that the identity written at the embed site never persists. It does — through a different carrier: the `loadGraphData` wrapper stamps `graphData.extra`, and `rootGraph.extra` carries it on the live canvas, which is what a save serializes. The promise in that comment **is** kept, just not by the path #945 is about. Merging my version would have sent the next reader hunting a data-loss bug that does not exist — and it is exactly why this issue has no user-visible symptom.

**P1 — "aspirational, never real" over-reached.** The evidence supports *no supported path reaches it*: the fields are absent from the observed frontend, the chain arrived as a guess about "older builds", and nothing in this repo has ever exercised it. It does not support a historical universal across every release the pack claims to support — nobody has an old frontend in hand. Narrowed, and the same narrowing applied to the guards: **unreached here**, not "dead code".

**P1 — the tests modelled private helpers, so they could not detect drift.** `workflowOwnedExtra` and `embeddedWorkflowUuid` are module-private, so the checks necessarily reimplement them — which means production could change and the suite would stay green on a stale copy. There is now a source assertion pinning the real chain, the `allowGraph:false` call site, the parameter name both guards read, and the graph-extra stamp the "no symptom" reasoning depends on. Mutation-verified: repointing the chain in the source fails it.

**And it turned up a second finding while I was fixing that.** The chain appears **twice** — it is also the fallback tail of `activeWorkflowExtra`, reached only when the root graph's `extra` cannot be taken, which always succeeds here. That is the mechanism behind the asymmetry in the issue title: the graph carrier answers first, so `allowGraph:true` works and the workflow-owned tail behind it has never had to. My first mutation attempt failed to apply because I assumed one occurrence, which is how I found it.

Still no runtime change, deliberately. 3409 unit tests pass, typecheck clean.

Note on verification: the dev rig's ComfyUI is serving another session's branch, so panel-mounting e2e is unavailable to me. #945's surface is pure functions plus source assertions, which is why it was a reasonable pick under that constraint.

PR #951, opened 2026-08-10
feat(a11y): a text-size setting that actually scales the panel (#753)

artokun:
Draft — claiming #753.

The panel's text is small and there is no way to enlarge it. The obvious user fix does not work: inner sizes are `rem`, which resolve against the PAGE root, so overriding `.cmcp-root { font-size }` scales almost nothing. Only `zoom` scales the whole panel today, which the reporter had to ship as a separate custom-node pack so panel updates would not revert it.

Refs #753


artokun:
Closing unmerged — #753 is already shipped (0.11.52), and I should have checked before opening this.

The UI-scale setting exists (Comfy MCP Agent → General → "Panel UI scale (%)", 100–250%), and a prior iteration already measured the rem→em follow-up, found 830 rem values across 11 files with confirmed compounding in the live panel, recommended against it, and left the decision with the maintainer. The custom-property variant I was reaching for is named in that comment as the option to do *if asked*. It has not been asked for.

No code from this branch is worth keeping.

PR #955, opened 2026-08-10
fix(install): tell the truth when a git URL is not installable, and correct the 0.11.75 note (#920)

artokun:
Third pass at #920. The first two both reshaped the install payload and **both did nothing**.

## Why the two previous attempts were inert

The other session's (shipped as **0.11.75**) and mine (PR #927) each sent `repository`, justified by ComfyUI-Manager's generated schema documenting it as *"required if selected_version is nightly"*. Read from Manager's **source** instead:

```py
async def do_install(params: InstallPackParams):
    node_id = params.id; node_version = params.selected_version
    channel = params.channel; mode = params.mode
    skip_post_install = params.skip_post_install
```

```
params.repository read 0 times  —  tags 4.2.2, 4.1, branch draft-v4
```

`install_by_id(node_name, version_spec, channel, mode, …)` takes no repository argument, and the nightly path resolves the clone URL from Manager's **own database**. The only `params.model_dump()` forwarding is in `do_install_model`, a different handler — checked, because it was the one hole in the argument.

**A generated OpenAPI model is the contract of what a server ACCEPTS, never of what it DOES.** With Pydantic's default `extra='ignore'`, a declared-but-unread field is indistinguishable from a working one on the client side. That is how two independent attempts landed on the same wrong fix within an hour.

## What this changes

**Nothing about the payload** — there is nothing to send. On a stock v4 an unlisted git URL is not installable at all: the legacy `/manager/queue/install` route supports it (`@unknown` + `files:[url]`), but `comfyui_manager/__init__.py` registers the legacy server only under `--enable-manager-legacy-ui`.

**The failure message.** `panel_node_queue_status` — the surface the reporter actually polled — now says this is a *registry* lookup, that the `repository` parameter is accepted and ignored, and names the three things that work: `--enable-manager-legacy-ui`, a manual clone into `custom_nodes/`, or asking the author to publish.

Deliberately **conditional** (`IF YOU PASSED A GIT URL` / `IF YOU MEANT A REGISTRY PACK`): that surface does not carry the original request, so asserting which mistake the caller made would be claiming an unobserved fact — precisely the defect that produced both wrong fixes.

## The 0.11.75 release note is corrected

It promised *"Installing a custom node from a GitHub URL actually clones it now (#920)"*. It does not clone anything. The entry now states the payload change is forward-compatibility, that the install still fails on a stock v4, and what works today. The code itself is kept — it is what the documented API asks for and costs nothing if upstream implements the field.

A user who upgraded to 0.11.75, retried, and hit the identical error would reasonably have concluded the tool was broken in a new way. That is the part worth correcting quickly.

## Verification

- 3352/3352 unit, typecheck clean, vocabulary gate clean, no control bytes
- four mutations, each confirmed applied: never recognise the miss, assert the cause instead of offering it conditionally, remove the call site, fire on every failure — all killed

## Upstream

Worth an issue against ComfyUI-Manager: v4's schema declares `repository` required for a nightly install and the handler discards it. That mismatch is what made two of us confident and wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,16 +12,24 @@ All notable changes to this project are documented here. This project adheres to
 
 ### Fixed
 
-- **Installing a custom node from a GitHub URL actually clones it now (#920).** Passing a
-  repository URL queued a search of the node registry instead, and failed with
-  *"Node 'ComfyUI-SolAttn_triton@nightly' not found"* — because the panel reduced your URL
-  to just the repo's name and threw the rest away, leaving ComfyUI-Manager nothing to
-  clone from.
+- **The install request now carries your repository URL (#920).** The panel used to reduce
+  a GitHub URL to just the repo's name and throw the rest away; it now sends the URL in the
+  `repository` field ComfyUI-Manager's API declares for it.
 
-  The URL is now passed through. Manager's own install parameters require it for a
-  nightly install, which is the case this was reported against.
+  **This does not yet make such an install succeed, and this entry originally said it did —
+  that was wrong.** ComfyUI-Manager v4 accepts the field and then ignores it: its install
+  handler reads only `id`, `selected_version`, `channel`, `mode` and `skip_post_install`,
+  and resolves the clone URL from its own database instead. A pack that is not in the
+  registry still cannot be installed by URL on a stock v4, and you will still see
+  *"Node '<name>@<version>' not found in [...]"*.
 
-  The workaround of putting the URL in `id` instead still works and is unchanged.
+  The change is kept because it is what the documented API asks for and it will start
+  working if Manager implements the field — but it is forward-compatibility, not a fix.
+
+  What works today for an unlisted pack: run ComfyUI with `--enable-manager-legacy-ui`
+  (which registers the legacy route that installs from a URL), `git clone` into
+  `custom_nodes/` and restart, or ask the pack author to publish to the registry.
+  Tracking in #920.
 
 ## [0.11.74] - 2026-08-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@ All notable changes to this project are documented here. This project adheres to
   `repository` field ComfyUI-Manager's API declares for it.
 
   **This does not yet make such an install succeed, and this entry originally said it did —
-  that was wrong.** ComfyUI-Manager v4 accepts the field and then ignores it: its install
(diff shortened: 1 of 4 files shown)


artokun:
**Review found this PR was ALSO inert — the third in a row on #920. Fixed in `0fa83ed`.**

## P0: the advice never fired

The handler read `f?.reason` from `collectRecentTaskFailures`, which emits **`result`**. Proven at runtime against the reporter's exact Manager history shape:

```
record => {"ui_id":"ui-1","kind":"install","id":"ComfyUI-SolAttn_triton",
           "result":"Node 'ComfyUI-SolAttn_triton@nightly' not found in [...]"}
PR-wired (.reason) advice length: 0
correct  (.result) advice length: 679
```

One word, and it made the whole change a no-op.

## P1: the wiring test could not fail — again

It matched `/unlistedGitUrlAdvice\(recentFailures\.map\(/` and **stopped before the field name**, so it passed against `.reason` (broken), `.result` (correct), *and* a garbage field. 3352 tests green with the defect in place.

That is the second time on this issue that a source-text wiring assertion hid an inert fix, and I wrote both. Replaced with an end-to-end test: real Manager history → `collectRecentTaskFailures` → the handler's own join → assert the advice is non-empty. The field-name and garbage-field mutations are now killed.

## P1: the advice sent users to a 404

It said to start ComfyUI with `--enable-manager-legacy-ui` and stopped. For the **only** case this advice appears — an unlisted pack — that is not enough: `get_risky_level` rates it `high+`, and the legacy route then requires `allow_git_url_install` (config default **False**) or answers `404 "A security error has occurred"`.

Following my advice would have been the **third wrong instruction** on this issue. It now names both steps, says the flag *replaces* the v2 API rather than adding to it, and puts the manual clone first as the least-effort route.

## P2

Matcher narrowed to Manager's literal `Node '<id>@<ver>'` prefix, so a custom node's own `ValueError: 'x' not found in [...]` can't be told its git URL is the problem. CHANGELOG narrowed from "v4 ignores it" to "v4's **v2 install handler** ignores it" — v4's legacy module does read the field in one arm.

## Verification note

Four mutations, each confirmed applied **and semantically effective**. The regex one initially reported "survived" — but stripping `Node\s+` left the quoted-`@` requirement intact, so the mutation wasn't actually the widening it looked like. Re-applied in its true loose form before trusting the result. "Survived" and "the mutation did nothing" are indistinguishable in the output, and this is the third time today that distinction mattered.

3354/3354 unit, typecheck clean, vocabulary gate clean.

PR #956, opened 2026-08-10
0.11.83 — the panel tells you what changed when it updates under you (#758)

### LINKED ISSUE #758: No in-panel changelog: users can't see what changed when their install updates under them

artokun:
Releases #758, and restores a release gate I claimed to have added and had not.

Cutting this release is what exposed it: `web/changelog.json` came out still naming 0.11.82 and nothing failed. The gate that exists to catch exactly that had been silently removed by an earlier scripted edit whose replacement never applied. It is now in the file, outside the best-effort catch, and verified to fire against a stale file.

Closes #758


artokun:
**Coordination, not a blocker — land this as-is.**

This branch was cut from `551f813`, so it does not include `15f7320` (#955, merged to main a few minutes ago): the honest-error fix for #920 plus a **correction to the 0.11.75 CHANGELOG entry**.

That matters slightly more than a normal missed commit, because 0.11.75 is published claiming *"Installing a custom node from a GitHub URL actually clones it now (#920)"* and it does not — Manager v4's `do_install` reads only `id`/`selected_version`/`channel`/`mode`/`skip_post_install`, and `params.repository` appears **0 times** across tags 4.2.2, 4.1 and branch draft-v4. #955 retracts that entry in place and explains what actually works.

So the retraction reaches users one release later than it could. Not worth re-cutting this one for — I will take **0.11.84** straight after this merges, and it will carry both.

I have discarded my own local 0.11.83 bump rather than race you for the number. Flagging the general pattern separately: this is the sixth collision today (four release numbers, plus both of us independently building and shipping the same wrong fix for #920 within an hour). I am now claiming issues on the issue thread before starting; claiming the release number the same way would remove this one entirely.

PR #963, opened 2026-08-10
fix(binding): a blank canvas can accept its first node (#833)

artokun:
Draft — claiming #833's remaining half.

Measured on 0.11.86: reads are fine on a 0-node canvas (`graph_outline`, `graph_get_state` both ok), but a write is refused —

```
graph_add_node -> [dirty-mutation-binding-unproven]
"The active tab has unsaved changes and the live canvas carries no identity stamp proving it belongs to this workflow"
```

So an agent asked to build a workflow from scratch cannot place the first node.

#944 established the seal point this needed: `loadGraphData` brackets the swap and its exit edge is a moment where the rebuilt graph and the workflow the store names both agree. Any fix must still fail closed on a rebuild it never bracketed.

Refs #833


artokun:
Closing unmerged — no code. The remaining write-half does not reproduce on 0.11.86 through any real path I could drive; the measurement is on #833.

PR #972, opened 2026-08-10
fix(connect): connecting stops marking the workflow modified (#911)

artokun:
Draft — claiming #911, implementing the fourth option worked out on the issue.

Connecting the Agent flips `activeWorkflow.isModified` to true on a canvas the user has not touched. That is not cosmetic: it is the `*` in the tab title, the close prompt, and what the panel's own dirty-mutation fence reads.

The approach: reuse `clearSpuriousOpenModified()`'s precedent, with the one guard the open-time version does not need —

> read `isModified` BEFORE the capture; clear it only if it was false before and true after.

A workflow the user had genuinely modified stays modified, because it was already true before.

Refs #911


artokun:
Closing unmerged — no code. Measurement disproved the approach this PR was opened to implement; the finding is on #911.

PR #973, opened 2026-08-10
fix(connect): establish the change-tracker baseline so a read stops dirtying the canvas (#911)

artokun:
Draft — #911, second attempt, with the design corrected by measurement.

The first approach (clear `isModified` on a false→true flip around the capture) is retracted: measured, an untouched canvas and a genuine node write have the SAME signature (both false before), so it would erase exactly what #874/#878/#882 protect.

What measurement did show: with a sound baseline, capturing an untouched canvas produces no modification. The dirt comes from the FIRST capture of a session diffing against a baseline that was never established. So establish it — the `clearSpuriousOpenModified()` shape — instead of asserting something false about the flag afterwards.

Refs #911


artokun:
Closing unmerged and reverted — the approach did not work and the measurements are on #911.

PR #974, opened 2026-08-10
fix(#911): find what the first read actually diffs against

artokun:
Draft — #911, third attempt, starting from a diagnosis rather than a candidate fix.

Two designs have failed: clearing the flag around a capture (unsafe — indistinguishable from #882's genuine case) and establishing the baseline at connect (measured, does not hold). My own note said the next attempt should instrument initialState vs activeState around the first read. Doing that.

Refs #911


artokun:
Closing empty — the diagnosis is on #911: the difference is cg-use-everywhere injecting ue_links into extra when the graph serializes, not a panel baseline problem.

Issue #757, opened 2026-08-07
panel_set_widget cannot create dynamic widget slots (Power Lora Loader rgthree lora_N)


artokun:
### What happened / root cause

`panel_set_widget` cannot populate LoRA slots on **Power Lora Loader (rgthree)** — one of the most widely used LoRA nodes in ComfyUI.

A freshly added node reports only these widgets:

```
divider, PowerLoraLoaderHeaderWidget, divider, ➕ Add Lora
```

The actual `lora_1`, `lora_2`, … entries do not exist until the user clicks the node's **➕ Add Lora** button, which is a DOM-only control the agent cannot activate. So:

```
panel_set_widget(node_id=153, widget="lora_1", value={...})
-> Error: Node 153 (Power Lora Loader (rgthree)) has no widget "lora_1"
   (available: divider, PowerLoraLoaderHeaderWidget, divider, ➕ Add Lora)
```

Once a slot exists (user-created, or present in a loaded workflow) setting it works fine — the value is a JSON object like
`{"on": true, "lora": "sub\folder\name.safetensors", "strength": 0.5, "strengthTwo": null}`.
So this is purely about *creating* a slot, not writing one.

### Steps to reproduce

1. `panel_add_node(class_type="Power Lora Loader (rgthree)")`
2. `panel_set_widget(node_id=<new>, widget="lora_1", value={"on":true,"lora":"x.safetensors","strength":0.5,"strengthTwo":null})`
3. Refused — no such widget.

### Exact error (scrubbed)

```
Error: panel_set_widget refused "lora_1" on node 153 (Power Lora Loader (rgthree)):
Node 153 (Power Lora Loader (rgthree)) has no widget "lora_1"
(available: divider, PowerLoraLoaderHeaderWidget, divider, ➕ Add Lora)
```

### Fix

`applied locally: no`
`upstream-only: yes`

Not patched locally — the fix belongs in the panel's node-pack JS where widget mutation is handled, and a local hack there would be reverted by the next panel update (this user has already lost two in-pack patches that way).

**Workaround used:** ask the user to click ➕ Add Lora N times, or fall back to chaining core `LoraLoaderModelOnly` nodes — which loses the stacking UI the user specifically wanted.

[...]

PR #977, opened 2026-08-10
feat(set_widget): create a dynamic widget slot via the node's own control (#757)

artokun:
Draft — #757's remaining half. The refusal improvement shipped in #776; this is the capability itself.

`Power Lora Loader (rgthree)` builds its `lora_1`, `lora_2`, … rows only when its "➕ Add Lora" control is activated, so `panel_set_widget({widget:'lora_1'})` correctly refuses — the slot does not exist yet. The node offers the control; the panel just cannot press it.

Verified installed on this rig, so this is testable end to end.

Refs #757


artokun:
Closing empty — structure probe is on #757; the creator control is a custom widget with a callback, but I did not establish that invoking it out-of-band materialises the row.

Issue #983, opened 2026-08-10
panel_set_widget reports success but rgthree Fast Groups Bypasser toggle remains true


artokun:
### What happened / root cause
panel_set_widget reports success when setting the composite rgthree Fast Groups Bypasser widget RGTHREE_TOGGLE_AND_NAV.toggled=false, but a following panel_query_graph still reports {"toggled":true}. The target ControlNetApplyAdvanced node remains active, so the requested default-off control cannot be reliably established through the panel tool.

### Steps to reproduce
1. Add a Fast Groups Bypasser (rgthree), set matchTitle to an exact group.
2. Call panel_set_widget with widget "RGTHREE_TOGGLE_AND_NAV.toggled" and value false (or full JSON string {"toggled":false}).
3. Call panel_query_graph for that node.
4. It still returns toggled:true.

### Exact error (scrubbed)
No error: panel_set_widget returned a successful response showing {"toggled":false}, followed by panel_query_graph showing {"toggled":true}.

### Fix
Applied locally: no. Upstream-only: yes.
The panel should either persist the composite widget's updated value to the live rgthree node or detect/reject the unsupported mutation instead of returning success. Temporary workaround: set each target node to bypass directly.

### Environment
Windows 11; RTX 4090 24 GB; ComfyUI 0.31.0; comfyui-mcp 0.50.93; panel 0.11.45.

artokun:
**Reproduced on a live canvas, and the mechanism is pinned.** ComfyUI 0.31.1 / frontend 1.48.7, rgthree installed.

```
before                     {"toggled":true}
panel_set_widget →         reports {"toggled":false}, NO warning
read back immediately      {"toggled":false}      ← verification passes here
after rgthree rebuilds     {"toggled":true}       ← reverted
KSampler .mode throughout  0 (active)             ← nothing was ever bypassed
```

So the write is not lost at write time — it lands, and the panel's read-back verification legitimately passes. rgthree then **regenerates that widget from the live group state**, and the regenerated value is the truth. `RGTHREE_TOGGLE_AND_NAV` is a *view*, not storage.

Two consequences worth separating:

- **The success report is not a bug in the verification.** At the moment it checked, the value really was `{"toggled":false}`. Any read-back-based check would have passed. The value's owner reasserts later.
[...]

PR #1005, opened 2026-08-10
spec: #983 — the persistence signal, and why my first implementation of it was reverted

artokun:
Investigation of #983. **No code change** — the reproduction and the fix signal are the deliverable.

## Reproduced (ComfyUI 0.31.1 / frontend 1.48.7, rgthree installed)

```
before                     {"toggled":true}
panel_set_widget →         reports {"toggled":false}, NO warning
read back immediately      {"toggled":false}      ← verification legitimately passes
after rgthree rebuilds     {"toggled":true}       ← reverted
KSampler .mode throughout  0 (active)             ← nothing was ever bypassed
```

The read-back check is **not** at fault: when it looked, the value really was `false`. rgthree regenerates that widget from live group state afterwards.

## The distinguishing signal

My first pass concluded no signal separated this from composites that work. That was wrong. **The node's own serialization** does it — the direct form of "will this survive a save?":

| node · widget | in `serialize().widgets_values`? |
|---|---|
| `Power Lora Loader` · `lora_1` (composite, custom, rgthree) | **yes** |
| `Fast Groups Bypasser` · `RGTHREE_TOGGLE_AND_NAV` (composite, custom, rgthree) | **no key at all** |
| `KSampler` · `steps` (control) | **yes** |

Both rgthree, both `type: "custom"`, both composite objects — and they land on opposite sides. Neither widget type nor compositeness nor authorship separates them; serialization does, with no node-specific methods involved.

## Why no code here

The check belongs in the widget-write verification path that every `panel_set_widget` goes through. That is not somewhere to land a change without a full review pass, and I would rather hand over a validated signal than a rushed edit to it.


artokun:
Closing as superseded. #983 is fixed and merged in #1106.

This branch was the record of the first reverted attempt and its analysis. That analysis was right about what it ruled out and is preserved on the issue; what it did not consider is that the problem never needed a general persistence signal at all.

The rgthree pack is installed on this machine, so the question was answerable from its own source: `BaseFastGroupsModeChanger` declares `serialize_widgets = false`, the refresh loop overwrites `widget.toggled` from `group.rgthree_hasAnyActiveNode`, and only the widget's own `toggle()` changes anything because it also calls `doModeChange()`. Three facts, any one sufficient, none requiring a probe on the write path.

The fix is a node-type-keyed refusal in the same shape as the LTXDirector (#314) and PromptRelay (#506) routes already in that function — which is exactly the pattern this branch's own conclusion was reaching for and did not find.

Issue #988, opened 2026-08-10
Scoped panel_run batch_count duplicates randomize seed and returns cached images


artokun:
### What happened / root cause
panel_run with to_node_id plus batch_count > 1 reports multiple accepted prompt_ids, but a KSampler using control_after_generate=randomize is serialized with the same seed for every scoped batch item. The first prompt renders normally; later prompts complete from cache in ~0.2-0.3s, return the same filename, and produce identical pixels. This forced manual seed changes and one-at-a-time runs.

The affected runtime path is graph_run -> dispatchScopedRun in comfyui-mcp-panel. The run result already flags the seed/control_after_generate inputs as volatile and excluded from drift coverage, but batch dispatch does not advance them between accepted prompts.

### Steps to reproduce
1. Open a workflow with KSampler control_after_generate=randomize and a SaveImage output.
2. Call panel_run({to_node_id:<SaveImage id>, batch_count:2 or 3}).
3. Observe distinct prompt_ids are returned.
4. First prompt renders normally; remaining prompt(s) finish almost instantly from cache.
5. All events reference the same output filename and identical image.

Observed twice:
- batch_count:3 -> first 22s, next 0.2s/0.3s, all same image.
- batch_count:2 -> first 19s, next 0.3s, same image.

### Exact error (scrubbed)
No thrown error. Tool reports queued:true with multiple prompt_ids and drift_coverage.partial for:
- "42 control_after_generate"
- "42 seed"
The failure is silent duplicate output.

### Fix
Applied locally: no (attempt reverted)
Upstream-only: yes

One focused local attempt changed the scoped branch to invoke dispatchScopedRun batch:1 in a loop, awaiting each call so beforeQueued could run between serializations. 114 related unit tests passed, but live verification still produced the same cached duplicate, so the patch was reverted.

Precise change needed: the scoped batch path must explicitly advance/apply beforeQueued volatile inputs between prompt serializations (or use the frontend queue API in a way that does), then recompute/serialize the prompt with the updated seed for each item. Add a regression test asserting distinct serialized seeds/content hashes for control_after_generate=randomize across scoped batch items, plus distinct output filenames.

Temporary workaround: set control_after_generate=fixed, manually set a new seed, and call panel_run batch_count:1 for each image.

### Environment
Windows 11
ComfyUI 0.30.0
ComfyUI frontend 1.47.12
GPU NVIDIA GeForce RTX 4070 SUPER 12 GB
[...]

Issue #984, opened 2026-08-10
panel_get_errors reports "clean" when loader nodes reference genuinely-missing model files (misses missing_models the frontend shows as red)


artokun:
### What happened / root cause
On a freshly-loaded workflow (MiniMax H3 reference-to-video, Pixaroma pack) the ComfyUI frontend showed **6 red nodes**. `panel_get_errors` returned fully clean:

```
{ "last_execution_error": null, "node_errors": null,
  "note": "no errors recorded since the last execution start" }
```

No `missing_models` / `missing_media` block was surfaced. But the graph has 4 loader nodes pointing at model files that are **genuinely absent** from the model folders (verified independently via `list_local_models` and `get_system_stats action:"health"`):

- UNETLoader → `h3\minimax_h3_ref2va_pruned_int8_convrot.safetensors` (not in diffusion_models)
- CLIPLoader → `qwen3vl_32b_minimax_h3_nvfp4_awq.safetensors` (not in text_encoders)
- VAELoader → `minimax_h3_video_vae_fp16.safetensors` (not in vae)
- VAELoader → `minimax_h3_audio_vae_fp32.safetensors` (not in vae)

Per the tool's own contract it should surface these missing models *on load, before any queue attempt*. It surfaced nothing, so an agent using `panel_get_errors` as the source of truth reports "no errors" while the user is staring at red nodes. The other 2 red nodes were two empty `PixaromaLoadImageMini` reference-image slots. So user-visible count (6) vs tool report (0) fully diverged.

### Steps to reproduce
1. Load a workflow whose UNETLoader/CLIPLoader/VAELoader widgets name `.safetensors` files that are NOT present in the corresponding models/ subfolder (subfolder-qualified paths like `h3\foo.safetensors` may be relevant).
2. Nodes render red in the frontend.
3. Call `panel_get_errors` before any queue attempt → returns `last_execution_error:null`, `node_errors:null`, no `missing_models`.

### Exact error (scrubbed)
No error thrown — the defect is a **false-clean / empty** result. (No secrets in payload.)

### Fix
applied locally: no
upstream-only: yes — needs maintainer investigation of the load-time missing-model detection path (I did not attempt a blind patch of the error-detection logic without the source checkout; no clean minimal fix identified). Suggested direction: cross-check each loader widget value against the frontend/available-model list on load and populate `missing_models`, rather than relying solely on the server validation map (which appears empty pre-queue for these loaders, possibly due to subfolder-qualified paths).

[...]

Issue #981, opened 2026-08-10
panel_refresh_nodes does not rehydrate missing subgraph nodes after classes are registered


artokun:
### What happened / root cause
After installing custom-node classes and restarting ComfyUI, the server's /object_info endpoint reported all formerly missing classes as registered. panel_refresh_nodes returned {ok:true, refreshed:true}, but panel_get_errors immediately continued reporting the same three missing_node_types inside workflow subgraph definitions.

The frontend appears to refresh node definitions without rehydrating already-instantiated missing-node placeholders in subgraphs, and/or panel_get_errors keeps a stale missing-node cache.

### Steps to reproduce
1. Load a workflow containing missing classes inside subgraphs.
2. Install/update the packs and restart ComfyUI.
3. Confirm /object_info/<class> returns the class.
4. Call panel_refresh_nodes; it returns success.
5. Call panel_get_errors; missing_node_types still lists the registered classes.

Classes reproduced:
- MiniMaxChunkFeedForward
- MiniMaxLowVRAMAttention
- OllamaImageList_CLIPGenerateText

### Exact evidence
Server /object_info: registered=true for all three.
panel_refresh_nodes: {"ok":true,"refreshed":true}
panel_get_errors: missing_node_types still contains all three.

### Fix
applied locally: no
upstream-only: yes

After refreshing /object_info, rehydrate missing-node placeholders in root and subgraph definitions, then recompute missing_node_types against the new definitions. If live rehydration is unsafe, return requires_reload:true instead of claiming a complete refresh and expose a safe workflow reload that preserves current widget values and unsaved state.

Temporary workaround: manually hard-refresh the ComfyUI browser tab, then rebind the workflow.

### Environment
Windows 11
ComfyUI 0.30.0 (v0.30.0-12-g6f7cd7fc)
ComfyUI frontend 1.48.6
ComfyUI-Manager 3.41 legacy
comfyui-mcp 0.50.60
comfyui-mcp-panel 0.11.41
Backend: Codex

artokun:
Fixed and released in **0.11.95**.

## What was happening

Both answers you got were correct, which is why neither of them helped. Measured on
ComfyUI 0.31.1 / frontend 1.48.7 by loading a workflow against an absent class,
registering that class exactly as an install makes it appear, and re-reading the node
that was already on the canvas:

```
registered in LiteGraph   : true
node constructor nodeData : false    (unchanged)
node constructor title    : null     (unchanged)
[...]

Issue #976, opened 2026-08-10
panel_set_widget reports exception after successful MiniMaxH3Director duration write


artokun:
### What happened / root cause
While editing the live canvas from the sidebar panel, `panel_set_widget` successfully changed MiniMaxH3Director node 2693 widget `duration` from 8 to 12, but returned a `write_warning` saying an exception was thrown while applying the write: `Cannot read properties of undefined (reading 'options')`. The tool also verified the requested value was in effect by read-back and did not roll back.

This looks like a panel/orchestrator widget side-effect refresh path assuming an options object exists after applying a widget write. The actual value persisted, but dependent side effects may have been skipped.

### Steps to reproduce
1. Open workflow `DaSiWa_MiniMaxH3_TurboV4_600_8Step_Crow_GrandCanyon`.
2. Call `panel_set_widget` on MiniMaxH3Director node 2693, widget `duration`, value `12`.
3. Observe returned warning despite successful read-back.
4. Re-read node 2693 with `panel_query_graph`; `duration` is 12.

### Exact error (scrubbed)
`write_warning`: `an exception was thrown while applying the write (Cannot read properties of undefined (reading 'options')); the requested value IS in effect -- verified present by read-back -- and was NOT rolled back. Side effects the write would normally trigger (refreshing dependent widgets, previews, thumbnails) may not have run or completed; inspect the node if dependents look stale.`

### Fix
applied locally: no
upstream-only: yes

Suggested fix: guard the widget side-effect/refresh path after a successful write so it does not assume `options` is present for this widget/custom node shape. If side-effect refresh cannot run, return a structured non-fatal warning without surfacing a raw JS exception.

### Environment
Windows 11, ComfyUI 0.31.0 local, NVIDIA GeForce RTX 5090 32 GB VRAM, CUDA 13.0, torch 2.13.0, python 3.10, comfyui-mcp 0.50.82, panel 0.11.51.

artokun:
**First pass, static: the throw does not come from an unguarded read in the panel.**

Searched every `.options` access in the panel and its libs. All of them are already optional-chained or explicitly type-checked before use:

```
asset-staleness.js        w.options?.values ; if (!w.options || typeof w.options !== "object") w.options = {}
control-after-generate.js const opts = w?.options ; Array.isArray(w?.options?.values)
input-asset.js            widget.options?.values ; same guard before writing
node-widget-materialization.js  widget.options && typeof widget.options === "object" ? … : null
```

[...]

PR #1011, opened 2026-08-10
fix(refresh): stop claiming a complete refresh that did not rehydrate anything (#981)

artokun:
Claims #981.

**Reported:** after installing the packs and restarting, `/object_info` reports all three formerly-missing classes as registered. `panel_refresh_nodes` returns `{ok:true, refreshed:true}`, and `panel_get_errors` immediately still lists the same three in `missing_node_types` — the classes being inside **subgraph definitions**.

The reporter offers the fix as an either/or, and the second half is the interesting one:

> If live rehydration is unsafe, return  instead of claiming a complete refresh

That is the same defect class as #976, #984 and #988 this session: a reply asserting more than the mechanism did. `refreshed:true` after a refresh that rehydrated nothing is exactly that.

**Two candidates, and they need separating by measurement before either is written:**
- the frontend refreshes definitions but leaves already-instantiated placeholders alone, so the nodes stay placeholders
- `panel_get_errors` reads a missing-node store populated at load and never re-evaluated

The second is very plausible given what I measured on #984 — the `missingNodesError` store is a load-time snapshot, and `collectMissingNodeTypeReasons` reads it. If that store is the source, refreshing definitions would never clear it no matter how well rehydration worked.

- [ ] reproduce with a class that is genuinely absent, then registered
- [ ] establish which of the two is reporting stale — the placeholders or the store
- [ ] fix, or report `requires_reload` honestly if live rehydration is not safe
- [ ] codex review
- [ ] browser verification

### DIFF
--- a/browser_tests/unit/stale-placeholders.test.mjs
+++ b/browser_tests/unit/stale-placeholders.test.mjs
@@ -131,7 +131,13 @@ test("#981 the note credits what the refresh DID do, and names the one thing tha
     { node_id: "2", type: "MiniMaxChunkFeedForward" },
     { node_id: "3", type: "MiniMaxLowVRAMAttention" },
   ]);
-  assert.match(note, /definitions ARE now current/, "the refresh is not described as having failed");
+  // codex r2: the predicate establishes that the CLIENT can instantiate the class now —
+  // not that the definition is current, not that this refresh registered it, and nothing
+  // at all about what the backend will do with the prompt. The note may claim only that.
+  assert.match(note, /This page can NOW instantiate/, "the refresh is not described as having failed");
+  assert.doesNotMatch(note, /definitions ARE now current/, "…but does not claim currency it never checked");
+  assert.doesNotMatch(note, /will still fail at queue time/, "nor a backend outcome it never measured");
+  assert.match(note, /does not serialize the values its class expects/, "what it CAN say about the dead node");
   assert.match(note, /MiniMaxChunkFeedForward/, "names the classes");
   assert.match(note, /still a PLACEHOLDER/, "and what did not happen");
   assert.match(note, /does not rehydrate nodes that were created while it was unknown/, "why");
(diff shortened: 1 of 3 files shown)


artokun:
## Codex review: SHIP (round 5)

Four NO-SHIP rounds got here. Each one removed a claim the mechanism did not establish.

**r1 — the detector was wrong.** `!constructor.nodeData` finds a node with no *backend*
definition, which is not the same as a placeholder: frontend-only nodes have none either.
Verified live on the running install, and **four false positives were confirmed** — `Note`,
`Reroute`, `PrimitiveNode`, `MarkdownNote` were all reported, so a canvas with a single
Note on it would have demanded a workflow reload after every refresh. The discriminator is
now membership in the frontend's own load-time missing-node record, and the registration
lookup is the **client** registry rather than `/object_info`: the server having a
definition does not mean this page can instantiate the class.

**r2 — the fix was dead on arrival.** Codex asked for a caller trace instead of an
assurance. `refresh_nodes` returns `{ok:true, refreshed:true}` on the success path, which
discards every other field — so the warning was produced and no caller could ever see it.
Forwarded explicitly, with `ok` and `refreshed` left true, because the reload flag is about
the canvas, not a failed refresh.

**r3, r4 — the note claimed four things it never tested**, one at a time: that the
definitions were *current*, that the prompt would *fail at queue time*, that the page could
*instantiate* the class (a registry entry is not a successful `createNode`), and that every
reported node had *no widgets* (only `nodeData` is tested per node; the widgets observation
came from one instrumented instance). Each is gone, and each is pinned by a `doesNotMatch`
assertion that fails if the wording comes back.

Every assertion mutation-verified. Full suite 3606/3606.

Deliberate limitation, documented in the module: a genuine placeholder whose miss the
frontend never recorded is skipped. Under-reporting costs a diagnosis; the over-reporting
it replaced sent people reloading workflows that were fine.

Issue #995, opened 2026-08-10
0.11.41: panel_graph_outline false root-workflow-uuid-mismatch after tab switch/reconnect


artokun:
### What happened / root cause
Immediately after the panel announced a manual switch from `KREA2_023_BUILD_BODY_ANCHOR_LOCAL_CANDIDATE` to `KREA2_031_BUILD_B_INJECTION_TEMPLATE_V1`, `panel_graph_outline` refused the active canvas with `[root-workflow-uuid-mismatch]`. `panel_list_workflows` independently confirmed the new workflow active at `workflows/KREA2_031_BUILD_B_INJECTION_TEMPLATE_V1.json` with workflow UUID `6548a06c-8244-4523-b44c-d03d505d04e7`, while its `last_open` receipt still named the prior KREA2_023 tab. The active workflow was modified, so no re-open/update was performed automatically.

Root cause matches existing panel issue #817: ComfyUI reuses `app.graph` across tab switches and leaves the previous workflow identity tag in `graph.extra`. Panel 0.11.41 lacks the later content-proven rebind escape, so the stale tag is stickier than an untagged but otherwise identical canvas.

### Steps to reproduce
1. On panel 0.11.41, keep multiple saved workflows open.
2. Switch manually from one populated workflow tab to another, with a reconnect/session resume in the same interval.
3. Let the panel announce the newly active workflow.
4. Call `panel_graph_outline`.
5. Observe `[root-workflow-uuid-mismatch]` even though `panel_list_workflows` reports the intended workflow active.

### Exact error (scrubbed)
`Error: [root-workflow-uuid-mismatch] The live canvas carries a different workflow's identity tag than the active workflow — a load, tab switch, or reconnect left a stale tag behind, so this command was NOT applied.`

### Fix
Applied locally: no. Upstream-only/current release already fixed: yes. The active workflow has unsaved modifications and the project permission boundary requires an explicit workflow-bound GO before updating/reloading/re-opening, so no risky local mutation was made.

The precise upstream fix is commit `4964486f0c7acb1146e3f6bb34987a2e02463953` (`fix(binding): panel_graph_outline rejects the active workflow after a tab switch (#817) (#843)`), shipped in panel 0.11.51. It adds `contentProvesActiveWorkflow` to `resolveGraphRootUuidRebind`, using the same clean/exclusive/current-state content proof as the untagged-root seal. Current upstream is 0.11.89.

Local pure-function verification:
[...]

Issue #565, opened 2026-08-03
Binding guard throws a false "live graph is out of sync" on every NEW blank workflow (stale root extra.comfyui_mcp.workflow_uuid)


artokun:
### What happened / root cause

Every graph command routed through `assertGraphBoundToActiveWorkflow()` throws a false `out of sync` error on a **brand-new blank workflow tab**. The tab is unreachable: `panel_graph_outline` (and any edit) fails permanently, while `panel_query_graph` on the same tab succeeds and correctly reports `0 nodes` — so the two read paths disagree, which is what made it look like a genuine desync.

Root cause is the **UUID branch**, `graphRootWorkflowUuidMismatches()`:

- ComfyUI reuses the `app.graph` object across workflow tabs, and its clear/configure does **not** reset `graph.extra`.
- So a newly created blank tab inherits the PREVIOUS workflow's `extra.comfyui_mcp.workflow_uuid` on the live root, while its own workflow object mints a **fresh** uuid via `workflowStableUuid()`.
- `rootUuid !== activeWorkflowUuid` → the guard fires, on a canvas with **zero nodes on both sides**, where there is no workflow content that could possibly be confused.

The error message was also undiagnosable — it names no predicate, so all four branches produce identical text and the only available response is a blind retry loop (I burned several reload/re-open cycles before patching in a reason code to find out which branch was firing).

A second, latent instance of the same class exists in `graphRootMismatchesActiveWorkflow()`: with both node arrays empty it falls through to the `graphShape()` surface comparison, which deliberately preserves present-vs-absent for optional surfaces (`groups`, `reroutes`, `config`, `extra`). A brand-new / just-cleared workflow's ChangeTracker state routinely OMITS those while a live `graph.serialize()` emits them as present-but-empty — plus the panel's own `extra.comfyui_mcp` tag only exists on the live side. That manufactures a false mismatch on any empty canvas.

### Steps to reproduce

1. Have at least one workflow open (so the root carries a uuid tag).
2. `panel_new_workflow` → a blank tab becomes active.
3. `panel_graph_outline` → throws `The live graph is out of sync with the active workflow: the workflow reports 0 node(s) …`
4. `panel_query_graph {fields:'ids'}` on the same tab → succeeds, `total: 0`.
5. `panel_open_workflow` and `panel_set_workflow_target({mode:'current'})` do **not** recover it; a frontend reload only "fixes" it by switching to a different, non-empty tab.

[...]

PR #1016, opened 2026-08-10
fix(binding): a tab switch onto a MODIFIED workflow still refuses on a stale tag (#995)

artokun:
Claims #995. Draft opened before starting.

The report concludes it is already fixed upstream by #817/#843 (`contentProvesActiveWorkflow`
in `resolveGraphRootUuidRebind`). **Measured on current main, that is true for a clean tab
and false for the reported one.** The report's own state says "the active workflow was
modified":

```
dirty=false   contentProof=true    verdict=rebind      <- the fix they credit
dirty=true    contentProof=false   verdict=conflict    <- the reported state
```

`rootContentProvesActiveWorkflow` requires a CLEAN tab, so on a modified workflow the
escape never fires and `panel_graph_outline` is still refused with
`[root-workflow-uuid-mismatch]`.

Next step is to establish whether the clean-tab requirement can be narrowed without
weakening the fence — the concern behind it (#545) is that a dirty tracker can LAG the
canvas, and a lagging snapshot can only make an equality proof fail, not falsely succeed.
Whether that reasoning survives contact with the duplicate-content case is what has to be
measured before any code changes.

### DIFF
--- a/browser_tests/unit/canvas-graph-divergence.test.mjs
+++ b/browser_tests/unit/canvas-graph-divergence.test.mjs
@@ -1373,6 +1373,11 @@ test("graphCommandBindingBar: reads get the reduced DISPATCH bar (they re-assert
     assert.deepEqual(graphCommandBindingBar(cmd), {
       includeBaselineReadGuard: false,
       requireDirtyMutationBinding: false,
+      // #995 — the opt-IN that lets a classified read past a STALE TAG on a canvas whose
+      // content equals the active workflow's own snapshot. Set here and nowhere else, so
+      // this list is the whole surface that can reach it; a fence call that omits it (or
+      // any future caller) gets no bypass.
+      staleTagReadBypass: true,
     });
   }
 });
--- a/browser_tests/unit/graph-binding.test.mjs
+++ b/browser_tests/unit/graph-binding.test.mjs
@@ -37,6 +37,8 @@ import {
   resolveGraphRootUuidRebind,
   sealProvenRootBinding,
   rootContentProvesActiveWorkflow,
+  rootContentProvesActiveWorkflowDespiteEdits,
+  contentProofExclusiveAmongOpen,
   serializedStateProvenEmpty,
 } from "../../web/js/lib/graph-binding.js";
 import {
@@ -277,6 +279,8 @@ function buildDirtyStaleRouteHarness({
     "postReconnectSettleWindow",
     "sealProvenRootBinding",
     "rootContentProvesActiveWorkflow",
+    "rootContentProvesActiveWorkflowDespiteEdits",
+    "contentProofExclusiveAmongOpen",
     "graphRootMatchesState",
     "sameWorkflowObject",
     "app",
@@ -300,6 +304,8 @@ function buildDirtyStaleRouteHarness({
     () => postReconnectWindow === true,
     sealProvenRootBinding,
     rootContentProvesActiveWorkflow,
+    rootContentProvesActiveWorkflowDespiteEdits,
+    contentProofExclusiveAmongOpen,
     graphRootMatchesState,
     sameWorkflowObject,
     { graph: rootB, extensionManager: { workflow: { openWorkflows } } },
--- a/browser_tests/unit/stale-tag-dirty-tab.test.mjs
+++ b/browser_tests/unit/stale-tag-dirty-tab.test.mjs
@@ -199,7 +199,7 @@ test("#995 (codex P1) source guard: READS only, and NOTHING is written", () => {
   // onto the OTHER workflow and the wedge would follow the user back to it. So the flag
   // is cleared for one call and the tag is left alone.
   const src = readFileSync(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url), "utf8");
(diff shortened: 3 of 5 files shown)


artokun:
## The constraint any fix has to satisfy, found before writing one

Relaxing `rootContentProvesActiveWorkflow`'s clean-tab requirement is not safe on its
own, and the reason is on the *other* side of the proof.

Exclusivity is what stops a content match from proving the wrong canvas — two separately
open tabs can carry byte-identical state, and equality alone cannot tell which one the
root belongs to. The enumeration that establishes it does this
(`comfyui-mcp-panel.js:5727`):

```js
if (other.isModified === true) continue; // unprovable — not evidence of ambiguity
```

**Dirty twins are skipped.** So if the active side stopped requiring clean, a dirty twin
holding identical content would be invisible to the ambiguity check, and the panel could
stamp the active workflow's identity onto a root that is actually the twin's canvas —
wedging THAT tab the moment the user switched to it. That is the mirror-image failure the
#833 analysis describes, and it is worse than the refusal being fixed.

A rule that could hold, to be measured rather than assumed:

- the root serializes EQUAL to the active workflow's own current tracker state (which for
  a dirty tab reflects its captured edits), **and**
- every OTHER open workflow is readable and provably different — with any other workflow
  being dirty making the whole thing unprovable, since a lagging tracker cannot establish
  that its canvas differs either.

That is strictly more conservative on the twin side than today while relaxing the active
side, and it covers the reported case (one modified tab among otherwise clean ones).

**What has to be measured before any of it is written:** whether a dirty workflow's
`changeTracker.activeState` actually equals the canvas right after a tab switch. If it
lags there, the proof simply never fires and the relaxation buys nothing — which is worth
knowing before changing a fence.

artokun:
## Live measurements, and one of them invalidates my own test route

Driving the real install (ComfyUI 0.31.1 / frontend 1.48.7) rather than reasoning.

**1. The stale tag is real and it does not decay.** Across three tab switches the root
tag never changed — `e66e531b` while workflow X was active, still `e66e531b` while Y was,
still `e66e531b` on the way back. That is the reported condition, standing indefinitely
rather than for a window.

**2. A dirty workflow's tracker state DOES equal its canvas at the switch moment**, so
relaxing the clean-tab requirement would buy something real rather than firing on
nothing:

```
B dirty, before switching away    rootMatchesTrackerState=true
switched to A                     (A active)
switched BACK to dirty B          rootMatchesTrackerState=true
…1.5s later                       rootMatchesTrackerState=true
```

**3. But my switch route is not the user's.** Switching with `store.openWorkflow(Y)`
programmatically left the canvas holding the PREVIOUS graph while the store reported Y
active and clean — measured against disk:

```
store says active   Anima Wojak Batch.json   isModified=false
tracker state       matches the file on disk
LIVE CANVAS         does NOT                 (different node ids and types entirely)
```

Stable at 4 seconds, so not a settling window. With the canvas genuinely holding another
workflow's graph, the panel refusing `panel_graph_outline` is **correct** — reading it
would have returned the wrong workflow. So this route cannot be used to judge the fence;
a user clicking a tab goes through the workflow service, which repaints. **The next
measurement has to drive the switch through the UI**, not the store API.

**4. The constraint from the previous comment still holds** and is now the shape of the
fix: exclusivity skips dirty twins (`comfyui-mcp-panel.js:5727`), so relaxing the active
side must be paired with making any other dirty workflow render the proof unprovable.

Nothing has been written to the user's workflows. One probe did nudge a node in a saved
workflow and leave a draft; the canvas was restored from disk and the workflow reads
clean and unmodified again.

artokun:
## Codex review: SHIP (round 3)

Two NO-SHIP rounds, and the first one changed the design.

**r1 — the relaxation could re-tag a different canvas.** My argument that "a lagging
tracker can only cause a false negative" was circular: it holds only once you know the
root IS the active canvas, which is what the branch was trying to establish. The
counterexample — A and B hold the same graph G, A is edited to H so its tracker lags at G,
the root is still B's canvas holding G, and B is absent from the `openWorkflows` snapshot —
makes equality a false POSITIVE, and the re-stamp would have moved the wedge onto B
permanently.

So the escape no longer rebinds. It clears the mismatch for one read and writes nothing.
`resolveGraphRootUuidRebind` keeps the clean-only proof.

**r2 — the read gate was default-permit.** `requireDirtyMutationBinding !== true` gave the
bypass to any caller that omitted the flag, including two fence calls that pass no options
at all. It is now a positive `staleTagReadBypass`, false by default at the fence and set in
exactly one place — the read branch of `graphCommandBindingBar` — with a test pinning that
it appears once, so a future fence call has to be classified read-only before it can reach
it.

**r2 also — the comments claimed ownership.** The branch establishes serialized-content
equality against a tracker snapshot, and in the A/B sequence the returned graph equals A's
snapshot while being B's mounted canvas. The wording says exactly that now.

Two things the existing suite caught that reasoning did not: an EMPTY canvas cannot
identify itself (every blank graph serialises alike — the #565 gate), and
`sealProvenRootBinding` must keep the clean-only bar, because writing a tag onto an
untagged root is a different decision from re-stamping one.

Twelve tests, every guard mutation-verified. Full suite 3631/3631.

Issue #982, opened 2026-08-10
panel_set_widget cannot validate subgraph widget after restart despite bound live graph


artokun:
### What happened / root cause
panel_set_widget cannot edit a widget exposed by a subgraph after a ComfyUI restart. Live graph reads work and panel_set_workflow_target reports graph_binding:"bound", but panel_set_widget resolves the exposed outer widget to inner VAELoader node 122 and refuses because object_info is allegedly unavailable. The ComfyUI API is healthy and direct /object_info/VAELoader is reachable, so the panel/orchestrator backend-definition cache is stale after reconnect.

### Steps to reproduce
1. Restart ComfyUI while a workflow containing subgraphs remains open.
2. Rebind with panel_set_workflow_target({mode:"current"}); it reports bound/already_current.
3. Call panel_refresh_nodes; it reports ok:true, refreshed:false.
4. Call panel_set_widget on the outer subgraph node for an exposed VAELoader combo.
5. It resolves to the inner node but refuses with object_info unavailable.

### Exact error
Cannot set widget on node 122 ("VAELoader"): cannot verify the node type against the ComfyUI backend (object_info is unavailable — the backend is unreachable or the fetch failed). Refusing to write rather than trust a possibly-stale node cache (#458).

### Fix
applied locally: no
upstream-only: yes

Precise change needed: after restart/reconnect, rehydrate the object_info cache used by mutation validation before reporting graph_binding:"bound"; panel_refresh_nodes should force that cache when writes are fenced by unavailable definitions. If the cache cannot be restored, report requires_hard_refresh explicitly.

Temporary workaround: manually hard-refresh the ComfyUI browser tab, rebind, then retry panel_set_widget.

### Environment
Windows 11
ComfyUI 0.30.0 (v0.30.0-12-g6f7cd7fc)
ComfyUI frontend 1.48.6
comfyui-mcp 0.50.60
comfyui-mcp-panel 0.11.41
Backend: Codex

Related (same root cause): #981

artokun:
Additional duplicate report confirms this on Windows 11 / ComfyUI 0.30.0 / frontend 1.48.6 with comfyui-mcp 0.50.60 and panel 0.11.41: `panel_refresh_nodes` did not recover the unavailable `object_info` validation cache after restart. Ctrl+Shift+R, rebind, and retry immediately allowed the resolved inner VAELoader change; the workflow then rendered successfully, `panel_get_errors` returned 0, and it saved normally.

[...]

PR #1018, opened 2026-08-10
fix(set-widget): stop refusing a write for object_info that a reconnect never restored (#982)

artokun:
Claims #982. Draft opened before starting.

After a ComfyUI restart with a subgraph workflow open, `panel_set_widget` resolves the
exposed outer widget to its inner node and then refuses:

```
cannot verify the node type against the ComfyUI backend (object_info is unavailable —
the backend is unreachable or the fetch failed). Refusing to write rather than trust a
possibly-stale node cache (#458).
```

…while the API is healthy, `/object_info/VAELoader` is reachable, reads work, and
`panel_set_workflow_target` reports `bound`. The reporter also observed
`panel_refresh_nodes` answering `ok:true, refreshed:false`, so the refresh did not restore
whatever the write path consults.

First step is to establish WHAT the write path actually reads and why it is empty after a
reconnect — the refusal names `object_info`, but that is the panel's phrasing, not
necessarily the source it consulted.

### DIFF
--- a/browser_tests/unit/object-info-oracle.test.mjs
+++ b/browser_tests/unit/object-info-oracle.test.mjs
@@ -156,7 +156,11 @@ test("#982 (codex) the fallback is consulted ONLY when the client returned nothi
   // answer — so the fallback must never be able to override an answer the client gave.
   const src = readFileSync(new URL("../../web/js/lib/object-info-oracle.js", import.meta.url), "utf8");
   const clientBlock = src.slice(src.indexOf("if (typeof getNodeDefs === \"function\")"), src.indexOf("// SECOND TRANSPORT"));
-  assert.match(clientBlock, /if \(usableDefs\(defs\)\) return \{ defs, failures \};/, "a usable client answer returns");
+  assert.match(
+    clientBlock,
+    /if \(usableDefs\(defs\)\) return \{ \[CACHE_OUTCOME\]: true, defs, failures \};/,
+    "a usable client answer returns",
+  );
   // …and the measured equivalence is recorded rather than assumed.
   assert.match(src, /MEASURED on ComfyUI 0\.31\.1 \/ frontend 1\.48\.7/, "the comparison is on the record");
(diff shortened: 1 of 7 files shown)


artokun:
## Codex review: SHIP (round 4)

Three NO-SHIP rounds, each finding something the previous version got wrong.

**r1 — the fallback could widen authorization.** If `api.getNodeDefs()` filtered the
schema, falling back to the raw route would authorize against a broader one. Measured both
routes back to back on the live install: 4183 types each, no type present in only one,
identical per-type field sets for the sampled types — recorded as evidence for that build
with the limit stated, not as a contract. Also: untrusted failure text was interpolated
verbatim into a caller-visible error, and the observation was module state a concurrent
request could overwrite.

**r2 — an empty map was treated as an absence.** A client that deliberately filters could
express deny-all as `{}`, and the raw route would have overruled it. `{}` is now terminal
and fails closed without a second attempt. Also: the burst cache coalesces reads, so a
second write joining a FAILED read would have refused while naming no routes at all — the
outcome now rides through the cache, and the usability test still inspects the payload so a
failure can never be cached as a success.

**r3 — a node type can be named `defs`.** The wrapper was detected structurally, so a bare
schema containing that type was misread. When its definition is empty the schema stopped
being cached entirely, which is the #716 re-fetch-per-write symptom returning. Detection is
a `Symbol.for` tag now. Also: sanitizing ran over every entry of an arbitrarily long array
before slicing, and one comment claimed `String(Symbol("x"))` throws, which it does not.

**r4 — SHIP**, with one correction taken: the slice-first order is not output-identical for
arbitrary caller input, only for what this module produces.

Twenty-one tests, every guard mutation-verified.

**What I could not do:** reproduce the original report. It needs a ComfyUI restart *and*
the panel's own tool surface, which this client cannot reach. The change is aimed at the
two defects visible in the report — one transport, and a refusal that asserted a cause it
never established — not at a reproduced mechanism.

Issue #978, opened 2026-08-10
panel_save_workflow save-as leaves agent fenced to previous workflow instance


artokun:
### What happened / root cause
After a successful `panel_save_workflow` with a new `name` (Save-As), the live canvas identity becomes unreadable. Subsequent graph tools fail with `workflow instance mismatch` fenced to the previous instance id. `panel_set_workflow_target({mode:""current""})` and `panel_open_workflow` could not rebind the fence on panel 0.11.38.

### Steps to reproduce
1. Open an unsaved Untitled workflow and build a graph with panel_* tools
2. Call `panel_save_workflow({name:""Ref-Image-to-Prompt-Grok-style""})`
3. Save succeeds (`saved_as: true`) but reply notes identity could not be read (needs panel 0.11.45+)
4. Any further panel_graph_outline / panel_set_workflow_target fails with workflow instance mismatch

### Exact error (scrubbed)
`workflow instance mismatch: this command targets a different workflow than the active canvas`
Previous instance: fcfec679-04a0-4084-b6ab-bcfe837372e9
panel_open_workflow: `workflow_open could not prove that the active canvas was rebound`

### Fix
applied locally: no
upstream-only: yes
Panel note already says 0.11.45+ reports new workflow identity on save reply. Likely need: after save-as, re-derive fence from the newly saved workflow identity without requiring a full canvas re-read that the fence blocks; and/or make panel_open_workflow always succeed at fence re-derive even when proof is partial.

### Environment
OS Windows 11 / ComfyUI 0.30.2 / frontend 1.47.12 / RTX 5070 12GB / comfyui-mcp 0.50.93 / panel 0.11.38

artokun:
Additional reproduction on panel 0.11.38 with comfyui-mcp 0.50.93: after Save As from an already saved workflow, both `panel_graph_outline` and `panel_get_errors` were rejected by the stale-instance fence. `panel_open_workflow` and `panel_set_workflow_target({mode:"current"})` did not restore the binding. Environment: Windows 11, ComfyUI 0.30.0, frontend 1.47.12, RTX 4070 SUPER.

artokun:
Fixed and released in **0.11.99** — though the fix is a diagnosis rather than the heal I
first wrote, and the reason is worth having.

## Two fences, and re-targeting only clears one

The **command** fence compares a command's stamp against the active workflow's uuid. A
Save-As makes a different workflow active, so refusing is correct — and since 0.11.45 the
[...]

Issue #941, opened 2026-08-10
panel_save_workflow (save-as) leaves session fenced to the old workflow instance; panel_open_workflow cannot recover


artokun:
### What happened / root cause

`panel_save_workflow({name: "<new name>"})` performed a Save-As correctly — the new file was written and the original was left intact (`saved:true, saved_as:true, copied_from:..., original_on_disk:true`). But the panel then could NOT report the new workflow's identity on the reply, so the orchestrator's workflow fence stayed bound to the PREVIOUS workflow instance id. Every subsequent `panel_*` graph call is refused with a workflow-instance mismatch.

The tool's own error text says the read is only returned on the reply from panel 0.11.45+, and points at `panel_open_workflow` as the exempt recovery path. **That recovery also failed:**

```
Error: workflow_open could not prove that the active canvas was rebound to the requested workflow.
The open may have switched the active workflow; reload the panel before graph edits.
```

So on panel 0.11.38 there is no in-band recovery: the documented escape hatch (`workflow_open`, which is supposed to be fence-exempt and to re-derive the fence from its own reply) does not re-derive it either. The only remaining options are a manual browser hard-refresh by the user, or a panel sync + ComfyUI restart.

That last part is the real cost here: **the recovery requires a ComfyUI restart, which is destructive to in-flight work.** At the time this fired there were 10 queued video renders draining (~19 min each, ~3h of GPU time). The agent had to choose between recovering its own tooling and preserving the user's overnight batch. An agent that follows the suggested fix blindly would have killed 3 hours of the user's renders to fix a bookkeeping problem.

### Steps to reproduce

1. Open a saved workflow (A).
2. Edit widgets on the live canvas.
3. `panel_save_workflow({ name: "B" })` — Save-As to a new name.
4. Observe: file B is written, but the reply reports it could not read the live canvas identity, and the session is still fenced to A's instance id.
5. `panel_open_workflow("B")` → fails with "could not prove that the active canvas was rebound".
6. Every `panel_*` graph tool is now refused for the rest of the session.

### Exact error (scrubbed)

From `panel_save_workflow`:

```
The workflow WAS saved. Could NOT read the live canvas identity — the panel did not answer,
so NOTHING about this session's graph binding was observed.
... It is still fenced to the previous workflow instance (<uuid>), which the panel will keep refusing.
This is not a rebind — it is an unknown.

UNDERLYING CAUSE: Error: workflow instance mismatch: this command targets a different workflow
[...]

Issue #708, opened 2026-08-06
panel_new_workflow reports a blank tab but reconnect persists the previous live graph into it


artokun:
### What happened / root cause
`panel_new_workflow` returned `{created:true, active:"Unsaved Workflow", key:"tmp:…"}` while a dirty 12-node workflow was active. The connection then dropped before any load or graph edit. After reconnect, `panel_list_workflows` showed a persisted `Untitled …` tab, and opening that tab followed by `panel_graph_outline` showed the previous dirty 12-node graph instead of an empty graph. This appears to be a reconnect/per-tab persistence race: the newly created tab identity is retained, but the previously active canvas serialization is saved or restored into that tab.

### Steps to reproduce
1. Open a modified persisted workflow with nodes.
2. Call `panel_new_workflow`. Observe `{created:true, active:"Unsaved Workflow"}`.
3. Drop/reconnect the panel session before loading or editing the new tab.
4. Open the resulting persisted `Untitled …` tab.
5. Call `panel_graph_outline`; it contains the prior workflow graph rather than 0 nodes.

### Exact error (scrubbed)
No exception was returned. The success result was misleading. The new tab later contained 12 nodes and 4 groups copied from the former active canvas.

### Fix
Applied locally: no. Upstream-only for this task: yes. The authorized workflow task explicitly forbids panel reloads/restarts or panel-code changes.

Precise change needed: make `panel_new_workflow` persist and verify an empty graph snapshot keyed to the new tab's routing key/workflow UUID before acknowledging success. During reconnect restoration, never use the global/current canvas serialization for an unsaved tab; restore only tab-local state. If the frontend cannot prove the new tab has 0 nodes after creation, return outcome-unknown instead of `created:true`. Add a regression test covering disconnect/reconnect immediately after `panel_new_workflow` while the previous workflow is dirty.

### Environment
Connected ComfyUI target: Linux, ComfyUI 0.29.0, NVIDIA A100-PCIE-40GB, CUDA 13.0, torch 2.13.0, Python 3.12. comfyui-mcp 0.49.8. comfyui-mcp-panel 0.11.41. Backend: Codex.

artokun:
the user needs to hit the + button to start a new session, please verify if my assumption is correct or of there is an actual issue here

artokun:
Fixed in `8d22140`, **merged but not yet released** — it ships in the next panel version, not 0.11.43.

[...]

PR #1021, opened 2026-08-10
fix(save): Save-As must leave a fence the next command can pass (#978)

artokun:
Claims #978. Draft opened before starting.

After a Save-As the reporter's next graph command was refused as a `workflow instance
mismatch` against the PREVIOUS instance, and neither `panel_open_workflow` nor
`panel_set_workflow_target({mode:"current"})` recovered it. Reported on panel 0.11.38.

Part of this was addressed by #941: `saveProducedIdentity` mints and publishes the
identity of the object the save produced, so the reply carries a `workflow_uuid` the
caller can re-stamp from. That is the half the report calls "needs panel 0.11.45+".

What is NOT established is the other half — whether the live ROOT graph carries that new
identity after a Save-As, or keeps the tag from before it. If it keeps the old tag, the
next command is refused exactly as reported even though the reply published the right
uuid, which is the same shape as #995.

First step is to measure that on the live install rather than reason about it.


artokun:
## Where this stands after the first pass

**Half of the report is already implemented, and it is the half the reporter identified.**
`saveProducedIdentity` (#941) mints and publishes the identity of the object the save
produced, precisely because a Save-As activates a brand-new object nothing has established
an identity for. Its comment names the exact wedge from this issue: "the panel knew the
value and would not publish it, and a caller fenced to the pre-save workflow had nothing
to recover with." That is the "needs panel 0.11.45+" note in the report.

**The unestablished half is what the live ROOT graph carries after a Save-As.** If it keeps
the tag from before the save, the next command is refused exactly as reported even though
the reply published the right uuid — the same shape as #995, which shipped in 0.11.97.

**I did not get a clean measurement this pass, and I am not reporting a dirty one.** Two
probing artifacts got in the way, both worth recording because they will bite the next
attempt:

1. `store.openWorkflow(wf)` called programmatically activates the workflow WITHOUT
   repainting the canvas — the store names the new workflow while `app.graph` still holds
   the previous one. A user clicking the tab goes through a path that does repaint. Any
   measurement driven through the store API alone is measuring that artifact, not the
   product.
2. `store.createTemporary(name, graph)` did not attach the graph I passed; the scratch
   workflow came up holding the active canvas's content instead. So the "new workflow" I
   built to Save-As was not the graph I thought it was.

Next pass: create the scratch workflow through ComfyUI's own new-workflow command, confirm
the canvas actually holds it, then Save-As under a scratch name and compare the root tag
against the identity the save reply publishes — deleting the scratch file afterwards.

No user workflow was written to. The scratch workflow was closed and never reached disk.

artokun:
## Codex review: PASS (round 5) — and the fix it produced is not the one I started with

**r1 killed the original change outright, correctly.** I had the save stamp the produced
identity onto the live root to clear the refusal. The panel's own Save-As adapter says why
that is wrong, in as many words: ComfyUI's `workflowStore.openWorkflow` moves the active
pointer and does **not** call `loadGraphData` — only `workflowService.openWorkflow` does —
and that is precisely why the adapter persists the copy from the SOURCE tab instead of
re-reading the shared canvas (#708). So the copy is active while the canvas still holds the
source's graph, and stamping would have made both fences accept the copy while every graph
tool operated on the source's canvas. My causal premise was false, and my test could not
catch it because it modelled a root already assumed to belong to the produced record.

**What that leaves is a diagnosis, not a heal.** The graph fence refuses *correctly* — the
canvas really is the other workflow's — and what was missing is that nothing said so. The
reply already told a caller to re-fence to the new `workflow_uuid`; a caller who did that
perfectly was still refused, which is exactly what the reporter did.

**r2** found a real counterexample: `workflow_save_as` flagged unconditionally, but a
first save is identity-continuous with the temporary predecessor, so that caller would have
been sent re-fencing and re-opening for a problem they do not have. It now follows
`outcome.saved_as`. It also caught that `canvas_not_repainted` asserted a reply-time state
nothing observes — a tab switch or reconnect can repaint during the save's awaits — so the
flag is `canvas_repaint_not_requested` and the consequence is conditional.

**r3 and r4** found the same overclaim surviving in the explanation: first the source
comment and test prologue, then the test title. On a change whose subject is claiming only
what was observed, those are exactly the places it matters.

Seven tests, mutation-verified, including one that keeps the unsound stamp out and one that
pins the store claim to the adapter that documents it.

Issue #890, opened 2026-08-09
A blocked registry yields a STALE Manager catalogue, and the panel cannot tell stale from current


artokun:
Split from #808, and now **observed rather than theorised**.

#808 made an empty catalogue distinguishable from a no-match. This is the case that fix explicitly did not close, and I have since measured it on a machine whose network genuinely blocks the registry.

## Measured

ComfyUI-Manager on that install raises `InvalidChannel` for `raw.githubusercontent.com/ltdrdata/ComfyUI-Manager` and reports `Cannot connect to comfyregistry`. The catalogue it then serves:

```
GET /customnode/getmappings?mode=cache  ->  HTTP 200, 1,570,773 bytes
packs: 5583
```

Populated — from the `extension-node-map.json` **bundled in the Manager package**, per `get_data_by_mode`'s except-branch. So the failure mode is not an empty list. It is a **full list that may be months out of date**, presented identically to a current one.

## Why it matters

A user searching for a pack released after their Manager version was packaged gets "no matches" — indistinguishable from "that pack does not exist". #808 fixed the *empty* case; this is the *stale* case, and it is the more likely one in the field, because Manager works hard to avoid ever returning empty.

## What is missing

Nothing in the payload carries provenance: no fetch timestamp, no indication of whether the data came from the network, the on-disk cache, or the bundled fallback. The panel therefore cannot say "this list is from the package, not the registry" — and cannot warn that a search over it is not authoritative.

## Not obvious how to fix

Manager does not expose which source answered. Options worth weighing rather than one obvious route:

- infer staleness from the bundled file's mtime versus the cache file's, if both are reachable;
- probe the channel host directly from the panel and report reachability alongside the result — honest, but adds a network call to a read;
- ask upstream for provenance on the response, which fixes it for every consumer rather than just us.

Filing so the gap is tracked on evidence. **Not** proposing a staleness claim inferred from anything the panel cannot actually observe — that would repeat the fault #808 was filed about.

artokun:
**Correcting my own report. The central claim here — that a blocked registry serves the BUNDLED `extension-node-map.json` — does not survive measurement.**

I filed that by reading `get_data_by_mode`'s except-branch, not by measuring it. Measured now on a rig with a *working* registry (ComfyUI-Manager V3.41):

```
served  /customnode/getmappings?mode=cache   5583 entries   1,473,770 bytes
bundled custom_nodes/ComfyUI-Manager/extension-node-map.json
[...]

PR #1023, opened 2026-08-11
fix(manager): report the provenance a catalogue answer can actually establish (#890)

artokun:
Claims #890. Draft opened before starting.

A blocked registry does not yield an empty catalogue — it yields a FULL one that may be
months old, presented identically to a current one. A user searching for a recently
released pack gets "no matches", which is indistinguishable from "that pack does not
exist".

The issue's own follow-up already withdrew the attractive fix: the served catalogue is NOT
the bundled `extension-node-map.json` (measured — 5583 vs 4884 entries, sharing only ~1800
keys), so a "is this the bundled map" discriminator would never fire and would ship as a
check that always passes.

What that comment DID establish is observable provenance of a different kind:

```
GET /api/customnode/getlist?mode=cache      -> HTTP 500
GET /api/customnode/getmappings?mode=cache  -> HTTP 200
GET /api/manager/channel_url_list           -> {"selected":"custom", ...}
GET /api/manager/version                    -> V3.41
```

So the direction to weigh is reporting what the panel can actually see — which endpoints
answered, which failed, and which channel is selected — WITHOUT inferring a staleness
claim it cannot support. First step is to measure those endpoints on this install and read
what the panel reports today.

### DIFF
--- a/browser_tests/unit/catalogue-no-match-provenance.test.mjs
+++ b/browser_tests/unit/catalogue-no-match-provenance.test.mjs
@@ -1,0 +1,111 @@
+/**
+ * #890 — a blocked registry does not yield an EMPTY catalogue. It yields a FULL one that
+ * may be months old, presented identically to a current one, so a user searching for a
+ * recently published pack gets "no matches" — indistinguishable from "that pack does not
+ * exist". #808 closed the empty case; this is the one it left open, and the likelier one
+ * in the field, because Manager works hard never to return empty.
+ *
+ * WHAT CANNOT BE CLAIMED, and the issue's own follow-up establishes it by measurement:
+ * the served catalogue is NOT the bundled `extension-node-map.json` (5583 served vs 4884
+ * bundled on a working rig, sharing only ~1800 keys), so a "is this the bundled map"
+ * discriminator would never fire and would ship as a check that always passes. Nothing in
+ * the payload carries a fetch time or a source, so staleness is not observable and is not
+ * asserted here.
+ *
+ * WHAT CAN: the panel asked for `mode=cache`. That is its own request, not an inference
+ * about the response, and it is exactly the fact a reader needs before concluding a pack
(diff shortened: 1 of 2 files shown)


artokun:
## Codex review: SHIP (round 2)

**r1 caught my fix committing the parent issue's own fault.** `catalogue_mode: "cache"` and
"the catalogue that was searched is ComfyUI-Manager's CACHED copy" derived a claim about
the RESPONSE from the panel's own REQUEST. Manager may ignore the parameter and serve a
live fetch — nothing in the answer says either way — so "reflects whatever that cache last
held" inherited an unsupported premise. That is exactly the class of claim #890 was filed
about, which makes it the worst possible place to make it.

The field is `requested_mode` now, and the note limits itself to what holds: the request
asked for `mode=cache`; Manager does not report whether it honoured that, when the data was
fetched, or whether it served the network, its on-disk cache or the bundled copy; so a
no-match cannot distinguish "no such pack" from "a pack too recent for whatever list was
searched". The remedy — refresh Manager's cache and search again — never depended on the
withdrawn claim.

Eleven tests, mutation-verified: renaming the field back, re-asserting that the response IS
the cache, firing on a match, dropping the mode gate, printing an absent size as "0 packs"
(`Number(null)` is 0 — the same trap as #982), asserting staleness, and swallowing the
#808 empty-catalogue answer are each caught.

**Deliberately not attempted**, per the issue's own measured warning: any staleness
inference. The served catalogue is not the bundled map (5583 vs 4884, ~1800 shared keys),
so that discriminator would never fire and would ship as a check that always passes.

**Not verified against a live Manager.** This install answers `/api/v2/manager/version` and
`/api/v2/manager/queue/status` but 404s every `/customnode/*` and `/api/manager/*` route the
issue measured, so the catalogue endpoint is unreachable here. Unit-tested against the
served payload shapes instead.

Issue #1019, opened 2026-08-10
panel_new_workflow leaves 0.11.43 session fenced away from newly created unsaved tab


artokun:
### What happened / root cause
On panel 0.11.43, panel_new_workflow successfully creates an unsaved tab but cannot read its identity from the reply. The session remains fenced to the previous workflow, so all subsequent graph writes are refused. Since the new workflow is unsaved, panel_open_workflow cannot recover it.

### Steps to reproduce
Call panel_new_workflow from a session fenced to an existing workflow on panel 0.11.43.

### Exact error (scrubbed)
workflow instance mismatch; blank workflow created but live canvas identity could not be read.

### Fix
applied locally: no
upstream-only: yes
The tool response states panel 0.11.45+ returns the new workflow identity directly and fixes the fence. Current recovery requires a browser hard refresh.

### Environment
Windows 11; ComfyUI v0.30.2; RTX 4080 SUPER 16 GB; comfyui-mcp 0.50.107; panel 0.11.43.

artokun:
Addressed in **0.12.1**, and the outcome is partly "already fixed since your version" and
partly a real gap you identified.

## Already fixed since 0.11.43, verified rather than assumed

Your tool response was right that 0.11.45+ returns the new identity, but a version-lag
claim on this cluster has twice turned out to cover only ONE of two independent fences
(#978, #995), so both were checked on current main:

- **command fence** — `panel_new_workflow` publishes `workflow_uuid` and `routing_key` for
  the tab it created (#755), so a caller has something to re-fence to.
- **graph fence** — the created tab's canvas carries a fresh panel identity. Measured live
  by running `Comfy.NewBlankWorkflow`, the same command the handler runs:

```
before   krea2_identity_edit   tag e66e531b   24 nodes   provenEmpty false
after    Unsaved Workflow      tag 2d7fa288    0 nodes   provenEmpty true
```

A changed tag on a proven-empty canvas is the creation stamp (#606/#708), so graph commands
compare against the new tab's identity rather than the previous tab's.

## What was still missing, and it is the sentence you wrote

> Since the new workflow is unsaved, panel_open_workflow cannot recover it.

You had to work that out yourself. The refusal offers two recoveries and one of them cannot
apply to a tab with no path — so it now says so:

> Note that the ACTIVE tab is unsaved, so panel_open_workflow cannot re-select THAT one —
> it resolves a workflow by path and this tab has none. If the canvas you want is this
[...]

Issue #755, opened 2026-08-07
workflow_new should return workflow_uuid on its reply


artokun:
Small one, from artokun/comfyui-mcp#932.

`workflow_new` mints the new canvas's identity eagerly:

```js
// Mint the per-instance tmp: routing id (and seed the transcript uuid) eagerly so
// the key exists BEFORE the first edit and is returned for pinning.
const key = workflowTabId(wf);
workflowStableUuid(wf);
```

…but the reply omits it:

```js
return { created: true, active: getWorkflowTitle(), key, routing_key: key, open_seq: receipt.seq };
```

So the orchestrator has to make a second `workflow_list` round trip to learn the identity of a canvas the panel just created and already knows. artokun/comfyui-mcp#1024 does exactly that — it works, and it fixed a real wedge (a session stayed fenced to the PREVIOUS workflow after `panel_new_workflow`, so every `panel_add_node` failed with `workflow instance mismatch`).

**Ask:** add `workflow_uuid` to the `workflow_new` reply, alongside `key`/`routing_key`. The orchestrator side will keep the `workflow_list` fallback for older panels either way, so this is a latency and robustness improvement rather than a fix — one fewer round trip, and one fewer chance for the canvas to change between creating it and reading it back.

Related: artokun/comfyui-mcp#932's reporter was on panel 0.11.38 with an *unsaved* workflow, where `active.workflow_uuid` is omitted unless an identity is established. Returning it directly from the command that establishes it removes that whole class of ambiguity.

artokun:
Extending this to cover **save** as well as new — a second reporter hit the same deadlock through a different door, and the save receipt has the identical gap.

artokun/comfyui-mcp#1045: a Save-As re-points the active workflow (the canvas is now the new file, with its own identity), the orchestrator's fence still names the pre-save instance, and every `panel_*` call afterwards fails with `workflow instance mismatch`. I've fixed the orchestrator half the same way as `workflow_new` — re-derive from `workflow_list` after the save.

**But that is a workaround, and artokun/comfyui-mcp#1043 shows why.** When the canvas-identity read does not answer, the refresh honestly reports that it could not establish the fence — and the documented recovery, `panel_set_workflow_target({mode:"current"})`, depends on *that same read*. So it can never break the deadlock. The reporter's only escape was asking a human to press F5, and `panel_reload` is contraindicated there (#803).

The panel already knows the answer at the moment it acts. `workflow_save_as` returns:

```js
[...]

PR #1025, opened 2026-08-11
fix(new): a created tab must leave a fence its own commands can pass (#1019)

artokun:
Claims #1019. Draft opened before starting.

Reported on panel 0.11.43: `panel_new_workflow` creates the tab but the session stays
fenced to the previous workflow, and because the new tab is unsaved,
`panel_open_workflow` cannot recover it. The reporter's own tool response says 0.11.45+
returns the new identity and fixes it.

Working #978 established that a version-lag claim on this cluster needs BOTH halves
checked, because they fail independently:

- the COMMAND fence needs the reply to publish the new tab's identity;
- the GRAPH fence needs the LIVE ROOT to carry it, and `workflow_new` stamps the root
  only when both sides are PROVEN content-free.

That second condition is the one to measure. ComfyUI's store activation does not repaint
the canvas (`workflowStore.openWorkflow` never calls `loadGraphData` — established while
fixing #978), so if the created tab is activated that way, the canvas can still hold the
PREVIOUS workflow's graph, the empty-proof fails, no stamp happens, and the graph fence
refuses exactly as reported.

First step is to read which path `workflow_new` takes and measure it live.

### DIFF
--- a/browser_tests/unit/unsaved-tab-fence-remedy.test.mjs
+++ b/browser_tests/unit/unsaved-tab-fence-remedy.test.mjs
@@ -20,9 +20,11 @@
  *     a consistent identity to compare against rather than the previous tab's.
  *
  * What remained, and is what this ships: the refusal's own remedy names
- * `panel_open_workflow`, which resolves a workflow BY PATH. For an unsaved tab — exactly
- * the canvas a `panel_new_workflow` just made — that half of the advice cannot be
- * followed, and the reporter had to say so themselves.
+ * `panel_open_workflow`, which resolves a workflow BY PATH — so it cannot re-select an
+ * unsaved ACTIVE tab, exactly the canvas a `panel_new_workflow` just made. The claim is
+ * kept that narrow (codex): a mismatch means the command's intended workflow is not the
+ * active one, and if that intended workflow is a SAVED one, open remains the right
+ * recovery. What is ruled out is re-selecting THIS tab, nothing more.
  */
 import test from "node:test";
 import assert from "node:assert/strict";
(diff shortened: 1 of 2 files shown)


artokun:
## Codex review: SHIP (round 2)

**What was already fixed, verified rather than assumed.** A version-lag claim on this
cluster has twice covered only one of two independent fences (#978, #995), so both were
checked on 0.12.0:

- the reply publishes `workflow_uuid` and `routing_key` for the tab it created (#755);
- the created tab's canvas carries a fresh panel identity — MEASURED by running
  `Comfy.NewBlankWorkflow`, the same command the handler runs:

```
before   krea2_identity_edit   tag e66e531b   24 nodes   provenEmpty false
after    Unsaved Workflow      tag 2d7fa288    0 nodes   provenEmpty true
```

A changed tag on a proven-empty canvas is the #606/#708 creation stamp, so the graph fence
compares against this tab's identity rather than the previous tab's.

**What shipped** is the sentence the reporter had to write for themselves: the refusal's
remedy names `panel_open_workflow`, which resolves BY PATH, and an unsaved tab has none.

**r1 caught a logical error in my first version.** I wrote "re-targeting is the route here",
which assumes the caller wants the *active* tab. A mismatch means the command's intended
workflow is not the active one — and if that intended workflow is saved, open is exactly
right. The clause is now scoped to this tab: open cannot re-select *that one*; re-target if
the canvas you want is this active tab; opening a different saved workflow still works.

Seven tests, mutation-verified: dropping the clause, treating a non-boolean as unsaved,
assuming the fact instead of reading it, claiming unsaved for an unreadable tab, re-asserting
that open is useless, and widening the clause from the tab to the tool are each caught.

One incidental fix: the test's function extractor had to find the BODY brace rather than
the first one — both functions here take a destructured parameter, so the naive scan
returned a signature with no body and failed as a `SyntaxError` that read like a defect in
the source under test.

Issue #952, opened 2026-08-10
panel_ask loses outcome when workflow tab disconnects and cannot be retried safely


artokun:
### What happened / root cause
`panel_ask` was used to present three recovery choices while diagnosing a third-party UI scan. The panel workflow tab disconnected mid-command. The tool returned an outcome-unknown error and provides no card id, delivery acknowledgement, pending-card lookup, or retry token, so the agent cannot know whether the question card exists and cannot safely reissue it without risking duplicate cards.

### Steps to reproduce
1. Call `panel_ask` with a connected workflow tab.
2. Let the tab/socket disconnect while the command is being delivered.
3. Observe the tool returns outcome unknown.
4. Attempt recovery: there is no `retry_of` argument or read-only way to query whether the ask card was created.

### Exact error
`Error: panel tab wf:728f6 disconnected mid-command ("ask_user") — OUTCOME UNKNOWN: the command was already sent, so the panel may have applied it ... Verify before retrying ... instead of re-issuing it blindly.`

The suggested verification examples concern render queues/media and do not apply to an ask card.

### Fix
Applied locally: no. Upstream-only: yes; this is in the long-lived panel orchestrator command-delivery path, not the workflow JS being diagnosed.

Precise proposed change: make `panel_ask` idempotent with a request/card id and `retry_of`, or add a read-only pending-card query. On disconnect, return whether the card was acknowledged by the tab. The recovery hint should be ask-specific rather than queue/media-specific.

### Environment
Windows 11; ComfyUI 0.27.0; ComfyUI frontend 1.45.20; Python 3.12.10; PyTorch 2.8.0+cu129; NVIDIA RTX 5070 12 GB; comfyui-mcp 0.50.49; comfyui-mcp-panel 0.11.45.

Related (same root cause): #404, #646

artokun:
Traced this on the panel side against 0.11.83. The report says "upstream-only", and the *messages* are — but there is a concrete panel-side reason a retry is unsafe here, and it is worth writing down before anyone builds a `retry_of` argument on top of it.

**The panel already honours `retry_of`.** Every command, `ask_user` included, goes through `commandRidLedger.begin(...)` / `settleRid(reply)`, and an inbound frame carrying `retry_of` is looked up before executing. On a hit the panel replays the original reply under the retry's rid rather than re-executing — so a duplicate card is already prevented, and the tool schema exposing `retry_of` for `panel_ask` would be a real fix.

[...]

PR #1027, opened 2026-08-11
fix(ask): mark a question card whose connection dropped, instead of leaving two live ones (#952)

artokun:
Claims #952 — the part of it that needs no design decision.

The two design questions this issue records are deliberately left alone: whether interactive
cards should dedupe on a scope that survives a reconnect, and what a retry should return
while the original card is unanswered. Both are still the owner's call, and the orchestrator
half shipped in comfyui-mcp 0.50.88.

What that fix has to tell the user today is this:

> Expect the stale card to remain on screen: retry suppression is keyed to the socket that
> dropped, so this counts as a new command and the user may see two. Tell them which one to
> answer.

The panel can say that itself. A card painted on a connection that has since been replaced
can no longer deliver its answer — the panel deliberately does not replay a reply of this
kind across a reconnect — so it is not merely stale, it is dead, and it still looks exactly
as clickable as the live one beside it.

Marking it needs no new scope, no retry semantics, and no ledger change: a card records
which connection painted it, and a later connect retires the ones from before.

### DIFF
--- a/browser_tests/unit/bridge-disconnect.test.mjs
+++ b/browser_tests/unit/bridge-disconnect.test.mjs
@@ -12,7 +12,7 @@ const panelPath = fileURLToPath(new URL("../../web/js/comfyui-mcp-panel.js", imp
 
 function bridgeStatusHandler() {
   const source = readFileSync(panelPath, "utf8").replace(/\r\n/g, "\n");
-  const start = source.indexOf("const client = createBridgeClient({\n    onStatus(state) {");
+  const start = source.indexOf("const client = createBridgeClient({\n    onStatus(state, socketId) {");
   assert.notEqual(start, -1, "could not locate the bridge status callback");
   const end = source.indexOf("\n    onSay(", start);
   assert.notEqual(end, -1, "could not locate the callback following onStatus");
--- a/browser_tests/unit/interactive-card-fence.test.mjs
+++ b/browser_tests/unit/interactive-card-fence.test.mjs
@@ -235,7 +235,10 @@ test("an unknown reason still produces a refusal, never an empty or partial stri
 
 /** Pull one 4-space-indented object-literal method out of the panel source. */
 function extractMethod(name) {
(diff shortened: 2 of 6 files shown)


artokun:
## Codex review: SHIP (round 4)

Three P1 rounds, and each found something I had not.

**r1 — the trigger was wrong in the dangerous direction.** `"connected"` is emitted on
every re-handshake (each `models` frame calls `markConnected`, and a workflow change
re-hellos the *live* socket), so counting status events would have retired question cards
that were still perfectly answerable — worse than the duplicate this fixes. The client now
mints an id per WebSocket and hands it to `onStatus`. r1 also caught that `paintSecret` was
not registered at all, and that it matters more than the question card: a live password
field whose reply has nowhere to go can still display "Token saved".

**r2 — the pre-handshake window.** A command frame is accepted before the handshake, so a
card can be painted on socket B while the UI still believes A is live. Splitting adoption
(any status carrying a new id) from retirement (only on `connected`) was half the answer;
the other half is that the open status is suppressed once the patience window has been
given up on. So the id now comes from the **command**: `onAsk`/`onSecret` receive
`thisSock.__cmcpSocketId` and thread it to the painter. No status emission is required for
correctness.

**r3 — the Settings token card.** `paintSecret` also serves the "Set … token" buttons,
which have no command behind them and can still send `set_secret` on the current socket
after a reconnect. Retiring that card disabled a working control and told the user to wait
for a request that was never coming. Registration is gated on a command now, and with that
gate the registry's belief-based fallback is gone: an entry names its own socket or the
card is not tracked.

**Deliberately untouched**, because this issue records them as its owner's decisions:
whether interactive cards should dedupe on a scope that survives a reconnect, and what a
retry should return while the original is unanswered. None of the above needed either.

Fourteen tests, mutation-verified — including one that only killed its mutant after the
guard **counted** both painters, because matching the spread once passed while the secret
card had lost it.

**Not verified end-to-end:** driving a real disconnect needs the bridge and an agent
session, which this client cannot reach. The retirement function, the registry and the
sequencing are unit-tested; the wiring is pinned by source guards.

Issue #1006, opened 2026-08-10
Panel should serve its own /object_info so a remote canvas can be converted (cause behind comfyui-mcp#1359)


artokun:
Split out of #1359, whose diagnosis half is fixed and released. **The cause is untouched and lives here.**

## The split authority

`panel_strip_workflow` (and anything else that pairs a panel-captured graph with node definitions) reads the graph from the **connected panel** and the definitions from the **global headless client**, which resolves `COMFYUI_URL`. Those are the same machine for a local session and two different machines for a connected remote panel — so a remote panel cannot strip its own live canvas at all.

#1359 now explains that clearly when it fails, and names the workaround (point `COMFYUI_URL` at the same ComfyUI). That is a diagnosis, not a fix.

## Why the obvious orchestrator-only fix is wrong

The orchestrator knows the panel's ComfyUI origin (`tabServerOrigin`, server-observed on the WS upgrade). Fetching `/object_info` from it directly is tempting and would have worked for the reporter, whose ComfyUI is behind a public RunPod proxy URL. It should not be done on its own:

- **it does not generalise.** In a tunnel or loopback-only topology the *browser* is the only thing that can reach that ComfyUI. The orchestrator fetching it would fail exactly as it does today, for a different reason — one silent failure traded for a narrower one.
- **it adds an outbound fetch to an origin chosen by the connected tab.** That is a new capability with its own review surface, and it should not arrive as a side effect of a strip fix.

## What the reporter proposed, which is right

A panel bridge command — `graph_get_object_info` — where the frontend fetches its **own** ComfyUI origin's `/object_info` and returns it. The browser is by definition able to reach that host: it is already talking to it. Optionally the same data could ride along with `graph_serialize`, though that makes every serialize pay for a large payload.

Consumers then choose by SOURCE: panel-sourced definitions for the live canvas, `COMFYUI_URL` for a pack/path/inline graph (which is correctly tied to it).

## Notes for whoever takes it

- Size: `/object_info` is large (4183 node types on one local install here). Whatever carries it should be cacheable per tab and per ComfyUI generation, not refetched per call.
- There is already a per-tab notion of the panel's ComfyUI identity to key that cache on.
- Reported against comfyui-mcp 0.50.100 / panel 0.11.89, orchestrator on Windows, ComfyUI 0.30.0 on Linux behind a proxy URL.

artokun:
[...]

PR #1029, opened 2026-08-11
feat(graph): serve the panel's OWN /object_info so a remote canvas can be converted (#1006)

artokun:
Claims #1006 — the panel half, which is the half that lives here.

Anything that pairs a panel-captured graph with node definitions reads the graph from the
CONNECTED PANEL and the definitions from the global headless client, which resolves
`COMFYUI_URL`. Same machine locally; two different machines for a remote panel — so a
remote panel cannot convert its own live canvas at all.

The issue rules out the tempting orchestrator-only fix (fetch `/object_info` from the tab's
observed origin): it does not generalise to a tunnel or loopback-only topology where the
BROWSER is the only thing that can reach that ComfyUI, and it adds an outbound fetch to an
origin chosen by the connected tab.

What it asks for instead is a panel bridge command: the frontend fetches its OWN origin's
`/object_info` and returns it. The browser is by definition able to reach that host — it is
already talking to it.

Two things the issue flags for whoever builds it, and both shape this:

- the payload is large (4183 node types on this install), so it must be cacheable per tab
  and per ComfyUI generation rather than refetched per call;
- there is already a per-tab notion of the panel's ComfyUI identity to key that on.

The panel already grew the pieces this needs while fixing #982: a two-transport
`/object_info` oracle that reports honestly when it cannot answer, and a burst cache that
never caches a failure.


artokun:
## Codex review: SHIP (round 3)

Three findings, and the last one was my own lesson repeating itself one level down.

**r1 — `served_by` named the wrong host.** It read `comfyuiUrlForAgent()`, which prefers the
user's Remote-URL override, while the fetch goes through the page's own `api` client. On an
install where those differ, the reply attributed the schema to a host that served nothing.
It reports the page origin now.

**r1 — the fingerprint claimed proof.** A single 32-bit FNV-1a over a 4000-type schema
collides far too readily to support "two payloads with the same fingerprint have the same
types" — and the reply built on it would then say `unchanged` about a genuinely different
set. Two independent bases plus the exact type count now, and the comment says plainly that
this is a hash and a cache key.

**r2 — and then I concatenated the two halves without a boundary.** Base-36 strings are
variable length, so `${a}${b}` loses exactly the agreement the second base was added to
establish: `"1"+"23"` and `"12"+"3"` both read as `"123"`. This is the same
non-injective-join defect the file already documents about `JSON.stringify(names)`, and I
walked into it while fixing the thing above it. Separated by a hyphen, which base-36 cannot
contain.

**The 5MB question I could not answer** was answered by the review rather than assumed: the
bridge's WebSocketServer sets no lower `maxPayload`, so `ws` uses its 100 MiB default, and
an ordinary rid reply is parsed and resolved directly rather than journaled.

Eleven tests, mutation-verified — dropping the second base, dropping the count, reinstating
the concatenation, restoring the override as `served_by`, and returning a partial map on
failure are each caught.

**This is the panel half of a cross-repo fix.** The executor exists for the orchestrator to
dispatch to; it names no MCP tool, so the vendored vocabulary is untouched. It is not
verified end-to-end because the other half does not exist yet, and I have said so rather
than implying a working path.

PR #1031, opened 2026-08-11
docs(983): what a persistence check can and cannot establish — no code

artokun:
Claims #983, building to the spec the previous attempt left behind after review reverted it.

The reproduction and the signal are already established on the issue: a write to the
rgthree Fast Groups Bypasser toggle reads back correctly and then persists nowhere, while
Power Lora Loader — same author, same `type: "custom"`, same composite object — persists
fine. Read-back verification is not at fault.

What the review killed was the GENERALISATION, and its four findings are the spec:

1. only reading `widgets_values` — a custom widget can persist through `properties` or a
   node-specific field, so "not in widgets_values" is not "not persisted";
2. assuming serialization preserves the value — a serializer may transform it, and
   JSON comparison trips on key order, NaN, undefined, toJSON, proxies;
3. matching anywhere in the array — which creates the complementary false negative;
4. `widgets_values: []` proving only that this serializer emitted no positional values.

Plus a cost concern: calling a plugin-overridable `serialize()` after every write, on the
path #716 exists to keep cheap.

The shape that answers all four is to stop looking for the VALUE and ask whether the node's
serialized form CHANGED at all across the write. That is immune to transformation, covers
every field rather than one, needs no per-widget matching, and asks the question that
actually matters: will this write survive a save or a queue.

First step is measuring what a per-node serialize costs on a real canvas, because the
previous attempt died partly on cost and I would rather know than argue.


artokun:
## Outcome: this PR ships no code, and that is the finding

The second attempt at #983 solved the four defects that reverted the first one, then hit two
of its own. Both came out of review; neither can be closed.

### What the design DID fix

Comparing the node's whole serialized form **before and after** the write, rather than
searching that form for the value, is immune to all four of the reverted attempt's failures:

| reverted attempt's defect | before/after comparison |
|---|---|
| only read `widgets_values` | a value in `properties` still changes the form |
| assumed no transformation | a transformed value still changes the form |
| matched anywhere in the array | nothing is matched at all |
| `widgets_values: []` proved nothing | the form is what decides |

**Cost is not the objection it was, either.** Measured on a 24-node canvas of the reporter's
kind: one node serializes in **≤0.1 ms**, all 24 in **0.4 ms** — against a write path that
already awaits an `/object_info` oracle measured at 818 ms cold.

The **promoted-write** hazard review raised was real and is fixed rather than argued away:
the value can serialize on the OUTER subgraph rail, so those writes get no claim.

### Why it still cannot ship

**1. Two matching samples are not proof of absence.** A plugin serializer may include its
widget value conditionally — the example that killed it gates on a clock phase — so a write
probed inside the excluding window yields byte-identical forms while the very next save
contains the value. The probe establishes that its two samples matched, not that later
serialization omits the write. A definitive negative on every direct `panel_set_widget`
reply cannot rest on that.

**2. The probe is an untracked plugin-execution path.** `serialize()` is overridable, so the
capture runs arbitrary code twice more per write, **outside** `applyWidgetWrite`'s undo
envelope. A serializer that mutates a property, fires a callback or switches workflow does
that twice — and the first effect survives even when the re-asserted fence then refuses the
real write. `try/catch` suppresses an exception; it cannot roll back a mutation.

That is the same shape as the judgement that reverted the first attempt: a wrong warning, or
an untracked side effect, on every widget write is worse than the narrow silent failure it
was meant to catch.

### What a third attempt needs

A persistence signal that does not invoke node-owned serialization, or a way to invoke it
that is provably pure. Neither exists today.

The reporter's workaround remains correct: **set the target nodes' modes directly** — the
bypasser is a control surface over them.

PR #1044, opened 2026-08-11
feat(753): scale every panel font size from one CSS variable

### LINKED ISSUE #753: Panel sidebar has no font-size / UI-scale setting; rem-based inner sizes make .cmcp-root font-size overrides ineffective

artokun:
Closes #753's option 2, by the route the evidence pointed at rather than the one it named.

## What this does

Every inner `font-size: Xrem` (214 values, 9 non-vendor files) becomes
`calc(var(--cmcp-fs, 0.8125rem) * k)`, k = X/0.8125, so each renders its original pixel
size at the default. `--cmcp-fs` is declared on **both** `:root` and `.cmcp-root`.

Setting `--cmcp-fs` in a user stylesheet now scales every panel font size at once —
including the CivitAI explorer, Apps and the modals.

## Why not `em`, which is what the issue asked for

This branch tried `em` first, and the record is in its commits. `em` resolves against the
**parent**, so a rule inside a block that sets its own size needs a different multiplier,
and a rule used under two different parents has no single correct value at all.

Measured, not predicted:

| | |
|---|---|
| elements drifted by the mechanical rescale | **529** |
| nested sites needing individual correction | ~25, across views that are expensive to render |
| rules with no expressible single value | **5** |
| independent sweeps that disagreed on values | 2 (mine and codex's, about what a parent renders at) |

A variable is flat: the multiplier is identical regardless of nesting depth. No per-parent
corrections, no multi-context exceptions, and no per-surface anchors — the `em` version
needed five, because the sub-modal overlay, CivitAI lightbox and media lightbox mount on
`<body>` outside `.cmcp-root` and would have rendered ~23% larger. `:root` reaches them.

## Verification

Measured on the live panel after the rewrite:

- **2471 elements compared** against a pre-change capture — **zero drift** beyond
  sub-0.01px rounding.
- Overriding `--cmcp-fs` to 1.5x moved **2104 elements by exactly 1.50**. The 367 that did
  not are PrimeIcons sized by ComfyUI's own CSS (331) plus a few of our rules that use
  literal `px`. For comparison, the `em` version moved 887 and left 1579 fixed.
- `var(--cmcp-fs, 0.8125rem)` everywhere, so a dropped custom property falls back to the
  panel's own default instead of erasing the declaration.

## Also

- The UI-scale tooltip named the trap; it now names the knob, and says plainly that it
  scales text but not spacing or icons. Its guard test asserted the warning, so it now
  asserts the working knob — a warning must not outlive the thing it warned about.
- One self-inflicted bug worth recording: a backtick inside a comment I added to the
  `PANEL_CSS` template literal closed the string, so the extension failed to register and
  the panel did not render at all. Neither the unit suite nor typecheck sees this — the
  tests read the panel as text and tsc does not cover `web/js`. `node --check` over every
  panel file reproduces it in a second and is now part of my routine here.

3722 tests, typecheck and tool-vocabulary clean.

Closes #753

### DIFF
--- a/browser_tests/unit/panel-ui-scale.test.mjs
+++ b/browser_tests/unit/panel-ui-scale.test.mjs
@@ -166,5 +166,9 @@ test("the tooltip names the knob that now DOES work, and what it does not cover"
   );
   // It scales TEXT. Spacing is still rem on purpose, and saying otherwise would have
   // users expect a denser panel that never arrives.
-  assert.match(entry, /does not scale spacing or icons/);
+  assert.match(entry, /does not scale spacing, icons, or/);
+  // ...and the fixed-pixel elements. A tooltip promising "every panel font size" would be
+  // wrong: a training caption and a few rules carry literal px (codex).
+  assert.match(entry, /fixed pixel size/);
+  assert.ok(!/scales every panel font/.test(entry), "no absolute claim");
 });
--- a/browser_tests/unit/panel-ui-scale.test.mjs
+++ b/browser_tests/unit/panel-ui-scale.test.mjs
@@ -151,11 +151,20 @@ test("onChange applies immediately and is NOT gated on settingsArmed", () => {
   );
 });
 
-test("the tooltip tells the user why their stylesheet override did not work", () => {
-  // The reported confusion was not just "text is small" — it was that the
-  // obvious fix silently does nothing. A setting that does not explain that
-  // leaves the next person to rediscover it.
+test("the tooltip names the knob that now DOES work, and what it does not cover", () => {
+  // The original confusion was that the obvious fix silently did nothing, and the
+  // tooltip existed to say so. #753 removed the trap: every inner font size is
+  // calc(var(--cmcp-fs) * k), so there is a working knob to name instead of a dead end
+  // to warn about. The warning must not survive the thing it warned about — a tooltip
+  // that still says overrides do not work is now the misleading one.
   const entry = PANEL.slice(PANEL.indexOf("id: SETTING_UI_SCALE"), PANEL.indexOf("id: SETTING_STALL_S"));
-  assert.match(entry, /rem/);
-  assert.match(entry, /font-size/);
+  assert.match(entry, /--cmcp-fs/, "names the variable");
+  assert.match(entry, /0\.8125rem/, "and its default, so a user can compute a target");
+  assert.ok(
+    !/does NOT \?" \+/.test(entry) && !/most text ignores it/.test(entry),
+    "the retracted claim that overrides cannot work is gone",
+  );
+  // It scales TEXT. Spacing is still rem on purpose, and saying otherwise would have
+  // users expect a denser panel that never arrives.
+  assert.match(entry, /does not scale spacing or icons/);
 });
--- a/web/js/cmcp-a2ui.js
+++ b/web/js/cmcp-a2ui.js
@@ -249,41 +249,41 @@ export const A2UI_CSS = `
 }
 .cmcp-a2ui { border: 1px solid var(--p-content-border-color, #3f3f46); border-left: 3px solid var(--p-primary-color, #3a7bd5);
   border-radius: 8px; padding: 0.6rem 0.7rem; margin: 0.35rem 0; background: var(--p-content-background, #1f1f23);
-  font-size: calc(var(--cmcp-fs) * 0.9846); position: relative; }
+  font-size: calc(var(--cmcp-fs, 0.8125rem) * 0.9846); position: relative; }
 .cmcp-a2ui.resolved { opacity: 0.85; }
-.cmcp-a2ui-title { font-weight: 600; font-size: calc(var(--cmcp-fs) * 0.8862); text-transform: uppercase; letter-spacing: 0.05em;
+.cmcp-a2ui-title { font-weight: 600; font-size: calc(var(--cmcp-fs, 0.8125rem) * 0.8862); text-transform: uppercase; letter-spacing: 0.05em;
   opacity: 0.7; margin-bottom: 0.4rem; padding-right: 1.2rem; }
 .cmcp-a2ui-x { position: absolute; top: 0.35rem; right: 0.4rem; background: none; border: none; cursor: pointer;
-  color: var(--p-text-muted-color, #a1a1aa); font-size: calc(var(--cmcp-fs) * 0.8615); line-height: 1; padding: 0.15rem; }
+  color: var(--p-text-muted-color, #a1a1aa); font-size: calc(var(--cmcp-fs, 0.8125rem) * 0.8615); line-height: 1; padding: 0.15rem; }
 .cmcp-a2ui-x:hover { color: var(--p-text-color, #e4e4e7); }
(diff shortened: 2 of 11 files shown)


artokun:
**Not merging this. Two independent sweeps now say the `em` mechanism has a long tail of
per-site corrections, and they disagree with each other on the values — which is the finding.**

### What has been done and works

- 214 `font-size` values converted; `.cmcp-root` anchor preserved.
- The five `document.body`-mounted surfaces anchored at `0.8125rem`. Verified on the real
  code path: the sub-modal overlay mounts on a 16px body and computes 13px, title 13.6px.
  Without this all 59 values inside those surfaces rendered ~23% larger.
- 529 drifted elements found by a 2466-element before/after capture and corrected to
  `originalPx / actualParentPx`; the chat panel measured clean afterwards.
- Codex round 1 found three P1s in views the browser measurement never rendered; all fixed.

### Why it stops here

Codex round 2 and a static sweep of my own each found **more** nested-parent errors, and
they do not agree. For `.cmcp-toolbtn .pi` my sweep computes `1.6114em` and codex computes
`1.1818em`, because we disagree about what the parent actually renders at (8.8px measured
live earlier vs 11px read from the stylesheet). At least one of us is wrong, and the only
way to settle it per site is to render that view and measure — for ~25 sites across Apps,
CivitAI, Training, RunPod and the tray, several of which are only reachable through UI
flows.

Codex also found cases invisible to both static reading and my capture: `.cmcp-rp-muted`
and `.cmcp-deps-muted` render in two different parent contexts, so no single `em` value
exists for them at all. That is now **five** rules kept absolute for that reason.

This is the hazard the original analysis on #753 predicted, holding up under test: `em`
compounds, and every nested site is an independent chance for a quiet visual break.

### The variant that removes the problem rather than managing it

```css
:root, .cmcp-root { --cmcp-fs: 0.8125rem; }
.cmcp-whatever { font-size: calc(var(--cmcp-fs) * 0.923); }
```

Flat by construction — every value resolves from one variable regardless of nesting depth.
No per-parent corrections, no multi-context rules that cannot be expressed, and no anchors
needed for the body-mounted surfaces, because `:root` reaches them too. The knob becomes
`--cmcp-fs` rather than `.cmcp-root { font-size }`: a better knob, and one that works in the
portals, but not literally the cascade this option asked for.

The branch is left intact and draft. Everything measured above is recorded in the commits.

PR #1050, opened 2026-08-11
feat(i18n): translate the Settings names and tooltips (52 ids x 4 languages)

artokun:
Unit 11 of the panel i18n fan-out: **Settings names & tooltips**. This surface uses a
different mechanism from the rest of the panel, and the deliverable is JSON, not code.

## How it works

ComfyUI's backend discovers `locales/<lang>/*.json` in any custom node and serves them
merged from `GET /i18n`; `SettingItem.vue` then looks each of our rows up as
`settings.<normalizeI18nKey(id)>.name` / `.tooltip`. `normalizeI18nKey` replaces dots with
underscores and **preserves hyphens**, so `comfyui-mcp.defaultModel.claude` is translated
by adding `comfyui-mcp_defaultModel_claude` to `locales/<lang>/settings.json` — with zero
JavaScript.

`tr()` in `panelSettingsList()` would be actively wrong, not merely redundant: that list is
evaluated as an *argument* to `app.registerExtension`, which runs **before** `setup()`
awaits the catalog, so every `tr()` there resolves against an empty catalog and freezes the
English fallback for the session. Text built inside a `type: () => …` render fn is the
opposite case — ComfyUI never looks at it and the fn runs after the catalog loads — so
those **do** use `tr()`.

## ⚠️ Correction to the unit brief: `settings.json` is FLAT

The brief specified `{ "settings": { "comfyui-mcp_x": … } }`.
`custom_node_manager.py::build_translations` does `node_translations["settings"] = json_data`
— it assigns the *file's contents* to the `settings` key. A wrapper would produce
`settings.settings.*` and translate nothing, silently. Cross-checked three ways: ComfyUI
0.31.1's source, ComfyUI's own `src/locales/*/settings.json`, and an installed custom node
(ComfyUI-Easy-Use) — all flat. **Other units taking key-shape guidance from the same brief
should re-check.**

## What changed

- `locales/{en,ko,zh,ja}/settings.json` — all **52** registered setting ids, name +
  tooltip, **104 strings** each. The 25 per-provider model/effort tooltips are interpolated
  in the code and ComfyUI's lookup has no placeholders, so each id carries its own
  fully-written literal.
- The `joinDiscord` tooltip was the one settings string routed through `main.json` via
  `tr()`. Restored to a plain literal — and deliberately kept **off** the `tooltip:` line,
  because `i18n-extract` proposes any literal preceded by `tooltip:` on the same line and
  `i18n-apply --write` would silently re-wrap it. Every other tooltip in the list escapes
  for exactly that reason by accident; now it is written down.
- Render-fn text converted with `tr()`: the what's-new button, the beta note, the store
  links, `— coming soon`, and the token button label. The token label is **one key per
  row** rather than one interpolated `"Set {friendly} {noun}…"` — those values are English
  constants, so an interpolated label could never agree with the fully translated row name
  ComfyUI paints directly above it.
- Where a translated tooltip names a dropdown **option**, the option stays in Latin
  (Panel / Workflow / Ask / Detect / Auto / Model default). ComfyUI *can* translate options
  via `settings.<id>.options.<label>` — unused here, and worth a follow-up — but a tooltip
  naming 「確認する」 would send the reader hunting for a choice the list does not offer.
- `main.json` regenerated with `npm run i18n:build` (picks up the new render-fn `tr()`
  sites, drops the key the `joinDiscord` un-wrap orphaned — the round-trip guard requires
  both directions). ko/zh/ja carry the same 11 keys, translated.

## New guard

`settings-i18n-keys.test.mjs` joins the code's registered setting ids to the catalog **in
both directions**, so a new row cannot ship untranslated and a typo'd key cannot sit there
silently resolving to nothing. This is the failure mode the Zod locale gate structurally
cannot see: it only checks that the four catalogs agree with *each other*, never that any
of their keys reaches a real setting. Verified by mutation — a typo'd key, a dropped row,
and a blank tooltip each fail it; the id extractor finds exactly 52, so the assertions are
not vacuous.

## Note on the SyntaxError P0

This branch originally also carried a fix for the spliced `lib/i18n.js` import (found
independently here while parse-checking the source as a module, since the settings work
could not otherwise be verified at all). `04d020bf` landed on the base first with an
equivalent fix and an equivalent module-parse guard, so **both were dropped from this
branch in favour of what shipped** — this PR no longer touches `scripts/i18n-apply.mjs`.
The landed guard was verified here rather than assumed: re-nesting the import fails it, and
the healthy tree passes.

## Verification

- `npm run test:unit` — 3716 pass, 0 fail (includes the Zod locale gate: `locales OK`,
  settings.json 100% translated in ko/zh/ja)
- `npm run test:e2e:list` — 68 tests in 29 files
- `npm run typecheck` — clean
- End-to-end without a browser: simulated `build_translations()` + `SettingItem.vue`'s
  lookup over the committed files — 18/18 probes resolve to translated text, 0 fell back to
  English, hyphen preserved in the key transform.
- English parity: 0 drift across all 52 rows (23 static compared against the verbatim
  literal, 29 factory rows reconstructed from the same template plus `BACKEND_TEXT`).
- Per the brief, no live browser run; the coordinator's consolidated pass covers it.


artokun:
### Verification evidence (no browser required)

**1. The mechanism resolves end-to-end.** Simulated `build_translations()` (assign each locale file's contents to its `settings`/`commands`/`nodeDefs` key, merge with `main.json` at top level) and then `SettingItem.vue`'s `t("settings." + normalizeI18nKey(id) + ".name", fallback)`, over the files as committed:

```
=== ko ===
  OK  comfyui-mcp.defaultModel.claude.name    -> 기본 모델
  OK  comfyui-mcp.defaultEffort.codex.tooltip -> ChatGPT 에이전트의 기본 추론 강도로, 해당 스케일(none–ultra)에서…
  OK  comfyui-mcp.bridgeUrl.single.name       -> 브리지 URL (고급)
  OK  comfyui-mcp.featureFlag.runpod.name     -> RunPod / 로컬 표시
  OK  comfyui-mcp.setCivitaiToken.name        -> CivitAI 토큰
  OK  comfyui-mcp.starGithub.name             -> GitHub에서 스타 주기
(zh and ja likewise)

key transform: comfyui-mcp_defaultModel_claude — present in en: true
hyphen preserved: true
fell back to English: 0 / 18
```

The hyphen probe matters: it is the one part of the key transform most likely to be wrong, and a wrong transform fails **silently** — the row just renders English.

**2. English says exactly what the code registers — 0 drift across all 52 rows.** English is served from the catalog too, so drift means English users read stale text while the source looks correct. The 23 static rows were compared against the verbatim literal in `panelSettingsList()`; the 29 factory rows have no literal in the source (their English is built from a template), so they were reconstructed from that same template plus `BACKEND_TEXT`:

```
factory rows compared: 29
static rows compared:  23
drift: 0
```

**3. Guards verified against the defect, not assumed.**

- `settings-i18n-keys` (new here) — mutation-tested: a typo'd key, a dropped row, and a blank tooltip each fail it; baseline and restore pass. Its id extractor finds exactly 52, so the assertions are not vacuous.
- The module-parse guard that landed in `04d020bf` — re-nesting the `lib/i18n.js` import into the specifier list above it fails the guard; the healthy tree passes. That is why the equivalent guard originally on this branch was dropped rather than kept alongside it.

**4. Suite state:** `npm run test:unit` 3716 pass / 0 fail (includes the Zod locale gate — `locales OK`, settings.json 100% translated in ko/zh/ja); `npm run test:e2e:list` 68 tests in 29 files; `npm run typecheck` clean.

artokun:
### Round-2 review follow-ups

Four findings addressed, one deferred, one reported upstream.

**Fixed — an English defect that was about to be frozen into four languages.** `tokenSetting`'s tooltip template hardcoded `API` in front of `noun`, but `noun` is already `"API key"` for the two key rows — so English users read *"Securely store your OpenRouter API API key."* today. This surfaced only because the catalog serves **English too**, and my translations had faithfully reproduced the doubling into ko/zh/ja. Fixed at the source (prefix only when missing, so `your CivitAI API token` still reads right) and regenerated all four catalogs. 8 strings corrected.

**Fixed — a locale seam this change created.** `d.toLocaleDateString()` with no argument formats in the *browser's* locale while the sentence around it now renders in the *panel's*: a Korean panel in an `en-US` browser read `🔒 8/10/2026 설정됨`. Now `toLocaleDateString(currentLocale())`.

**Fixed — CJK spacing.** `安全地保存你的 自定义端点…` and `カスタムエンドポイント の API` carried an ASCII space inherited from the English `your ${friendly}` templating. Spacing is now decided per term (a space against a Latin name, none against a CJK one), so `OpenRouter の API キー` and `カスタムエンドポイントの API キー` are both right.

**Fixed — the test could truncate silently.** `settings-i18n-keys` slices the id set out of the source between `function panelSettingsList() {` and the next column-0 `}`; a stray column-0 `}` inside would drop ids and leave both assertions vacuous. The `>= 50` floor only catches a total collapse. The slice is now asserted to reach the last registered row and the returned array's terminator — verified by injecting a `}` that drops **exactly one** id (51 left, comfortably above the floor): fails now, would have passed before.

**Deferred, with reasoning.** Render-fn text (token buttons, `🔒 set …`, store links) is computed when ComfyUI mounts the Settings dialog, so changing the panel language leaves that dialog half-translated until it is reopened — ComfyUI's own rows are reactive vue-i18n, ours are plain DOM. Re-mounting third-party dialog widgets is out of scope for this unit and riskier than the seam. The `comfyui-mcp.language` tooltip already states the contract ("Takes effect when the panel is reopened or reloaded"), which covers it honestly.

**Reported, not fixed here — `scripts/i18n-apply.mjs` is not idempotent.** Its "already converted" guard is `/\btr\(\s*$/`, which only matches when the literal is the *first* argument, but the script always emits the English text as the *second* (`tr("key", "text")`) — so it never recognises its own output. Measured on the current tree: `node scripts/i18n-apply.mjs --dry` reports **"would apply: 319"**. A second `--write` would nest 319 call sites as `tr("k", tr("k2", "text"))`. That file belongs to the base (`04d020bf`) and this PR deliberately no longer touches it, so flagging rather than editing — but it is a live footgun for the rest of the fan-out.

Suite after all of the above: **3716 pass / 0 fail**, `locales OK`, typecheck clean, 0 English drift across all 52 rows, 18/18 end-to-end lookup probes still resolve translated.

Issue #1014, opened 2026-08-10
A saved workflow loses its identity across a reconnect — adoption needs per-load sanitizer provenance


artokun:
Split out of #1001 (bug 1). #1013 fixes bug 1b only; this half is not fixed and the
reason is a design constraint worth recording rather than a missing patch.

## The wedge

A workflow created in-session and then SAVED has a real path. After a ComfyUI
drop/reconnect the canvas is on the correct workflow and the panel says so — yet every
mutating command is refused for `no trusted identity`, while reads keep working.

`establishedWorkflowReplyIdentity` publishes only when the uuid is already in the
live-object WeakMap, because a reply must never be able to initialize an identity for a
canvas the panel has never seen (#716). A reconnect can REPLACE the active
ComfyWorkflow object; the successor is not in that map, so the hello publishes no
`workflow_uuid`.

The identity is derivable at the moment it goes missing — the reporter's own recovery
was `panel_open_workflow` on the workflow that was ALREADY active, and an open mints
through `workflowStableUuid`, which seeds the map from the path.

## Why the obvious fix was built and then removed

Adoption from the persisted path→uuid alias was implemented, reviewed, and pulled.

- **A path is not proof of incarnation.** Delete workflow A, save a new workflow B at
  A's former path, reconnect: a path-only rule hands B the identity A established, and a
  stale A-scoped command then passes the fence against B.
- **Corroborating with the panel's root tag does not close it.** The tag TRAVELS WITH
  THE GRAPH. A copy of A — tag included — saved at A's former path makes the alias and
  the tag agree on A's uuid for what is now a different workflow. Two correlated
  records are not two independent ones.
- **`_loadGraphDataForkInstalled` is capability evidence, not provenance.** It proves
  the creation-boundary sanitizer is installed NOW, not that this mounted graph was
  created or replaced through it. A copied tagged graph loaded before installation keeps
  its tag, and once the flag flips true every check passes.

## What would make it sound

A **per-load marker minted per session** — not persisted, not derived from anything a
file can carry — stamped by the `loadGraphData` wrapper on every load it processes.
Adoption then requires the mounted root to carry the CURRENT session's marker, which a
copied file cannot fake and a pre-installation load does not have.

That changes what every load writes into `extra`, with implications for what a save
serializes, so it wants its own design pass rather than riding along in a fix for a
different defect.

## Until then

`panel_open_workflow` on the already-active workflow is the recovery, and it works
[...]

PR #1055, opened 2026-08-11
wip(1014): per-load provenance — design note, not a fix

artokun:
**Do not merge.** Left open as the record of a design pass that got further than the issue
did and then stopped at a specific, named hole.

**What it establishes.** The reason #1014 was parked — that a per-load marker "changes what
every load writes into `extra`, with implications for what a save serializes" — is
avoidable. Key the marker off the live root graph OBJECT instead. Nothing new is serialized,
and object identity is something a file fundamentally cannot carry, so it is also the
stronger form.

**Where it stops.** The marker proves *"this root was processed by this session's wrapper"*,
but adoption needs *"this root's tag was minted by this session's sanitizer"*. #570 mints on
creations and deliberately not on reload-restore — so marking every load lets the
delete-A/save-B-at-A's-path case through (B is marked, B's file carries A's tag), while
marking only minted loads leaves the reconnect case — the one this issue is about —
unmarked. The two requirements conflict precisely on the path that matters.

Shipping a fence-widening change on an argument with that hole in it would be worse than the
wedge it fixes: the wedge refuses mutations, and a wrong adoption lets a stale command
through against the wrong workflow.

**Direction worth trying next**, unbuilt and unmeasured: hand the identity forward from the
predecessor object at the moment a reconnect replaces it. The panel knows both sides then,
and an identity passed from an object it already established involves no file at all — so it
needs no marker and no tag.

Refs #1014


artokun:
Closing this rather than leaving a do-not-merge branch open indefinitely. **Nothing is lost:
its entire content is a design record, and that record now lives on #1014 where the next
person will actually look for it.**

What it established, in short:

- The reason #1014 was parked — that a per-load marker "changes what every load writes into
  `extra`, with implications for what a save serializes" — is **avoidable**. Key the marker
  off the live root graph OBJECT instead. Nothing new is serialized, and object identity is
  something a file fundamentally cannot carry, so it is also the stronger form.
- But the marker proves *"this root was processed by this session's wrapper"*, and adoption
  needs *"this root's tag was MINTED by this session's sanitizer"*. #570 mints on creations
  and deliberately not on reload-restore, so marking every load lets the
  delete-A/save-B-at-A's-path case through, while marking only minted loads leaves the
  reconnect case — the one the issue is about — unmarked. **The two requirements conflict
  precisely on the path that matters.**
- The direction worth trying instead, unbuilt: hand the identity forward from the predecessor
  object at the moment a reconnect replaces it. The panel knows both sides then, and an
  identity passed from an object it already established involves no file at all — so it needs
  neither a marker nor a tag.

No code is being discarded; the branch never contained a working implementation, only the
argument for why the obvious one is unsound. Reopen or re-branch freely if the predecessor
handoff turns out to be the answer.

Issue #968, opened 2026-08-10
Instance fence reports bound to requested workflow after panel_open_workflow, but graph commands still target the previously active tab (silent wrong-graph edits) after reconnect storm


artokun:
### What happened / root cause
During a session with repeated orchestrator connection drops/reconnects, `panel_open_workflow(path:"idle-rpg-char-anima-v1")` returned `opened` with `workflow_uuid 53b8038f-1b22-45ed-be9c-7d4fab076c78` (open_seq 2, then again open_seq 3), and `panel_set_workflow_target(mode:"current")` reported `graph_binding: bound` / `already_current` on that same uuid. However the live canvas continued to serve the PREVIOUSLY active workflow (`idle-rpg-char-sheet-qwen-edit-v1`, a 16-node Qwen Edit graph): `panel_graph_outline` returned the sheet graph, and `panel_add_node` allocated node id 17 (next id of the 16-node sheet graph) and `panel_connect` from node 4 reported output `patched_model` (CFGNorm — a node type that exists only in the sheet graph; the requested workflow's node 4 is LoraLoaderModelOnly with output `MODEL`). Net effect: a silent wrong-graph edit — the exact failure mode the workflow-instance fence exists to fail closed against.

The on-disk JSON for the requested workflow was verified intact (contains `anima-base-v1`/`er_sde`, no `UnetLoaderGGUF`), so this is a frontend tab-activation/identity desync, not file corruption: the tab registry/fence adopted the requested workflow's identity while the canvas never actually switched tabs — most likely the reconnect/tab-restore logic racing panel_open_workflow's activation.

### Steps to reproduce
1. Open two saved workflows in tabs (A and B); A active.
2. Cause repeated orchestrator reconnects (reconnect storm) while the session works against A.
3. `panel_open_workflow(B)` -> returns opened + B's uuid.
4. `panel_set_workflow_target(current)` -> `already_current` on B's uuid.
5. `panel_graph_outline` / `panel_add_node` -> operate on A's graph (A's next node id, A's node types), no mismatch error raised.

### Exact error (scrubbed)
No error raised — that is the bug. Evidence trail:
- panel_open_workflow -> {"opened":{"path":"workflows/idle-rpg-char-anima-v1.json"},"workflow_uuid":"53b8038f-1b22-45ed-be9c-7d4fab076c78","modified":true,"open_seq":2} (and open_seq 3 on retry)
[...]

PR #1056, opened 2026-08-11
feat(968): record WHAT last moved the active workflow

artokun:
**Do not merge.** The wiring is complete and the suite is green, but codex found four ways
the diagnostic can LIE — and a diagnostic that lies is worse than none, because #968 is
exactly a case where people are trying to tell a stale binding from a fresh one.

## Confirmed good (codex verified, do not re-litigate)

- **It is genuinely diagnostic-only.** The fence decision precedes observation and throws
  unconditionally afterwards; nothing here can change whether a command is accepted.
- **A refusal with no note is byte-identical.** `movedNote` defaults to null and only ever
  appends a non-blank line.

## Open findings — all four make the record wrong, none are cosmetic

1. **P1 — a failed or superseded open can mislabel a LATER external move as panel-made.**
   `workflow_open` claims before `await s.openWorkflow()`, and a failure only records
   `openFailed`; the claim is never cleared. The observer keeps it while the key is
   unchanged, then applies it to the next different key. My comment saying an unfulfilled
   claim is discarded by the next observation is **false**. This mislabels precisely the
   case the issue is hunting.
2. **P1 — `NEW_EXECUTOR` is declared but never claimed.** `workflow_new` creates a tab
   without claiming, so its move is reported as `external` — the opposite mislabel. The test
   inserts that cause directly into the recorder, so it proves the string, not the wiring.
3. **P2 — `from` is not reliably the previous ACTUAL workflow.** Observation happens only on
   the refusal path, so several real switches between refusals collapse into one invented
   A→C transition.
4. **P2 — "no active workflow" is recorded as a tab id.** `workflowTabId(null)` falls back to
   `getTabId()`, a browser-session id, so the observer's no-destination branch is never
   reached in that state and a browser tab id can be reported as though it were a workflow.

## What each needs

(1) clear the claim on every failure path and expire it if no move follows; (2) either claim
in `workflow_new` or drop the cause from the enum until it is wired; (3) either observe on a
path that sees every switch, or say in the note that `from` is the last OBSERVED workflow
rather than the previous one; (4) treat a null active workflow as no destination rather than
letting the tab-id fallback stand in for one.

(3) is the interesting one: fixing it properly may mean the observer cannot live on the
refusal path at all, which reopens the call-site question this branch worked around.

Refs #968

### DIFF
--- a/browser_tests/unit/active-workflow-provenance.test.mjs
+++ b/browser_tests/unit/active-workflow-provenance.test.mjs
@@ -1,0 +1,130 @@
+/**
+ * #968 — WHAT last moved the active workflow.
+ *
+ * The issue is three reports of the fence saying "bound to the requested workflow" while
+ * graph commands keep hitting the previous one. They have not converged because, after the
+ * fact, a STALE binding and a FRESH one are the same observation: "the active workflow is
+ * X", with nothing recording how it came to be X.
+ *
+ * Ruled out before building this, so it is not another guess: `panel_open_workflow` forces
+ * the repaint and verifies it, both its skip paths fail closed, and the report where
+ * `panel_run` queued the wrong workflow was on a build that had both protections.
+ *
+ * DIAGNOSTIC ONLY — nothing here decides whether a command may run. These tests pin that
+ * property as hard as they pin the behaviour, because widening trust on an unknown entry
+ * route is how a refusal becomes the silent wrong-graph edit the issue reports.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+
+import {
+  MOVE_CAUSES,
+  createActiveWorkflowProvenance,
+} from "../../web/js/lib/active-workflow-provenance.js";
+
+test("#968 a move the panel did NOT make is the one worth naming", () => {
+  const p = createActiveWorkflowProvenance();
+  p.record({ cause: MOVE_CAUSES.OPEN_EXECUTOR, from: "wf:a.json", to: "wf:b.json", at: 1, detail: "open_seq 2" });
+  p.record({ cause: MOVE_CAUSES.EXTERNAL, from: "wf:b.json", to: "wf:c.json", at: 2 });
+
+  const note = p.describeLast();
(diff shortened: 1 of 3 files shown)


artokun:
**Codex second pass: NO-SHIP, and I agree. Recording the verdict and the fix it names.**

## Confirmed fixed

- **(1)** every observation clears the claim before comparing keys, and a thrown native open
  releases it explicitly;
- **(3)** the note reads `to X (last seen as Y)`;
- **(4)** a null active state is no destination and never reaches `workflowTabId(null)`.
- **No acceptance regression**: the recorder runs only after the target predicate has already
  refused, and with no non-blank note the refusal is byte-for-byte unchanged.

## Why it still must not ship

> The remaining blind spot makes a known panel action (`panel_new_workflow`) assert "the
> panel did not make that move." That is not merely incomplete telemetry; it gives a false
> exclusion precisely on one of the reported entry paths. For a diagnostic intended to
> identify the desync's source, confidently blaming an external actor when the panel created
> the tab is worse than providing no attribution.

That is right, and it is the same standard applied to every other claim in this repo: a
record that asserts more than the mechanism establishes is worse than one that says less.

## The fix, which is in codex's own phrasing — "or reported as unknown"

The false exclusion comes from the DEFAULT, not from the missing claim. An unclaimed move is
currently recorded as `external`, which asserts the panel did not do it. It should be
recorded as **`unknown`**, which asserts nothing:

> The active workflow last moved to X (last seen as Y), and NO PANEL COMMAND CLAIMED IT.
> That does not prove the panel did not do it — not every executor can register a claim, so
> this covers a `panel_new_workflow` as well as a tab click, a reconnect restore or a file
> reopened at a new path. What it does establish is that a binding taken before this move is
> stale.

That removes the false exclusion **without** needing a claim inside `workflow_new` — which
is what ran into the #606/#708 guards. The "binding taken before this is stale" half, which
is the actually useful part, survives intact.

I started that change and applied it only partially before stopping; the branch is reverted
to its last green commit rather than left half-converted. It is a contained edit — rename the
cause, invert the default, rewrite the one sentence, re-premise the two tests that assert the
old wording.

PR #1059, opened 2026-08-11
fix(646): bound a duplicate's wait at the caller's own deadline

artokun:
**Do not merge.** The mechanism is sound and the double-apply guarantee holds, but the bound
is set above the deadline it exists to beat — which makes the whole change a no-op for the
reported failure.

## What survived review

- **Double-apply: PASS.** The helper races `Promise.resolve(prior)` against a timer and holds
  no ledger or resolver reference. A losing timer, or the original settling after the timer
  fired, only resolves its own race participant. The entry is still settled solely by
  `settleRid`, which is once-only.
- **Reply shape: PASS.** `{rid, ok:false, error}` composes with the retry-rid rewrite, and the
  synthetic reply carries no user payload, so redaction would pass it through if journaled.
- **No false execution result.** The original still settles and replies on its own rid; if its
  socket died that reply is journaled and replayed through the existing liveness path.

## P0 — the bound is 25s and the caller's deadline is 20s

The report this is built on shows `workflow_list` timing out at **6000 ms** and `graph_query`
at **20000 ms**. A duplicate does not receive the synthesized reply until **25000 ms**. The
caller has already given up. This preserves the exact failure it was written to remove.

Worse, my own test encoded the contradiction: it asserts the bound must be *"at or under the
long ones (20s)"* and then accepts anything up to 30s. That is a false claim I wrote and did
not catch — recorded rather than quietly corrected, because it is the reason this got as far
as review.

## P1 — the bound mislabels commands that legitimately block on a human

Dedupe runs before dispatch, and `ask_user` / `request_secret` await human input
indefinitely. After 25s their duplicate receives a graph-oriented "still running, read the
graph" notice, when the truthful answer is "a person has not answered yet".

## What it actually needs

1. **A bound below the relevant caller deadline** — or an honest admission that the panel does
   not know the caller's deadline, and a mechanism that does not depend on guessing it.
2. **Interactive commands excluded**, or given a message that matches what they are waiting
   for.

(1) is the real design question: the panel cannot see the orchestrator's per-command timeout,
so any fixed number is a guess about someone else's contract. A protocol-level progress or
result-follow-up signal is the alternative codex names, and it is a larger change than this.

Refs #646

### DIFF
--- a/browser_tests/unit/duplicate-await-bound.test.mjs
+++ b/browser_tests/unit/duplicate-await-bound.test.mjs
@@ -37,25 +37,25 @@ function namedFunctionSource(src, name) {
 const buildHelper = () => {
   const fn = namedFunctionSource(SRC, "awaitDuplicateReply");
   assert.ok(fn, "awaitDuplicateReply not found");
-  const ms = SRC.match(/const DUPLICATE_AWAIT_MS = (\d+);/);
-  assert.ok(ms, "DUPLICATE_AWAIT_MS not found");
-  // Rebuilt with a tiny bound so the timeout path is testable in milliseconds.
-  return new Function(`const DUPLICATE_AWAIT_MS = 20; ${fn}; return awaitDuplicateReply;`)();
+  const margin = SRC.match(/const DUPLICATE_AWAIT_MARGIN_MS = (\d+);/);
+  assert.ok(margin, "DUPLICATE_AWAIT_MARGIN_MS not found");
+  // Rebuilt with a tiny margin so a millisecond-scale caller deadline still leaves a budget.
+  return new Function(`const DUPLICATE_AWAIT_MARGIN_MS = 5; ${fn}; return awaitDuplicateReply;`)();
 };
 
 test("#646 a settled original is returned UNCHANGED — the bound never rewrites a real reply", async () => {
   const awaitDuplicateReply = buildHelper();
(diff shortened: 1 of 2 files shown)


artokun:
**Found where the deadline lives, which is the blocker this PR stalled on.**

`ui-bridge.ts:3632` computes `timeoutMs` per command, and `ui-bridge.ts:4003` builds the
frame as `{ ...cmd, rid }` (+ optional `workflow_uuid`) — the timeout is never sent. So the
panel genuinely cannot know the caller's deadline, and every fixed bound is a guess about
another process's contract. That is exactly how 25s ended up preserving a 20s failure.

**The fix this branch needs is therefore two-part**, and the panel half becomes trivial and
guess-free:

1. orchestrator — add `timeout_ms` to the frame (additive; older panels ignore it);
2. panel — bound the duplicate await at `msg.timeout_ms` minus a margin **when present**, and
   fall back to today's unbounded wait when absent, so nothing regresses against an
   orchestrator that does not send it.

That also disposes of the `ask_user` / `request_secret` objection without a special case:
they carry long or absent deadlines, so a short bound never fires for them.

Leaving this branch as the record of the mechanism and the double-apply argument (both of
which survived review), with the bound itself unshipped. It should not merge as-is — codex is
right that at 25s it changes nothing for the reported failure.

PR #1060, opened 2026-08-11
chore(release): 0.13.7

artokun:
Reconciles 0.13.7 into protected main. Tag `v0.13.7` is pushed.

**artokun/comfyui-mcp#938** — remove ONE dynamic widget row.

Verified against this machine's live ComfyUI: the real Power Lora Loader def declares only `model`/`clip`, so `lora_N` classifies as dynamic; removal runs through rgthree's actual `removeWidget` override and leaves `lora_1, lora_3` (deliberately not renumbered); `KSampler.steps` is refused as backend-declared; unreadable node definitions refuse rather than authorize.


artokun:
Heads-up before this lands: **main moved after this branch was cut, so the changelog is
missing one entry.**

`024df78b fix(646): bound a duplicate's wait at the caller's own deadline (#1059)` merged
after 0.13.7's changelog was written. As it stands this release ships that code with no note
for it — the 0.13.7 section covers #938 only.

Suggested addition (or just re-run `gen-changelog-json.mjs` after regenerating the section):

> ### Fixed
> - a DUPLICATE delivery of a request the panel is already running no longer waits forever
>   on it. In-flight ledger entries are never evicted (dropping one would let a replay apply
>   the same mutation twice), so an executor that never returned left every redelivery
>   awaiting a promise that could not resolve, and the panel answered nothing at all. The
>   duplicate now gets a real answer: still running, nothing applied twice, do not retry.
> - the wait is bounded by the CALLER's own deadline, never a number the panel invents — and
>   the panel cannot see that deadline today, so absent it the behaviour is unchanged. This
>   half activates when the orchestrator sends the timeout it already computes at
>   `ui-bridge.ts:3632` but omits from the frame at `:4003`.

Not touching your branch — you may prefer to regenerate rather than hand-edit. I have left
this release to you deliberately since it is yours; flagging only because the omission would
be invisible after merge.

Issue #1062, opened 2026-08-11
panel_add_node cannot add core SaveGLB: union input type MESH,FILE_3D_* misclassified as widget-needing


artokun:
### What happened / root cause

`panel_add_node({class_type:"SaveGLB"})` always fails on a stock ComfyUI install with:

```
Cannot add "SaveGLB": 1 required input type had no widget after 5.0s waiting for node extensions to register.
  - input "mesh" (declared type "MESH,FILE_3D_GLB,FILE_3D_GLTF,FILE_3D_OBJ,FILE_3D_FBX,FILE_3D_STL,FILE_3D_USDZ,FILE_3D_PLY,FILE_3D_SPLAT,FILE_3D_SPZ,FILE_3D_KSPLAT,FILE_3D_SPLAT_ANY,FILE_3D_POINT_CLOUD_ANY,FILE_3D"): no installed node outputs "..." and no frontend widget is registered for it.
```

Core `SaveGLB` declares its `mesh` input as a comma-joined UNION of `MESH` plus a list of `FILE_3D_*` types. The add-node precheck resolves that union as a single opaque type string: it looks for a node whose output type equals the FULL joined string, finds none (the producer `VoxelToMesh` outputs plain `MESH`), and then waits 5s for a widget that never registers, so the add is refused. The error text itself says a comma-joined union of socket types is supposed to be added immediately — that path evidently does not SPLIT the union before matching, so a union whose FIRST member (`MESH`) is a real socket type is treated as unknown.

Net effect: SaveGLB — the only core node that writes a .glb — cannot be placed on the canvas by an agent at all, which blocks building any image-to-3D / text-to-3D workflow via panel tools.

### Steps to reproduce

1. Stock ComfyUI (no extra 3D packs needed; SaveGLB / VoxelToMesh / Hunyuan3D nodes are core).
2. `panel_add_node({class_type:"SaveGLB"})`.
3. Fails as above. `panel_refresh_nodes` then retry: same failure. Retry after a workflow switch: same failure.

Note the same graph loads fine with SaveGLB present when the workflow JSON is loaded from disk — so only the ADD path is affected, not the node itself.

### Exact error (scrubbed)

See above (verbatim, no secrets present).

### Fix

applied locally: no   upstream-only: yes

[...]

PR #1063, opened 2026-08-11
fix(1062): treat a comma-joined input type as a UNION, not one opaque type

artokun:
Claiming #1062.

Core `SaveGLB` declares its `mesh` input as `MESH,FILE_3D_GLB,FILE_3D_GLTF,…` — a comma-joined UNION. The add-node precheck resolves that as a single type string, looks for a node whose output equals the whole joined string, finds none (the real producer `VoxelToMesh` outputs plain `MESH`), then waits 5s for a widget that will never register and fails.

So a stock ComfyUI node cannot be added at all, and the error blames a missing widget rather than the classification.

Planned: split a declared type on commas and satisfy the input if ANY member has a producer or a widget — with the union spelled out in the failure text when none does, so the next such node reports which member it looked for rather than a 300-character joined string.

Draft first, per the loop. Codex review before merge.

Refs #1062


artokun:
Closing: the one-quantifier change is unsafe. An empty config `{}` counts as socket-shaped, so `["MESH,ZIPN_STYLE_GALLERY", {}]` would be admitted under `some` with its unregistered custom widget skipped — a #580 false accept. `every` prevents a proven socket member from laundering an unproven custom-widget member. Full reasoning on #1062; the fix needs a narrower trusted-union rule or a better widget-vs-socket discriminator.

Issue #1066, opened 2026-08-11
panel_save_workflow 500s forever on a URL-derived temp tab (directoryOf accepts a URL as a directory) + panel_new_workflow wedges the 0.11.43 fence unrecoverably


artokun:
### What happened / root cause

Two distinct defects hit in one session on a live ComfyUI canvas.

**(1) `panel_save_workflow` fails with a 500 for any workflow tab whose `directory` is a URL.**

ComfyUI mints a TEMPORARY workflow whose `path` is the URL an asset was opened from. Opening/dropping an output image produces a tab pathed like:

    workflows/http://127.0.0.1:8188/api/view?filename=x.png&type=output&subfolder=anima%5Cpreset%5C10-framing-basic

(the user had several such tabs open, titled `view?filename=...`). Renaming that tab replaces only its FILENAME, so the URL survives as the tab directory.

`directoryOf()` in `web/js/lib/workflow-save.js` only redirects a directory to the workflows root when it is empty or matches `isExternalWorkflowPath()` (drive letter / leading separator). A URL matches neither: `http://...` has no `^[a-zA-Z]:` (the colon is at index 4) and no leading slash. So the directory was accepted verbatim and `targetPath()` built:

    workflows/http://127.0.0.1:8188/api/AnimaPltMg001.json

which /userdata rejects. The tab could not be saved under ANY name — from the panel or from ComfyUI own Save/Save-As.

**(2) `panel_new_workflow` permanently wedges the session fence on panel 0.11.43.**

Working around (1) by copying the nodes into a fresh tab, `panel_new_workflow` created the tab but could not read back the new canvas identity (the panel only publishes the workflow uuid on the reply from 0.11.45+). The session stayed fenced to the OLD workflow instance and every graph command was refused from then on. The documented recoveries both failed:

- `panel_set_workflow_target({mode:"current"})` -> could not read the canvas (the fence blocks the very read it needs).
- `panel_open_workflow(<a real persisted path>)` -> ran and proved the canvas is bound, but reported `FENCE: NOT cleared (unreadable)` because the panel publishes no workflow_uuid on that verdict, and told us to try `panel_list_workflows` — which is itself fenced and failed with the same instance mismatch.

Net effect: an unrecoverable-in-session dead end whose only exit is a manual browser refresh. The escape hatch is very narrow: the fence-exempt `workflow_open` cannot address an UNSAVED tab (no path), so the tab the user actually wants is unreachable by the one command that still works.

[...]

PR #1073, opened 2026-08-11
wip(1066b): reverted — the 'proven absence' was disproven by my own comment

artokun:
**Reverted. Do not merge.** The claim this rested on is false, and the counterexample was
already in the file — written by me, hours earlier, in 0.13.9:

> on POSIX a managed directory may syntactically contain `://` (`workflows/notes://draft`)

I documented the exact case that refutes the theorem, then asserted the theorem anyway. A
path containing `://` does **not** prove no `/userdata` file can exist at it, so
`groundingIsSafe` must not return `true` on that basis.

Two further findings, both correct:

- **`isPersisted === true` is not a sufficient guard.** The surrounding code already
  recognises that persisted files can be flagged temporary after the open race — which is
  exactly the window where this early return would have bypassed the disk fact.
- The commit message claimed 4 tests where the diff added 3, and "does not need the oracle"
  was imprecise: a callable oracle is still required, it is merely not invoked.

**No live data loss was shipped**, and I want to be exact about why: review traced the
downstream path and found no route from this return to an in-place overwrite or source
rename — the save layer redirects the target and re-probes the source. That is the save
layer catching it, not this change being safe. It changed a SAFETY predicate to return true
on a condition it does not establish and wrote the false proof into a comment, which is the
part that would have outlived the code.

**What defect (2) actually needs:** a signal that separates a URL-derived TEMPORARY tab from
a real file in a POSIX folder whose name contains `://`. The path shape alone cannot do it —
that is the whole finding. Candidates worth assessing: the tab's own `isTemporary`
provenance, or asking the oracle a question it *can* answer for this shape rather than
skipping it.

Defect (1) remains shipped in 0.13.9 and is unaffected.

Refs #1066


artokun:
Closing as obsolete. This branch was the RECORD of a reverted attempt at #1066 defect (2), kept open so the reasoning would not be lost — it was never mergeable and its own title says so.

Defect (2) shipped in **0.14.1** via #1075, by a different route than anything attempted here: rather than trying to prove the URL-derived source absent, it scoped the #226 guard to the hazard it names. The three signals eliminated along the way — the `://` path shape, the `isTemporary` flag, and the disk oracle (which answers **500**, not 404, measured against a live ComfyUI) — are recorded on #1066 itself, which is the durable place for them.

#1066 is closed. Nothing here is needed any more.

Issue #1065, opened 2026-08-11
Relay panel handshake omits browser Origin so workflow fence cannot bind


artokun:
### What happened / root cause
A connected sidebar tab is readable, but all graph mutations fail because the connection carries no server-observed Origin while the live workflow identity is origin-bound. panel_set_workflow_target({mode:'current'}) reads the identity but the orchestrator validator refuses to adopt it. The tool explicitly identifies the relay protocol as needing to forward the browser handshake Origin.

### Steps to reproduce
1. Connect the sidebar through the relay backend.
2. Open a workflow and call panel_set_workflow_target({mode:'current'}).
3. Observe graph reads may work but every mutation remains fenced.

### Exact error
`Read the live canvas identity (...) but could NOT adopt it ... connection carries no server-observed Origin, and the workflow identity is bound to one ... relay protocol has to forward the browser's handshake Origin for this path to work at all.`

### Fix
applied locally: no; upstream-only: yes. Forward the browser handshake Origin across the relay protocol and include it in the server-observed connection metadata used by workflow identity fencing.

### Environment
Windows 11; comfyui-mcp 0.50.114; connected sidebar panel version unknown; panel/orchestrator required floor 0.13.0. Headless COMFYUI_URL is intentionally not the browser panel origin.

artokun:
**Confirming "upstream-only" from the panel side, with the reason it cannot be fixed here —
which also explains why it must not be.**

The panel already sends its origin in the handshake. `comfyuiUrlForAgent()` hands the
orchestrator either the advanced Remote-URL override or, failing that,
`window.location.origin` — the URL the browser was actually served from. So the information
exists and is already on the wire.

**But that is a CLAIM, not an OBSERVATION, and the fence deliberately requires the second.**
A value in the hello payload is whatever the client typed; the `Origin` the server reads off
the WebSocket handshake is attested by the browser and cannot be set by page script. Workflow
identity is bound to the observed one precisely because a forgeable origin would let any
connection assert ownership of any workflow — which is the failure the fence exists to
prevent.

So the panel cannot substitute for the missing header, and adding a louder client-side claim
would be worse than the current refusal: it would look like a fix while removing the
property that makes the binding meaningful. The refusal you quoted is correct, and its
wording already names the right remedy — the relay has to forward the browser's handshake
`Origin` into the server-observed connection metadata.

**One thing worth checking on the relay side**, since it decides how much work this is: the
[...]

PR #1077, opened 2026-08-11
fix(1065): relay handshake omits the browser Origin so the workflow fence cannot bind

artokun:
**Claiming #1065** (P1). A sidebar connected through the **relay** backend is readable but every
graph mutation stays fenced: the workflow identity is bound to a server-observed Origin, and a
relay socket has no handshake for the orchestrator to observe one on.

## Initial findings

**This is a two-repo fix, and the panel owns the half that does not exist yet.** The panel's
handshake carries no origin at all — `buildHelloPayload` (`web/js/lib/session-rebind.js:290`)
builds `tab_id`, `title`, `panel_version`, `backend`, `tab_session_id`, `comfyui_url`,
`comfyui_path`, the capability flags — and nothing about the page's own origin. So even a relay
willing to forward one has nothing to forward.

The orchestrator half is already articulated on the other side:

- `comfyui-mcp/src/services/secure-bridge.ts:263` — *"server-observed Origin, and a relay socket
  has no handshake to observe one"*
- `comfyui-mcp/src/orchestrator/fence-refusal.ts:70` — the exact refusal the reporter pasted
- `comfyui-mcp/src/__tests__/orchestrator/fence-refusal-reason.test.ts` already pins that reason

**The design question this turns on, before any code:** an origin the panel puts in its own
hello is *self-asserted*, not server-observed. Those are not interchangeable, and the fence's
whole purpose is to bind identity to something the server saw. The panel already carries
self-asserted claims (`enforces_workflow_stamp`, `accepts_agent_notes`) and the file is explicit
about what that is worth — *"self-asserted, so a modified panel can spoof it, but that only
leaks the user's own workflow into their own workflow (self-attack). No attestation."*

So the question is whether the origin belongs in that same honest-local-correctness class, or
whether the relay must observe and stamp it instead — which would put the whole fix in the relay
and leave this repo unchanged. The reporter's own note says *upstream-only: yes*. I want to
establish which of those is actually true rather than assume, because picking wrong here means
either a fence that a modified panel can talk its way past, or a fix in the wrong repo.

Findings to follow in this PR.

Refs #1065


artokun:
Closing unmerged — the premise was wrong, and checking it first would have caught that.

I opened this to add the browser origin to the panel's `hello`. The orchestrator explicitly forbids exactly that: `workflowIdentityParts` requires a **server-observed (unspoofable)** origin and names `hello.comfyui_url` as the thing it must never be. A panel-asserted origin would have satisfied the presence gate with a value the panel picked for itself — which is the one case that gate exists to refuse.

The real defect is upstream and is a stale diagnostic, not a missing capability: comfyui-mcp #1240 fixed the relay origin in v0.50.67, but only updated the refusal's *reason* text. The *remedy* text in `panel-tools.ts` still tells users the relay backend is structurally broken and asks them to report that the handshake Origin needs forwarding — which is the sentence the reporter quoted as this issue's proposed fix.

Full analysis on #1065. The panel repo needs no change.

PR #1082, opened 2026-08-11
fix(1339): rgthree seeds are invisible to the batch-repeat warning

artokun:
Upstream half of artokun/comfyui-mcp#1339.

**Measured** on the live rig with `/prompt` intercepted so nothing queued ([full results](https://github.com/artokun/comfyui-mcp/issues/1339#issuecomment-5257682765)):

| run | posted seeds |
|---|---|
| `app.queuePrompt(0, 3)` | 742732484087382, 27923863108977, 114376582494196 — differ |
| `app.queuePrompt(0, 3, ["29"])` | 77161535712866 ×3 — repeat |
| rgthree handler driven 3× with `seed: -1` | 1028465986822020, 98533269447704, 557498712106716 — differ |

`findRepeatingControlWidgets` matches widgets named `control_after_generate`. **rgthree deletes that widget** — a live `Seed (rgthree)` node carries `seed` plus three buttons and nothing else — so the #988 disclosure is structurally blind to the most widely used custom seed node. The reporter ran `batch_count: 10`, got ten identical images, and was told `queued: true` with ten prompt ids.

**What this will NOT do:** claim an rgthree seed repeats because the batch is scoped. The measurement above says it does not — rgthree substitutes in its own `api.queuePrompt` patch, which fires per item even when scoped. Extending the existing warning to cover it would be a confident wrong sentence, which is the failure mode #988's module was written to avoid.

**What it should say instead** is the thing that actually distinguishes the two outcomes: an rgthree seed is *armed* (widget holds a `-1`/`-2`/`-3` sentinel → a fresh value per item) or *fixed* (a concrete number → every item uses it). For a batch of more than one, that is the fact the user needs and currently cannot see.

Refs artokun/comfyui-mcp#1339.

### DIFF
--- a/browser_tests/unit/scoped-batch-seed.test.mjs
+++ b/browser_tests/unit/scoped-batch-seed.test.mjs
@@ -270,6 +270,17 @@ test("#1339 — an ARMED node with a degenerate random range still repeats", () 
   }
 });
 
+test("#1339 — a MUTED or BYPASSED seed node is not named", () => {
+  // rgthree's handler returns early for exactly these modes, so the node substitutes
+  // nothing and contributes nothing. Naming it points the user at a node that is not in
+  // the run, in a warning whose entire value is naming the right one.
+  for (const mode of [2, 4]) {
+    assert.deepEqual(findRgthreeSeedNodes([{ ...rgthreeSeed(649, 12345), mode }]), [], `mode ${mode}`);
+  }
+  // …and an ordinary mode (0 = ALWAYS) is still reported.
+  assert.equal(findRgthreeSeedNodes([{ ...rgthreeSeed(649, 12345), mode: 0 }]).length, 1);
+});
+
 test("#1339 — silent for a batch of one, where a repeated seed is not a surprise", () => {
   const found = findRgthreeSeedNodes([rgthreeSeed(649, 12345)]);
   assert.equal(rgthreeFixedSeedNote(found, 1), "");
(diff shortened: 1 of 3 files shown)


artokun:
**Review status, disclosed:** codex reviewed round 1 (findings accepted and fixed). The round-2 re-review **wedged** — nine minutes silent in `starting` phase, then the companion exited 1 without producing anything — so I cancelled it and substituted a Claude adversarial review against the same probe list. Calling that out rather than implying an independent second opinion.

Working those probes myself found two more real defects, both mine:

**The step is part of the degeneracy condition.** `min >= max` was incomplete. rgthree draws `randomRange = (randomMax - randomMin) / (step / 10)` then `floor(random() * randomRange) * (step/10) + randomMin`, so `randomRange <= 1` returns `randomMin` every time. Measured over 200 draws:

```
min=0 max=5     step=100  ->   1 distinct value   <- min < max, looks healthy, repeats
min=0 max=5     step=1    ->  50 distinct
min=0 max=1e15  step=1    -> 200 distinct
min=5 max=5     step=1    ->   1 distinct
```

**A muted or bypassed node must not be named.** rgthree's handler returns early for mode 2 (NEVER) and 4 (bypass), so the node substitutes nothing and is not in the run; naming it points the user at the wrong node.

Also checked, nothing found:
- **Default fallback** — `Number(props.randomMin || 0)` mirrors rgthree's own `|| 0` / `|| 1125899906842624` exactly, including how it treats falsy and string values, so the note describes what the node will actually use.
- **Fragment assembly** — rendered the fixed-only, degenerate-only, mixed, and >5 (`and 3 more`) cases: no double spaces, no stray commas, each remedy matches the cases present, every variant terminates correctly.
- **Removing the link guard** cannot make the note wrong, because rgthree writes `outputInputs[seed]` from its own widget regardless of the link.

3858 panel unit tests green, typecheck clean, and every fix mutation-verified — reverting each guard kills a test.

PR #1123, opened 2026-08-12
fix(1448): say whether the workflow list was actually re-read, and stop blaming the folder

artokun:
Claims artokun/comfyui-mcp#1448.

`panel_open_workflow` refused with *"it isn't among the saved/open workflows even after a refresh. For a file outside the workflows folder, load it with panel_load_workflow"* — for a file the reporter had confirmed on disk **inside** that folder, twice.

Two defects in one sentence:

1. **It asserted a refresh it had not checked.** The lookup only refreshes when the frontend exposes `syncWorkflows`, and a throw from it was swallowed by a `console.warn` no agent session reads. On a frontend without the method, or when the call failed, the message claimed a re-read that never happened — fabricating the one fact that could have pointed somewhere.
2. **Its remedy named the wrong cause.** "For a file outside the workflows folder" reads as a diagnosis, and sent someone away from a file that was exactly where they thought it was.

**What I could not reproduce, and therefore do not claim to have fixed.** Measured on the live rig (ComfyUI 0.32.0 / frontend 1.48.7): `syncWorkflows()` genuinely re-reads — the store went 109 → 107, dropping two stale entries — every file on disk was present afterwards, and a bare `<name>.json` matches a saved record via its `key` (records expose `path` = `workflows/X.json`, `filename` = `X` with no extension, `key` = `X.json`). So the refresh path works on that build. This fixes the message, which was making a claim it had no evidence for in either direction.

### DIFF
--- a/browser_tests/unit/open-workflow-not-found.test.mjs
+++ b/browser_tests/unit/open-workflow-not-found.test.mjs
@@ -100,7 +100,16 @@ test("#1448 WIRING: the caller records the outcome instead of assuming one", () 
   assert.match(panel, /refresh = "unavailable"/, "a frontend without syncWorkflows");
   assert.match(panel, /refresh = "ok"/, "a successful re-read");
   assert.match(panel, /refresh = `failed: \$\{err\?\.message \?\? err\}`/, "a thrown re-read");
-  assert.match(panel, /openWorkflowNotFoundMessage\(\{ path, refresh, known: knownSelectorSample\(all\) \}\)/);
+  // The sample must be taken AFTER the refresh (codex review). A successful re-read
+  // removes stale entries — measured 109 -> 107 on a live rig — so a snapshot taken
+  // before it can offer a workflow that no longer exists as an example.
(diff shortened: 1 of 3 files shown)


artokun:
**Round 2 blocked** — `codex exec` returned `Selected model is at capacity` on four consecutive attempts. Staying unmerged; a self-review does not gate a merge here.

Round 1 found a real P2 and it is fixed: the failure sample was taken from a snapshot captured *before* the refresh. That is not hypothetical — while investigating this issue I measured `syncWorkflows()` taking the store from 109 to 107 entries, so a pre-refresh snapshot can offer, as an example of an addressable workflow, a file the re-read had just deleted. The sample is computed at the failure site now, from the current store, and both mutations (removing it, not passing it) fail.

State: full panel suite green (3957), typecheck clean, vocabulary gate exit 0. Nine mutations verified in total — claiming a re-read unconditionally, reporting a failed re-read as proof of absence, restoring the outside-the-folder diagnosis, offering an unsaved tab as an addressable example, unbounding the sample, either call-site branch misreporting its outcome, and the two above — with a capitalisation control surviving.

Issue #886, opened 2026-08-09
panel_open_workflow reports unconfirmed failure despite binding correct workflow


artokun:
After a managed restart, panel_open_workflow on workflows/3D Character Pose Style Generator.json returned isError with an UNCONFIRMED per-node-fields difference, while it also said the canvas was bound to the intended workflow and every same-id node/type was present. Subsequent panel_query_graph read the expected workflow correctly. This creates a false failure and requires redundant recovery. Windows 11, ComfyUI 0.29.0 local, RTX 5080 16GB.

Related (same root cause): #696, #825

artokun:
Triaged. Some of what you describe is deliberate and some of it is probably already fixed — and I can't tell which without your version numbers, because this exact wording has changed twice.

## The part that is by design

An open reports `UNCONFIRMED` when the **binding** is proven — the canvas IS the workflow you asked for — but the graph on it does not byte-match what was loaded. That is not the panel being timid. `loadGraphData` transforms the payload before the canvas serializes again: it grows every node to at least `computeSize()`, rewrites combo and `control_after_generate` values, may substitute a schema-validated copy, and runs every installed extension's `loadedGraphNode` hook on every node. A perfectly rebound canvas therefore differs from the bytes handed in.

Reporting that as a clean success was implemented here and then withdrawn, for a reason worth knowing: LiteGraph creates every node with its id and type and *then* configures each one, and `loadGraphData` catches a `configure()` failure and returns. A throw in that second pass leaves the **complete node id/type set** over nodes that silently lost their widget values. That is byte-for-byte the same observation as "the frontend normalized the fields", and no discriminator available to the panel separates them.

So your two observations — *"every same-id node/type was present"* and *"`panel_query_graph` read the expected workflow correctly"* — are exactly what a partly-applied load also looks like from here. Reporting success on that evidence would fabricate a success over possible data loss, which is the worse failure.

## The part that has already changed twice

Your complaint is really about the **cost** of it — a false failure requiring redundant recovery — and that is fair, and has been narrowed:

- **#696** — the reassurance used to be hostage to an allowlist of field names, so one unrecognised display flag from any node pack (`showAdvanced` was the reported one) sent a healthy open down the "the load may only have partly applied" path. Now the headline rests on the **node set** being intact, whatever fields differ, with the differing fields named so you can judge.
[...]

PR #1125, opened 2026-08-12
fix(886): a definitions difference that is only link renumbering is not a content change

artokun:
Claims #886's residual, with a spec measured on a live rig rather than inferred.

**Measured** (`Anima Wojak Batch.json`, 4 subgraph definitions, panel 0.14.14 / frontend 1.48.7), raw disk JSON vs serialized-after-load:

- node count, node ids, node types inside each subgraph: **identical**
- only differing node field: `inputs` (its `localized_name`/`name`/`type` heads are byte-identical — the difference is the link references)
- `links`: differ
- `state.lastLinkId`: **2092 → 2106**

The frontend regenerates link identity inside subgraph definitions on load. Nothing semantic moves.

Today `graphRootReproducesStateContent` refuses any second surface:

```js
if (surfaces.length !== 1 || surfaces[0] !== "nodes") return NOT_PROVEN;
```

so a faithful open of any workflow containing subgraphs reports CONTENT_UNVERIFIED — binding proven, node comparison perfect, refused on a surface nobody had characterised. That is why the 0.13.5 node-geometry work did not close this: the difference is not in `nodes`.

**Rule being implemented — deliberately narrow:**

> a `definitions` difference is benign when, for every subgraph, the node SET and TYPES are identical and the only differences are link ids (`links`, node `inputs` link references) plus `state.lastLinkId`.

A node added, removed, retyped, or a changed widget value must still fail, exactly as today. Same discipline as the node-geometry allowlist: name the normalisation, tolerate only it, keep the guard for everything else — because this guard's failure mode is a wrong-graph open reported as success (#968).


artokun:
Holding this as a draft — the codex gate found two ways it is wrong, and one of them is that it may not fix the bug at all.

**P1, and the reason this is not a patch:** `graph-binding.js:903` still ends with `if (!surfaces.includes("nodes")) return NOT_PROVEN;`. My change tolerates `definitions` as an *extra* surface alongside `nodes`, but the reported #886 case is a graph where `definitions` is the ONLY surface that differs — link identity regenerates inside `definitions.subgraphs` while the nodes are untouched. That path still returns NOT_PROVEN. I verified the measurement (lastLinkId 2092→2106) and then wired the tolerance into a branch the measured case does not reach. Same mistake as [the unreachable-mechanism note] — a green suite proving the predicate works, never that production runs it.

**P0: the predicate accepts genuinely different graphs.** `RENUMBER_FIELDS` contains `inputs` and `outputs`, and `differsOnlyIn` `continue`s on every allowed key without further checking. Those are the subgraph's interface ports — renaming, retyping, adding or removing one is a semantic change to the graph, and this returns `true` for it. Accepting a wrong graph as bound is strictly worse than the false refusal I was fixing, and it is the failure mode I asked the review to hunt for.

Also flagged and credible: `links` is waived wholesale with no endpoint comparison; node ids are neither required nor unique so `String(undefined)` and duplicates collapse in the Map; `JSON.stringify` throws on a cyclic field rather than failing closed; and the duplicate-surface case (`["nodes","nodes"]`) is now admitted where the old exact-one rule refused it.

Reworking against the measured case rather than against my reading of it — starting by capturing the real before/after `definitions` pair off the rig and asserting the predicate on THAT, so the surface list is derived from evidence instead of assumed.

Issue #310, opened 2026-07-30
panel graph tools lose connected tab after panel_free_vram


artokun:
### What happened / root cause
After a successful panel_free_vram call, the next read-only panel_get_errors call failed with `no connected tab with id ... Connected: none`, even though the same panel graph tools had just inspected and edited the live canvas.

### Steps to reproduce
1. Use panel_graph_outline and panel_set_widget on a connected ComfyUI tab.
2. Call panel_free_vram.
3. Call panel_get_errors (or another graph read tool).

### Exact error (scrubbed)
Error: no connected tab with id "tmp:[REDACTED]". Connected: none

### Fix
applied locally: no  upstream-only: no
Workaround: continue after the panel reconnects; no workflow data was changed beyond the requested widget edit.

### Environment
OS: Windows 10; ComfyUI 0.27.0; GPU: NVIDIA RTX 3090 24 GB; comfyui-mcp 0.48.5; panel version: unknown.

Issue #1096, opened 2026-08-12
Panel socket remains disconnected after reconnect notification, blocking graph recovery


artokun:
### What happened / root cause
The panel announced a resumed session and active Unsaved Workflow, but panel_graph_outline repeatedly returned Connected: none. This prevents verifying or completing partially-applied live graph mutations.

### Steps to reproduce
1. Start a graph mutation sequence on a new workflow.
2. Panel connection drops.
3. Receive reconnect/resume event.
4. Call panel_graph_outline.

### Exact error (scrubbed)
`no connected tab with id tmp:2806... Connected: none`.

### Fix
Applied locally: no. Upstream-only: yes. Panel reconnect state should not announce resume until graph RPC socket is available, or should automatically restore it.

### Environment
Linux; ComfyUI 0.30.1; NVIDIA RTX 3090 24GB; comfyui-mcp 0.50.25; panel 0.11.38.

artokun:
**These four (#1095, #1096, #1097, #1098) are one cascade from one session, not four
independent bugs — and the session is on panel 0.11.38, where current is 0.14.9.** Saying so
up front because it changes what would be useful next, not to wave them away: I am leaving
all four open.

## They chain

Read in order, the errors are a single failure propagating:

1. **The bridge socket dropped mid-command** — `panel tab tmp:2806 disconnected mid-command
   (graph_set_widget) — OUTCOME UNKNOWN` (#1095).
2. **It never came back for RPC** — `no connected tab with id tmp:2806 … Connected: none`,
   even though a resume was announced (#1096, this issue).
3. **So `workflow_open` had nobody to answer it** — `Panel tab did not reply to workflow_open
   within 15000 ms` (#1097, #1098).
4. **And the switch state never cleared**, because the thing that clears it is the reply that
   never arrived (#1097).

Everything from step 2 on is downstream of a tab that is announced as present and is not
answering. Fixing step 3 or 4 in isolation would only change the wording of the wedge.

## The shape of step 2 is one we have seen resolve a specific way

A tab that announces a resumed session while `Connected: none` persists — with a `tmp:` id
rather than a `wf:` one — has repeatedly turned out to be **the browser running a stale panel
bundle**: it advertises the old capabilities, the orchestrator fences the newer tools, and the
socket identity does not line up. The remedy there is a hard refresh (Ctrl+Shift+R), not a
code change, and it is recorded in this repo's changelog against a different report that
looked like a detection bug and was not.

[...]

Issue #278, opened 2026-07-30
Reconnect active workflow session after tool-triggered ComfyUI reboot


artokun:
### What happened / root cause
A tool-triggered panel_restart_comfyui reboot completed, but the sidebar did not reconnect its active workflow tab. Subsequent panel_add_node calls returned: Error: no connected tab with id "wf:workflows/Cyberpunk.json". Connected: none.

The frontend reconnect guard only resumes when REBOOT_KEY or AUTOCONNECT_KEY is set. In this path the reboot completed without REBOOT_KEY surviving, even though the current thread still had a resumable active session.

### Steps to reproduce
1. Start an active sidebar conversation bound to a workflow.
2. Invoke panel_restart_comfyui.
3. Wait until /system_stats is healthy again.
4. Invoke a live graph tool such as panel_add_node.
5. Observe "Connected: none" until manually refreshing/reconnecting the sidebar.

### Exact error
Error: no connected tab with id "wf:workflows/Cyberpunk.json". Connected: none

### Fix
Applied locally: yes
Upstream-only: no

```diff
-    if (!ssGet(REBOOT_KEY) && !lsGet(AUTOCONNECT_KEY)) return;
+    if (!ssGet(REBOOT_KEY) && !lsGet(AUTOCONNECT_KEY) && !resumableSessionId(thread)) return;
```

This lets an active resumable workflow conversation recover even when the explicit reboot marker is missing, while preserving the existing bridge-alive guard below.

Verification: node --check web/js/comfyui-mcp-panel.js and git diff --check both exit 0.

### Environment
Windows 11
ComfyUI 0.29.0
GPU NVIDIA GeForce RTX 5090 32 GB
comfyui-mcp 0.48.11
panel 0.11.9

PR #1134, opened 2026-08-12
fix(1096): a resume must not announce a tab the graph route cannot reach

artokun:
Refs artokun/comfyui-mcp-panel#1096.

> **Status: BLOCKED. Do not merge, and do not read the green suite as readiness.**
> Two CONFIRMED correctness defects from the 2026-08-13 review are still open, and
> **neither can be fixed from this branch** — see
> [What still blocks this](#what-still-blocks-this). What has changed since that
> review is the merge to `main`, and the two findings that *were* actionable (the
> mutation-green test, the session source mismatch).
>
> **Re-verified against the merged tree on 2026-08-14** (this branch + `origin/main`
> at v0.14.41): both defects are still live, at the line numbers cited below. Merging
> `main` did **not** resolve either one. The prerequisite that would make this branch
> correct — a re-advertise carrying no session semantics — **has no branch and no PR
> open against it**. Until that lands, this PR is a correct-looking change that ships
> two regressions, and it stays a draft.

A ComfyUI `reconnected` where the **bridge survived** leaves the orchestrator's tab
mapping dropped, and nothing re-establishes it. Every graph tool then answers
`Connected: none` while a resumed session sits on screen saying otherwise — the
reporter's `panel_graph_outline` calls against a tab id the orchestrator no longer held.

## The uncovered case

On `reconnected` the panel asks `shouldResumeAfterComfyReconnect` whether to respawn:

- **Bridge DOWN** → it reconnects, and the fresh connect sends a hello, so the mapping is
  re-established as a side effect.
- **Bridge UP** → that guard returns `false`, correctly — bouncing a live session is the
  spurious "you reconnected" #278 removed — and the panel then returns having done
  **nothing**. The mapping stays dropped.

That second branch is #1096. This branch re-advertises the tab on the **existing** socket.

## What the announcement is gated on

Only on things that were **observed**, never on what is expected:

| Gate | Observation |
| --- | --- |
| the socket is up | `client.isConnected()`, read at the moment of the attempt — not a cached flag |
| the frame will carry a session | `Boolean(ssGet(SESSION_KEY))` — the *same* source `getResume` reads |
| the hello actually reached the wire | `sendHello` resolving **literally `true`**; only then is the debounce window armed |

Nothing is announced to the user: no chat line, no reconnect, no session frame.

## Changed since the review

**The debounce is now a unit, not an inline block.** The review's one CONFIRMED test
finding was that the source scan stayed green when the guard was negated and the debounce
inverted. The window, the in-flight latch and the arm-only-when-landed ordering moved into
`createReconnectRehelloGate` in `session-rebind.js`, and the tests assert on the **send**
rather than on a returned boolean — so *"the socket is not up, therefore nothing left the
panel"* is a claim about the frame. Eleven mutations were applied and **all eleven are
killed**: inverted window, stamp ahead of the send, stamp on a refused hello, dropped
predicate, truthy-instead-of-`true`, removed latch, un-dispatched call site, wall clock,
assumed socket, stubbed send, and the anchor check that proves each mutation applied.

**The session source mismatch (review: PLAUSIBLE) is fixed.** The guard read
`hasResumableSession` from the active thread **record** while the hello it gates carries
`resume` from `SESSION_KEY`. When those disagree the guard passes and the frame goes out
with `resume: null`, which by `sendHello`'s own contract **spawns a clean agent** — the
spurious start the input exists to prevent. It now asks the source the frame will read.

**The reassuring comment is gone.** The predicate's header used to file the epoch bump
under "worth knowing"; it now records both open defects as open.

## What still blocks this

Both are inside the hello path, not in the decision to re-advertise, and both are
**already reachable on `main`** through #310's post-`free_vram` re-advertise. Both were
re-read in the merged tree on 2026-08-14 and are unchanged:

1. **The re-advertise disarms the mid-task resume nudge.** The orchestrator answers a
   hello with a `ready` ack even on a live socket, and the ack handler consumes
   `MID_TASK_KEY` *before* it decides whether to nudge. The marker is spent with
   `outageMs: 0`, so a later genuine restart finds nothing and the resumed agent sits idle
   over half-finished work. Regresses #1138 / #1145 / #1163.

   *Evidence, `web/js/comfyui-mcp-panel.js` (merged tree):* the handler opens at **28574**
   with `if (ack?.kind === "ready" && ssGet(MID_TASK_KEY)) {` and clears it
   unconditionally on the very next line, **28575** — `ssSet(MID_TASK_KEY, null);` — while
   the decision that should have gated that clear is **31 lines later, at 28606**,
   `if (!shouldNudgeAfterMidTaskReconnect({ outageMs: bridgeOutage.outageMs() })) return;`.
   A re-hello on a live socket reaches the clear with `outageMs()` at `0`; the guard then
   returns `false` (`session-rebind.js:91`, `if (outageMs <= 0) return false;`) and the
   handler returns — but the marker is already spent.

2. **The re-advertise re-injects the canvas-tool disclosure.** A landed hello bumps
   `agentSessionEpoch`, invalidating `canvasToolsProvenEpoch`, so a session that had
   *proven* its canvas tools by executing them is re-asked and the user's next message is
   silently prefixed with the disclosure paragraph.

   *Evidence:* `sendHello`'s `.then((sent) => {…})` bumps the generation at **19472**
   (`agentSessionEpoch++;`, guarded on 19471 only by `if (sent)`) — so it fires on
   **every** landed hello, with no notion of whether that hello was a first advertise or
   a re-advertise of a session that already exists.
   `canvasToolsProvenEpoch` is only ever stamped to the *then-current* epoch (**28203**)
   and is read back against it (**19686**), so the bump strands the proof.

The remedy named in review is a re-advertise that does **not** carry the hello's session
semantics — no `ready` ack, no epoch bump. **That work is not in this branch, has no
branch of its own, and no PR is open for it.** It removes the same exposure #310 already
carries, and it gates this one. The debounce here bounds how *often* these happen, never
*whether*; it must not be read as a mitigation.

**Why this cannot be patched here.** Both defects are properties of the hello path itself,
which this branch only *calls*. Fixing them from inside `shouldRehelloAfterComfyReconnect`
or `createReconnectRehelloGate` would mean special-casing the ack handler and the epoch
bump for one caller while #310's post-`free_vram` re-advertise keeps hitting both
unguarded — a second, divergent code path for the same event, which is a worse defect than
the one it papers over. The honest sequencing is: land step 1 (a session-semantics-free
re-advertise), then rebase this branch onto it, at which point this PR becomes a few lines
of gating.

Also unaddressed, from the first review and not re-raised in the second: the window has
**no trailing edge**, so a second genuine bounce inside the 5s gap is dropped rather than
deferred; and the gate is still constructed per call site, so the three sibling re-hello
paths can still double up on one physical event.

## Verified

- `npm run test:unit` in full (chains i18n, i18n-render, tool-vocabulary and panel-scope
  gates) on the merged tree: **4421 pass / 0 fail** (4422 tests, 1 todo).
  `session-rebind.test.mjs` alone is **78/78**.
- `npm run typecheck` clean.
- Merged `origin/main` (now through **v0.14.41**); no conflict markers in `web/` or
  `browser_tests/`.

**Green is not readiness here.** The suite passes because the code does what this branch
intends; the two blockers above are behaviours of the hello path that no test on this
branch asserts against, and that the previous review found by reading, not by running.
See the status note at the top.

### DIFF
--- a/browser_tests/unit/session-rebind.test.mjs
+++ b/browser_tests/unit/session-rebind.test.mjs
@@ -1024,8 +1024,10 @@ test("#1096 wiring: the re-advertise is debounced and does not disturb the sessi
   // across lines and a future one would break a stricter locator for no reason.
   const at = src.indexOf("shouldRehelloAfterComfyReconnect({");
   assert.notEqual(at, -1, "the reconnect handler must consult the predicate");
+  // The predicate's ARGUMENTS, sliced once and reused by both argument assertions below.
+  const callArgs = src.slice(at, src.indexOf("})", at));
   assert.match(
-    src.slice(at, src.indexOf("})", at)),
+    callArgs,
     /bridgeConnected: client\.isConnected\(\)/,
     "the live bridge state must come from the client, not a cached flag",
   );
@@ -1060,10 +1062,11 @@ test("#1096 wiring: the re-advertise is debounced and does not disturb the sessi
     "…and only when the hello actually landed, so a refused one leaves the window open",
   );
   // A hello with no session spawns a clean agent, so the session input must be WIRED,
-  // not left to its default.
-  const callAt = src.indexOf("shouldRehelloAfterComfyReconnect({", 0);
-  const callBlock = src.slice(callAt, src.indexOf("})", callAt));
-  assert.match(callBlock, /hasResumableSession/, "the session input must be passed at the call site");
+  // not left to its default. Asserted against the argument slice computed ONCE above:
+  // this used to recompute `src.indexOf(needle, 0)`, which is the same call as
+  // `src.indexOf(needle)`, so it re-derived a byte-identical slice under a second name
+  // and read as a broader check than it was.
+  assert.match(callArgs, /hasResumableSession/, "the session input must be passed at the call site");
   // MONOTONIC: a wall-clock adjustment must not be able to make the gap look negative.
   assert.match(block, /monotonicNow\(\)/);
   // It re-advertises and nothing more — no connect, no session frame. Bouncing a live
--- a/browser_tests/unit/session-rebind.test.mjs
+++ b/browser_tests/unit/session-rebind.test.mjs
@@ -11,6 +11,7 @@ import { readFileSync } from "node:fs";
 import {
   shouldResumeAfterComfyReconnect,
   shouldRehelloAfterCommand,
+  shouldRehelloAfterComfyReconnect,
   shouldNudgeAfterMidTaskReconnect,
   createBridgeOutageTracker,
   planSoftReloadRecovery,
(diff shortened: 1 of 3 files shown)


artokun:
## NO-SHIP — the approach is wrong, not the details. Holding.

The review invalidated the premise rather than the implementation: **a full hello is not a silent repair.** My commit claimed "the session, the thread and the agent are untouched", corrected that once for the clean-agent case, and it is still wrong for three more reasons — each CONFIRMED.

### A benign blip now injects a false "you dropped mid-task" turn into a live agent

A hello is answered with a full handshake, including a `ready` ack — this file even notes at 18190 that *"'ready' repeats on a live socket"*. `onAck`'s mid-task branch then fires, and its only bridge-survived guard is `Date.now() - lastBridgeDownAt < 6000`, which **cannot** fire here because `lastBridgeDownAt` is only stamped when the bridge socket closes and the bridge never closed. So on a mere tab refocus, with the agent mid-turn:

- the durable transcript gets "Reconnected — picking up where we left off."
- `sendUserMessage("✅ Your connection dropped mid-task…continue exactly what you were doing")` is injected into an agent that is *still working*

The user reads a false statement, and the agent restarts or duplicates what it is doing. Before this diff the branch sent nothing, so no ready ack could be manufactured on a surviving bridge — I created this.

### It re-injects the canvas-tool disclosure into a session that already proved its tools

The landed hello bumps `agentSessionEpoch`, invalidating `canvasToolsProvenEpoch`, so the user's next message is silently prefixed with the 1.2KB live-canvas-tools paragraph. Mid-conversation the model can act on it and tell the user their session lacks the tools and to update comfyui-mcp. #633's comment describes precisely this as the thing that module exists to prevent — *"invalidated a WORKING agent's proof and re-asked a session that demonstrably had the tools whether it had them"*.

### The re-greeting noise #588 banned

A hello re-greets ("agent ready") and re-runs handshake side effects, now on every ComfyUI blip more than 5s apart.

### And the debounce is structurally wrong twice over

- **No trailing edge**: a second genuine bounce at t=3s is dropped *forever*, not deferred. The flap settles, no further `reconnected` arrives, and the mapping stays dropped for the page's life — the rate limiter reproducing the exact bug the fix exists to prevent.
- **Armed before the send**: `sendHello` returns false without sending in three cases, one of which (#640 route identity unestablished) is a *reconnect-window* condition — so the first attempt plausibly refuses while the window is already stamped, suppressing the repair for 5s in the one scenario that matters.
- Private to this call site, so #310's `free_vram` hello and the socket-open hello can still double up on the same bounce.

### The test proved nothing (mutation-verified)

A verifier copied the branch, inverted the debounce and relocated the send, and my test still passed. Token-presence assertions cannot express the ordering and nesting claims the test names.

## What this actually means for #1096

The important consequence is bigger than this PR: **#310's `free_vram` rehello has the same side effects today.** The spurious mid-task injection and the disclosure re-injection are reachable on current main via free_vram, not only via my change — I would have widened a latent bug from one trigger to every ComfyUI reconnect.

So the honest ordering is:

1. Make a re-advertise **silent** first — distinguish a re-hello ack on a live socket from a fresh-connect handshake, so it does not drive the mid-task branch, the greeting, or the epoch bump. That is panel-side and it fixes #310's existing exposure too.
2. Only then use a re-advertise to repair the dropped mapping on reconnect, with the #607 shape (in-flight latch, `sent === true`, failure backoff) rather than a hand-rolled clock.

Step 1 is the real work and is worth its own issue. Holding this branch rather than shipping a change that turns tab refocus into a false mid-task claim.

artokun:
## Review outcome: this should not merge in its current form

A high-effort review of the rebased branch returned two CONFIRMED correctness defects, and both come from the same root: **a hello is not a cheap frame.** The change treats "re-advertise the tab" as a light repair, but a landed hello has three established side effects, and firing one on every benign ComfyUI reconnect drags all three along.

### 1. The rehello disarms the mid-task resume nudge (CONFIRMED)

`ready` repeats on a live socket — the panel's own comment says a re-advertise draws a fresh handshake. That ack's handler clears `MID_TASK_KEY` **before** it decides whether to nudge.

So: agent is mid-build, the user opens an asset preview, ComfyUI's WS blips, this code rehellos, the orchestrator answers `ready`, and the marker is consumed with `outageMs: 0` — no nudge, no re-arm. Thirty seconds later ComfyUI genuinely restarts and the bridge really drops; that `ready` finds no marker and skips the nudge entirely. The resumed agent sits idle over a half-finished workflow.

This directly regresses #1138 / #1145 / #1163, which are already on main. #310's `free_vram` rehello has the same hole, but only behind an explicit user action — this fires it on every blip.

### 2. The rehello re-injects the canvas-tool disclosure (CONFIRMED)

Every landed hello bumps `agentSessionEpoch`, and the disclosure is suppressed only while `provenEpoch === agentEpoch`. So each blip discards proof a session earned by actually executing `panel_*` commands, and the user's next message is silently prefixed with the ~2KB disclosure paragraph.

Concretely: twenty turns into a working session, a blip happens while the user reads a reply; they type *"make the sampler 30 steps"* and get a tool-availability lecture and an upgrade checklist. Once per blip.

### Also raised

- **PLAUSIBLE** — `hasResumableSession` is computed from the active thread record's sessionId, while the hello it gates carries `resume` from `SESSION_KEY`. Two different sources: the guard can pass while the hello goes out with `resume: null`, which by `sendHello`'s own contract spawns a clean agent — the spurious agent start this guard exists to prevent.
- **PLAUSIBLE** — the 5s debounce is private to this call site and invisible to the three sibling rehello paths that can fire on the same physical event.
- **CONFIRMED (tests)** — mutation testing showed the source-scan test **stays green when the guard is negated or the debounce inverted**. That is the technique this very test file cites an earlier review for disproving.

### Fixed on the branch already

- The debounce armed its window **before** the send — the "recorded before it happened" defect the panel names one screen up at the #607 re-hello, with the same 5s constant. A refused or dropped hello spent the window without re-advertising anything, leaving the mapping this change exists to repair dropped for the full gap. Now armed inside the send's resolution and only when the hello landed; the assertion pins the ordering and is mutation-checked.
- The wiring test derived one identical slice twice under two names, reading as two independent checks when it was one.

### What I think this needs

Not a patch. The premise — repair the mapping with a full hello — is what carries the side effects, so guarding each one individually would mean three new guards around a frame that was chosen for being cheap and is not.

Worth considering instead: a frame that re-advertises the tab **without** the hello's session semantics, so no `ready` ack, no epoch bump, and no resume ambiguity. That is an orchestrator-side contract change, which also matches where the reported symptom (`Connected: none`) is decided.

Leaving this draft rather than merging it. The rebase itself is sound and verified — 65 → 69 tests with nothing lost, all 10 main exports intact — so the branch is a good base for whichever direction is chosen.

One rebase blemish worth recording: one line was dropped from an intermediate commit message (`"#884) — so the session, the thread and the agent are untouched. A test pins that this"`), leaving a broken sentence. It is confined to a commit message a squash-merge will not carry, and repairing it means history surgery for prose that never ships, so I left it.

artokun:
sounds like this is the wrong apprioach closing as reference only

Issue #1136, opened 2026-08-12
Status chip reports "disconnected" on a working session: it reflects the per-backend bridge URL while the session runs on the shared one


artokun:
## What happens

The panel's connection chip shows a red dot and `disconnected` while the session is actually working. The `connecting` state never appears either — the chip goes straight to `disconnected` and stays there.

## Measured mechanism

The chip is reporting a **different socket** from the one the session uses.

Settings on the affected machine:

```
comfyui-mcp.defaultBackend        = claude
comfyui-mcp.bridgeUrl.claude      = ws://127.0.0.1:52727
comfyui-mcp.panel.bridgeUrl       = ws://127.0.0.1:9180
comfyui-mcp.bridgeUrl.single      = ws://127.0.0.1:9180
comfyui-mcp.externalOrchestrator  = true
```

Port state, measured with `netstat` and a raw TCP connect (NOT with `fetch()` from the page — see the note below):

| port | state | serves |
|------|-------|--------|
| 9180 | LISTENING (pid 74404), HTTP `426 Upgrade Required` | WebSocket bridge — healthy |
| 9181 / 9182 / 9183 | LISTENING, same pid 74404 | same orchestrator process |
| **52727** | **ECONNREFUSED** | `bridgeUrl.claude` — nothing listening |

So with `defaultBackend: "claude"`, the panel's status client targets `ws://127.0.0.1:52727`, which is dead, while the orchestrator is alive on 9180 and the session works over it. The chip is faithfully reporting a socket that is genuinely down; it is just not the socket doing the work.

`52727` looks like an ephemeral port captured once and left behind — `externalOrchestrator: true` means the panel does not own that process, so its port can move across restarts while the stored per-backend URL does not.

## Why "connecting" never shows

A refused connection to a closed port fails immediately with `ECONNREFUSED`. The `connecting` state only holds while a connect is genuinely in flight, so an instant refusal drops straight through to `disconnected`. The missing intermediate state is a symptom of the same wrong target, not a separate defect.

## What it is NOT

Initially suspected to be an i18n regression. It is not:

- `web/js/comfyui-mcp-panel.js` parses clean as an ES module with no duplicate top-level declarations.
- The render boundary is intact — `STATUS_LABEL[state]?.() ?? state` with all three states present as literal keys.
- No panel init errors in the console on a fresh load.

An i18n break would produce a *wrong word*; this produces a *frozen, technically-true-but-irrelevant* state.

[...]

PR #1137, opened 2026-08-12
fix(1136): the status chip must report the socket the session actually uses

artokun:
Claims #1136 — reported live by the maintainer: the chip reads `disconnected` on a working session, and the `connecting` state never appears.

**Measured, not inferred:**

| port | state | role |
|---|---|---|
| 9180 | LISTENING, `426 Upgrade Required` (pid 74404) | shared bridge — healthy, and `panel.bridgeUrl` |
| 9181/9182/9183 | LISTENING, same pid | same orchestrator |
| **52727** | **ECONNREFUSED** | `bridgeUrl.claude`, and `defaultBackend` is `claude` |

So the chip faithfully reports a socket that is genuinely down — it is just not the socket doing the work. `externalOrchestrator: true` means the panel does not own that process, so an ephemeral port recorded once goes stale across a restart while the stored per-backend URL does not.

The missing `connecting` state is the same defect, not a second one: `ECONNREFUSED` on a closed port returns instantly, so the in-flight state never holds long enough to paint.

**Ruled out** — it is not i18n, which was the first suspicion. The module parses clean with no duplicate top-level declarations, the render boundary (`STATUS_LABEL[state]?.() ?? state`) is intact with all three states as literal keys, and there are no init errors on a fresh load. An i18n break yields a wrong *word*, not a frozen *state*.

**Also ruled out** — upgrading. Triage suggested 0.14.15 → 0.14.16; that range is one commit (`fix(1089)`, workflow-open wording) and touches none of this.

Scope to be set once I can name where the two clients diverge. The fix must not be "show connected more often" — a chip that under-reports is annoying, but one that claims a live session over a dead socket is worse.


artokun:
Mechanism confirmed in code, and it matches the measured ports exactly. Three pieces combine.

**1. Where 52727 came from.** `comfyui-mcp-panel.js:21801-21808` migrates the pre-per-backend key ONCE into the Claude group:

```js
const lu = getSetting(LEGACY_SETTING_BRIDGE_URL);   // "comfyui-mcp.bridgeUrl"
… getSetting(SETTING_BRIDGE_URL.claude) == null
  setSetting(SETTING_BRIDGE_URL.claude, lu);
```

Its own comment says why: *"migrated ONCE into the Claude group so a returning user's custom port isn't lost"*. That is a reasonable intent — but the value it preserves is an **ephemeral orchestrator port**, which stops being valid the moment that process restarts. It is preserved forever regardless.

**2. Why it beats the healthy single-port URL.** At `:28851`:

```js
const manualOverride =
  !!wanted && wanted !== defaultBridgeUrlFor(selectedBackend) && wanted !== lastAutoUrl;
```

`52727 !== 9180`, so the stale migrated value is classified as a **deliberate manual override** and honoured over `bridgeUrl.single`. The panel cannot distinguish "the user typed a custom port on purpose" from "a migration copied a dead ephemeral port here in 2026" — both look identical to this predicate.

**3. Why nothing recovers it.** With `externalOrchestrator: true` (this setup), `:28858` onward dials the configured per-backend URL **directly** and never POSTs `/connect`, so the orchestrator's advertised `bridge_url` — which would have carried the live 9180 — is never consulted. The retry then surfaces the "start it locally" hint for a bridge that is already running, one port over.

**This is the exact failure the design forbade.** `:3141-3146`, defining the single-port key:

> a FRESH key — deliberately NOT the per-backend or legacy ones, whose stale pre-single-port values (e.g. a migrated custom port) must not leak in and **make the panel dial a dead port**.

The guard was written; the migration at :21808 walks around it by writing the legacy value into a per-backend key that :28851 then treats as sacred.

**Direction, not yet a patch.** The honest discriminator is provenance: a URL the panel *migrated* is not a URL the *user chose*, and only the latter should outrank a live single-port bridge. Marking the migrated value as non-manual (or dropping it once single-port is in play, which is what :3146 already assumes) would let 9180 win without weakening a genuinely hand-set port.

What the fix must NOT do is make the chip claim connected while dialing a dead socket — the label is currently telling the truth about the wrong target, and the repair belongs at the target, not the label.

artokun:
Gate returned **NEEDS-REVISION** with two P1s. Both are right, and I am holding this rather than merging.

**1. My docblock lied.** It stated *"The record is cleared as soon as the value diverges"* — and no clearing path exists in the diff. A comment asserting behaviour the code does not have is the same defect class this PR fixes, written into the fix for it. Corrected in `307feac1`: the limitation is now stated plainly instead of denied.

The real consequence: a user who changes the URL and later **deliberately retypes** the value the migration once wrote is still classified as inherited, and their genuine choice is disregarded at both call sites.

**2. Provenance lives in the wrong store.** The record is in `localStorage` while the bridge URL is a **synced ComfyUI setting**. On a second tab or device the setting can arrive without its record (wrongly manual), or a stale local record can suppress a legitimately re-entered URL. That makes the fix non-deterministic across a sync boundary — which is not a property I want in the thing deciding which socket the panel dials.

**Neither makes the reported bug worse**, and the measured case (a dead 52727 outranking a live 9180) is fixed with all five mutations killing tests. But "fixed for the reporter, undefined across tabs" is not a merge.

**What this needs before it lands:**
- Clear the record wherever the user's own edit is committed, so provenance dies the moment it stops describing reality.
- Move provenance beside the setting rather than into `localStorage`, so it syncs with the value it describes. That likely means storing the migrated marker as a setting, or recording the migration in a way that is derivable rather than remembered.

The second is the more honest fix and probably subsumes the first: if provenance travels with the value, there is nothing to desynchronise and nothing to clear.

Worth noting the gate also earned its keep on the mechanical side earlier: the two `manualOverride` call sites were byte-identical, and my first patch attempt asserted a unique anchor and **refused to apply** — fixing one would have left the other dialing the dead port. And one mutation initially survived because the whitespace test padded the argument the caller already trims, leaving the stored-value side untested; the case that actually exercises it is in now.

artokun:
Round 2 also **NEEDS-REVISION**, and taken together the two rounds say something more useful than either finding alone: **provenance is the wrong frame for this fix.**

Round 1 killed the *remembered* marker — it never cleared, and it desynchronised from the synced setting it described. Round 2 kills the *derived* one:

> If migration copied legacy URL `A` into the Claude setting, then the legacy setting later changes to `B` while Claude remains `A`, `A` is reclassified as a manual override and again blocks discovery of the live bridge. The migration's authorship is lost as soon as these two independently addressable settings diverge; being synced does not make them coupled.

That is correct and I had not considered it. Both approaches fail the same way: they try to infer **who chose a URL**, and that fact is simply not recoverable from what the panel stores. Every encoding of it either rots or over-claims — and the second P1 in both rounds is the same residual, a user who deliberately picks the old value being silently overridden.

**The framing that does not require guessing intent: liveness.**

The user-visible defect is not "the panel honours the wrong URL", it is "the panel dials a dead port forever and never recovers". That needs no theory of authorship:

- honour the configured URL, exactly as today — no reclassification, no suppression of anyone's choice
- when it fails to connect, fall back to the single-port bridge (or the orchestrator-advertised one) and **say so**
- a URL that works is never second-guessed, and a URL that does not work stops being a dead end

That fixes the reported case (52727 refuses → fall back to the live 9180) without ever deciding whether 52727 was typed or migrated. It also fixes cases neither provenance version touches: any bridge URL that becomes stale for any reason, including ports the user genuinely did type and then moved.

It also sits better with what external-orchestrator mode already does — it deliberately skips `/connect` because the host may have no agent, but that is what removes the one channel that could have corrected the port. A liveness fallback restores self-healing without reintroducing the POST.

I am stopping here rather than starting a third attempt at the wrong shape. The diff on this branch (derived provenance, 9 tests, 4 mutations killing) is a working fix for the measured case, but it is the wrong architecture and I would rather not merge a design the gate has now refused twice for sound reasons.

**Recommendation:** keep this branch as the investigation record, and rebuild against liveness. The measurement, the mechanism at :21808/:28851, and both refuted approaches are captured here and in #1136 so none of it has to be rediscovered.

The immediate unblock for anyone hitting this remains: clear `comfyui-mcp.bridgeUrl.claude` and the panel falls through to the working single-port bridge.

Issue #291, opened 2026-07-30
Codex sidebar session did not expose live panel_* graph editing tools


artokun:
### What happened / root cause
A Codex-backed sidebar session received the panel system instructions but no live `panel_*` graph tools were callable or discoverable. Repeated tool discovery for `panel_graph_outline`, `panel_new_workflow`, `panel_add_node`, `panel_connect`, `panel_set_widget`, and `panel_open_workflow` returned only headless comfyui-mcp tools. This forced an awkward fallback to `save_workflow` and prevented the requested workflow from being opened/edited on the live canvas.

Likely root cause: panel tool registration/exposure is missing for this backend/session, while the system prompt assumes those tools are present.

### Steps to reproduce
1. Open the ComfyUI sidebar panel with Backend: Codex.
2. Ask the agent to build a workflow on the current/new canvas.
3. Agent searches for or attempts to call `panel_graph_outline` / `panel_new_workflow` / `panel_add_node`.
4. No panel graph tools are exposed; only headless comfyui-mcp tools are discoverable.

### Exact error
No direct call could be made because the tool schemas were absent. Tool discovery queries for exact panel tool names returned unrelated headless tools.

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: ensure Codex sidebar sessions register and expose the full connected panel_* tool set (at minimum outline/query/new/open/load/add/connect/widget/title/layout/save/canvas/todo) before injecting instructions that require them; if the panel is disconnected, expose stubs that return the documented connection error.

### Environment
Windows 11; NVIDIA RTX 4060 Laptop GPU 8 GB; ComfyUI 0.27.0; comfyui-mcp 0.48.13; panel 0.11.4; Backend: Codex.

artokun:
Fixed in comfyui-mcp#594 (merged; ships next mcp release): a Codex sidebar session dropped the panel_* tools because comfyui's ~250 tools saturated Codex's tool budget — the non-Claude HTTP lane now spawns comfyui in compact mode, freeing budget so panel_* tools are advertised + callable (Claude unaffected; set COMFYUI_MCP_TOOL_MODE=full to opt out). Reopen if panel_* still don't appear for Codex on the next mcp release.

artokun:
[...]

Issue #419, opened 2026-07-31
panel_reload can leave all open sidebar tabs permanently disconnected


artokun:
### What happened / root cause
panel_reload returned "soft reload (orchestrator) scheduled" during a live panel session, but the active ComfyUI tab never reconnected. The chat agent continued running, while every panel graph call failed with "no panel connected".

This appears to be a reconnect/resume defect at the orchestrator-panel boundary. The documented recovery panel_set_workflow_target({mode:"current"}) cannot run because it also requires a connected panel.

### Steps to reproduce
1. Start from a working connected ComfyUI Agent sidebar.
2. Call panel_reload({}).
3. Wait 30+ seconds.
4. Call panel_set_workflow_target({mode:"current"}) or panel_get_errors.
5. Both fail because connected tabs are empty.

### Exact error
panel_set_workflow_target:
Panel not reachable: no panel connected - open ComfyUI with the comfyui-mcp-panel pack installed and check the Agent sidebar tab

Immediately after reload, panel_get_errors also reported:
graph_get_errors could not reach the ComfyUI panel - it is still reconnecting after a restart/reload ... (no connected tab ... Connected: none)

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: after a soft orchestrator reload, preserve or re-establish the bridge registration for already-open sidebar tabs, or give panel_set_workflow_target(mode:current) an out-of-band recovery route that does not itself require a registered tab. The current recovery advice is circular once Connected is none.

### Environment
Windows 11; ComfyUI 0.29.1 LOCAL; RTX 5090 D 32GB; comfyui-mcp 0.48.21; panel 0.11.20.

artokun:
Recurs on a newer release: Windows 11 / ComfyUI 0.30.1 / comfyui-mcp 0.49.8 / panel 0.11.38. `panel_reload({scope:"frontend"})` reported `soft reload (frontend) scheduled`, but the active workflow did not reconnect; `panel_graph_outline` and `panel_set_todo` then failed with the reconnecting/no-connected-tab message and `Connected: none`, blocking canvas edits.

artokun:
Fixed: performSoftReloadRecovery in web/js/lib/session-rebind.js stops THEN restarts the bridge on every reload (unconditional reconnect), with a #419 lifecycle suite in browser_tests/unit/session-rebind.test.mjs (accepted/503/thrown reloads, N tabs each reconnect).

PR #1139, opened 2026-08-12
fix(1138): a live-socket re-advertise must not read as a fresh connect

### LINKED ISSUE #1138: A re-advertise on a live socket injects a false "you dropped mid-task" turn and re-asks a proven agent for its tools
### What happens

A re-advertise (`client.rehello()`) is answered by the orchestrator with a **full handshake**, including a `ready` ack. The panel's own comment records that `'ready' repeats on a live socket` (comfyui-mcp-panel.js ~18190). So on a socket that never dropped, a re-advertise drives three side effects that only make sense for a *fresh* connection:

1. **A false mid-task resume.** `onAck`'s ready branch injects `✅ Your connection dropped mid-task…continue exactly what you were doing` as a user message, plus `Reconnected — picking up where we left off.` into the durable transcript. Its only bridge-survived guard is `Date.now() - lastBridgeDownAt < 6000`, and `lastBridgeDownAt` is **only stamped when the bridge socket closes** — so on a surviving bridge it is still `0` and the guard cannot fire. If the agent is mid-turn (`MID_TASK_KEY` armed by `turn:working`), it is told to resume work it is already doing, and the user reads a claim about a drop that never happened.

2. **The canvas-tool proof is discarded.** `sendHello` bumps `agentSessionEpoch`, invalidating `canvasToolsProvenEpoch`, so the user's next message is silently prefixed with the ~1.2KB live-canvas-tools disclosure. Mid-conversation the model can act on it and tell the user their session lacks the tools and to update comfyui-mcp / switch backends. #633's comment names exactly this as what that module exists to prevent: *"invalidated a WORKING agent's proof and re-asked a session that demonstrably had the tools whether it had them"*.

3. **Re-greeting noise.** The handshake re-greets ("agent ready"), which #588 banned for reconnect cycles.

### How it is reachable today

`shouldRehelloAfterCommand` (#310) re-advertises after every successful `free_vram`, on a live bridge, by design — that fix is correct about needing the re-advertise. So a user who frees VRAM mid-task can get a false "your connection dropped" turn injected into a working agent.

The same mechanism blocks #1096: the tab mapping after a ComfyUI bounce can only be repaired by a re-advertise, and a re-advertise is currently too loud to use on a live session. An attempt to use it for #1096 (PR #1134) was withdrawn for exactly this reason.

### Why the panel can fix this

The panel knows which kind of hello it sent. A re-advertise over a live socket that carries the **same** session id is a routing change, not a new agent — the orchestrator itself carries the routing state across it (`same-socket re-hello … the shared session continues`, #884). So:

- a `ready` ack arriving in response to a re-advertise must not drive the mid-task branch or the greeting;
- and a re-advertise that keeps the same session must not invalidate the canvas-tool proof, because the agent it was proven against is the one still running.

Neither weakens the fresh-connect path, which is where those behaviours belong.

### Fix

Applied locally: no. Upstream-only: yes (panel).

Distinguish a live-socket re-advertise from a fresh-connect handshake, and make the ready-ack handling, the greeting and the epoch bump conditional on the latter.

### Notes

Found while reviewing PR #1134 (#1096). Not reported from the field — the mid-task injection would read as the agent behaving oddly after a `free_vram`, which is hard to attribute.

artokun:
Closes #1138. Unblocks #1096; fixes an exposure that exists on main today via #310.

A re-advertise is answered with a full handshake including a `ready` ack — the panel's own comment says `'ready' repeats on a live socket`. On a socket that never dropped, that drives three behaviours which only make sense for a fresh connection: the mid-task resume injection, the re-greeting #588 banned, and the `agentSessionEpoch` bump that throws away a working agent's canvas-tool proof.

The mid-task one is the sharp edge. Its only bridge-survived guard is `Date.now() - lastBridgeDownAt < 6000`, and `lastBridgeDownAt` is only stamped when the bridge socket **closes** — so on a surviving bridge it is still 0 and the guard cannot fire. A user who runs `free_vram` mid-task (which #310 correctly re-advertises after) can therefore have `✅ Your connection dropped mid-task…continue exactly what you were doing` injected into an agent that is still working, and read a claim about a drop that never happened.

Found while reviewing #1134, where I tried to use a re-advertise for #1096's dropped tab mapping and would have widened this from one trigger to every ComfyUI reconnect. Withdrawing that and fixing this first is the right order: the mapping repair needs a re-advertise, so the re-advertise has to be safe on a live session before anything else can use it.

## What I intend to establish first

**That the same session really does survive a re-advertise**, rather than assuming it. The orchestrator logs `same-socket re-hello … routing state carried; the shared session continues` (#884), which is why skipping the epoch bump looks correct — the agent the proof was taken against is the one still running. But #291 documents the opposite case: a hello with the session key cleared spawns a clean agent, and then the proof MUST reset. So the condition is not "is this a re-advertise" but "does this hello carry the same session id it was bound to", and I want that read off the payload rather than inferred from the call site.

**Whether the ready ack can be attributed reliably.** Suppressing the mid-task branch needs to distinguish an ack answering *this* re-advertise from one answering a genuine reconnect that raced it. If that cannot be done from what the panel holds, the honest fix is narrower — gate on the bridge never having dropped rather than on which hello it answers.

Draft while I read onAck's ready branch and the hello payload.

### DIFF
--- a/browser_tests/unit/session-rebind.test.mjs
+++ b/browser_tests/unit/session-rebind.test.mjs
@@ -4,6 +4,9 @@
 
 import test from "node:test";
 import assert from "node:assert/strict";
+// #1138 adds one structural test (the call site's polarity), which the rest of this
+// pure-logic file does not need.
+import { readFileSync } from "node:fs";
 
 import {
   shouldResumeAfterComfyReconnect,
@@ -601,4 +604,22 @@ test("#1138: the guard is INVERSION-SENSITIVE, unlike a source scan", () => {
   );
   assert.equal(neverDropped, false);
   assert.equal(realRestart, true);
+});
+
+test("#1138 wiring: the call site NEGATES the predicate and returns", () => {
+  // The predicate is tested by behaviour above; this covers the one thing a pure test
+  // cannot see — that the call site consults it in the right POLARITY. Dropping the
+  // negation would invert the fix into "nudge only when nothing dropped", which is worse
+  // than the original bug. An exact-substring claim on purpose: #1096's review proved by
+  // mutation that loose token-presence scans over this file assert nothing at all.
+  const src = readFileSync(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url), "utf8");
+  assert.ok(
+    src.includes(
+      "if (!shouldNudgeAfterMidTaskReconnect({ bridgeDroppedAt: lastBridgeDownAt, now: Date.now() }))",
+    ),
+    "the ready-ack mid-task branch must NEGATE the predicate and read the live drop stamp",
+  );
+  // The reverted sessionStorage twin must not return without its own review: every
+  // confirmed leak on this branch came from persisting that timestamp.
+  assert.doesNotMatch(src, /BRIDGE_DOWN_AT_KEY/, "the persisted drop timestamp stays reverted");
 });
--- a/browser_tests/unit/session-rebind.test.mjs
+++ b/browser_tests/unit/session-rebind.test.mjs
@@ -8,6 +8,7 @@ import assert from "node:assert/strict";
 import {
   shouldResumeAfterComfyReconnect,
   shouldRehelloAfterCommand,
+  shouldNudgeAfterMidTaskReconnect,
   planSoftReloadRecovery,
   performSoftReloadRecovery,
   isTransientReconnectError,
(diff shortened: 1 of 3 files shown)


artokun:
## Review returned 22 verified findings — and most of them are already reverted

Important framing: the run reviewed the branch **including** the persisted-timestamp commit (5dadb04b), which I reverted before it finished. Its own synthesis says the findings are "dominated by one" defect — the persistence having a single consume site — and that is the commit that is gone. So the majority describe code no longer on the branch.

What the review found there is still worth recording, because it is a much sharper account of *why* that commit was wrong than my revert message managed:

- **One consume, seven exits.** I retired the stamp in the mid-task branch; six other paths retire `MID_TASK_KEY` and leave the stamp standing (reboot resume, soft reload, turn end, fresh session, idle drops). So the invariant I wrote — "a drop authorizes ONE nudge" — held on **one of seven** paths. The concrete sequence is worse than I described: a restart clears `MID_TASK_KEY` at 28056 without touching the stamp, the agent takes the resume message and re-arms `MID_TASK_KEY`, then a mid-turn `free_vram` re-hello (#310) draws a second `ready` and the panel injects "your connection dropped mid-task" into a session that never dropped again — restarting a render already running. That is reachable in *any session that has restarted or soft-reloaded once*, which is the common case, not the corner.
- **Duplicated tabs** inherit the stamp (sessionStorage is copied), so a fresh tab nudges a live agent mid-turn against a drop it never saw. #419 treats multi-tab as supported.

Both confirm the revert was right, and for a stronger reason than I gave: the persistence did not merely add uncleaned exits, it broke its own stated invariant on six of seven paths.

## What survives against the branch as it now stands

**One CONFIRMED defect, and it is PRE-EXISTING — not introduced here.** Every FAILED reconnect attempt re-stamps `lastBridgeDownAt`: `connect()` assigns `sock` before the connection resolves, so a refused attempt during a ComfyUI restart runs the close handler in full and re-stamps. With backoff at 1s/3s/7s/15s, the gap the predicate weighs is *the last attempt's delay*, not the outage. A genuine restart that returns within ~7s therefore reads as a fast reconnect and loses its nudge.

My predicate inherits that faithfully — it does not cause it and does not worsen it, and the guard is strictly better than the sentinel bug it replaces. But it means the long-gap heuristic is measuring the wrong interval in a whole class of real restarts, which deserves its own issue rather than a patch smuggled in here.

**A false claim in my reverted comment**, already corrected: `Number(ssGet(...))` of an absent key is `0`, not `NaN`, so the `Number.isFinite` guard I credited is not what suppresses it — the `<= 0` check is. Behaviour was right, stated reason wrong.

**Test coverage** for the panel-side wiring: addressed by the polarity test just pushed, which asserts the whole guard line rather than its tokens.

## Where this leaves #1139

Ready for merge as one substantive commit plus its tests: require that the bridge actually dropped, with a behaviourally-tested predicate and a structural test for the call site's polarity. 3976 tests pass, typecheck clean.

Not re-running the review for the reverted state: the revert strictly REMOVES the code the findings target, and the one live finding is pre-existing and documented above rather than fixed here.

PR #1143, opened 2026-08-12
feat(i18n): Arabic (ar) panel catalog — all six plural categories

artokun:
Adds the Arabic (`ar`) translation of the ComfyUI Agent Panel. `ar` was already
in the language picker (`web/js/lib/i18n.js` `LOCALES` + `RTL`), so selecting
العربية rendered English in full; it now renders Arabic. Nothing outside
`locales/ar/` is touched.

## Final counts

| file | expected | shipped |
|---|---|---|
| `locales/ar/main.json` | 1269 keys (945 ordinary + 54 × 6) | **1269** (945 ordinary + 324 plural) |
| `locales/ar/settings.json` | 52 entries / 104 leaves | **52 / 104** |

All **54 counted strings carry six plural forms** — `zero`, `one`, `two`, `few`,
`many`, `other`. Verified programmatically: `plural bases: 54, incomplete: 0`.

## Plural forms

`Intl.PluralRules("ar")` allows exactly six categories and the gate enforces it,
so the skeleton's six keys per base were filled, never added to or removed from.
The wording follows Arabic numeral agreement *as it reads with a rendered digit*,
which is the register the panel is actually in — `{count}` is mandatory in every
form (`i18n-check` enforces placeholder parity per form, and the render check
asserts the digit appears), so the numeral-less citation forms («عقدة واحدة»,
a bare «عقدتان») are not expressible here by design:

```
panel.nodes    0 عقدة | 1 عقدة | 2 عقدتان | 3 عقد | 5 عقد | 11 عقدةً | 21 عقدةً | 100 عقدة
n_models       0 نموذج | 1 نموذج | 2 نموذجان | 3 نماذج | 5 نماذج | 11 نموذجًا | 21 نموذجًا | 100 نموذج
chars_count    0 حرف | 1 حرف | 2 حرفان | 3 أحرف | 5 أحرف | 11 حرفًا | 21 حرفًا | 100 حرف
```

`few` (3–10) takes the broken plural, `many` (11–99) the accusative *tamyīz*
(`عقدةً`), and `zero`/`one`/`other` the singular genitive — which is why those
three coincide rather than having been copied. Abbreviated units (`{count} د`,
`{count} س`, `{count} ث`) do not inflect, so their six forms are identical by
grammar, not by omission.

## RTL

`applyDirection()` owns `dir="rtl"`, so **no bidi control characters are embedded
anywhere** (LRM/RLM/isolates: 0, verified). What *is* handled in the strings is
glyph direction, since Unicode does not bidi-mirror arrows:

- state-change and flow arrows `→` → `←` (`الجهد ← {effort}`, `النموذج ← {model}`, `وُصل {from} ← {to}`)
- back arrows `←` → `→` (`→ مجموعة البيانات`, `→ تطبيقاتي`)
- collapsed disclosure `▸` → `◂`; revert marker `↩` → `↪`
- `«»` and the `›` breadcrumb are left as authored — those *are* in
  BidiMirroring.txt and the browser flips them itself
- `▶` (run/play) is left alone, per the usual convention that media controls are
  not mirrored

## Deliberately left in Latin script

Brands and product names per the batch rules (ComfyUI, CivitAI, RunPod, LoRA,
Claude, ChatGPT, Gemini, Grok, Kimi, MiniMax, Ollama, LM Studio, OpenRouter,
HuggingFace, Antigravity, GitHub Copilot, PowerShell, TestFlight, GPU, VRAM), plus
`Pod` as the RunPod unit noun. 15 values are byte-identical to English and all 15
are names or pure format strings, not gaps:

`civitai_ui.civitai`, `sidepanel_ui.civitai`, `panel.claude`, `panel.gemini`,
`panel.openrouter`, `panel.chatgpt_codex`, `panel.github_copilot`,
`panel.comfy_mcp_agent`, `panel.command_prompt`, `panel.powershell`,
`panel.macos_linux`, `panel.ios_testflight`, `training_ui.preflight_docker`,
`training_ui.preflight_label` (`{label}:`), `runpod_ui.duration_pair`
(`{first} {second}`).

## Verification

- `node scripts/i18n-merge-parts.mjs ar` — 1269 keys, 52 settings, no drift
- `node scripts/i18n-render-check.mjs ar` — **999/999 rendered (100%), 0 fall back to English**, every plural base across counts 0/1/2/3/5/11/21/100
- `npm run test:unit` — **3985 pass, 0 fail** (locale gate + render check + vocabulary + suite)
- `node scripts/check-tool-vocabulary.mjs` run **after `git add`** — OK. Mutation-tested it rather than trusting the pass: gluing an Arabic letter onto `panel_graph_outline` in the staged `locales/ar/main.json` made it FAIL with the exact line, so the clean run is evidence the file is in scope.
- Additional parity audits, all zero mismatches: backticked shell commands (`` `claude` ``, `` `codex login` ``, `` `ollama serve` ``), URLs, underscore identifiers (`ZAI_API_KEY`, `MOONSHOT_API_KEY`, `MINIMAX_API_KEY`, `OPENROUTER_API_KEY`, `COMFYUI_MCP_BRIDGE_PORT`, `download_model`, `prompt_id`, `train_doctor`, `train_*`, `panel_graph_outline`/`panel_find_nodes`), and literal `\n` line breaks in the provider setup blocks.

No gate was weakened, no plural form or placeholder dropped.


artokun:
Rebased onto `origin/main` (now includes #1144) and repaired.

**Escape repair.** `node scripts/i18n-check.mjs` named the same 9 values the
coordinator listed. Each carried a two-character backslash-`n` copied faithfully
from the pre-#1144 English; they now carry a real line break in the same
positions. Verified by parity rather than by eye — the real-newline **count** per
key now matches English for all 1269 keys (0 mismatches), and **0 values in
either file contain a backslash at all**. Key set unaffected: still 1269 / 52.

**Bidi sweep** (per the Persian unit's finding). Scanned both files at raw byte
level for `U+200B–U+200F`, `U+202A–U+202E`, `U+2060–U+2069`, `U+061C`, `U+FEFF`,
`U+00A0`: **0 occurrences**. I had hit this trap while drafting — a few RLMs went
in around Latin fragments — and stripped them before the first push, so the
rule held.

**`Bidi_Mirrored` measurement matches yours**, and was checked mechanically:
`→ ← ↩ ↪ ▸ ◂ ▾` are all `Bidi_Mirrored = false` and so were hand-mirrored in all
38 places they appear; `« » › ( ) [ ]` are all `true` and were left exactly as
authored.

## Review findings fixed in the same push

An adversarial review of the diff (every English leaf read against its Arabic,
plus all 54 plural families × 6 forms) found no placeholder, fragment, escape,
HTML, or negation defects — it explicitly re-read every value containing
NOT/never/won't/can't/WARNING/⚠/safe/cannot-be-undone and found zero flips. It
did find these, all now fixed:

- `civitai_ui.checkpoints` — `نقاط التحقق` ("verification points") is the wrong domain sense for a CivitAI model-file category, and it broke the policy its own neighbour `loras` (`ملفات LoRA`) sets. Now `ملفات Checkpoint`, and `training_ui.cancel_this_training_run_saved_checkpoints_stay` was aligned to the same term.
- `runpod_ui.cost_per_hour` — `cmcp-runpod-ui.js` states the currency symbol is deliberately **not** translated; I had replaced `$` with `دولار`, which is also ambiguous across USD/CAD/AUD. Now `${cost}/ساعة`, and the prose blurb was aligned to `$0.30–0.70 للساعة`.
- `runpod_ui.paste_pod_id_from_console_runpod_io` — I had generalised `console.runpod.io` to "the RunPod console", losing the only place a user is told *where* to get a pod ID. Hostname restored.
- Three `_two` forms used the nominative dual in a slot governed by a verb, requiring the oblique: `panel.checked_errors_none_unchecked_two` (`عقدتان`→`عقدتين`), `training_ui.n_skipped_wrong_extension_two` and `n_skipped_not_a_real_image_two` (`ملفان`→`ملفين`). The third was self-inconsistent — it already wrote `صورتين` correctly later in the same value. The other 51 families were correct, including nominative where the noun is the subject.
- `comfyui-mcp_chatScope.name` — `نطاق محادثة الدردشة` was a pleonasm and disagreed with the runtime toasts `panel.chat_scope_*`. Now `نطاق المحادثة`.
- `civitai_ui.current`/`legacy` → `الحالية`/`القديمة` (agreement with the implied النماذج); `civitai_ui.negative` → `الموجّه السلبي`; `training_ui.new_character_lora` → `LoRA جديد لشخصية` (masculine `جديد` had been sitting against feminine `شخصية` on that screen's primary CTA).
- `panel.thinking_tokens` — the accusative *tamyīz* `رمزًا` is correct only for 11–99, but `fmtThinkTokens` emits 100–999 or `12.5k`. Now `رمز`.

## Two adjacent bugs the review surfaced — NOT in this diff, not mine to touch

Filing rather than fixing, since my unit owns `locales/ar/` only:

1. **`web/js/comfyui-mcp-panel.js:24492`** passes a hard-coded English literal
   `what: "question"` — no `tr()` — into `panel.the_connection_that_asked_this_dropped_so`.
   An Arabic user reads «…الذي أرسل question، لذا…». The translated
   `panel.question` is dead on that path; the `??` fallback further down never
   fires. This hits **every** non-English locale, not just Arabic.
2. **Two disclosure carets are not RTL-mirrored in code**: the CSS
   `content: "▸"` around :18826 and `btn.textContent = collapsed ? "▸" : "▾"`
   around :23871. The catalog mirrors `panel.advanced` to `◂`, which now makes
   the hard-coded ones visibly inconsistent in RTL.

## Re-verification after the rebase

- `node scripts/i18n-check.mjs` — `ar/main.json` 930/945, `ar/settings.json` 104/104, **locales OK** (the escape rule now passes)
- `node scripts/i18n-render-check.mjs ar` — **999/999 rendered (100%), 0 fall back to English**
- `npm run test:unit` — **4033 pass, 0 fail**
- `node scripts/check-tool-vocabulary.mjs` run after `git add` — OK
- 54 plural bases, 0 incomplete; 1269 main leaves; 52 settings / 104 leaves

One correction to my own PR body: the byte-identical-to-English count is **30**,
not 15 — the gate's counter requires `/[a-z]{4}/` so it skips `GPU`, `Pod`,
`" ×{n}"`, `"{first} {second}"`, `"{label}:"` and friends. All 30 were checked
individually; every one is a brand name or a pure format string, so the
substance of the claim stands.

artokun:
One more piece of evidence, since a clean instrument run only means something if
the instrument can fail on *this* file. I fired `i18n-render-check.mjs` at a
fixture copy of the shipped `locales/ar/main.json` with three injected defects,
one per class I was relying on it to police:

```
✗ ar: panel.command_failed: {cmd} was never substituted — rendered "فشل ⟦v1⟧"
✗ ar: panel.connected_auto_matched: English starts with a space; this joins straight onto the text before it
✗ ar: panel.nodes @count=2: the number is missing from "عقدتان"

3 rendering problem(s).
```

All three caught. The third is the useful one: it is the *bare dual* — the
citation form «عقدتان» with no numeral, which is what a native Arabic UI would
normally print for 2. It fails here by design, and `i18n-check`'s per-form
placeholder-parity rule rejects it independently. That is the constraint behind
the paradigm in the PR body: `{count}` is mandatory in all six forms, so the
numeral-less forms are not expressible and the wording is the numeral-agreement
register instead (`2 عقدتان`, `3 عقد`, `11 عقدةً`). Recording it here so the
choice reads as measured rather than as a translator shortcut.

PR #1146, opened 2026-08-12
fix(1145): stamp the outage once, not every failed retry

### LINKED ISSUE #1145: The mid-task nudge measures one backoff step, not the outage — a fast-returning restart loses its nudge
### What happens

The mid-task resume nudge decides whether a reconnect followed a REAL ComfyUI restart by the gap since the bridge dropped: a long gap means a real restart, a short one means a blip and the turn kept running. #1138 fixed the case where no drop had happened at all. The interval it measures is still wrong.

`connect()` assigns `sock = thisSock` immediately after construction, before the connection resolves. So a REFUSED attempt during a restart dispatches `close` on a socket that `isActive()` still accepts, the close handler runs in full, and it re-stamps `lastBridgeDownAt = Date.now()`.

With backoff doubling from `RECONNECT_BASE_MS` the retries land at roughly t=1s, 3s, 7s, 15s. Whichever attempt finally succeeds is separated from the previous failed close by only that attempt's delay — so the gap the guard weighs is **one backoff step**, not the outage.

Consequence: a genuine ComfyUI restart that comes back within ~7s is measured as a fast reconnect and **loses its nudge**. The session resumes with full context but no pending turn, so the agent sits idle after exactly the event the nudge exists for.

### Steps to reproduce

1. Start a long agent turn (mid-build or mid-render).
2. Restart ComfyUI so it comes back in under ~7 seconds.
3. The bridge drops, retries at ~1s and ~3s are refused (each re-stamping the drop time), the ~7s attempt connects.
4. The `ready` ack arrives with the mid-task marker set; the gap reads ~4s, under the 6s threshold, so no nudge.
5. The agent is idle with a resumed session and no turn.

### Also unaudited: the other ready-ack branches

#1138 fixed the mid-task branch only. The reboot-resume branch and the soft-reload branch read the same kind of state and were never checked for the equivalent "sentinel means never, but subtracts like 1970" exposure. Worth auditing together rather than one at a time — the class is what bit here, not the instance.

### Fix

Applied locally: no. Upstream-only: yes (panel).

Measure the OUTAGE, not the last retry: stamp the drop once per outage (on the first close that begins it) and leave it alone until a connection actually succeeds, rather than re-stamping on every failed attempt. Then audit the sibling ready-ack branches for the same class.

### Notes

Confirmed during review of #1139 (#1138); recorded in 0.14.18's release notes as deliberately left rather than patched there, because it is a different defect in the same guard and worth its own change.

artokun:
Closes #1145. Follow-up to #1138 (shipped in 0.14.18), which fixed a different defect in the same guard.

The mid-task nudge decides "real restart vs blip" from how long the bridge was gone. The interval it weighed was not that.

## What the investigation established

**The re-stamp is per-attempt, not per-outage — confirmed at the source, not inferred.** `connect()` does `sock = thisSock` immediately after `new WebSocket(url)`, before the connection resolves (`web/js/comfyui-mcp-panel.js`). `isActive()` is `sock === thisSock`, so a REFUSED attempt is still the active socket: its close handler runs in full and re-stamped the drop time. A failed WebSocket handshake fires `error` and then `close` (code 1006) — `close` is not conditional on the connection ever having opened, so the path is reached on every refused retry, not just some. With backoff doubling from `RECONNECT_BASE_MS`, retries land near t=1s/3s/7s/15s and the successful one is separated from the previous failed close by only that attempt's delay. The guard read its own retry cadence.

**Where the outage begins — a clean definition was available.** The first close of a sequence begins it; the handshake that ends it is `markConnected()`, the `models` frame. That is the same line the replay already draws: an open socket only proves something holds the port, a `models` frame proves an orchestrator is behind it. No definition of "when the sequence ends" is needed that depends on the thing being measured — the end is an event, not a timeout. So the honest alternative floated in the draft (drop the clock entirely) was not needed here.

**A second reading of the same guard, found while defining the first.** `Date.now() - stamp` answers "how long since the last drop, whenever that was", and it kept answering after the outage it described had ended. `ready` repeats on a live socket and the panel re-advertises after every successful `free_vram` (#310) — so a turn running ten minutes and two reconnects after a real drop could be told its connection had just dropped. That is #1138's defect with a real timestamp in place of the `0` sentinel; requiring the stamp to be positive does not catch it. Closed here rather than deferred, because it is the same read of the same variable.

**The sibling branches, as a class — audited, and clean.** The reboot-resume branch already decides on an observed drop rather than elapsed time, and says so: *"Elapsed time is not evidence of a new connection; an observed drop is."* The soft-reload branch reads a marker it sets and clears itself, with no clock at all. The mid-task branch was the only one deriving a restart from a clock, so the class turns out to be one instance — this change brings the odd branch in line with the pattern the other two already use.

## The change

`createBridgeOutageTracker()` in `web/js/lib/session-rebind.js`, wired at three points:

| event | wiring | rule |
|---|---|---|
| socket `close` | `noteBridgeClosed()` | the FIRST close opens the outage; refused retries after it do not move its start |
| `markConnected()` | `noteHandshake()` | measures the outage that ENDED here; a handshake that ended none records nothing |
| `onTurn("working")` | `noteTurnStarted()` | evidence is scoped to the turn now running |

`shouldNudgeAfterMidTaskReconnect` now takes a measured `outageMs` instead of `{bridgeDroppedAt, now}`. A duration cannot be silently reinterpreted as "time since something", which is what let one variable mean the wrong thing twice.

## Why a tracker, and why tests over it

The defect is a SEQUENCE — four closes in a row, each resetting the clock. No single call distinguishes the drop from the third refused attempt, so nothing about it is visible in a pure per-call predicate, and a source scan over the panel could not have caught it: #1096's review proved by mutation that token-presence scans over that file stay green when the logic under them is inverted. The tracker is driven through the real order of events and asked what it measured.

Each guard was mutation-checked — the mutant is named, and the test that dies with it:

- re-stamp on every close → *refused retries during a restart must not reset the outage clock*
- a re-advertise inherits the previous outage → *a handshake that ended NO outage inherits nothing*
- turn start stops clearing settled evidence → *an outage that ended BEFORE this turn is not this turn's evidence*
- the outage is never closed out → *a second outage measures itself, not the gap since the first*

## Verification

- `npm run test:unit` — 4012 pass, 0 fail (14 new tests)
- `npm run typecheck` — clean
- `node --check` over all of `web/js/**/*.js`, plus the panel copied to `.mjs` and re-checked as a module (the check the file's own header asks for)
- `python -m py_compile __init__.py`

One neighbouring assertion was bounded to `markConnected`'s BODY rather than a fixed 1600-character window: the window stood in for "inside this function" and stopped meaning that the moment the function grew a comment — a passing assertion turning into a failing one for a reason unrelated to what it checks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,40 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- **The mid-task resume nudge now measures the outage instead of one backoff step, so a
+  ComfyUI restart that comes back quickly keeps its nudge (#1145).** The nudge tells a
+  resumed agent its connection dropped mid-task and to continue what it was doing, and it
+  fires only for a REAL restart — judged by how long the bridge was gone. But the panel
+  assigns its socket before the connection resolves, so a REFUSED reconnect attempt is
+  still the active socket and ran the close handler in full, re-stamping the drop time.
+  With backoff doubling, the retries land near one, three, seven and fifteen seconds, so
+  the attempt that finally connected was separated from the previous failed close by only
+  that attempt's delay. The guard weighed its own retry cadence, not the outage: a genuine
+  restart returning inside about seven seconds measured roughly four and lost its nudge,
+  leaving the agent idle with a resumed session and no pending turn — after exactly the
+  event the nudge exists for. The outage is now stamped once, by the first close that
+  begins it, and its duration measured at the handshake that ends it.
+  A second reading of the same guard is closed with it: a drop stamp answers "how long
+  since the last drop, whenever that was", and kept answering after the outage it
+  described had ended — so a `ready` repeating on a live socket (the panel re-advertises
+  after every successful `free_vram`, #310) could be told about a drop from ten minutes
+  and two reconnects earlier. That is #1138's defect with a real timestamp in place of the
+  zero sentinel. What the guard reads is now a measured duration, scoped to the turn now
+  running, and a handshake that ended no outage records nothing rather than inheriting the
+  previous one.
+  The accounting moved into a tracker with unit tests that drive it through the real
+  sequence — drop, refused retry, refused retry, handshake — because no single call can
+  distinguish a drop from the third refused attempt, and a source scan over the panel
+  could not have caught this (a review previously proved by mutation that such scans stay
+  green through an inversion). Each guard was mutation-checked against those tests.
+  The two sibling `ready`-ack branches were audited for the same class and are clean: the
+  reboot-resume branch already decides on an observed drop rather than elapsed time
+  ("elapsed time is not evidence of a new connection; an observed drop is"), and the
+  soft-reload branch reads a marker it sets and clears itself. The mid-task branch was the
+  only one deriving a restart from a clock.
+
 ## [0.14.21] - 2026-08-12
 
 ### Added
--- a/browser_tests/unit/command-liveness.test.mjs
+++ b/browser_tests/unit/command-liveness.test.mjs
@@ -501,8 +501,15 @@ test("#508 wiring: an undelivered reply is journaled, replayed after hello, and 
     false,
     "replay must not fire at bare socket-open",
   );
-  const markConnected = src.slice(src.indexOf("function markConnected() {"));
-  assert.match(markConnected.slice(0, 1600), /replayLostReplies\(sock\);/, "replay rides the handshake");
+  // Bound to markConnected's actual BODY, the way the #508 R6 test below bounds
+  // handleUndeliveredReply's. A fixed character window stood in for "inside this
+  // function" and stopped meaning that the moment the function grew a comment
+  // (#1145 added one) — a passing assertion turning into a failing one for a reason
+  // unrelated to what it checks.
+  const mcStart = src.indexOf("function markConnected() {");
+  assert.notEqual(mcStart, -1, "could not locate markConnected in the panel source");
+  const markConnected = src.slice(mcStart, src.indexOf("\n  }", mcStart));
+  assert.match(markConnected, /replayLostReplies\(sock\);/, "replay rides the handshake");
 });
 
 test("#508 codex R6: an outcome journaled AFTER the replacement socket replayed is delivered anyway", () => {
--- a/browser_tests/unit/command-liveness.test.mjs
+++ b/browser_tests/unit/command-liveness.test.mjs
@@ -508,7 +508,12 @@ test("#508 wiring: an undelivered reply is journaled, replayed after hello, and 
   // unrelated to what it checks.
   const mcStart = src.indexOf("function markConnected() {");
   assert.notEqual(mcStart, -1, "could not locate markConnected in the panel source");
-  const markConnected = src.slice(mcStart, src.indexOf("\n  }", mcStart));
+  // Both ends are asserted: a -1 from either indexOf would slice a window that is not
+  // the function — an unfound END silently degrades `slice` into a near-whole-file scan,
+  // which would pass no matter where replayLostReplies actually sits.
+  const mcEnd = src.indexOf("\n  }", mcStart);
+  assert.notEqual(mcEnd, -1, "could not locate the end of markConnected");
+  const markConnected = src.slice(mcStart, mcEnd);
   assert.match(markConnected, /replayLostReplies\(sock\);/, "replay rides the handshake");
 });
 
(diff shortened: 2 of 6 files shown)


artokun:
## Question 1 answered: the per-attempt re-stamp is real, confirmed by reading

Not inferred from assignment order — read directly:

- `sock = thisSock` (17304) is assigned **immediately after construction**, before the connection resolves.
- the close handler's only guard is `if (!isActive()) return;` (18134), and `isActive()` compares against that same `sock` — so a **refused** attempt passes the guard, runs the handler in full, and re-stamps the drop time at 18146.

So every failed retry re-stamps, and the gap the nudge weighs is the last attempt's backoff delay rather than the outage. Confirmed.

## Question 2 changes the shape of the fix

"Stamp once on the first close" is not sufficient on its own, and the reason is the trap I already fell into on #1138.

The value has to **survive the successful reconnect**, because the `ready` ack is what reads it — but it must **not** be re-stamped by the intermediate failures. That combination means the stamp becomes state with a lifetime: set once when an outage begins, and cleared when something consumes it.

Consume-on-every-exit is exactly what I got wrong in #1138's reverted commit: the review found my consume covered **one of seven** exits (reboot resume, soft reload, turn end, fresh session, idle drops all retire `MID_TASK_KEY` and left the stamp standing), which let an old drop authorize a later false nudge. So this fix cannot be "stamp once" — it is "stamp once **and** enumerate every exit", and the enumeration is the hard half.

## Which suggests the alternative is worth costing first

The panel may not need a clock here at all. The question the nudge actually asks is "did the agent's turn survive?" — and the panel has more direct evidence of that than an elapsed interval: whether turn frames are still arriving for the session it resumed. A guard built on that needs no stamp, no lifetime, and no exit enumeration, which removes the entire class of defect that has now bitten this guard twice (#1138's sentinel, and this).

Before writing either version I want to know whether that evidence is actually available at ready-ack time, or whether the ack necessarily precedes the first turn frame. If it does precede it, the clock is unavoidable and the exit enumeration has to be done properly and tested per exit.

Stopping here rather than starting an implementation I cannot finish and verify in this session. Next step is that availability question, not code.

artokun:
## Question 3 answered: the evidence exists, so the clock is avoidable

From the panel's own comment at 26209:

> a turn that IS still alive re-announces via `turn:working` on reconnect

That settles the design question. The nudge does not need to infer "did the turn survive?" from an elapsed interval at all — a surviving turn **announces itself**. The panel already has the direct signal, and the whole clock is a proxy for something it is told.

So the fix for #1145 should not be "stamp the outage once and enumerate every consume exit". That path was only forced by keeping the clock, and it is the path that produced #1138's reverted commit (consume covered one of seven exits) and the sentinel bug before it. Twice now the defect has been in the bookkeeping around this heuristic rather than in the heuristic's intent.

**Shape to build instead:** on a `ready` ack with the mid-task marker set, do not decide immediately. A live turn re-announces via `turn:working`, so give it that chance and nudge only if no such announcement arrives — the absence of a re-announcement is the actual evidence that the turn did not survive. That removes:

- the drop timestamp entirely (no sentinel, no 1970 arithmetic — #1138's bug class)
- its lifetime (no stamp-once, no consume, no seven exits — the reverted commit's bug class)
- the backoff-vs-outage confusion (this issue)

all three at once, because none of them are questions the panel has to answer any more.

**What still needs checking before writing it:** whether `turn:working` re-announcement is reliable enough to gate on — specifically whether it is guaranteed for a resumed session or only best-effort, and what the ordering guarantee is against the `ready` ack. The straggler-suppression machinery around 26148/26299 (`localEndAt`, "briefly ignore a straggler turn:working from this turn") is directly relevant, since it exists precisely because these frames arrive out of order relative to local state.

Also worth noting: #1138's shipped guard remains correct and is not affected. It answers "did anything drop at all", which stays a genuine precondition under either design.

Next step is that reliability/ordering question, then the implementation.

artokun:
## The re-announcement is load-bearing, not incidental — good news for the design

The reaper at 26205-26215 already depends on it: when it decides a turn is not live it clears `agentWorking` AND `MID_TASK_KEY`, and justifies doing so precisely because "a turn that IS still alive re-announces via turn:working on reconnect (outside the stale-working window), which re-shows it."

So the signal is not best-effort trivia — an existing mechanism already treats it as reliable enough to reap a turn on its absence. Gating the nudge the same way is consistent with shipped behaviour rather than a new assumption, which was the risk worth checking.

It also means the two mechanisms would agree by construction: the reaper clears the mid-task marker for a dead turn, and the nudge fires only when no re-announcement arrives. Same evidence, same direction.

One caveat carried forward: the parenthetical "outside the stale-working window" points at the `localEndAt` suppression near 26148/26299, so an implementation has to respect that window rather than treat any absence as death. That is the remaining detail, not a blocker.

Next session can implement from here.

PR #1150, opened 2026-08-12
feat(i18n): add the French (fr) panel catalog — 1107 keys, one/many/other plurals

artokun:
The 12-language picker has offered **Français** since the i18n foundation landed, but no `fr` catalog existed — selecting it rendered English. This adds both files, written from scratch.

## Final counts

| file | count |
|---|---|
| `locales/fr/main.json` | **1107 keys** — 945 ordinary + 54 counted strings × 3 plural forms |
| `locales/fr/settings.json` | **52 settings / 104 leaves** |

`i18n-check` reports `fr/main.json 900/945 (95%)` and `fr/settings.json 103/104 (99%)`. Nothing is truncated: the gap is values that are *legitimately identical* to English, enumerated below.

## Plural forms

Modern CLDR gives French three categories and the skeleton emitted exactly those: **`one`, `many`, `other`**. No `_<category>` suffix was added to or removed from what `i18n-skeleton.mjs fr` generated. `many` covers large/compact values and is **not** the English `one`/`other` split, so every counted base carries all three forms — 54 bases × 3 = 162 keys, complete, none half-formed.

French treats 0 like 1, which the render check confirms across counts 0/1/2/3/5/11/21/100:

```
0 => 0 nœud       | 0 téléchargement
1 => 1 nœud       | 1 téléchargement
2 => 2 nœuds      | 2 téléchargements
1000000 => 1 000 000 téléchargements
```

## Verification

Against `origin/main` @ **0.14.19** (this branch is rebased onto it, not onto the tree I started from):

```
node scripts/i18n-render-check.mjs fr   ✓ fr — 999/999 rendered (100%), 0 fall back to English
node scripts/i18n-check.mjs             locales OK
node scripts/check-tool-vocabulary.mjs  OK (run AFTER git add — see below)
npm run test:unit                       4033 pass, 0 fail
```

Four things beyond just running the gates:

1. **The vocabulary gate was mutation-tested against this file.** `check-tool-vocabulary` enumerates with `git ls-files`, so a green run before staging proves nothing. I staged the catalog, injected `panel_graph_outliné` into `panel.nothing_is_selected_on_the_canvas_ask`, and confirmed the gate fails on `locales/fr/main.json:661` (rule 2b *and* the near-miss rule), then restored. The green is therefore about my file, not about my file being invisible.

2. **The render-check instrument was fired at a broken `fr` fixture.** A dropped `{model}`, a dropped trailing space on `civitai_ui.prompt`, and a dropped leading space on `panel.coming_soon` were all caught. The instrument can see the defect classes French is most likely to introduce.

3. **Independent per-key parity audit** (not relying on the gates), re-run after the rebase so it is measured against the *current* English: placeholder *hole sets* — **0 mismatches**; leading/trailing-space presence — **0 divergences**; newline count — **0 divergences**.

4. **No invisible characters.** 0 NBSP/narrow-NBSP, 0 C0/C1 control bytes, 0 zero-width characters, no BOM, trailing LF present. French typography wants a narrow no-break space before `: ; ! ?`; I used ordinary spaces instead, matching ComfyUI's own French and keeping the files free of characters that break regexes silently.

## Deliberately left in English

73 values are byte-identical to English, all of them intentionally:

- **Brands / product names** (per the brief): ComfyUI, CivitAI, RunPod, LoRA, Claude, ChatGPT, Gemini, Grok, Kimi (K3), MiniMax, Ollama, LM Studio, OpenRouter, GitHub Copilot, PowerShell, TestFlight, GPU, VRAM, Antigravity, llama.cpp, Pi (pi.dev), GLM (z.ai), Comfy MCP Agent.
- **Words spelled the same in French**: Description, Images, Standard, Local, Auto, Image, Internet, Prompt/Prompts, Max, Minimal, Ultra, Effort, question, segment, Checkpoints, Workflows.
- **Format-only strings** with no prose: `{first} {second}`, `{status} · pod`, `Plan · {done}/{total}`, `max {max}`, ` ×{n}`.
- **Adopted loanwords** I chose to keep as the French term, for terseness in a narrow sidebar and because they are what French ComfyUI users say: *workflow*, *dataset*, *job*, *prompt*, *seed*, *checkpoint*, *backend*, *endpoint*, *reroute*, *token*, *pod*.

## Glossary held across all ~1100 keys

node → **nœud** · graph → **graphe** · subgraph → **sous-graphe** · canvas → **canevas** · queue → **file d'attente** · snapshot → **instantané** · revert → **restaurer** · rewind → **rembobiner** · provider → **fournisseur** · bridge → **pont** · turn → **tour** · feed → **flux** · deafen → **sourdine** · blind → **aveugle** · composer → **zone de rédaction** · effort → **effort**.

Register is vouvoiement throughout, matching ComfyUI's own French.

## A wrong call I made, and the gate that caught it

Worth recording rather than hiding, because the gate did its job.

When I branched, `locales/en/main.json` still carried **extractor artifacts** — 13 keys whose value held a literal two-character `\n` (or `\"`) that the source JS did not:

```js
// web/js/comfyui-mcp-panel.js — a REAL newline
"…(create one at https://openrouter.ai/keys):\n" + "  • Settings → OpenRouter…"
```

I judged that mirroring the artifact was the right call (the already-shipped Korean catalog mirrors it, and one extractor fix would then correct all 12 languages at once) and shipped nine French values carrying literal `\n`.

That was wrong, and **CI said so on the first push**: `#1144 — fix(i18n): decode string escapes when extracting` had landed on `main` while I worked, adding an `i18n-check` rule that rejects undecoded escapes by name. All nine of my keys failed it.

The fix is what the gate asked for, not a way around it: this branch is **rebased onto `origin/main` @ 0.14.19** and all nine values now carry **real newlines**, proven by a per-key newline-count diff against the decoded English (**0 mismatches**, and **0 fr values** still holding any `\n`/`\t`/`\r`/`\"` escape). Quotation marks throughout use French guillemets **« »**, which is the correct French mark regardless of what English uses.

The lesson is the ordinary one: a green local check is only green for the `main` it ran against.

## Not done here

No e2e run: `/i18n` is `@lru_cache(maxsize=1)` so a locale change needs a ComfyUI restart, and there is one shared instance on `:8188`. Per the coordinator's recipe, `i18n-render-check.mjs` is the end-to-end verification for this unit; the live 12-language DOM walk happens once after merge.

Only `locales/fr/` was touched — nothing under `web/`, `scripts/`, `browser_tests/`, `package.json`, or any other language's directory.


artokun:
Code review pass — two findings applied in fb5208c6, and one "correction" checked and declined with evidence.

**Applied**

1. `panel.signs_in_as_vs_code_against_github` / `panel.experimental_signs_in_as_vs_code_against` — real weakening, good catch. EN says signing in as VS Code is *against* GitHub's Copilot API terms; my "contrairement aux conditions" means *unlike / in contrast to*, i.e. a neutral remark where English gives a compliance warning. The chip tooltip carries no risk clause at all, so that was the entire warning. Both now read **"en violation des conditions"**.
2. `training_ui.1_2_h_on_a_4090_the` — "La vraie chose." was a calque of "The real thing." Now **"Du sérieux."**

**Declined, with the measurement**

The review reported that the extractor artifacts I describe above "do not exist" — that zero values in any locale contain a literal backslash, and that my premise came from `node -e` backslash mangling in Git Bash.

That warning about `node -e` is a good one, so I re-measured with no shell in the path: a `.mjs` file, reading git blobs through `execFileSync` (argv, not a shell string), classifying by **codepoint 92** rather than by regex.

```
en @ 300d8359 (my base, pre-#1144): 13 value(s) contain a literal backslash
    apps_ui.hide_the_workflow_for_this_deletes_the -> backslashes=2  newlines=0
en @ origin/main (post-#1144):        0
fr @ 67d9b866 (the commit CI rejected): 9
fr @ HEAD (shipping now):               0
```

The artifacts were real at the commit I branched from; **#1144 removed them while this branch was open**. A third, independent witness agrees: CI on 67d9b866 failed `i18n-check` with *"contains an undecoded escape (a backslash before n/t/r/quote)"* naming exactly those 9 keys — a Linux runner, the project's own gate, reading the committed bytes.

So the likely explanation is the mirror image of the one proposed: the review measured English at *current* `main` (post-#1144, genuinely 0) and concluded the artifacts never existed. The PR body was already rewritten after CI and describes this accurately, so nothing there needs changing.

Everything else in the review re-verified independently and matched — plural families, symmetric hole-set diff, exact whitespace comparison, and the gender split on `n_selected`. Suite green against `origin/main` @ 0.14.19: 4033 pass, 0 fail.

PR #1162, opened 2026-08-12
fix(i18n): translate the "why can't I use this provider" hints in all 11 languages

artokun:
## What

`notReadyHint()` is the panel's entire answer to "why can't I use this provider" — *Ollama not running — run: `ollama serve`*, *No OpenRouter API key — add it via API Keys*, *Not signed in — run: `claude auth login`*. #1160 wired those 22 strings (20 hints + the two disclosure carets) for translation, which added 22 keys to `locales/en/main.json`. Every non-English catalog then fell back to English on exactly those keys.

This fills all 22 in all eleven languages. Purely additive: 22 keys per catalog, no existing value touched, key order preserved, nothing outside `locales/<lang>/main.json`.

## Coverage, before → after

`node scripts/i18n-check.mjs`, `main.json`:

| lang | before | after |
|---|---|---|
| zh | 930/967 (96%), 22 untranslated | 952/967 (98%), 0 untranslated |
| zh-TW | 929/967 (96%), 22 untranslated | 951/967 (98%), 0 untranslated |
| ru | 927/967 (96%), 22 untranslated | 949/967 (98%), 0 untranslated |
| ja | 932/967 (96%), 22 untranslated | 954/967 (99%), 0 untranslated |
| ko | 929/967 (96%), 22 untranslated | 951/967 (98%), 0 untranslated |
| fr | 900/967 (93%), 22 untranslated | 922/967 (95%), 0 untranslated |
| es | 905/967 (94%), 22 untranslated | 927/967 (96%), 0 untranslated |
| pt-BR | 903/967 (93%), 22 untranslated | 925/967 (96%), 0 untranslated |
| tr | 915/967 (95%), 22 untranslated | 937/967 (97%), 0 untranslated |
| ar | 930/967 (96%), 22 untranslated | 952/967 (98%), 0 untranslated |
| fa | 929/967 (96%), 22 untranslated | 951/967 (98%), 0 untranslated |

Every language gains exactly 22 keys and `i18n-check` no longer reports *"N not yet translated (renders English)"* for any of them. The residual difference against 967 is plural categories the language does not use (`_one` in zh/zh-TW/ja/ko), which is correct by design — a raw key-set diff against `locales/en` is now **0** for ar, es, fa, fr, pt-BR, ru and tr, and 54 plural-category entries for the CJK four.

## Left in English on purpose

Everything the user types or reads literally, inside an otherwise translated sentence:

- commands — `ollama serve`, `codex login`, `claude auth login`, `gemini`, `grok`, `kimi`, `agy`, `pi`, `/login`, `llama-server -m model.gguf --jinja -c 16384`
- env vars — `ZAI_API_KEY`, `MINIMAX_API_KEY`, `MOONSHOT_API_KEY`, `ANTHROPIC_API_KEY`
- domains/paths — `ollama.com/download`, `lmstudio.ai`, `github.com/ggml-org/llama.cpp/releases`, `pi.dev`, `PATH`
- the LM Studio menu path `LM Studio → Developer → Start Server`, because that is what the untranslated LM Studio UI actually says
- brand names — Ollama, LM Studio, llama.cpp, OpenRouter, Antigravity, Moonshot, MiniMax, Grok, Kimi, Codex, Gemini, Claude, DeepSeek, vLLM, Google

The `{label}` placeholder in `panel.not_ready_cli_not_installed` is preserved in all eleven. In `ar` and `fa` the word "CLI" itself is rendered as واجهة سطر الأوامر / ابزار خط فرمان, matching what `panel.install_the_antigravity_cli_agy` already ships in those catalogs.

## Two things that are easy to get wrong

**RTL carets.** `panel.caret_collapsed` is `◂` in `ar` and `fa`, `▸` everywhere else. A collapsed caret points at the text it would reveal, and that is leftwards in RTL — it matches what `panel.advanced` already ships in both catalogs. `panel.caret_expanded` stays `▾` in every language.

**Rule 2b.** `check-tool-vocabulary` scans locale JSON for a non-ASCII *letter* inside an underscore-joined word, and its character class swallows fullwidth `（）：、，`, the em dash and the Persian ZWNJ — any of which drags a Han, Hangul, Arabic or Cyrillic letter into the identifier's word and fails CI. Every underscored identifier here is separated from adjacent non-Latin script by an ASCII space or ASCII punctuation. Concretely, `zh`/`zh-TW` end the pi hint with the env var at end-of-string (`…（例如 ANTHROPIC_API_KEY）。`), `ja` uses an ASCII `例: ` before it, and `ko` relies on the ASCII `)` to break the run before the particle.

## Verification

```
node scripts/i18n-check.mjs         # locales OK
node scripts/i18n-render-check.mjs  # 11/11 at 1021/1021 rendered, 0 fall back — rendering OK
node scripts/check-tool-vocabulary.mjs   # run AFTER git add — OK
npm run test:unit                   # 4078 pass, 0 fail, 1 todo
```

`i18n-render-check` is the real end-to-end: it loads each catalog into the actual `web/js/lib/i18n.js` and inspects the painted string. Spot-checked `tr('panel.not_ready_cli_not_installed', …, {label:'Moonshot'})` through the real module in ko/ar/zh/ja/fa — substitutes correctly — and confirmed the caret keys render `◂`/`▾` in ar and fa, `▸`/`▾` elsewhere.

No live ComfyUI/browser run: the shared instance is down and is the coordinator's to verify.


artokun:
## Review pass — 1 fix applied, 10 findings triaged as out of scope

Fixed in c4acea6f — **fr was the only locale that rendered "Not signed in" as "Non connecté"**, which is already its phrase for the bridge being down (`not_connected_reconnect_the_panel_first` = "Non connecté — reconnectez d'abord le panneau."). A French user with a live bridge and a signed-out Codex would have seen the status pill say "connecté" and the hint say "Non connecté" — self-contradicting, and word-for-word the message meaning the panel lost its connection. Now "Non authentifié" for the five signed-out hints; the action verb stays "connectez-vous" to match `sign_in` = "Se connecter".

The remaining findings are all real but **pre-existing in the English source or in `web/`**, which this locale-only PR must not touch. Recording them so they are not lost:

**Worth a follow-up issue (product-side, `web/js/comfyui-mcp-panel.js`):**

1. **The hint is unrecoverably truncated.** `item({small})` at :22445 sets `s.textContent = small` and never sets a `title`, while :19594 is `.cmcp-popover-item.cmcp-provider small { overflow:hidden; text-overflow:ellipsis; white-space:nowrap }`. The payload is terminal in every hint — `ollama serve`, `/login`, `llama-server -m model.gguf --jinja -c 16384`, `ollama.com/download` — so the ellipsis eats exactly the part the user needs, with no hover to recover it. This already bites the 109-char English string; translations run up to 40% longer. **The fix is one line (`s.title = small`), not shorter translations** — trimming eleven catalogs to fit a broken layout would be papering over it.
2. **`not_ready_antigravity_install_cli` is returned from both branches** (:20779 under `r.cli === false` and :20785 in the auth branch), so a user who already has `agy` installed but no Google session is told to install it. There is no `not_ready_antigravity_signed_out` key.
3. **The auth branch has no `{label}` fallback.** The CLI branch escapes to `not_ready_cli_not_installed` with `{label}` (:20781); the auth branch just falls through to `return tr("panel.not_ready_claude_signed_out", …)` (:20797), so any future backend is told to run `claude auth login` under a non-Claude chip.
4. **Hosted, CLI-less providers may never reach their `*_no_key` hint.** openrouter and custom are special-cased out of the `r.cli === false` branch (:20773, :20775); moonshot/glm/minimax are equally CLI-less but are not, so they may render "the MiniMax CLI is not installed". Depends on how the orchestrator populates `b.cli` — worth confirming.
5. **The literal ▾ points at a chevron that is not ▾.** The status pill renders `<i class="pi pi-angle-down">` (:20589-20594), a thin PrimeIcons outline, but 8 strings tell the user to look for a filled ▾.
6. **`not_ready_openrouter_no_key` is the only `*_no_key` hint that never names its env var** — its three siblings name ZAI_API_KEY / MINIMAX_API_KEY / MOONSHOT_API_KEY.

Each of 2-6 is a one-line change in `locales/en/main.json` or the panel source **today**, and an eleven-locale change once translated — so they are cheapest to fix before the next translation batch, not after.

**Declined:**

- *Carets belong in CSS, not the catalog* (`[dir="rtl"] … { transform: scaleX(-1) }`) — correct critique, but of #1160's design, not of this fill. Filing it as a follow-up beats leaving eleven locales falling back to English in the meantime.
- *zh/zh-TW/ko fold "takes effect immediately" into a conditional* — "添加即可立即生效" / "추가하면 바로 적용됩니다" does assert immediacy. Restoring a separate clause would put fullwidth punctuation straight after `ZAI_API_KEY`, which is exactly the rule-2b failure this PR was careful to avoid.
- *RTL bidi isolation around the doubled `llama-server`* — I worked the UAX#9 resolution through: the em dash between an R run and an L run takes the paragraph level, so the runs lay out right-to-left in logical order and **the reading order is correct**. The reviewer concedes this. Injecting U+2068/U+2069 would add invisible \p{Cf} characters to the catalog for an aesthetic gain.
- *The caret keys inflate the coverage metric* — true, and it points at `locales/en/main.json`, which is out of bounds here.

PR #1174, opened 2026-08-13
fix(654): the panel must re-register its tab after panel_restart_comfyui

### LINKED ISSUE #654: Panel does not reconnect after panel_restart_comfyui

artokun:
Closes #654.

After `panel_restart_comfyui` reports `server_ready:true`, the sidebar tab did not reconnect: `panel_tab_reconnected:false`, `panel_set_workflow_target` stayed deferred, and `panel_graph_outline` failed with `no connected tab` until the browser tab was refreshed by hand.

## First question, before any change

**Whether this still reproduces.** The report is from panel **0.11.39**; main is **0.14.24**, and the reconnect path has since grown a good deal of #654-tagged machinery — a refused-hello retry budget, the tab-route lease, and the #278 resume guard among them. Some or all of this may already be fixed, and "fixing" a repaired path is how a working guard gets inverted.

So this starts by establishing which half is actually broken today: the socket not coming back at all, or the socket returning while the tab's ROUTE registration does not — those have different causes and the report's symptoms (`graph_tools_ready:false` with a healthy server) point at the second.

Draft while I read the restart → reconnect → hello → route-registration chain against current main.


artokun:
Closing unmerged: #654 does not reproduce on current main, and I am not going to write a fix for a path I cannot exercise.

Full evidence on the issue. In short: I killed and restarted ComfyUI under a live panel with the browser tab open, and the panel stayed connected throughout — no close, no new socket, no re-hello. The bridge never dropped because this deployment runs the orchestrator **out-of-band** on its own port, so a ComfyUI restart does not kill it. #654 needs the orchestrator to die *with* ComfyUI, which the reporter's older `comfyui-mcp 0.49.6` likely did.

Two #654-tagged mechanisms have also landed since the reported 0.11.39 — a retryable identity lease and a bounded refused-hello retry — either of which may already cover the reported wedge. Neither was written up in the changelog, which is likely why this stayed open.

Guessing at the remaining gap would mean choosing blind between two different causes (socket not returning vs. route registration not restored), against code I cannot exercise. That is precisely how a correct guard gets inverted, which this cluster has already suffered twice. The issue now asks for the one fact that decides it.

Issue #517, opened 2026-08-02
Timed-out graph mutations can still apply and create duplicate nodes on retry


artokun:
### What happened / root cause
A panel_add_node mutation timed out and reported an error, but the frontend applied the node after the timeout. Retrying the apparently failed call created a second node. The first node remained as an orphan duplicate inside a native subgraph. A later cleanup panel_remove_node also timed out, so its final mutation state must be re-read before another retry.

### Steps to reproduce
1. Enter a native subgraph.
2. Call panel_add_node for LoraLoaderModelOnly while the wf:workf tab is intermittently slow.
3. Receive: Panel tab wf:workf did not reply to "graph_add_node" within 6000 ms.
4. Refresh graph binding and retry panel_add_node.
5. Inspect the subgraph: both the timed-out mutation and retry exist; one duplicate is orphaned.
6. Call panel_remove_node on the orphan; the mutation also times out and its applied state is ambiguous.

### Exact error
Panel tab wf:workf did not reply to "graph_add_node" within 6000 ms — the ComfyUI tab may be backgrounded or frozen
Panel tab wf:workf did not reply to "graph_remove_node" within 6000 ms — the ComfyUI tab may be backgrounded or frozen

### Fix
Applied locally: workflow cleanup only. Upstream-only: yes.
Precise change needed: graph mutation requests need request IDs/idempotency and late acknowledgements. After a timeout, reconcile the request ID or query the resulting node/mutation before reporting failure; retries with the same request ID must not duplicate the mutation. For destructive cleanup, return an explicit unknown/committed state rather than a plain failure.

### Environment
Windows 11; RTX 4080 SUPER 16GB; ComfyUI 0.29.2; comfyui-mcp 0.48.22; panel 0.11.32.

artokun:
Resolved in shipped 0.11.33 by #521 (8e1434b) plus follow-ups 479f0a7, 41f45e4, and 2e2e6c7. Current focused validation: command-dedupe 13/13; MCP ui-bridge 122/122. Closing as stale rather than reintroducing an unsafe retry ledger.

artokun:
[...]

Issue #508, opened 2026-08-01
Panel command bridge times out for all live-canvas commands


artokun:
### What happened / root cause
The sidebar chat remained connected and accepted the user request, but every live frontend command timed out without a reply from the registered panel tab. This blocks graph inspection and the real Cloud Run frontend flow. Root cause appears to be a wedged frontend command bridge/tab registration: the orchestrator still targets `wf:462cc`, while the tab never acknowledges commands.

### Steps to reproduce
1. Open a workflow in ComfyUI with the sidebar panel connected.
2. Send a panel-agent request.
3. Invoke `panel_set_todo`, `panel_graph_outline`, then `panel_list_workflows`.
4. Each command waits 5–6 seconds and times out.

### Exact error (scrubbed)
`Panel tab wf:462cc did not reply to "set_todo" within 5000 ms — the ComfyUI tab may be backgrounded or frozen`
`Panel tab wf:462cc did not reply to "graph_outline" within 6000 ms — the ComfyUI tab may be backgrounded or frozen`
`Panel tab wf:462cc did not reply to "workflow_list" within 6000 ms — the ComfyUI tab may be backgrounded or frozen`

### Fix
Applied locally: no. Upstream-only/runtime-state issue: yes. The frontend bridge should detect/re-register a stale command handler (or surface a reconnect action) when chat remains alive but all graph command acknowledgements stop.

### Environment
macOS darwin; GPU MPS 16 GB; RAM 16 GB; ComfyUI 0.29.0 LOCAL; torch 2.12.1; Python 3.13; comfyui-mcp 0.48.18; panel 0.11.20.

artokun:
Additional reproduction context: on Windows 11 with an RTX 4080 SUPER, ComfyUI 0.29.2, comfyui-mcp 0.48.22, and panel 0.11.32, this occurred after a successful save of a 92-node workflow following a long render/postprocess sequence. The registered tab remained `wf:workf`, but `panel_set_widget` timed out at 6s, followed by `panel_get_errors` at 20s and `panel_save_workflow` at 15s, all without a frontend reply. The reporter's comfyui-mcp is behind the provided latest version (0.48.32); the panel version is current (0.11.32).

PR #1176, opened 2026-08-13
fix(1095): graph mutations must not lose the tab route after creating a workflow

### LINKED ISSUE #1095: Panel graph mutations can lose connection mid-batch after creating a new workflow
After panel_new_workflow created a fresh tab, panel_set_widget reported OUTCOME UNKNOWN and subsequent panel_set_widget/panel_connect calls reported no connected panel tab. The new nodes had already been added.

### Steps to reproduce
1. Create a new workflow with panel_new_workflow.
2. Add several nodes.
3. Apply sequential widget/link mutations.

### Exact error (scrubbed)
`panel tab tmp:2806 disconnected mid-command (graph_set_widget) — OUTCOME UNKNOWN`; then `no connected tab ... Connected: none`.

### Fix
Applied locally: no. Upstream-only: yes. The panel should preserve/recover the connection or make the post-new-workflow routing state reliable before graph mutations.

### Environment
Linux; ComfyUI 0.30.1; NVIDIA RTX 3090 24GB; comfyui-mcp 0.50.25; panel 0.11.38.

artokun:
Closes #1095.

Creating a workflow and then running graph mutations loses the panel connection mid-batch. The reporter's log names the mechanism precisely:

```
panel tab tmp:2806 disconnected mid-command (graph_set_widget) — OUTCOME UNKNOWN
then: no connected tab ... Connected: none
```

## The diagnosis, which still holds

`rehelloForWorkflow()` re-advertises the tab when the active workflow changes, and its own comment says why that is destructive:

> Re-target the socket to the current workflow's tab id, then resume that workflow's agent session. **The backend drops the socket's prior tab mapping** so a background workflow's output can't leak here.

Dropping the prior mapping is deliberate and correct. The defect is *when* it fires. The workflow change is detected by a **600 ms poll** (`_wfPoll` → `onWorkflowMaybeChanged`) that knows nothing about whether a command is executing. `panel_new_workflow` changes the tab's identity to a fresh `tmp:<uuid>`, the next tick re-hellos, and the orchestrator drops the mapping the in-flight `graph_set_widget` is routed to.

It matches every detail of the report: a **temp** route id (a brand-new unsaved workflow), the drop landing **mid-command** rather than at a boundary, `OUTCOME UNKNOWN` rather than a failure (the command really did apply — "the new nodes had already been added"), and the follow-on `Connected: none`. It is a race, not a wedge, which is why the reporter saw it "mid-batch": a batch of sequential mutations is exactly the window where a poll tick lands between two commands.

The existing #508 journal-and-replay machinery does not cover it: that triggers on a **socket** close or supersession. Here the socket is fine — it is the orchestrator's *routing entry* that was withdrawn, by this tab's own hello.

## What changed since the last review

The previous cut put the guard at the workflow poll. Review rejected it — **the lead finding was that it could make the reported scenario more likely, not less** — and this rewrite answers all three findings by moving the wait into the mechanism every re-advertise already passes through.

**1. The guard covered one re-advertise out of four.** `rehelloForWorkflow` is not the only thing that re-advertises: the #607/#570 workflow-instance fence fires a full `sendHello()` **from inside the command branch**, and the #310 `free_vram` re-advertise and the #508 self-heal re-registration call it too. Worse, the interaction ran backwards — deferring the poll's hello keeps the orchestrator's cached `workflow_uuid` stale for *longer*, which trips the fence more often, which fires the one re-hello that was still ungated, mid-batch.

`sendHello` is now the **gated entry point** and `advertiseHello` the raw send. Nothing else may reach `advertiseHello`, and that is asserted structurally off the syntax tree rather than by convention, because a guard at one caller cannot be completed by patching callers — the next caller added would be ungated again.

The **one** deliberate bypass is the FIRST hello on a socket. That creates the mapping rather than replacing one, so it can strand nothing, and deferring it would leave the tab unregistered — strictly worse than the race. It is keyed on `advertisedSock`, recorded only on the send-LANDED path, so a hello that never left the tab cannot arm the gate and suppress the retry that would actually register it.

**2. The budget was shorter than the commands that most need it.** Five poll ticks is ~3 s, and `ask_user`/`request_secret` hold the window for a *human* wait — so for exactly the commands most likely to still be in flight across a workflow change, the fix stalled three seconds and then withdrew the route anyway. The bound is now contributed by the commands themselves, in two classes because there are two clocks: **4 s machine-paced, 30 s human-paced**. Measured on a **monotonic** clock (`monotonicNow`), never `Date.now`.

**3. `applyChatScope`'s inline re-bind was swallowed.** That caller nulls `currentWorkflowId`, calls the poll, and then repaints the context ring and announces the scope change — so a deferred early return announced a re-bind that had not happened. `onWorkflowMaybeChanged` now gates nothing: every commit (`currentWorkflowId`, the storage key, the workflow ref, `ssSet(SESSION_KEY, …)`) happens inline and unconditionally, exactly as before #1095, and only the wire send waits.

Also fixed on the branch during the earlier review and retained here: the in-flight mark leaked on five dedupe/retry paths that return without reaching `deliverReply`, and the pairing test could not fail on the bug it was written for.

## Two generation-scoping defects found in review, since fixed

**P1 — a release could free another connection's command.** The gate counted marks, and `cancel()` (setUrl / stop / destroy) zeroed that shared counter. But a command already executing on the **retired** socket still finishes and still calls `deliverReply`, whose release was unconditional — so it decremented whatever the **replacement** socket had marked. A parked hello could then flush while a live command was mid-flight: the exact route-loss race this PR exists to close, recreated by its own bookkeeping.

The count is now a **Set of mark ids**. `began()` returns the id, `deliverReply` takes it as a fourth argument and releases exactly that one, and `cancel()` clears the set — which invalidates every outstanding mark, so a late release from a retired generation finds nothing and is inert. The same mechanism makes a **double release** inert instead of discounting an unrelated command, and makes an **unnamed release fail closed**: it can only delay a re-advertise (the budget still expires), never let one through mid-command.

Keying on `rid` was considered and rejected — rids are not unique here, because the #517 retry path re-delivers a previously-seen rid, so two marks would collapse into one entry and the first reply would release both.

**P2 — an ordinary close inherited the gate.** `stop()`/`setUrl()`/`destroy()` null `sock` synchronously, so their late close fails the `isActive()` guard and never reaches the close handler. An ordinary drop (network, server restart) recovers through `scheduleReconnect` instead, and was therefore the one teardown that carried the previous connection's gate state across: a parked waiter or its timer could fire `advertiseHello()` against the **replacement** socket, and the dead socket's outstanding marks would make the next legitimate re-advertise wait out a budget for commands that can never reply. The active close now cancels the gate and clears `advertisedSock` — placed **after** the `isActive()` guard, so a superseded socket's close cannot withdraw the route from the live one.

## A cross-workflow leak the deferral opened, since closed

Found in review, and more serious than the routing bug this PR started from.

`onWorkflowMaybeChanged` commits the new workflow **inline** — that is deliberate, it is what fixed the swallowed-rebind finding above — while the re-hello is parked behind an in-flight command. So for the length of the deferral the panel is on workflow B and the orchestrator still has this socket bound to A.

A `user_message` frame carries **no tab id** (only `sendFrame` stamps one), so the orchestrator can route it only by the binding the socket already has. Typed in that window, the user's text, context and images for B are delivered into **A's conversation**. The 30 s human budget that `ask_user`/`request_secret` correctly earn is exactly the window that makes it worst.

`sendFrame` was unsafe for the mirror reason: its `tab_id` comes from `bridgeRouteId()`, which is derived from the **live** active workflow, so during a deferral it stamps a route the orchestrator has not been told about. `new_session`/`resume_session` in particular hand the next turn to a different agent over this socket.

Both send paths now compare the route the last **landed** hello actually carried against the live one (`routeIsStale`, kept pure so it can be driven) and hold the frame behind the hello when they disagree. The hold **expedites** that hello rather than waiting out the budget, so the wait is the hello's own round trip.

That expedite is the deliberate trade, and it is stated rather than hidden: a user action that needs the new route outranks an in-flight command's reply. Losing that reply reports `OUTCOME UNKNOWN`, which is honest and recoverable; delivering a message to the previous workflow's agent is silent and is not. The #1095 case this PR is actually about — an agent running a batch of mutations while nobody is typing — never reaches the hold at all.

Frames are **queued, not refused**: FIFO, drained as one batch after the same advertise settles, and drained even if that advertise **fails**, because a frame held forever is a mute panel.

## A test that could not observe its own property

The AST reachability scan returned at every `TryStatement` and examined only the `catch` and `finally`. A `return` inside the marked executor's `try` — which skips the catch, never reaches `deliverReply`, and leaks the in-flight mark — was invisible, and the test would have gone on passing. That is the same defect class as the token-counting version this scan replaced.

It now traverses the `tryBlock` and excludes only **caught** throws, and is driven against five fixtures including exactly that shape. Mutation-checked both ways: reverting the traversal turns the fixture red, and injecting a `return` into the real panel's marked `try` is caught and reported by line number.

## Three narrower defects, all in the deferral's own seams

**The `#517` duplicate-replay path was the one command path with no mark.** It awaits a ledger promise and then answers with a direct `thisSock.send`, so it holds the route exactly like an executing command — and it reads as "the reply already exists, nothing to protect". That is wrong when the retry lands on a **replacement** socket while the original is still running: `cancel()` invalidated the retired socket's mark, so the client believed nothing was in flight, and a workflow switch during the await let a re-hello withdraw the replacement's route before the reply was written. The OUTCOME UNKNOWN this PR exists to prevent, reached through the one path that had no mark.

Now marked, and released in a **`finally`** — the branch has three exits and answers on two of them without ever reaching `deliverReply`, so a release before any single `return` is one the next exit added would miss. The mark is taken outside that `try`, or a throw before the assignment would release `undefined`.

**The canvas-tool disclosure was decided for the wrong agent generation.** A held frame is sent after a hello the hold expedited, and a landed hello **advances** `agentSessionEpoch` — so `planCanvasToolDisclosure` running at call time decided for a generation the message would never reach. If the outgoing generation had already proven its tools, `disclose` was false, the new generation received no disclosure, and `canvasToolMessagedEpoch` then marked it as prompted. That generation never got the paragraph at all: silent, and total for the session. It is now planned **inside** the deferred send, along with everything derived from it. The one-shot context deliberately stays outside — `pendingContext` is consumed when the user presses send, and re-reading it at drain time would lose it or attach it to a later message.

**The gate started a second registration for one switch.** A hello is asynchronous — it awaits the tab-identity lease before it can even build its payload — so "a hello is happening" has real duration. On the most common path in the whole feature (an ordinary workflow switch, existing session, nothing in flight) `rehelloForWorkflow` starts the hello and immediately sends `resume_session`; that frame found the route still stale, asked to be held, and the hold's flush started a whole second hello — duplicate registration, duplicate "agent ready" greeting, and **two `agentSessionEpoch` increments per switch**. The doubled epoch is not cosmetic: it feeds the disclosure above, whose entire job is to run once per generation.

`flush()` now **joins** the advertisement already on its way. Joining is correct rather than merely cheaper: `makePayload` reads the identity *after* the lease resolves, so an advertisement started a microtask ago has not yet decided what it carries. The join clears once it **settles**, so a genuine later switch still gets its own hello, and `cancel()` drops it so a frame held on a replacement socket cannot conclude the new route is published because the retired socket's hello landed.

## The expedite is gone — a simpler rule that refuses more

The hold used to **expedite**: it pushed the parked hello onto the wire immediately. That was argued as a trade ("a person is waiting, and their action outranks an in-flight command's reply") and the argument was sound for the case it was written about and wrong for the traffic it covered. `rehelloForWorkflow` **also runs automatically** — when workflow-scoped history supplies a session id it re-hellos and immediately sends `resume_session`. Nobody is waiting for that frame. It reached the hold, expedited the hello the gate had correctly parked, and let the backend drop the old mapping before an in-flight command's reply: the OUTCOME UNKNOWN race this PR exists to close, re-opened by a convenience added to protect against a different failure.

The fix is deliberately **not** a user-vs-automatic flag. Nothing in the gate can know which is which, every call site that gets it wrong is another P1, and the flag would have to be threaded through `sendUserMessage` — which the restart-resume nudges also call with no person involved. So the expedite is removed and the rule is uniform: **a frame that needs the new route waits for the hello the deferral will produce, and cannot force one.**

The cost is bounded and visible: a frame can wait up to the deferral budget (4 s, or 30 s behind a human-paced command) before the hello goes out and it follows, sitting in the pending tray meanwhile with its own delivery timeout still covering it. What it buys is that no outbound frame can withdraw a route from a command that has not replied — by construction rather than by case analysis. This removes a special case rather than adding one.

**Held frames are also retired with their connection.** `cancel()` cleared `waiters` and `inFlight` but left `held` and its already-scheduled `then` intact, so a retired advertisement would settle, the drain would run against the mutable current socket, and a frame composed for the old connection would be written to the **replacement** one — or dropped after `true` had already been returned. A never-settling stale promise also wedged every later frame behind `held.length > 1`. Drains are now generation-fenced and the batch is cleared on cancel: the same generation-scoping rule as the mark ids, one layer out.

## One rule, replacing four rounds of point fixes

Rounds 2–5 each produced the same family of defect: **a frame reaching the wrong workflow's agent.** Each fix was correct and each round found another instance, which is what a missing abstraction looks like. The rule now is:

> An outbound frame is **stamped with the route it was composed for**. It leaves only when *that* route has actually been advertised. It **never causes an advertisement** — it waits for one. If the panel commits to a different workflow first, it is **discarded** rather than delivered there.

**Why "never causes an advertisement" is the load-bearing half.** `bridgeRouteId()` is derived from the live canvas and moves the instant the user clicks a workflow tab, but `SESSION_KEY` is not rebound until the 600 ms poll — and the hello reads `SESSION_KEY` for its spawn-time `resume`. `onWorkflowMaybeChanged` is the only place that commits a switch as a whole; its own comment says *"Order is the whole fix"*. So anything that advertises inside that gap pairs the **new route** with the **previous workflow's session**, which on a route with no live agent spawns a resume-**fork** of the previous conversation.

Earlier cuts had the hold call `request()`, which advertises immediately when nothing is in flight — so an outbound frame arriving in the gap was *itself* what published the half-committed switch. Removing the trigger removes the class: the poll stays the only author of a workflow re-advertisement and everything else observes. Same move as removing the expedite, applied to the remaining trigger.

**The route stamp closes the other half.** A→B queues B's `user_message`/`resume_session`; the user switches B→C before it drains; the hello advertises C — and the untagged queue delivered B's frames onto C.

**Cost.** A frame composed in the gap waits for the poll (~600 ms) plus any deferral, bounded at `REHELLO_DEFER_HUMAN_MS`. Past that, or if the panel commits elsewhere, it is discarded and the caller's delivery timeout surfaces it in the pending tray. Refusing a write is acceptable; sending it to the wrong workflow's agent is not.

**What this does *not* unify**, stated plainly: the round-3 findings (the unmarked replay wait, the disclosure epoch) were about *command lifetime* and *agent generation*, not frame routing — genuinely different seams that keep their own fixes. And `callTool`/`uploadMedia` still read the route live: during the gap they stamp a route the orchestrator does not hold yet, so those calls **fail** rather than misdeliver. Same family, opposite failure direction, left as-is and written down rather than quietly widened.

**A guard removed rather than kept.** The held queue briefly carried a generation stamp, from when the drain was scheduled on the advertisement's promise and could outlive a `cancel()`. The drain is synchronous now, so `cancel()` clearing the queue is total and the stamp became unreachable — mutation testing proved it by deleting the check and failing no test. It is gone, with a note that it must return if the drain ever becomes asynchronous again. An unreachable guard implying a protection nobody exercises is the same defect as a test asserting a property it cannot observe.

## A queued frame is not a delivered one

The unified rule changed frames from "sent immediately" to "sent later, or discarded" — and the **return contract did not follow**. `holdForRoute` answered `true`, which propagates through `sendFrame` as a successful write, and `runCompletion` acts on that **irreversibly**: `framePushed` → `markDelivered` retires the prompt from pending so it is never recovered from `/history`. A frame reported as delivered and then discarded is neither refused nor delivered but **invisible**, which is worse than either.

`holdForRoute` now returns `false` always — queued is not sent, and that is simply true. Two alternatives were considered and rejected:

- **A tri-state** (`sent`/`queued`/`refused`): a truthy `"refused"` would make every existing `if (!ok)` silently never fire. The blast radius is every call site, and the failure mode is invisible.
- **`false` while still queueing**: the tray shows failed, the user presses *Send now*, and the queued copy goes out too — a duplicate message.

So a frame may only be queued where **nobody reads the answer**. That is exactly `resume_session` and `new_session`: fire-and-forget, and meaningless before the hello anyway since they hand the next turn to an agent on a route the orchestrator has not been told about. Everything else is **refused** while the route is stale, which every caller already handles correctly — `runCompletion` re-pends and recovers from `/history`, which is what that path was built for ("sendFrame returns false when the bridge socket is down"). A user message is refused rather than queued for the same reason, so one retry sends exactly one message.

**The hold expiry now tracks the deferral** instead of being a fixed window from the first enqueue. A command starting *after* a frame queued pushes the advertisement's deadline out, and the old timer would clear the queue while the hello it was waiting for was still legitimately pending — a self-inflicted loss rather than a timeout.

## Scope of the guarantee — read this before trusting the rest

**This PR guarantees two things:**

1. A **`user_message` never leaves on a stale route.** That is the #1095 harm — the frame carries no tab id, so the orchestrator routes it by the socket binding, and before the new route is advertised it reaches the *previous* workflow's agent, carrying the user's text, context and images. It is refused instead, and the pending tray shows it so one retry sends exactly one message.
2. **`resume_session` / `new_session` arrive after the hello** that establishes their route, rather than naming a route the orchestrator has not been told about.

**It does NOT guarantee that no frame ever leaves on a stale route.** These are **not** protected and keep their pre-#1095 behaviour:

| Frame | Callers |
| --- | --- |
| `cancel_message` | message cancellation |
| `interrupt` | Stop, and Send-now's requeue |
| `rewind` | conversation rewind |
| `set_options` | model / effort changes (5 sites) |
| `agent_event` | run-completion reporting |
| pairing frames | remote-control pairing |
| `callTool`, `uploadMedia`, `title` | direct tool calls, media upload, title updates |

A brief cut *did* refuse all of these, and it was more correct in the abstract and **worse in practice**: every one of those callers ignores `sendFrame`'s return, so refusing did not stop them — it made them fail silently while their callers announced success, and `rewind` resubmits immediately after. Refusing a frame nobody checks converts a routing problem into a lying UI, which is worse than the routing problem. They now behave exactly as they did before this PR, so the regression that cut introduced is gone and no new guarantee is claimed.

Making all ~50 `sendFrame` call sites handle refusal is worthwhile, but it is a **different change** — error-handling UX across many features rather than a routing fix — and attempting it inside this PR is what kept producing the next round's defect. It should be its own issue.

## The armed one-shot survives a refusal

`pendingContext` is a one-shot transcript replay or workflow instruction, merged into the outgoing frame and cleared. The refusal used to sit *below* that merge, so a refused send consumed the armed replay — and `trackSend` stores the payload it was **given**, so the retry went out without it, silently. Same class as reporting a discarded frame delivered: the user's intent evaporates with no signal.

The refusal now sits **above every one-shot consumption**, so a refusal consumes nothing *structurally* — rather than by a restore that every future edit would have to remember. The ordering is asserted in both directions.

## Where the code lives

- `web/js/lib/rehello-gate.js` — the gate. Injectable clock and timers, so it is driven rather than asserted about.
- `web/js/comfyui-mcp-panel.js` — `sendHello` (gated) in front of `advertiseHello` (raw); the mark/release pair around the command branch; `setUrl`/`stop`/`destroy` drop a parked hello.

## How it was verified

`npm run test:unit` — **4452 tests: 4451 pass, 0 fail, 1 todo.** Fully green (after merging current `main`). 36 of them are new, in `rehello-gate.test.mjs`.

One caveat worth recording: `manager-install.test.mjs` → `#671 verifyInstalled … when the queue never drains` failed on two earlier runs of this suite and passed on the final ones. It is **pre-existing and not this branch's** — it asserts `elapsed < 10000` on a **wall clock** while the whole suite runs in parallel, and its own comment already concedes the hazard ("loaded machines overrun timers — the suite has seen a 5s budget take 10.4s"). Measured rather than assumed: unmodified `origin/main` (d7c7497c) fails the same test the same way under full-suite load (4399 tests, 4397 pass, 1 fail — `#671` at 34.8 s). It passes in isolation on this branch (122/122). Nothing here touches the Manager install path. It is a real flake in the suite and deserves its own issue.

The gate's behaviour is **driven**, not pattern-matched: the invariant is temporal ("no hello reaches the wire between a command starting and its reply") and `if (false) advertise()` matches any source pattern you could write for it. What remains structural is the mark/release pairing — an **AST reachability scan**, because review previously proved by execution that counting `++`/`--` tokens cannot tell the leaking arrangement from the fixed one — plus the new closure check on `advertiseHello`.

**Mutation-tested: 31 mutations, 31 killed.** Every one is re-run against the current tree by a harness (restore from an in-memory snapshot, never `git checkout`; each anchor asserted to occur exactly once and normalised to the file's own line ending; a harness timeout scores INCONCLUSIVE rather than as a kill):

| Mutation | Result |
| --- | --- |
| `request()`: `live.size === 0` → `>= 0` (never defer — the original bug) | **killed** |
| `ended()`: drop the drain-triggered flush | **killed** |
| `REHELLO_DEFER_HUMAN_MS` 30000 → 4000 (collapse the two clocks) | **killed** |
| `sendHello`: always `return advertiseHello()` (remove the gate) | **killed** |
| Take the mark above the #517 dedupe region (reinstate the leak) | **killed** — named all five leaking exits by line |
| Remove `rehelloGate.cancel()` from `destroy()` | **killed** |
| `ended()` releases the first mark it finds, ignoring the one it was given | **killed** |
| `cancel()` leaves outstanding marks in place | **killed** |
| Remove `rehelloGate.cancel()` from the close handler | **killed** |
| Move that cancel ABOVE the `isActive()` guard | **killed** |
| **`user_message` no longer waits for the route** (the leak) | **killed** |
| **The hold sends immediately instead of after the advertise** | **killed** |
| **The reachability scan skips `try` bodies** (its own blind spot) | **killed** |
| **A `return` inside the real panel's marked `try`** | **killed** — reported by line number |
| The duplicate-replay wait takes no mark | **killed** |
| The replay mark released outside the `finally` | **killed** |
| The disclosure planned at call time instead of at send time | **killed** |
| `flush()` always starts a new advertisement (no join) | **killed** |
| The join is never cleared, so a later switch is never announced | **killed** |
| `cancel()` leaves the retired connection's advertisement joinable | **killed** |
| **A held frame forces the parked hello out** (the expedite, restored) | **killed** |
| **The hold advertises instead of waiting** (the round-5 P1) | **killed** |
| **A frame is released regardless of the route it was composed for** | **killed** |
| **The hold is unbounded** | **killed** |
| **Held frames released before the hello lands** | **killed** |
| **sendUserMessage holds without stamping a route** | **killed** |
| **A cancelled generation's drain still runs** | **killed** |
| **`cancel()` keeps the held batch** | **killed** |

## What a reviewer should look at

- **The fence is now gated too, and that is a trade.** During a deferral the #607 recovery hello also waits, so a mutation aimed at the new canvas is cleanly **refused** ("Nothing was applied", safe to retry) rather than losing its route. That is better than `OUTCOME UNKNOWN`, but it lasts up to the budget — and **30 s is the number to argue with**, since it is the one place a person's pace and a wedge risk trade against each other directly.
- **Long-running commands are deliberately not covered.** A `graph_run` that renders for minutes gets the machine budget and then loses its route exactly as it does today. No regression, no cover. Extending the deadline to reach it would hold a stale route for the length of an arbitrary render.
- **Frame ordering across the deferral.** `rehelloForWorkflow` sends `resume_session` immediately after requesting the hello. `sendHello` was *already* asynchronous (it awaits the tab-identity resolve), so `resume_session` already reached the wire ahead of the hello today; the deferral widens an existing gap rather than creating one. Unchanged here on purpose, but worth a second opinion.
- **Not reproduced live.** The mechanism is read off the panel's own comment and the reporter's log, not observed against a running orchestrator — the orchestrator side of "the backend drops the socket's prior tab mapping" is not something the panel can see. The report is from panel 0.11.38; this branch is on current main.
- **An outbound frame can now wait up to 30 s** (behind a human-paced command) before it reaches the wire. That is the price of removing the expedite, and it is the one deliberate UX regression in this PR. The frame is queued and visible in the pending tray, never dropped or silently sent on a stale route.
- **`routeIsStale` fails open on an unreadable route id.** Following the #607 fence's rule ("an identity we cannot read is not one we know to be new"), because the alternative holds every frame for as long as the id stays unreadable. It does mean a leak could slip through if `bridgeRouteId()` ever returned null mid-deferral.
- **Marks are never garbage-collected by anything but a reply or a teardown.** That is deliberate (the budget bounds the wait either way), but it means a command path that somehow reached neither `deliverReply` nor a socket teardown would hold an id forever. The AST reachability scan exists precisely to make that unreachable, and it is the assertion to keep honest if the command branch is restructured.
- **Five test assertions across four files now match `deliverReply`'s leading arguments rather than its exact arity.** That was necessary — the mark is a fourth argument — but it does loosen them slightly. Each still asserts the ordering claim it was written for.

## Deliberately not done

No `CHANGELOG.md` entry and no version bump — releases are a separate PR in this repo, and the previous cut's changelog text was mangled anyway ("the tab 's", "workflow 's"). The merge with `main` resolved `CHANGELOG.md` to main's version.

Left as a **draft**.

### DIFF
--- a/browser_tests/unit/binding-recovery.test.mjs
+++ b/browser_tests/unit/binding-recovery.test.mjs
@@ -645,17 +645,24 @@ test("#607 an unreadable live identity re-hellos nothing (unknown is not new evi
   assert.equal(hellos, 0, "an unknown identity must not be treated as a changed one");
 });
 
-test("#607 sendHello reports whether the hello reached the wire, and publishes only then", () => {
-  // The hook's budget is only meaningful if `sendHello()` answers truthfully: a
+test("#607 the hello reports whether it reached the wire, and publishes only then", () => {
+  // The hook's budget is only meaningful if the hello answers truthfully: a
   // closed socket must resolve false rather than an ignored undefined, and the
   // advertised identity must be published on the success path, never at payload
   // time — publishing early would credit a hello that did not happen.
-  const src = balancedFrom(SRC, "function sendHello()");
+  //
+  // #1095 — this reads `advertiseHello`, the payload builder, which is what these
+  // assertions have always been about. `sendHello` is now the GATED entry point in front
+  // of it (it holds a re-advertise until in-flight commands have replied), so pointing
+  // this at `sendHello` would silently slice a five-line wrapper and pass on the absence
+  // of everything it checks — the same "a passing assertion stops meaning what it checks"
+  // trap this suite has corrected three times for character-window slices.
+  const src = balancedFrom(SRC, "function advertiseHello()");
   assert.match(src, /readyState !== WebSocket\.OPEN\) return Promise\.resolve\(false\)/);
   assert.match(src, /return sendBridgeHello\(/);
   assert.match(src, /return sent === true;/);
   const publishAt = src.indexOf("lastAdvertisedWorkflowUuid = advertisedWorkflowUuid;");
-  assert.notEqual(publishAt, -1, "sendHello must publish what it advertised");
+  assert.notEqual(publishAt, -1, "the hello must publish what it advertised");
   const successAt = src.indexOf("if (sent) {");
   assert.ok(successAt !== -1 && publishAt > successAt, "…only inside the success branch");
   const payloadAt = src.indexOf("(advertisedWorkflowUuid = workflowStableUuid())");
--- a/browser_tests/unit/canvas-tool-disclosure.test.mjs
+++ b/browser_tests/unit/canvas-tool-disclosure.test.mjs
@@ -480,11 +480,16 @@ test("the README does not promise parity it cannot guarantee — anywhere in the
   assert.match(readme, /identical \*\*as served\*\*/);
 });
 
-/** The whole body of the bridge client's sendHello(). */
+/** The whole body of the bridge client's hello PAYLOAD builder.
+ *
+ *  #1095 — `advertiseHello`, not `sendHello`. `sendHello` is now the gated entry point in
+ *  front of it (a re-advertise waits for in-flight commands to reply), so slicing from
+ *  `sendHello` would return a five-line wrapper containing none of what this file asserts —
+ *  and `assert.equal(count, 1)` style checks would keep passing on an empty window. */
 function sendHelloBody(source) {
-  const start = source.indexOf("  function sendHello() {");
+  const start = source.indexOf("  function advertiseHello() {");
   const end = source.indexOf("  // When the workflow title changes", start);
-  assert.ok(start >= 0 && end > start, "could not locate sendHello");
+  assert.ok(start >= 0 && end > start, "could not locate advertiseHello");
   return source.slice(start, end);
 }
 
--- a/browser_tests/unit/change-tracker-snapshot.test.mjs
+++ b/browser_tests/unit/change-tracker-snapshot.test.mjs
@@ -46,9 +46,14 @@ test("#581 wires the deferred snapshot after delivering a successful command rep
   const here = dirname(fileURLToPath(import.meta.url));
   const source = readFileSync(join(here, "../../web/js/comfyui-mcp-panel.js"), "utf8").replace(/\r\n/g, "\n");
   const capture = source.indexOf("changeTrackerToSnapshot =");
-  const deliver = source.indexOf("if (deliverReply(reply, msg.cmd, superseded))", capture);
-  const defer = source.indexOf("deferChangeTrackerSnapshot(changeTrackerToSnapshot)", deliver);
+  // #1095 — matched on the leading arguments, not the exact arity. The claim here is about
+  // ORDER (capture → deliver → defer the snapshot); pinning the full call made it fail when
+  // the in-flight mark became a fourth argument, which is a passing assertion breaking for a
+  // reason unrelated to what it checks.
+  const deliver = source.slice(capture).search(/if \(deliverReply\(reply, msg\.cmd, superseded[,)]/);
+  const deliverAt = deliver === -1 ? -1 : capture + deliver;
+  const defer = source.indexOf("deferChangeTrackerSnapshot(changeTrackerToSnapshot)", deliverAt);
   assert.ok(capture >= 0, "successful executor path captures its tracker");
-  assert.ok(deliver > capture, "reply delivery follows the successful executor");
-  assert.ok(defer > deliver, "snapshot is scheduled only after the reply is delivered");
+  assert.ok(deliverAt > capture, "reply delivery follows the successful executor");
+  assert.ok(defer > deliverAt, "snapshot is scheduled only after the reply is delivered");
 });
--- a/browser_tests/unit/command-dedupe.test.mjs
+++ b/browser_tests/unit/command-dedupe.test.mjs
@@ -525,7 +525,14 @@ test("#694 wiring: the panel handler falls back to retry_of and REWRITES the rid
   // falling through to begin + execute.
   const fpAt = src.indexOf("const fingerprint = commandFingerprint(msg);");
   assert.notEqual(fpAt, -1, "the command handler must fingerprint the frame");
-  const block = src.slice(fpAt, fpAt + 3600);
+  // Bounded by the executor's `try {`, which is the structural end of the dedupe region,
+  // not by a fixed character count. The count stood in for "the dedupe block" and stopped
+  // meaning it the moment the block grew a comment (#1095 added one) — a passing
+  // assertion failing for a reason unrelated to what it checks. Third instance of this
+  // trap in the suite, after markConnected and #508's codex R3.
+  const blockEnd = src.indexOf("\n        try {", fpAt);
+  assert.notEqual(blockEnd, -1, "could not locate the end of the dedupe block");
+  const block = src.slice(fpAt, blockEnd);
   assert.match(block, /const commandEpoch = thisSock\.__cmcpBridgeEpoch;/, "the command snapshots its socket epoch");
   assert.match(
     block,
--- a/browser_tests/unit/command-liveness.test.mjs
+++ b/browser_tests/unit/command-liveness.test.mjs
@@ -846,8 +846,14 @@ test("#1095 codex P1: BOTH outbound send paths consult the advertised route firs
     }
     ts.forEachChild(n, walk);
   })(sf);
-  assert.equal(holdSites.length, 2, "sendUserMessage and sendFrame must both be able to hold");
-  assert.equal(staleReads.length, 2, "…and each hold must be gated on the check");
+  // ONE hold site, not two. A queued frame's outcome is not yet known, so it may only be
+  // queued where nobody reads the answer — which is `sendFrame`'s session-ordered control
+  // frames alone. `sendUserMessage` REFUSES instead: queueing it and answering `true` would
+  // report it delivered while it can still be discarded (runCompletion's markDelivered
+  // retires a prompt permanently on exactly that answer), and answering `false` would leave
+  // the tray showing failed while the queued copy still goes out on the next advertisement.
+  assert.equal(holdSites.length, 1, "only the session-ordered control frames may be queued");
+  assert.equal(staleReads.length, 2, "…but BOTH send paths must consult the route first");
   assert.equal(
     (src.match(/^\s*function advertisedRouteIsStale\(\) \{/gm) || []).length,
     1,
@@ -862,11 +868,6 @@ test("#1095 codex P1: BOTH outbound send paths consult the advertised route firs
     const end = src.indexOf("\n    },", at);
     assert.notEqual(end, -1, `could not bound ${name}`);
     const body = src.slice(at, end);
-    assert.match(
-      body,
-      /if \(advertisedRouteIsStale\(\)\) return rehelloGate\.holdForRoute\(send, [^;]+\);/,
-      `${name} must hold the frame, stamped with its route, when the advertised route is stale`,
-    );
     const guardAt = body.indexOf("if (advertisedRouteIsStale())");
     const plainAt = body.lastIndexOf("return send();");
     assert.ok(guardAt !== -1 && plainAt > guardAt, `${name} must only send directly AFTER the guard`);
@@ -875,6 +876,35 @@ test("#1095 codex P1: BOTH outbound send paths consult the advertised route firs
     const buildAt = body.indexOf("const send = () => {");
     assert.ok(buildAt !== -1 && buildAt < guardAt, `${name} must build the frame before deciding`);
   }
+
+  // sendUserMessage refuses outright; only sendFrame may queue, and only for the frames
+  // whose senders ignore the answer.
+  const umAt = src.indexOf("    sendUserMessage(");
+  const umBody = src.slice(umAt, src.indexOf("\n    },", umAt));
+  assert.match(umBody, /if \(advertisedRouteIsStale\(\)\) return false;/, "a user message is refused, never queued");
+  // The CALL, not the word — the comment above it names holdForRoute to explain why this
+  // path deliberately does not use it.
+  assert.equal(
+    /rehelloGate\.holdForRoute\(/.test(umBody),
+    false,
+    "…and it must not reach the queue at all",
+  );
+
+  const sfAt = src.indexOf("    sendFrame(frame) {");
+  const sfBody = src.slice(sfAt, src.indexOf("\n    },", sfAt));
+  assert.match(
+    sfBody,
+    /if \(!SESSION_ORDERED_FRAMES\.has\(frame\?\.type\)\) return false;/,
+    "every other control frame is refused, so runCompletion re-pends instead of retiring",
+  );
+  assert.match(sfBody, /return rehelloGate\.holdForRoute\(send, routeId\);/, "…and the queued one carries its route");
+  // The set must stay exactly the fire-and-forget session frames. Adding a frame whose
+  // sender reads the return would reintroduce "reported delivered, then discarded".
+  assert.match(
+    src,
+    /const SESSION_ORDERED_FRAMES = new Set\(\["resume_session", "new_session"\]\);/,
+    "only fire-and-forget session frames may be queued",
+  );
 
   // The comparison must be against what the hello ACTUALLY carried, recorded on the landed
   // path only — the same rule as lastAdvertisedWorkflowUuid and advertisedSock beside it.
--- a/browser_tests/unit/command-liveness.test.mjs
+++ b/browser_tests/unit/command-liveness.test.mjs
@@ -637,4 +637,10 @@ test("#1095: the in-flight command count is paired, and read before a re-target"
     "the deferral must be bounded so a stuck command cannot strand the tab");
   assert.match(body, /^\s*workflowRetargetDeferrals = 0;/m,
     "…and the budget must reset once the re-target proceeds");
+  // …and the budget must be released on the no-change path too, not only when a deferred
+  // switch proceeds. A counter that only ratchets one way leaves an ABANDONED switch
+  // (deferred a tick, then the id settles back) part-spent, so the next genuine switch
+  // gets fewer deferrals than it is owed.
+  const resets = (body.match(/^\s*workflowRetargetDeferrals = 0;/gm) || []).length;
+  assert.ok(resets >= 2, "the deferral budget must reset on the no-change path as well");
 });
--- a/browser_tests/unit/command-liveness.test.mjs
+++ b/browser_tests/unit/command-liveness.test.mjs
@@ -476,7 +476,14 @@ test("#508 codex R3: a command arriving on an ALREADY-superseded socket is refus
   assert.ok(parseAt !== -1 && guardAt !== -1, "the listener must parse before the instance guard");
   assert.ok(parseAt < guardAt, "the frame must be parsed BEFORE the superseded early return");
   // …and the refusal must be rid-correlated, explicitly NOT-APPLIED, and delivered/journaled.
-  const guarded = listener.slice(guardAt, guardAt + 1600);
+  // Bounded by the guard block's OWN closing brace, not a fixed character count. The
+  // count stood in for "inside this block" and stopped meaning it the moment the block
+  // grew a comment (#1095 added one) — a passing assertion turning into a failing one for
+  // a reason unrelated to what it checks, the same trap corrected for markConnected in
+  // #1146. `\n      }` is this block's indent level, so it cannot match an inner brace.
+  const guardEnd = listener.indexOf("\n      }", guardAt);
+  assert.notEqual(guardEnd, -1, "could not locate the end of the superseded guard");
+  const guarded = listener.slice(guardAt, guardEnd);
   assert.match(guarded, /if \(isCommandFrame\) \{/, "only command frames get a refusal");
   assert.match(guarded, /rid: msg\.rid/, "the refusal must correlate to the command");
   assert.match(guarded, /ok: false/, "a stale command is refused, never reported as done");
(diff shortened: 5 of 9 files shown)


artokun:
## Diagnosis

Reproduced by reading, not guessing — the mechanism is stated in the panel's own comment.

`rehelloForWorkflow()` re-advertises the tab when the active workflow changes, and says why that is destructive:

> Re-target the socket to the current workflow's tab id, then resume that workflow's agent session. **The backend drops the socket's prior tab mapping** so a background workflow's output can't leak here.

Dropping the prior mapping is deliberate and correct — it is what stops a background workflow's output leaking into this tab. The defect is *when* it fires.

The workflow change is detected by a **600 ms poll** (`_wfPoll` → `onWorkflowMaybeChanged`), which knows nothing about whether a graph command is currently being executed. `panel_new_workflow` changes the tab's identity to a fresh `tmp:<uuid>`, the very next poll tick re-hellos, and the orchestrator drops the mapping the in-flight `graph_set_widget` is routed to. That is the report verbatim:

```
panel tab tmp:2806 disconnected mid-command (graph_set_widget) — OUTCOME UNKNOWN
then: no connected tab ... Connected: none
```

It matches every detail: a **temp** route id (a brand-new unsaved workflow), the drop landing **mid-command** rather than at a boundary, `OUTCOME UNKNOWN` rather than a failure (the command really did apply — "The new nodes had already been added"), and the follow-on `Connected: none` until the new mapping settles.

This is a race, not a wedge, which is why the reporter saw it "mid-batch": a batch of sequential mutations is exactly the window where a poll tick can land between two commands.

## Why the existing machinery does not cover it

The panel already journals and replays a reply it cannot deliver (#508), but that path triggers on a **socket** close or supersession. Here the socket is fine — it is the orchestrator's *routing entry* that was withdrawn, by this tab's own hello. Nothing on the panel side observes that as a lost reply, so nothing replays it.

## Shape of the fix

The panel is the only party that knows both facts — that a command is executing, and that the workflow changed — so the deferral belongs here rather than in the orchestrator.

Two parts:

1. **The bridge client counts commands in flight.** It already owns the window between receiving a command frame and delivering its reply; nothing else does. The panel gets a read-only accessor.
2. **`onWorkflowMaybeChanged` defers the whole change while that count is non-zero** — critically *before* committing `currentWorkflowId`. Advancing the id and deferring only the hello would swallow the change permanently, because the next tick's `wfid === currentWorkflowId` early-return would treat it as already handled. Deferring the whole thing means the next 600 ms tick retries, which bounds it naturally: commands have their own timeouts, so the wait cannot outlive them.

Not adding a new timer or a lock. The poll is already the retry mechanism.

artokun:
## Review outcome: not mergeable — the guard is at the wrong level

The diagnosis in this PR holds. The fix does not, and the lead finding is that **it can make the reported scenario more likely, not less.**

### The guard covers one re-hello path out of several (CONFIRMED)

`rehelloForWorkflow` is not the only thing that re-advertises. The #570 workflow-instance fence calls `noteWorkflowInstanceMismatch()`, which fires a full `sendHello()` — the same re-advertise, dropping the same prior mapping — **synchronously from inside the command branch**, while `commandsInFlight > 0`.

And the interaction runs the wrong way. Deferring the poll's re-hello keeps the orchestrator's cached `workflow_uuid` stale *longer*, which makes the fence *more* likely to trip, which fires the ungated re-hello mid-batch. On the issue's own sequence — `panel_new_workflow`, then an immediate batch of mutations — that reproduces `disconnected mid-command (graph_set_widget) — OUTCOME UNKNOWN` followed by `Connected: none`. The exact symptom, through a path this change made more reachable.

That is not a patchable gap. It says the deferral belongs **at the re-advertise mechanism**, where every caller passes, rather than at one caller.

### The budget is shorter than the commands that most need it (CONFIRMED)

`ask_user` and `request_secret` hold the count for a *human* wait. Five ticks is ~3s. So for exactly the commands most likely to still be in flight across a workflow change, the fix adds a three-second stall and then withdraws the route anyway — the user answers a question and the orchestrator reports OUTCOME UNKNOWN for the answer they just gave.

A bound is still right (unbounded turns a race into a wedge), but a fixed tick count cannot be the whole answer when one class of command is human-paced.

### `applyChatScope`'s inline re-bind is swallowed (CONFIRMED)

I previously judged this as "delayed, not lost", because that caller nulls `currentWorkflowId` first and the poll retries. That was too generous. The caller expects the **inline** commit — id, key, ref, `ssSet(SESSION_KEY, …)`, `rehelloForWorkflow` — and then continues to repaint the context ring and announce the scope change. With a command in flight it announces a re-bind that did not happen, and the state is internally inconsistent until a later tick repairs it.

### Fixed on the branch during review

- **The pairing test could not fail on the bug it was written for.** Review proved it by running this byte-identical file against the leaking revision: green. `incs === 2` / `decs === 1` are true of both arrangements. It now reads the reachability property off the syntax tree, and restoring the pre-fix arrangement fails with the five leaking exits named.
- **The count leaked on five dedupe/retry paths** that return without ever reaching `deliverReply` — two of them replying via a direct `thisSock.send`. A leaked count never returns to 0, so every later switch would pay the full budget.
- **The deferral budget never reset on the no-change path**, so an abandoned switch permanently shortened the protection for the next real one.
- **Two mangled newlines in the changelog** shipped as "the tab 's" and "workflow 's".

### What I think this needs

Move the guard into the re-advertise mechanism so `rehelloForWorkflow`, the #570 fence path, and the `free_vram` re-advertise are all covered by construction — instead of one caller being gated while another stays open and gets *more* traffic because of it.

The in-flight counter itself is sound and now properly tested; it is the placement of the consumer that is wrong. Leaving this draft rather than merging a change that can worsen the case it was written for.

artokun:
## Merge-readiness pass

Three things checked, two changed.

### The session-scope invariant holds — this restores the ROUTE, not the session

The standing invariant is that an agent session is **orchestrator-scoped**; `tab_id` is a routing key only. A fix for "the tab route is lost" that made the conversation follow the workflow would be a bug against the design, not a fix.

Measured rather than argued. Across the whole net diff of `web/` against `main`, the only line matching `SESSION_KEY|sessionId|session_id|getResume|resume|threadForWorkflow|loadThread|agentSessionEpoch` is a **comment** inside `rehello-gate.js`:

```
+//     ref and `ssSet(SESSION_KEY, …)` INLINE and unconditionally. The first cut returned
```

No executable session-binding line changed, and `rehelloForWorkflow` is byte-identical to `main`. The gate defers *when the hello reaches the wire*; it does not touch which session is bound, and because the payload is built at send time a deferred hello advertises the identity that is live when it goes out. The panel's pre-existing per-workflow `ssSet(SESSION_KEY, …)` is untouched here — that is a separate concern, not one this PR introduces or deepens.

### Fixed: the body would have auto-closed the issue on merge

It opened with `Closes #1095`, which is a same-repo closing reference. Replaced with an owner-qualified, non-closing `Refs artokun/comfyui-mcp-panel#1095`. Re-scanned the whole body for `close|fix|resolve` followed by an issue reference — zero remain. **The issue now needs closing by hand after merge**, which is the right default given the body's own "Not reproduced live" caveat.

### Fixed: an assertion 71 characters from breaking

`#508 wiring: an undelivered reply is journaled…` sliced the socket-open handler as `slice(0, 2000)`. Measured: `sendHello();` sat at offset **1929 of 2000**, in a handler 3308 characters long. One more comment line in that handler and it goes red for a reason unrelated to what it asserts — and this branch had already added a comment there. Now bounded by the handler's own closing brace, with both ends asserted so an unfound end cannot silently degrade the slice into a near-whole-file scan. Verified by mutation: deleting `sendHello();` from the handler turns it red, so the new bound is not vacuous.

This is the third time this file has had to correct the same trap (markConnected in #1145, the superseded guard earlier in #1095).

### State

`npm run test:unit` on the pushed tip: **4423 tests, 4422 pass, 0 fail, 1 todo** — the full chain, including the i18n, tool-vocabulary and panel-scope gates. `main` is merged in (an ancestor of HEAD), no conflict markers in `web/` or `browser_tests/`, `mergeStateStatus` CLEAN. Still a draft; not merged.

Issue #1098, opened 2026-08-12
Saving a new workflow can trigger UUID mismatch and unresponsive workflow_open


artokun:
### What happened / root cause
A newly built workflow saved successfully as video_first_frame_upscale. The following panel_run returned root-workflow-uuid-mismatch. panel_open_workflow for the saved file then timed out with no correlated reply.

### Steps to reproduce
1. Create/edit a new workflow.
2. Save As through panel_save_workflow.
3. Call panel_run.
4. Attempt panel_open_workflow recovery.

### Exact error (scrubbed)
`[root-workflow-uuid-mismatch]`; then `Panel tab did not reply to workflow_open within 15000 ms`.

### Fix
Applied locally: no. Upstream-only: yes. Save-As must update/rebind the active workflow identity atomically, and workflow_open must recover without a timeout.

### Environment
Linux; ComfyUI 0.30.1; NVIDIA RTX 3090 24GB; comfyui-mcp 0.50.25; panel 0.11.38.

artokun:
Triaged with #1095, #1097 and #1098 as **one cascade from a single session on panel 0.11.38** (current is 0.14.9), rather than four independent bugs — the bridge socket dropped mid-command, never came back for RPC, and everything after that is `workflow_open` having nobody to answer it.

Full analysis, the list of what has shipped since 0.11.38 that touches this, and the two things that would settle it are on #1096. Short version: a retest on 0.14.9 after a hard refresh of the ComfyUI tab, and — if the socket still drops — whether the browser console shows a **clean** close (1000/1005, something closed it deliberately) or an **abnormal** one (1006, the process or network went away). Those have completely different causes and the tool output cannot tell them apart.

Leaving this open.

artokun:
Triaging #1095–#1098 together, because they are one incident rather than four defects: a panel socket that dropped mid-build, and everything downstream of it. They share a session and a version — comfyui-mcp **0.50.25**, which is roughly ninety patch releases behind current (0.51.11).

[...]

PR #1177, opened 2026-08-13
feat(757): panel_set_widget can create a dynamic widget slot (Power Lora Loader)

### LINKED ISSUE #757: panel_set_widget cannot create dynamic widget slots (Power Lora Loader rgthree lora_N)

artokun:
Closes #757.

`panel_set_widget` cannot populate LoRA slots on **Power Lora Loader (rgthree)**, one of the most widely used LoRA nodes. A freshly added node exposes only `divider, PowerLoraLoaderHeaderWidget, divider, ➕ Add Lora` — the `lora_1`, `lora_2`… entries do not exist until the user clicks a DOM-only control the agent cannot reach.

## Why this one, now

Deliberately picked because it is **not** the workflow-identity cluster. #1095, #1096, #1098 and #1014 all turn on re-advertising the tab, and review has now shown twice that fixing those one caller at a time makes things worse — they need one design (see the writeup on #1098). This is self-contained graph work with no lifecycle coupling.

## Before changing anything

**Whether the slot can be created without simulating a click.** The ➕ control is DOM-only, but the node almost certainly creates the slot through a method on the node object. If so the executor can call it directly, which is both reachable and far more robust than synthesising pointer events.

**Whether it generalises.** rgthree is the reported node, but "a node whose widgets appear only after a user gesture" is a category. A fix keyed to one node title would be a fix that stops working the moment the node is renamed or another pack does the same thing, so the first thing to establish is what the general shape is and where it belongs.

**What happens when it cannot.** If a node exposes no programmatic way to add the slot, the refusal must say so precisely rather than listing the widgets that happen to exist — the current message reads like the widget name was wrong when the real answer is that the slot has to be created first.

Draft while I read how rgthree adds a slot and where the executor's widget lookup happens.


artokun:
Closing unmerged: the remaining half of #757 needs a tool-surface change in `comfyui-mcp`, not panel code.

The disclosure half already shipped (`pressable-widget.js`). Pressing the control requires either a new tool name or a new argument on the existing one, and the panel depends on a vocabulary it does not own — `check-tool-vocabulary.mjs` says so in its header. The one panel-only alternative, auto-pressing when a widget is missing, would mutate the graph on a guess with no undo, and would fire on ordinary typos.

Full reasoning on the issue. The primitives it needs (`isPressableWidget`, `pressableWidgets`) are already exported and correct.

Issue #1161, opened 2026-08-12
graph_set_widget hangs for 30s on EVERY node after a ComfyUI restart, while all other panel commands respond instantly


artokun:
### What happened

After ComfyUI was restarted mid-session (and the tab reconnected), **`graph_set_widget` became the only panel command that does not respond.** Every call times out at 30 s regardless of node or widget; every other command on the same tab, in the same seconds, replies instantly.

Observed on one workflow, interleaved:

| command | node | result |
|---|---|---|
| `graph_set_widget` frame_load_cap=120 | 29 `VHS_LoadVideo` | **30 s timeout** |
| `graph_set_widget` ref_image_size="match" | 28 `MiniMaxH3ReferenceToVideo` | **30 s timeout** |
| `graph_set_widget` steps=9 | 20 `BasicScheduler` | **30 s timeout** |
| `graph_query` / `graph_outline` | — | instant, correct |
| `graph_edit_node` (title) | 29 | **instant, succeeded** |
| `list_workflows` | — | instant, `active_confirmed: true` |
| `run` (queue prompt) | — | queued fine |

So the bridge, the tab binding and the mutation path in general are all healthy — `graph_edit_node` is a mutation on the same node in the same tab and it works. It is specific to the `graph_set_widget` handler.

The failure is also **not a lost delivery**: re-reading the node after each timeout shows the value unchanged (`frame_load_cap` stayed 240 across four attempts), so the handler is not applying-then-failing-to-ack; it appears to never complete.

Error text returned each time:

```
Panel tab wf:<uuid>:workflows/<name>.json did not reply to "graph_set_widget" within
30000 ms — the ComfyUI tab may be backgrounded or frozen. This command MUTATES and was
already delivered to the tab, so it may have been applied despite the missing reply —
check the current state before re-issuing it; a blind retry can apply it twice
```

That message is misleading in this state: it blames a backgrounded/frozen tab, but the tab demonstrably answers other commands in the same second, and the mutation was NOT applied. Worth distinguishing "tab silent" from "this one handler never returned".

### Steps to reproduce

1. Session working normally — `graph_set_widget` succeeded many times on this same workflow earlier.
2. ComfyUI is restarted (here: via `panel_restart_comfyui`; the server went down and was later brought back from Pinokio).
3. Tab reconnects; reads and `graph_edit_node` work.
4. Any `graph_set_widget` → 30 s timeout, value unchanged.

### Impact / workaround

[...]

Issue #1163, opened 2026-08-12
The mid-task nudge can fire on a turn the user just started, and never fires when the panel itself respawns the orchestrator


artokun:
Follow-up to #1145 (merged in #1146). A high-effort review of that change confirmed two defects it left behind — one it introduced, one it inherited.

## 1. A turn started during an open outage is charged with the outage that preceded it

`noteTurnStarted()` zeroes the measured outage but leaves `startedAt` set, so the outage keeps running across the turn boundary and is later measured from before the turn began. The turn-scoping the change advertises does not hold in that direction.

Reachable because `sendUserMessage` only requires an OPEN socket, not a completed handshake, and the socket reopens well before the `models` frame — up to 45s earlier on a cold backend.

1. The bridge drops at t=0.
2. The socket reopens at t=5s; `models` has not arrived, so the outage is still open.
3. The user types a new message. The orchestrator answers `turn:working` → `MID_TASK_KEY` is armed and the measured outage is zeroed, but `startedAt` still points at t=0.
4. The `models` frame lands at t=35s and measures 35s — from before the turn existed.
5. The `ready` ack sees a mid-task marker and a 35s outage, and injects "Your connection dropped mid-task … continue exactly what you were doing before the drop" on top of the message the user sent seconds ago.

The agent restarts or duplicates the work it was just asked to do. That is the false-nudge-into-a-live-session harm #1138 exists to prevent, in the branch whose comment claims it is now covered.

**Fix:** `turn:working` is a frame FROM the orchestrator, so it is positive evidence the bridge is carrying traffic and the orchestrator is alive — strictly better evidence than any clock. A turn start should CLOSE the outage, not merely zero the total.

## 2. The panel's own respawn/reclaim/redial never opens an outage at all

`noteBridgeClosed()` sits behind the superseded-socket guard `if (!isActive()) return;`, and `stop()` / `setUrl()` / `destroy()` all null `sock` synchronously before the browser delivers `close`. So no panel-initiated teardown ever opens an outage.

For Disconnect, provider switch and soft reload that is harmless — each clears `MID_TASK_KEY` on its own path, so no nudge is pending. **The automatic recovery paths do not**: `tryAutoReclaim`, `tryAutoRespawn` and `tryHandshakeRedial` all restart the orchestrator through `setUrl`/`stop` while a turn is still armed.

1. The agent is mid-task; the orchestrator wedges (socket open, agent dead, no `models`).
[...]

PR #1178, opened 2026-08-13
test(1178): pin why the /object_info bound cannot live in the burst cache

artokun:
> **This is not a fix for anything.** The issue it was opened against is closed and shipped.
> What is left here is three regression guards plus the branch history that records three
> approaches ruled out on the way to the real fix. Draft on purpose; nothing here needs to
> merge for the product to be correct.

`Refs artokun/comfyui-mcp-panel#1161` — no longer open as of 2026-08-13, via #1179
(`1759e2a9`), released in 0.14.25. The oracle-layer bound is in `main` and is the shipping
behaviour.

## What is actually in this branch now

```
 browser_tests/unit/object-info-cache.test.mjs | 200 ++++++++++++++++++++++++++
 1 file changed, 200 insertions(+)
```

Test-only against `origin/main`. Every production hunk the earlier revisions carried — the
`OBJECT_INFO_READ_WAIT_MS` bound in `objectInfoCache.read()`, and a `CHANGELOG.md` entry for a
version that never shipped it — has been dropped. `web/js/lib/object-info-cache.js` is
byte-identical to `main`, and still carries no `#1161` reference, because the whole fix lives
one layer down in `object-info-oracle.js`.

`npm run test:unit`: **4411 pass, 0 fail, 1 todo** (4412 tests), merged up to `main` at
`afb8db64`.

## What the three guards pin

They are written against the cache's *behaviour under a call*, not against its source text, so
a rename cannot slip past them.

1. **`#1178/#1179: a hung client route still reaches the fallback THROUGH the burst cache.`**
   Builds the production loader shape exactly as `graph_set_widget` builds it — the oracle read
   through this cache — with `getNodeDefs` that never settles and a working `GET /object_info`.
   Asserts the fallback route is *asked* (`httpCalls === 1`), that its schema reaches the fence,
   and — the load-bearing part — that the release happens at the client route's share of the
   oracle budget. Elapsed time is asserted on an injected clock, so it proves *which layer's*
   bound answered rather than merely that an answer arrived.

2. **`#1178: the burst cache arms NO timer of its own.`** The decisive one. Counts real
   `setTimeout` calls across an issuer read and a joining read. Re-introducing a bound in
   `read()` — for either — arms a timer here and fails with the reason attached.

3. **`#1178/#1223: a fully silent backend rides through the cache as SILENCE, not as an
   answer.`** With neither route answering, the *oracle's* outcome must survive the cache
   intact — `defs: null`, every unanswered route named, and the transport tags still readable by
   `noBackendAnswerEstablished()`. #1223's snapshot fallback is licensed on exactly that
   distinction, so a cache that substituted a note of its own would disable that fallback for
   every caller *and* make the refusal name a cause that never happened.

### Mutation-verified, both directions

- Re-adding the third revision's bound verbatim (an 8s `Promise.race` wrapping both the issuer
  and the joiner in `read()`) → **guard 2 fails.** Guards 1 and 3 survive this one, and honestly
  so: their subject settles in simulated time, so a real 8s timer never fires inside a
  millisecond-long test.
- Stripping the #982 outcome wrapper (`return payload` in place of `return defs`) → **guards 1
  and 3 both fail.**

Each guard is therefore killed by at least one mutation; neither mutation is caught by the
`#716` tests that were already there.

## The record: three approaches that were ruled out

Preserved as commits rather than as prose, because the diffs are the evidence:

1. `90990a39` — **bound the issuer, release the `inflight` slot.** Missed every *joining*
   reader: the early return handed out the raw unbounded promise, so a burst after a restart —
   the case #716 built this cache for — was unfixed. Also broke the one-request-per-generation
   invariant and returned a bare `null`, which computes an empty failure list at both call sites.
2. `8689f2bc` — **bound both, retire by generation.** Retiring discarded the fetch's own late
   payload, so a merely *slow* backend (remote, tunnelled, large `/object_info`) became a
   permanent refusal — worse than the bug.
3. `084eab8c` — **bound both via `bounded-step.js`, do not retire.** No longer discards the late
   payload, but converts a 30s hang into a permanent 8s-per-call refusal: 29 widget writes cost
   29 × 8s and all refuse, with refusal text advising a retry guaranteed to fail.

### Why the layer is wrong

`fetchWholeObjectInfo` has **two** transports: `getNodeDefs()` (the frontend client route) and a
direct `GET /object_info`, added by #982 for exactly this failure — a client route that stops
working while the HTTP endpoint answers fine.

A bound *above* the oracle can only choose **how to fail**. It cannot advance a hung
`getNodeDefs()` to the fallback, because the oracle still awaits that first transport. And to
fire at all it must be shorter than the oracle's deadline — which is precisely what pre-empts
the fallback before that route is ever asked. Measured on this branch's code against the
production loader shape: **refused at 8003ms, where `main` answers with a usable schema at
10010ms.** The refusal also printed `Tried one route` while the second route had never been
attempted.

Bounding **inside** the oracle, per transport, both stops the hang and makes the fallback
reachable, so the widget write *succeeds* rather than being refused. That is #1179:
`OBJECT_INFO_DEADLINE_MS = 20000` spent down per step, with a reserved floor guaranteeing the
fallback is still issuable when the client route is the thing that hung. Its live verification
measured the real #1161 state end to end at 11.7s, fallback issued, 4304 types returned — which
an 8s cache-level bound would never have seen.

## Related work that did land

- #1179 — the oracle-layer bound (the actual fix; `Refs artokun/comfyui-mcp-panel#1161`)
- #1186 — the sibling `api.getNodeDefs()` call sites, `graph_add_node` / `panel_refresh_nodes` (#1180)
- #1201 — the remaining unbounded network awaits (#1188)
- #1232 — a transient `/object_info` timeout must not refuse a safe widget edit (#1223)

## Disposition

Two defensible outcomes, and the maintainer should pick:

- **Land the guards** — cherry-pick the single test file onto `main` as a `test(...)` commit. It
  is 200 lines, adds no production surface, and gives the next person who reaches for a
  cache-level bound a failing test with the reason instead of a fourth rediscovery.
- **Close the PR** and keep the narrative in the issue thread. The guards are then lost, and the
  three ruled-out approaches survive only as an unreferenced branch.

Recommendation: land the guards. The mutation results above show they catch the exact regression
the three attempts each walked into, and #1178's real cost was that nothing in the tree said the
bound could not live in the cache.


artokun:
## Diagnosis — an unbounded fetch, made permanent by request coalescing

Found by reading the two awaits on the write path. It accounts for every detail of the report, including the ones that make it look strange.

### The mechanism

`graph_set_widget` is the only executor that consults `/object_info` before writing (added by #718, so the workflow-instance stamp can be re-checked after the await). It does so through `objectInfoCache.read(...)`, and that read is **unbounded** — `web/js/lib/object-info-cache.js` has no timeout, no `AbortController`, no `Promise.race`; the TTL only governs caching, not waiting.

After a ComfyUI restart the tab can hold a half-open connection, so `api.getNodeDefs()` never settles. The first call parks forever.

**Then coalescing makes it permanent.** The cache joins concurrent misses:

```js
if (inflight && inflightGeneration === generation) return inflight;
```

The hung request stays `inflight`, so every later `graph_set_widget` — any node, any widget — attaches to that same dead promise instead of issuing its own. Nothing ever settles it, so nothing ever clears it.

### Why it matches the report exactly

| observation | explained by |
|---|---|
| only `graph_set_widget` affected | it is the only command that fetches `/object_info` on its path |
| `graph_edit_node` instant on the same node | a mutation with no metadata fetch |
| every node, every widget | the fetch is per-call, not per-node |
| always exactly 30s | the caller's timeout — nothing panel-side ever fires |
| value unchanged after four attempts | execution never reaches the write |
| persists rather than self-heals | the hung promise remains `inflight` and is joined forever |

### The repo already states the principle

The startup seed bounds its wait at 8s, and its comment says why in as many words:

> The seed's own retry loop only advances once each `api.getNodeDefs()` SETTLES, so a request that never settles (a hung/half-open connection) would otherwise block the awaiting tool FOREVER. Bounding the wait converts that hang into the correct outcome.

That is this bug, on the sibling path, already understood and already solved. The write path's fetch simply never got the same treatment — and a bound there degrades to the same correct outcome the seed already produces: refuse **this one call** with "retry in a moment", leaving the fence's fail-closed behaviour intact.

### What I still want to establish before changing anything

- **Where the bound belongs** — inside `objectInfoCache.read` (covering every caller by construction) or at this call site. The #1095 review is a fresh lesson that gating one caller while siblings stay open can be worse than gating none, so the mechanism is the strong default.
- **Whether a bounded-out request must also be evicted from `inflight`.** Bounding the *wait* alone would leave the dead promise joinable, so the next call would bound out too — a permanent 8s tax instead of a permanent hang. Better, but not fixed. The generation counter already exists for exactly this kind of retirement.
- **That a late response still helps.** The seed's design is explicit that giving up on the wait must not latch anything, so a response arriving afterwards still establishes the baseline. The same must hold here.

artokun:
## Review outcome: the diagnosis holds, this implementation does not

16 CONFIRMED findings. The root cause in the diagnosis above is unchallenged — an unbounded `/object_info` read made permanent by coalescing — but the fix breaks invariants this file is built on, and needs reworking rather than patching.

### It breaks the one-request-per-generation invariant (CONFIRMED, several finders)

The timeout arm releases the `inflight` slot but never advances `generation`. That invariant is what `issuedAt === generation` exists to enforce, and the file's own header explains why it matters: a generation counter is how an invalidation *retires* an in-flight fetch rather than merely forgetting its value.

With the slot released and the generation unchanged, an abandoned request and its replacement are both "current". Both can populate `value`, whichever lands last wins arbitrarily, and each re-stamps a fresh TTL. I intended "a late response still helps"; what I actually built lets a stale response overwrite a newer one.

### `return null` is a new return type, and it destroys the diagnostics (CONFIRMED, both call sites)

`read()` previously returned defs or threw. Both consumers do:

```js
oracleFailures = defs ? [] : (outcome?.failures ?? []);
```

so a `null` yields an **empty** failure list. The refusal the user sees then misdiagnoses a hung fetch as an unreachable or failing backend, with no routes named. The #982 `{[CACHE_OUTCOME], defs, failures}` wrapper exists precisely to carry that cause, and the timeout arm should use it.

### Retiring without cancelling accumulates parked fetches (CONFIRMED)

Coalescing only holds while the slot is occupied. Once the timeout empties it, every subsequent read issues *another* request against a backend that is not answering, so an outage accumulates parked fetches — removing the one-request-per-burst property #716 added this cache for.

### A slow-but-healthy backend becomes a permanent refusal (CONFIRMED)

The bound is 8s and the TTL is 1500ms. A backend that consistently answers in, say, 10s now fails **every** call spaced beyond the TTL, where previously it succeeded slowly. Fixing a hang by converting slow into never-works is the wrong trade, and it is the same shape as the mistakes on #1163 and #1095: the repair creating a worse steady state than the bug.

Related: at page load against a half-open backend, `graph_set_widget` now pays the seed's 8s **and** this 8s sequentially before refusing.

### My tests were wrong in two ways I should have caught

- **They use the real wall clock.** `boundedCache()` drops the injected `now` that all 13 pre-existing tests use — against the rule stated in this very file's docblock, that a cache test depending on wall-clock timing is how it becomes flaky.
- **Both mutants HANG rather than fail.** I presented them as proof. A hanging test is not a failing test; in CI it is a timeout with no signal about which invariant broke. The tests await a read that may never settle, which is a defect in the tests independent of the fix.

Also confirmed: `setTimer` is armed inside a Promise executor, so a throwing timer rejects `read()` and discards a fetch result that had already succeeded; and no test covers the `clearTimer` cleanup — deleting it leaves all five new tests green.

### What the corrected design needs

1. **Advance the generation when abandoning**, so the abandoned request is genuinely retired and cannot re-populate the cache behind its replacement.
2. **Return the outcome wrapper with a real failure note**, so the refusal says the definitions could not be read in time.
3. **Keep coalescing across the abandonment** — later readers should join the *replacement*, not each spawn their own.
4. **Reconsider the bound against the TTL**, so a slow-but-healthy backend still succeeds. The seed's 8s was chosen for a startup wait with different consequences; reusing the number because it is the same endpoint was reasoning by analogy, not by consequence.
5. **Drive the tests from an injected clock and never await a possibly-never-settling promise**, so the mutants fail loudly instead of hanging.

Leaving this draft. The bug is real and the P1 stands; this fix is not the one to ship for it.

artokun:
## Addendum — I under-reported this. The headline finding is worse than what I posted.

The full review adds a defect I missed entirely, and it is the most important one:

### The bound never applies to a coalescing reader (CONFIRMED, verified by execution)

`read()` returns the in-flight promise at line 128 — **before** the bound is installed further down. So only the reader that ISSUES the fetch is bounded; every reader that joins gets the raw, unbounded promise. The timeout path then clears the slot without settling the promise those joiners are awaiting, so they park forever.

Verified by running the shipped module: three reads against a never-settling fetch, and the joiners stay pending indefinitely with **zero timers armed for them**.

That means **the fix does not fix the reported bug** in the case it most matters. #716 added this cache for bursts of `graph_set_widget` calls; in a burst after a restart, the first call now answers "retry in a moment" at 8s and every overlapping call still parks until the orchestrator's 30s timeout — the original symptom, for any call arriving inside the 8s window.

I claimed this change fixed a P1. It does not, and I should have tested a concurrent burst rather than only the single-reader path.

### Also confirmed, and also missed by me

**A read can refuse while holding a perfectly good schema.** The timeout path returns `null` without re-checking the TTL cache. If an invalidation and a successful refetch happen while an earlier read is still waiting, that read still refuses — telling the user to reconnect ComfyUI at the exact moment the panel holds a fresh authoritative schema.

### Corrected design

Adding to the list above:

6. **Bound the joiners too** — the bound belongs where the promise is handed to a caller, not only where it is created. Every reader of a hung request must give up, not just the first.
7. **Re-check the TTL before refusing**, so a payload that arrived while this call waited is used rather than discarded.

Still draft, and now for a stronger reason than I first gave: this does not solve the P1 it claims to.

artokun:
## Round 2: 16 CONFIRMED, and the decisive one is that I should not have written this by hand

The rework fixed the headline defect from round 1 (joiners are bounded now, verified) but introduced its own, and the review found the reason underneath all of them.

### The repo already has this primitive

`web/js/lib/bounded-step.js` exists, and its header says exactly what I did wrong:

> A second timeout helper written alongside the first is how this repo keeps producing near-duplicate bugs, so there is one here and both callers import it.

I wrote a third by hand, across two attempts, and re-derived its hardening badly both times. What it guarantees is precisely what my version got wrong:

| bounded-step guarantees | my hand-rolled version |
|---|---|
| always settles, **never rejects** | a throwing `setTimer` degrades to *no bound*, reinstating the hang |
| settles **at most once** — a late fulfilment after the bound is **dropped, not raced** | a joiner still inside its bound receives the very payload the generation bump was meant to retire |
| timer cleared on settle, so a page cannot accumulate them | a timer is allocated **per `boundRead` invocation**, so N joiners arm N timers |
| non-positive `ms` disables the bound cleanly | same, but reimplemented |

That third row is the hole I created while fixing the previous hole: `generation++` protects the **cache** from a stale payload, and does nothing about the **caller** already awaiting the abandoned promise directly.

### Also confirmed

- **The blunt retire.** `generation++` on the timeout path is unconditional, so it invalidates the abandoned request permanently — which is what makes a healthy-but-slow backend never populate the cache at all.
- **A test still awaits a promise that may not settle** (line 320), so one mutant still wedges `node --test` rather than failing. I claimed the tests no longer detect bugs by hanging; I had verified that for the three mutants I ran, not in general.
- **Releasing the slot without cancelling** turns one un-abortable hung fetch per session into one per read.

### What this needs

Rewrite the bound in terms of `bounded-step.js` rather than continuing to patch a parallel implementation. Its at-most-once contract removes the joiner hole and the timer accumulation by construction, and its never-rejects contract removes the throwing-timer regression — three of the confirmed defects disappear rather than being individually guarded.

Then re-examine the two questions the helper does **not** answer, which are genuinely about this cache rather than about timeouts: whether a bounded-out read should retire the shared request at all (it drops the one-request-per-burst property #716 added this cache for), and what the bound should be given a 1500ms TTL and the caller's 30s timeout — 8s was chosen by analogy with the startup seed, which is reasoning from similarity rather than from consequence.

Still draft. The diagnosis has held through two rounds; the implementation has not, twice, and the reason is now clear enough to be worth stating rather than iterating past.

artokun:
## Round 3: 25 verified findings, and the actual answer — the bound is in the wrong layer

Three attempts, three permanent-refusal cases. The review found why they all failed, and it is not a tuning problem.

### The fix belongs in the oracle, not the cache

`fetchWholeObjectInfo` has **two transports**: `getNodeDefs()` (the frontend client route) and a direct `GET /object_info`, added by #982 **for exactly this failure mode** — a client route that stops working while the HTTP endpoint answers fine.

Bounding the *cache* means a hung `getNodeDefs()` never advances to that fallback. The oracle awaits it with no bound of its own, so it never moves on, and no joiner runs a loader of its own either. The route that would have answered is never asked — and the refusal we print says `Tried one route`, asserting a single route was tried when the second was never attempted.

**Bounding inside the oracle, per transport, both stops the hang and lets the fallback answer.** The widget write then *succeeds* instead of being refused. That is why every round of mine traded one permanent refusal for another: I kept choosing how to FAIL, when the fix is not to fail.

### What that means for this branch

Every remaining defect is downstream of that layering choice:

- **Not retiring** turns the 30s hang into a permanent 8s-per-call refusal — confirmed with an injected-timer harness: 29 widget writes cost 29 × 8s and all refuse, for the rest of the session, and the refusal text instructs a retry guaranteed to fail.
- **Retiring** (rounds 1 and 2) discarded the fetch's own late payload and made a merely-slow backend refuse forever.
- Neither is right, because with the bound at the oracle the question does not arise: the second transport answers and there is nothing to retire or refuse.
- Also confirmed: the timeout arm no longer re-checks the cache before refusing (a guard round 2 had and round 3 dropped), and `graph_add_node` has the same unbounded await, untouched by all of this.

### Fixed while the review ran

- The refusal text was 220 chars against the oracle's 200-char per-entry cap, so it ended mid-word. Now 146.
- The changelog asserted round 2's behaviour after round 3 replaced it — the **second** time this entry described a version that does not ship. Rewritten to claim only what the code does.
- A comment claimed `graph_set_widget` is the only `/object_info` consumer; `graph_remove_widget` is another.
- The test fixture hardcoded its bound instead of importing the shipped constant, so a production value of 0 (which disables the bound entirely) would have stayed green; and `gaveUp()` ignored the failures list, so four tests passed with the cause missing.

### Recommendation

**Do not merge this branch.** The P1 is real and the diagnosis has held across all three rounds, but the fix belongs in `object-info-oracle.js` — a per-transport bound so a hung first route falls through to the one that works — not in the cache. That is a different change to a different file, and it should be written fresh rather than reached by a fourth revision of this one.

I am leaving the branch as the record of what was ruled out, with the reasoning, so the next attempt starts from the answer instead of re-deriving three dead ends.

artokun:
## Round 4: main answered this, and shipping the branch as-written would now undo it

Round 3 recommended putting the bound in the oracle rather than the cache. That happened: **#1179** landed `fix(1161): bound each /object_info transport so a hung route falls through to the one that works` (`1759e2a9`), and it is **released in 0.14.25** — the changelog entry this branch also carries is already on `main`.

So the question is no longer "does this fix #1161". It is "what does this branch still do on top of a fixed main", and I measured it rather than reasoning about it.

### Measured: the cache bound pre-empts the fallback route

Same input both runs — the #1161 shape, `api.getNodeDefs()` never settling while `GET /object_info` answers — through the production loader shape both graph commands build:

| | outcome |
|---|---|
| this branch (8s cache bound) | **refused at 8003ms**, `defs: null` |
| `main` (no cache bound) | **answered at 10010ms** with a usable schema — write authorized |

This is round 3's finding, now with a number on it. A cache-level bound has to be *shorter* than the oracle's deadline to fire at all, so it can only ever fire **before** the fallback route #982 added for exactly this failure is asked. It does not choose how fast to fail; it chooses to fail where main succeeds.

### Measured: it also disables #1223's snapshot fallback, and misreports why

The branch's timeout outcome carries no `outcomes` tags, so `noBackendAnswerEstablished()` reads `false` and `objectInfoSnapshot.authorize()` refuses with:

> the backend ANSWERED the schema probe with something unusable rather than failing to answer at all

which in a timeout is the one thing that did not happen. That is #982's own defect — a refusal asserting a cause it never established — committed a layer up.

### What I did with the branch

Merged `main`, reverted the code change and the duplicate changelog entry, and kept what this branch actually established as three guards in `object-info-cache.test.mjs`, written against the production loader shape:

1. **a hung client route still reaches the fallback THROUGH the burst cache** — asserted on *elapsed* time (the answer arrives at the client route's share of the budget, which is what proves the oracle's bound released the call, not something else);
2. **the cache arms NO timer of its own** — re-applying this branch's own `7de88a67` fails this test;
3. **a fully silent backend rides through the cache as SILENCE**, tags intact, never as a usable schema.

Each was mutation-tested by breaking the thing it guards: re-applying the branch's cache (kills 2), removing the oracle's per-transport bound (kills 1 and 3), dropping the `outcomes` tags (kills 3). `npm run test:unit` is green.

### Recommendation

Land these guards or close the PR — either is fine, and **the code that was under review here should not ship**. The value left in this branch is that the next person to reach for a cache-level bound now gets a failing test with the reason, instead of a fourth rediscovery.

Refs artokun/comfyui-mcp-panel#1161

Issue #1166, opened 2026-08-13
hardRestart() retires the mid-task marker only on success, so a failed restart leaves a turn armed with no way to resume


artokun:
Surfaced by review of #1164; **pre-existing**, independent of #1145/#1163. Not reproduced in the wild — found by reading the failure branches.

`hardRestart()` clears `MID_TASK_KEY` only inside its `if (ok)` branch. Every other exit leaves the marker armed:

1. **The POST fails or throws.** `data.ok` is falsy, so the marker is never cleared. The bridge reconnects to whatever orchestrator is still alive — quite possibly one whose turn finished while the panel was stopped, so its `turn:done` was never delivered. The panel now believes a turn is in flight that has already ended.

2. **The invalidate early-return.** When `data.ok` is true but `invalidateDurableAgentSession()` returns false, the function returns before BOTH `ssSet(MID_TASK_KEY, null)` and `client.start()`. The marker stays armed and the bridge stays stopped — indefinitely, until the user reconnects by hand. That path exists precisely to avoid restoring a wedged shell, and its comment says the reconnect "spins up a brand-new agent" and must not arm the resume nudge — which is exactly what the marker left behind does arm.

Both leave the panel asserting a pending turn that nothing will ever complete or retire.

## Why it is worth fixing on its own

The mid-task marker is the arming condition for the resume nudge, so a stale one is not inert — it decides whether a later reconnect injects "your connection dropped mid-task … continue exactly what you were doing" into the session. A user who hard-restarts is deliberately discarding the current agent state, which is the case where a resume nudge is least wanted.

It also interacts badly with any future change that measures restart downtime: an outage recorded across a hard restart would find this marker still armed and nudge. That interaction is what surfaced it — an earlier attempt at #1163 recorded an outage in `client.stop()`, and these branches turned it into a false nudge. That attempt was reverted, so **nothing currently exploits this**; the marker handling is simply wrong on its own terms.

## Suggested shape

Retire the marker on EVERY exit from a deliberate restart, not just the successful one — the same invariant `performSoftReloadRecovery` already enforces for the bridge ("reconnect on every path, even if the decision or effects threw"). A deliberate restart means the user gave up on the current turn regardless of whether the restart succeeded, so the marker should be cleared up front rather than as a success side effect.

The second path additionally leaves the bridge stopped with no auto-recovery, which is the #379/#419 no-dead-bridge invariant; worth checking whether that early return should also route through a restart.

Issue #379, opened 2026-07-31
panel_reload wedges the session: softReload() stops the bridge before the POST and never restarts it on failure


artokun:
## Summary

`softReload()` stops the bridge **before** it knows whether the reload can happen, and every failure path `return`s without restarting it. Against this pack — whose `/comfyui_mcp_panel/reload` route is a hardcoded `503` **by design** — that failure path is the *only* path. Result: `panel_reload` permanently disconnects the panel, with no auto-recovery, every single time.

Panel version 0.11.20, ComfyUI 0.29.2, Windows 11.

## Reproduction

100% reproducible. Call `panel_reload` (agent-side) or hit the reload control:

```
soft_reload → "soft reload (orchestrator) scheduled"      ← MCP tool reports success
Soft-reloading the agent (new code, no ComfyUI restart)…
Restart the orchestrator to pick up new code: npx -y comfyui-mcp@latest connect
Not connected — click Connect (in the Connection panel) and try again.
Not connected — click Connect (in the Connection panel) and try again.
```

The connection chip still reads **connected** (the ComfyUI websocket is fine) while the agent bridge is dead — so the UI actively contradicts itself and the user has no signal that Connect is the fix beyond one easily-missed line.

## Root cause

`web/js/comfyui-mcp-panel.js`, in `softReload()` (~line 15305–15330):

```js
reloading = true;
ssSet(SOFT_RELOAD_KEY, origin === "agent" ? "agent" : "user");
setSoftReloadGuard();
appendSystem("Soft-reloading the agent (new code, no ComfyUI restart)…");
try {
  client.stop();                    // <-- bridge dropped FIRST, unconditionally
  const res = await api.fetchApi("/comfyui_mcp_panel/reload", { method: "POST" });
  const data = await res.json().catch(() => ({}));
  if (!data?.ok) {
    ssSet(SOFT_RELOAD_KEY, null);
    clearSoftReloadGuard();
    appendSystem(data?.message || "Soft reload failed — try Disconnect then Connect.");
    return;                         // <-- returns with the client STILL STOPPED
  }
} catch (err) {
  ssSet(SOFT_RELOAD_KEY, null);
  clearSoftReloadGuard();
  appendSystem(`Couldn't reach ComfyUI to reload the agent: ...`);
  return;                           // <-- same: STILL STOPPED
} finally {
  reloading = false;                // does not restart the client either
}
client.start();                     // <-- only reached on success
```

[...]

PR #1183, opened 2026-08-13
fix(1171): hold the reload guard across hardRestart's tail

### LINKED ISSUE #1171: The reload re-entrancy guard is released before hardRestart's tail, so a second restart can run concurrently
Found by review of #1170. **Pre-existing**, and deliberately not fixed there — `reloading` is shared with `softReload()`, so changing when it is released alters that path too, and #1170 is meant to stay a one-call fix.

`reloading` is the mutual-exclusion flag between `softReload()` and `hardRestart()`. `hardRestart` sets it, then releases it in a `finally` that closes **before** the rest of the function:

- `reloading = false` runs in the finally
- `await invalidateDurableAgentSession()` runs after it
- `client.start()` runs after that

So from the moment the POST settles, the guard is open while the function is still working. A second restart — or a soft reload, since they share the flag — invoked in that window runs concurrently with the first one's tail: two `invalidateDurableAgentSession()` calls racing, and two `client.start()` calls against the same client. The window is as long as the invalidate takes, which is a round trip, so a double-click is enough.

The guard's own intent is stated by its first line (`if (reloading) return;`): one reload at a time. It just stops guarding early.

## Why not fixed in #1170

The obvious shape is to wrap the existing try/catch in an outer try/finally so the flag is released at the true end, covering the invalidate early-return too. That is mechanical, but it holds a SHARED flag longer and so changes when a concurrent `softReload()` is refused. That is a behaviour change to a path #1170 does not otherwise touch, and folding it in would repeat the scope creep that made #1163 take three review rounds.

Worth doing on its own, where the interaction with `softReload()` can be reasoned about and tested directly.

artokun:
Closes #1171.

## The defect

`reloading` is the mutual-exclusion flag between `softReload()` and `hardRestart()`. `hardRestart` set it, then released it in a `finally` that closed **before** the rest of the function:

- `reloading = false` ran in that `finally`
- `endTurnLocally()` and three marker retirements ran after it
- `await invalidateDurableAgentSession()` ran after those
- `client.start()` ran last

So from the moment the POST settled, the guard was open while the function was still retiring session state and restarting the bridge. A second restart — or a soft reload, since they share the flag — could run against that tail.

## The fix

The `try/finally` now spans the tail and releases just before the reconnect. That is **`softReload`'s existing shape**: its `beforeStart` hook clears the same flag at the same point, for the same reason, so the two paths finally agree. The `finally` is the backstop for a throw, and covers the invalidate-failure early return.

`connectBackend`'s `if (!await invalidateDurableAgentSession()) return;` was a **bare, silent return** — the user picked a backend, nothing happened, no transcript line, old backend still selected. It now says what `hardRestart` says. The existing key is reused, so the frozen English catalog (#1135) is untouched.

## What was tried and removed

An earlier version of this branch also **bounded `historyStore.flush()`**, reasoning that widening a guard over an unbounded await trades a race for a wedge. Review and a closer read retired it:

- **The wedge is not reachable.** `flush()` awaits the store's serial `_writePromise`; every write resolves on all three terminal transaction events (`oncomplete`/`onerror`/`onabort`), `db.transaction(...)` is wrapped in try/catch, and `openDb` is capped by `IDB_OPEN_TIMEOUT_MS` and resolves `null`. No modelled path leaves it unsettled.
- **The bound cost real behaviour.** `flush()` awaits the *whole queued chain*, including the full-snapshot write `persistThreads()` enqueues one line earlier — so a timeout meant "the queue was busy", not "the store is broken". That gave the function a new way to answer `false` for a healthy store, and landed on two exits that cannot absorb it: `connectBackend` abandons a switch whose chips, prefs, `STORAGE_KEY_BACKEND`, `endTurnLocally()` and armed replay have **already** committed to the new backend, and `hardRestart` skips `client.start()` and strands the bridge for a write that lands a moment later.

My own measurement did not catch that: I timed persist+flush on a **fresh store with an empty queue** (35/40/61ms across 52KB–1MB), which is not the quantity the bound covered. The removal is asserted by a test, so reinstating a bound has to come with a reason that survives both problems.

## Verified

- `reloading` released in exactly one place, inside a `finally`, after the tail and before the reconnect — asserted structurally, and the mutation that drops the `finally` (latching the flag on the early-return path) fails the suite.
- Nothing in the tail can re-enter the guard: `reloading` has exactly two readers, and the tail is synchronous flag sets, `sessionStorage` writes, a `clearInterval` and two variable nulls, plus the one await.
- **A throw in the tail leaving the bridge stopped is pre-existing, not a regression** — compared against `main`, which also wraps neither the tail nor `client.start()` in a `catch`. Worth its own look under the #379/#419 "a reload never leaves a bridge dead" invariant; deliberately not folded in here.
- Panel loads and registers cleanly (`comfyui-mcp.agent-panel` present in `app.extensions`), no panel errors in console.
- Suite 4132 pass / 0 fail; scope check and typecheck clean.

## Why the tests are structural

This pack's `/comfyui_mcp_panel/hard_restart` answers `{"ok": false}` unconditionally — the orchestrator runs out-of-band — so the `if (ok)` tail never executes locally and the interleaving cannot be raced on this machine. Same approach #1166's test takes for the same function: *where* the release sits relative to the tail and the reconnect is the entire content of the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,16 +6,19 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
-### Fixed
+### Changed
 
-- **Restarting the agent can no longer overlap with a second restart or a reload (#1171).**
+- **Restarting the agent holds its exclusion flag until the reconnect (#1171).**
   The panel keeps one flag so that restarting the agent and reloading it cannot run at the
-  same time — but the restart released that flag as soon as the request came back, while it
-  was still finishing: ending the current turn, clearing the markers that say a turn is in
-  flight, invalidating the old session, and reconnecting. Anything started in that window
-  ran straight through the middle of it, tearing down the same state twice.
-  The flag is now held until the reconnect, which is what the reload path already did, so
-  the two finally agree.
+  same time, but the restart released it as soon as the request came back — while it was
+  still ending the current turn, clearing the markers that say a turn is in flight,
+  invalidating the old session, and reconnecting. The flag is now held until the reconnect,
+  which is what the reload path already did, so the two agree.
+  **No visible change when the panel manages ComfyUI's agent the usual way**: this
+  distribution's restart route always answers "not restarted" (the orchestrator runs
+  out-of-band), so the work this now protects is skipped anyway. It matters for setups
+  whose restart route really does restart the agent, and it removes the window before one
+  of those hits it.
 
 ## [0.14.25] - 2026-08-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,21 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
-## [0.14.25] - 2026-08-13
+### Fixed
+
+- **Restarting the agent can no longer overlap with a second restart or a reload (#1171).**
+  The panel keeps one flag so that restarting the agent and reloading it cannot run at the
+  same time — but the restart released that flag as soon as the request came back, while it
+  was still finishing: ending the current turn, clearing the markers that say a turn is in
+  flight, invalidating the old session, and reconnecting. Anything started in that window
+  ran straight through the middle of it, tearing down the same state twice.
+  The flag is now held until the reconnect, which is what the reload path already did, so
+  the two finally agree.
+- **Switching to a different agent backend no longer fails silently (#1171).**
+  If the old provider's session could not be invalidated, the switch stopped with nothing
+  said: no message, and the previously selected backend still showing. It now says the same
+  thing a restart says when it pauses for the same reason. (The panel can still be left
+  showing a backend it did not finish switching to — that is tracked separately in #1184.)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,11 +16,8 @@ All notable changes to this project are documented here. This project adheres to
   ran straight through the middle of it, tearing down the same state twice.
   The flag is now held until the reconnect, which is what the reload path already did, so
   the two finally agree.
-- **Switching to a different agent backend no longer fails silently (#1171).**
-  If the old provider's session could not be invalidated, the switch stopped with nothing
-  said: no message, and the previously selected backend still showing. It now says the same
-  thing a restart says when it pauses for the same reason. (The panel can still be left
-  showing a backend it did not finish switching to — that is tracked separately in #1184.)
+
+## [0.14.25] - 2026-08-13
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


artokun:
## Verification of the risk this change actually carries

Widening a mutual-exclusion guard is only safe if nothing inside the newly-guarded region can re-enter it, and only non-wedging if that region cannot hang. Both checked rather than assumed.

### Nothing in the tail can re-enter the guard

`reloading` is read in exactly two places — `softReload()` and `hardRestart()`. Every statement now inside the widened guard was traced:

| statement | what it does | can re-enter? |
|---|---|---|
| `endTurnLocally()` | sets `agentWorking`/`localEndAt`, one `ssSet`, `hideThinking()` | no — synchronous, no reload path |
| `ssSet(SOFT_RELOAD_KEY, null)` / `ssSet(REBOOT_KEY, null)` | sessionStorage writes | no — synchronous |
| `stopRebootWatch()` | `clearInterval` | no |
| `forgetRebootResumeAttempt()` | nulls two variables | no |
| `await invalidateDurableAgentSession()` | the one async step | no — bounded, see below |

So holding the guard across the tail blocks exactly what it is meant to block: a second `hardRestart` or a `softReload` racing this one's teardown.

### The one await is bounded, and the bound is not a guess

`invalidateDurableAgentSession()` awaits `historyStore.flush()`, which awaits the store's in-flight write promise. Unbounded, one stalled flush would latch `reloading` for the rest of the session — no further restart, no soft reload, no affordance to clear it. That is the wedge this issue's history warns about: two prior attempts at the sibling bug "each introduced a worse bug."

`SESSION_INVALIDATE_FLUSH_MS = 4000` is sized against the store's own decision: `chat-history-store.js` already caps IndexedDB open at `IDB_OPEN_TIMEOUT_MS = 2000` and resolves `null` past it, so the open half is bounded there and twice that leaves the same margin again for the write.

A timeout returns `false` = **not durably invalidated**, which is the question this function exists to answer and cannot answer if the flush did not finish. Both callers already handle `false` by pausing and disclosing it — the fail-closed direction, since the risk of proceeding is restoring the session the restart exists to discard. The write is not cancelled and may still land; the next call sees a settled store.

This also fixes a latent hang on the backend-switch path (`if (!await invalidateDurableAgentSession()) return;`), which awaited the same call and simply never proceeded if the flush stalled.

### What local verification can and cannot show

This pack's `/comfyui_mcp_panel/hard_restart` answers `{"ok": false}` unconditionally — the orchestrator runs out-of-band — so the `if (ok)` tail never executes on this machine and the interleaving cannot be raced locally. That is why the tests are structural, matching how #1166's test covers the same function: *where* the release sits relative to the tail and the reconnect is the entire content of the fix.

What was verified live: the served panel parses and registers (`comfyui-mcp.agent-panel` present in `app.extensions`), with no panel errors in console.

Mutations verified caught: releasing the guard before the tail, removing the flush bound, and reporting a timeout as a successful invalidation.

Issue #1191, opened 2026-08-13
Every release since 0.14.22 credits itself with work that already shipped — gen-changelog bounds on tags, and tagging stopped at v0.14.21


artokun:
Found while cutting 0.14.27.

## What happens

`scripts/gen-changelog.mjs` computes its commit range from the newest **git tag**. The panel's tags stop at `v0.14.21`:

```
v0.14.18  v0.14.19  v0.14.20  v0.14.21
```

while `pyproject.toml` has run on to **0.14.26**. So every release cut since then re-lists everything back to v0.14.18.

Cutting 0.14.27 produced **32 entries** for a release containing exactly one change. Among them, credited to 0.14.27:

- "Brazilian Portuguese (pt-BR) panel catalog" — shipped 0.14.24
- "Russian (ru) panel catalog" — shipped earlier
- eight more language catalogs, likewise

0.14.22 through 0.14.26 all carry the same inflation, and it is visible to users: `web/changelog.json` is generated from `CHANGELOG.md` and rendered in the panel's in-app changelog.

## Why it matters beyond tidiness

A changelog is the artifact users check to decide whether upgrading fixes their problem. One that lists ten features under a version that did not add them makes "was my bug fixed in X?" unanswerable — and someone on 0.14.24 reading the 0.14.27 notes sees their own already-installed features presented as new.

## Two candidate fixes

1. **Tag the releases.** The generator's assumption is reasonable; the tags are what stopped. This also restores `git tag --contains` for triage, which several closure workflows lean on.
2. **Bound on the previous `## [x.y.z]` heading in CHANGELOG.md** instead of on tags, so the generator does not depend on a step that can be skipped.

(1) is the smaller change and fixes the triage problem too. (2) is more robust to the same thing happening again. They compose.

## Not fixed here

0.14.27 was trimmed by hand rather than fixing the tooling inside a release commit. The already-published 0.14.22–0.14.26 entries are still inflated and would need a backfill pass (`gen-changelog.mjs --backfill` exists — worth checking whether it reconstructs these correctly once tags are restored).

Found while shipping artokun/comfyui-mcp#1520.

artokun:
it's the "v" probably, just strip it so it matches up properly

artokun:
Reopening: this was closed incidentally by the 0.14.28 release commit that *referenced* it, not fixed. `origin/main` still carries the two-arm predicate verbatim.

Diagnosis measured at pinned refs since:

[...]

Issue #1188, opened 2026-08-13
34 more unbounded network awaits can hang a command on the same half-open connection as #1161/#1180


artokun:
## Summary

#1161 bounded the `/object_info` oracle; #1180 bounded the sibling `getNodeDefs` sites and the log read on `graph_add_node`'s refusal path. Both were found the same way — a user reported one hanging command. **An audit of the panel while finishing #1180 found 34 more unbounded network awaits**, on other commands, with the identical failure mode.

## The failure mode, restated

After a ComfyUI restart the tab can hold a half-open connection: a request neither answers nor fails. `try/catch` does not help — there is nothing to catch — so any bare `await fetch(...)` or `await api.fetchApi(...)` parks until the caller's own 30s command timeout. The agent then gets `did not reply within 30000 ms`, naming nothing, which is strictly worse than a worded refusal.

## What the audit found

34 network awaits in `web/js/comfyui-mcp-panel.js` with no bound anywhere near them (comments stripped; a site counts as bounded if `withTimeout`, a budget, or a timeout appears within three lines). Spread across many commands — the sample includes the orchestrator API calls around line 1173-1223 and the `/userdata` reads around 4592-4655.

`web/js/lib/` is clean by comparison: the only bare network await is in `single-node-def.js`, and **both** of its call sites bound it — `graph_add_node` via #1180's `withTimeout`, and `graph_get_errors` via the `errorsStepBudget` machinery from #589/#610.

## Why this is filed rather than fixed

Bounding 34 call sites is not a mechanical sweep. #1180 established that each one needs its own decision:

- **What a timeout MEANS there.** `graph_add_node`'s fallback degrades to "no defs" and fails closed; `widenSocketProof` maps it to "keep the proof I have", because the alternative weakens a proof to an empty set and produces the #695/#700 false-cause message; the refresh's retry treats it as a failed attempt.
- **What the bound must FIT INSIDE.** `widenSocketProof` gets half `CUSTOM_WIDGET_REGISTRATION_TIMEOUT_MS` because it runs inside that wait. The refresh budget is sized so *two serialised runs* fit `BRIDGE_READ_DEFAULT_MS`, because `makeRefreshCoalescer` guarantees a forced call runs after the in-flight one. Both were wrong in the first draft.
- **Whether a bound REGRESSES a slow install.** A bounded call can refuse something an unbounded one eventually served. #1180 documents that trade for the refresh with its measured numbers.

[...]

Issue #1180, opened 2026-08-13
graph_add_node and panel_refresh_nodes still hang on the half-open socket #1161 fixed for set_widget


artokun:
## Summary

#1161 bounded the `/object_info` oracle so `graph_set_widget` survives a half-open connection after a ComfyUI restart. **The same unbounded `await api.getNodeDefs()` remains on three other paths**, so the identical failure still hangs other commands. Fixing one command and leaving the siblings makes the behaviour inconsistent in a way that is hard to report: after a restart, setting a widget works and adding a node hangs.

## The failure

After a ComfyUI restart the tab can hold a connection that never answers and never fails. `api.getNodeDefs()` then never settles — it does not throw. Any unbounded `await` on it blocks until the caller's own 30s command timeout.

## The remaining call sites

| file:line | path | consequence |
|---|---|---|
| `web/js/comfyui-mcp-panel.js:10109` | `graph_add_node` — fresh whole-schema fetch when the single-class fast path does not confirm | **`graph_add_node` hangs** |
| `web/js/comfyui-mcp-panel.js:10164` | `widenSocketProof` — #821's whole-schema widen, on the path that is about to refuse | **hangs while producing a refusal** |
| `web/js/comfyui-mcp-panel.js:859` | `fetchNodeDefsWithRetry` in the node-def refresh | **`panel_refresh_nodes` hangs** — the source comment already says so: *"the three requests themselves are unbounded — codex"* |

`comfyui-mcp-panel.js:767` (the startup baseline seed) is **not** affected in the same way and is listed only to rule it out: nothing awaits the seed directly, and `awaitObjectInfoHistorySeed()` already bounds the wait at 8s and fails closed with a recoverable "retry in a moment". I previously described this as four unbounded sites; it is three that can hang a command, plus this one that cannot.

## Why this is separate from #1161

Two of the three are in `graph_add_node`, which uses the **per-class** `/object_info/<Type>` route as its fast path and only falls back to the whole schema. `set_widget` cannot use the per-class route at all (#716/#821 — it authorizes two types before resolving which it writes to). So the two commands have genuinely different transport shapes, and #1161's oracle is not a drop-in for either add_node site. Bundling them would have meant reworking `add_node`'s fast path inside a fix for a different bug.

[...]

Issue #1175, opened 2026-08-13
panel_run times out before a successful late acknowledgement


artokun:
### What happened / root cause
`panel_run` returned outcome-unknown after a 20s reply timeout, but the panel acknowledged the same mutation 5.3s later and had successfully queued it. An immediate queue list returned empty, making safe reconciliation confusing. The retry token later recovered the original success without duplicating.

### Steps to reproduce
1. Restart/reconnect ComfyUI with an active workflow tab.
2. Call `panel_run({to_node_id: 8})` while the frontend is slow to answer.
3. Observe timeout at 20s, followed by a late acknowledgement.

### Exact error (scrubbed)
`Panel tab ... did not reply to graph_run within 20000 ms ... mutation may have been applied.`

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: reconcile late acknowledgements for mutation commands before returning outcome-unknown, or extend the mutation acknowledgement window and expose a receipt-status lookup keyed by rid so callers can resolve the result without querying the execution queue during its propagation race.

### Environment
macOS Darwin; ComfyUI 0.28.0; frontend 1.47.10; Apple MPS / 16 GB unified memory; comfyui-mcp 0.51.19; panel 0.14.14.

artokun:
Fixed in comfyui-mcp v0.51.57 (artokun/comfyui-mcp#1630): on a bridge-tagged reply timeout, panel_run now polls up to 10s for the late graph_run ack and takes the identical success path with a [RECOVERED] disclosure, instead of reporting outcome-unknown.

PR #1201, opened 2026-08-14
fix(1188): bound the remaining unbounded network awaits

artokun:
Works #1188. **Scoped — this does not close it.**

Continuation of #1161 (the `/object_info` oracle) and #1180 (the sibling `getNodeDefs` sites and the log read). Same failure: after a ComfyUI restart the tab can hold a half-open connection where a request neither answers nor fails, so `try/catch` cannot help and a bare `await` parks until the caller's 30s timeout.

## What #1180 taught that this must not repeat

- **`withTimeout` treats a non-positive `ms` as NO bound**, so passing `0` silently restores the hang.
- **It does not cancel.** An abandoned attempt keeps running, so bounds that retry can stack concurrent downloads.
- **Per-site bounds do not compose.** Sizing each one alone produced a path summing to ~39.6s against a 20s read default. Where several run in sequence, they need one shared deadline.
- **Bound the whole read, not the handshake.** `fetch` resolves when headers arrive; the stall lives in `res.json()`.
- **A guard can cause what it reports** — a diagnostic added during #1191 dropped the changelog base to the first commit.

## Work in progress

- [x] re-derive the site list against current `main` rather than trusting the issue's count
- [x] bound a scoped subset, stating which and why
- [x] tests that fail when each bound is removed *and* when it is passed `0`
- [x] verify by mutation

## The site list, re-derived

The issue's figure of 34 is wrong, and so was this PR's first correction of it. Counting **request awaits only** across `web/js` (excluding `vendor/`) gives **39 sites, 29 of them with no bounding primitive in scope**. Counting body reads as well — which is the honest unit, since `fetch` resolves at the head and the stall lives in `json()`/`text()` — gives **92 sites, 71 unbounded**.

This is recorded as a **heuristic, not a measurement**: the detector treats "a bounding primitive within ±40 lines" as bounded, and proximity does not prove coverage. Two were hand-verified in the earlier commit. The number is here to size the remaining work, not to be quoted as a count.

Bounded so far across this branch: **6** — the credentials console's three, and the three `/upload/image` sites below.

## This commit: the three `/upload/image` sites

Chosen because they share one endpoint, one failure, and one already-documented symptom.

**The wedge.** The composer awaits `Promise.all(pending.map((a) => a.ready))` before sending (`comfyui-mcp-panel.js:31607`). `att.ready` catches everything internally, so it never **rejects** — it simply never settles. After a restart leaves a half-open connection, the user cannot send a message at all and nothing on screen says why. The training wizard already hit this from the other side and worked around it locally (`cmcp-training-ui.js:704-719`): *"ONE hung upload left `uploadsPending > 0` permanently and deadlocked wizard navigation"*.

**Sized by payload, not flat.** The other #1188 sites carry a fixed number because their payload is a token or a small JSON body. This endpoint takes whatever the user dropped on it — `handleMediaUpload` exists specifically for video — so a flat bound either cuts off a legitimate large upload or waits absurdly long for a small one. The bound is a 30s floor plus an allowance at a 50 KB/s throughput floor, so it cannot fire on a transfer that is actually moving bytes.

**Stated plainly: this is a DURATION bound, not a STALL bound.** A transfer that dies halfway through a 2 GB file still waits out the whole allowance. Detecting that needs upload progress events, which `fetch` does not emit — it would mean moving the path to `XMLHttpRequest` (`upload.onprogress`), which is larger than #1188's scope and is **not** done here. What this does guarantee is that the wedge is finite, which is the property the composer's `Promise.all` needs.

**Each site keeps its own failure contract**, so no branch is added. `uploadBlobToInput` collapses the sentinel to the `null` it already returns for every other failure. There are **five** callers — `cmcp-apps-ui.js:1697`, `cmcp-civitai-ui.js:1196`, `cmcp-training-ui.js:754`, `lib/media-preview.js:893`, `lib/run-completion-frame.js:407` — and an earlier draft of this sentence said "all four … branch on that today", which was wrong on both halves. See the review record below; CivitAI's missing branch was a real bug and is fixed. The two composer sites report through `describeUploadFailure`'s `error` branch, whose #756 wording — *"Whether any bytes reached the server is unknown"* — is precisely true here, because the bound does not cancel.

**Composition, checked rather than assumed.** The storyboard pipeline already bounds the whole `sample → upload → HEAD` sequence at `MEDIA_PREVIEW_TIMEOUT_MS` (20s), so the inner bound is not what saves that path; it gives the other three callers, which have no outer bound, a specific failure instead of a hang.

**Placement.** Lands in `lib/attachment-upload.js` rather than the monolith: that module already owns upload failure semantics, it is unit-testable directly, and it holds the monolith diff to the three call sites while five other PRs contend for that file.

## #756's wiring test was updated, not relaxed

Moving the exchange into a bounded callback changed the expression that test pinned. The same contract is now pinned at **both ends** of the new shape — the description is built **and** surfaced — so neither half can be dropped silently.

## Mutations, all caught

Nineteen, listed in the review record below.

Two harness notes worth keeping, because both make a harness lie in the *safe-looking* direction: the working tree is CRLF, so multi-line anchors written with `\n` never match and a missed anchor reads exactly like a caught mutation; and the `missing-withTimeout` mutation makes the suite **hang** rather than fail, because the tests feed it a never-settling promise on purpose. Both are now reported as distinct outcomes.

## Recorded, not fixed here

`cmcp-training-ui.js:711` defines a **second `withTimeout`** with an incompatible contract — it rejects rather than degrading, and it does not guard a non-positive `ms`. `bounded-step.js`'s own header warns that *"a second timeout helper written alongside the first is how this repo keeps producing near-duplicate bugs"*. Its 30s outer bound is now redundant for this path but harmless; retiring it is a separate change.

## Scope boundary

#1188 as re-derived covers **HTTP-level** awaits (`fetch` / `api.fetchApi`). `graph_run`'s `graphToPrompt()` and `queuePrompt()` awaits are unbounded awaits of the same class but are **frontend-API**, not HTTP, and belong to #1175/#1192.

---

## Review record

Five Codex passes. Four returned NO-SHIP, and every one of them found something real — including three defects where **the bound cost more than the hang it replaced**, which is the specific way this class of fix goes wrong.

| # | Finding | Resolution |
|---|---|---|
| 1 | A `File` whose `size`/`type` getter throws threw *again* inside the reporting catch, escaped the handler, and **rejected `att.ready`** — breaking the send path for every attachment, not just the bad one | `readFileFacts()` reads each once, defensively |
| 1 | A stalled refusal body dragged a **known HTTP status** past the outer bound and reported it as "no response" — #756's own defect, reintroduced | `readErrorBody()` bounds the body separately and shorter |
| 1 | `setTimeout` coerces delays above 2³¹−1 ms to **1 ms**, so an over-generous allowance became an *instant* refusal | clamped to `MAX_TIMER_MS` |
| 1 | Slow-but-progressing uploads can be cut off | **Argued, not changed** — accepted as sound in pass 2 |
| 2 | The body's bound runs *inside* the outer one, so a head arriving near the outer deadline still lost the status | status recorded the instant it is known; `describeTimedOutUpload` consults it first |
| 3 | The new guard still fabricated refusals: `0`, `false`, `[]` rendered as `HTTP 0` / `HTTP false` / blank | `isHttpStatus()` — an integer in [100, 599] |
| 3 | The comment claiming "all four callers branch on null" was **false on both halves** — five callers, and CivitAI's did not branch at all | comment names all five with file:line; CivitAI's real bug fixed |
| 4 | A `2xx` is a status but not a refusal — `"200"` produced "upload REFUSED … Nothing was written" | refusal wording gated on `>= 400` |
| 4 | The caller-coverage test was **vacuous** — an unscoped search matched unrelated `!ref` ~340 lines away and passed with the guard deleted | scoped to 400 comment-stripped chars; verified red at all five sites |

Two of those are worth calling out because they are about *this repo's own stated rules*, broken in the act of enforcing them:

- `Number(null)` is `0`, which is finite — so "no status was ever observed" rendered as a confident `HTTP null`. That is `describeSize`'s ABSENT-before-COERCED rule, violated three functions below where it is written down.
- The false "four callers" comment had **already been read back as established fact** by a later review before anyone counted. It now names each site with file:line, and a test counts them so the claim cannot rot back into a guess.

`check-panel-scope` also caught two `ReferenceError`s from an import a CRLF-mangled patch had skipped — which all 4208 unit tests walked straight past.

## Verification

- **19 mutations, all confirmed red.** Two harness traps are recorded in the commits because both make a harness lie in the *safe-looking* direction: CRLF makes multi-line anchors silently miss (a missed anchor reads exactly like a caught mutation), and removing a bound makes the suite **hang** rather than fail.
- Full unit suite green: **4209 tests, 0 failures**.

### DIFF
--- a/browser_tests/unit/attachment-upload.test.mjs
+++ b/browser_tests/unit/attachment-upload.test.mjs
@@ -110,10 +110,19 @@ test("#756 WIRING: both upload paths record the status and the exception", () =>
   // The helper being right proves nothing if the call sites still discard. Pin the
   // two sites that had the defect, and that neither bare form survives.
   const src = readFileSync(PANEL_JS, "utf8");
+  // #1188 moved the exchange inside a bounded callback, so the non-200 description is now
+  // BUILT there and assigned by the caller. Same contract — the status must still be
+  // captured on both paths — pinned at both ends of the new shape so neither half can be
+  // dropped: a `failure:` nobody reads, or an `att.uploadError` fed by something else.
   assert.equal(
-    (src.match(/att\.uploadError = describeUploadFailure\(\{\s*\r?\n?\s*status: res\.status/g) || []).length,
+    (src.match(/failure: describeUploadFailure\(\{\s*\r?\n?\s*status: res\.status/g) || []).length,
     2,
     "both upload paths must record a non-200 status",
+  );
+  assert.equal(
+    (src.match(/att\.uploadError = outcome\.failure;/g) || []).length,
+    2,
+    "…and both must surface it as the attachment's error",
   );
   assert.equal(
     (src.match(/catch \(err\) \{[\s\S]{0,200}?att\.uploadError = describeUploadFailure\(\{ error: err/g) || []).length,
--- a/browser_tests/unit/attachment-upload.test.mjs
+++ b/browser_tests/unit/attachment-upload.test.mjs
@@ -124,10 +124,21 @@ test("#756 WIRING: both upload paths record the status and the exception", () =>
     2,
     "…and both must surface it as the attachment's error",
   );
+  // Window widened from 200 to 500: #1188 added a defensive re-read of the file's
+  // measurements inside each catch, which pushed the assignment further from `catch (err) {`.
+  // The contract is unchanged — the caught exception must still be what gets reported.
   assert.equal(
-    (src.match(/catch \(err\) \{[\s\S]{0,200}?att\.uploadError = describeUploadFailure\(\{ error: err/g) || []).length,
+    (src.match(/catch \(err\) \{[\s\S]{0,500}?att\.uploadError = describeUploadFailure\(\{ error: err/g) || []).length,
     2,
     "both upload paths must record the caught exception",
+  );
+  // #1188 — and neither catch may read `file.size`/`file.type` directly. A throwing getter
+  // there throws a SECOND time while reporting the first failure, escapes the handler, and
+  // rejects `att.ready` — which the send path awaits and cannot handle.
+  assert.equal(
+    (src.match(/catch \(err\) \{[\s\S]{0,500}?describeUploadFailure\(\{ error: err, name, size: file\.size/g) || []).length,
+    0,
+    "the reporting catch must not re-read a property that may be what threw",
   );
   assert.ok(
     !src.includes("/* upload failed — the chip still references it by name as a fallback */"),
--- a/browser_tests/unit/cmcp-secrets-bound.test.mjs
+++ b/browser_tests/unit/cmcp-secrets-bound.test.mjs
@@ -1,0 +1,104 @@
+// #1188 — the credentials console's three requests, bounded.
+//
+// Same failure as #1161 and #1180: after a ComfyUI or orchestrator restart the tab can hold
+// a half-open connection where a request neither answers nor fails, so there is nothing for
+// `try/catch` to catch. Here it wedges the UI rather than a command — the button stays
+// disabled reading "Saving…" forever, and because the panel only re-enables it in the
+// catch, the user cannot retry without reopening the frame.
+//
+// These are SOURCE assertions. The credentials frame is built inside the 1.7MB panel IIFE
+// from DOM handlers that cannot be constructed here, so what is pinned is the shape the
+// runtime behaviour depends on. Each assertion below corresponds to a mutation that passed
+// before it existed.
+
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+
+const SRC = readFileSync(
+  fileURLToPath(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url)),
+  "utf8",
+);
+
+/** The helper's body, anchored at both ends so prose cannot drift the window. */
+function helperBody() {
+  const at = SRC.indexOf("async function cmcpSecretsRequest(init) {");
+  assert.ok(at > 0, "cmcpSecretsRequest must be findable");
+  const end = SRC.indexOf("\n}", at);
+  assert.ok(end > at, "…and terminated");
+  // CODE ONLY. The comments in this helper NAME the calls they explain — `resp.json()`,
+  // `withTimeout`, `throw` — so a raw scan matches the prose and a mutation that moves real
+  // code into a comment slips through. One did: replacing the throw with a return while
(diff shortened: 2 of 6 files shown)


artokun:
## Review landed — DO NOT MERGE as-is

Four confirmed defects, two of them user-facing and **created by this change**. Recording them here so they are not lost; I am at my context limit and stopping rather than attempting the fixes badly.

### 1. No `AbortSignal` — an abandoned write can destroy a key

`withTimeout` does not cancel. So at 8s the helper throws, the catch **re-enables the button**, and the user can issue a second request that races the first.

Concretely: Clear stalls → button re-enabled, row still reads "set" → user pastes a new key and Saves, succeeding in ~1s → the original `POST {slot, clear:true}` lands at ~12s and **wipes the slot**. The card says the key is stored; the orchestrator has none.

**This window did not exist before this PR** — the button stayed disabled, so no second request was possible. I made a stall recoverable and introduced a race doing it.

The repo already solves this twice: `fetchImageBytes` uses `AbortController`+`signal`, and another site composes `withDeadline` with `onTimeout: () => readAbort?.abort()` under a comment saying the deadline stops *us* waiting, not the request.

### 2. A write bound applied to a read

The 8s constant is justified in its own comment as sized for *a user-initiated write*. It is also applied to the initial **load**, where giving up early empties the card — no rows, no Save buttons, no retry control. After a restart (the scenario this issue is about) a 9s cold start makes the dialog unusable; previously the rows simply appeared at 9s.

### 3. The tests do not hold the fix

Verified mutations that ship **5/5 green**: zeroing the bound, swapping `withTimeout`'s argument order, resolving a stall as success, and a raw `fetch` written across multiple lines. The assertion advertised as "the whole fix in one line" is line-based, so multi-line evades it. Same vacuous-assertion class already fixed twice on #1180 and #1184 — and I introduced it again here.

### 4. The stall string escaped the i18n guard

`credentials-card-i18n.test.mjs` slices from `function cmcpOpenCredentialsFrame(`; the helper was inserted **before** that anchor, so the enumerated-literal guard cannot see it. That test's header names this card specifically. On a non-English screen a stalled save renders English as the card's only text.

### What a corrected version needs

- an `AbortController` per request, aborted in `onTimeout`, so a bound write cannot land late
- a **separate, longer or unbounded** treatment for the initial load, or a retry affordance if it is bounded
- the helper moved **inside** `cmcpOpenCredentialsFrame` so the i18n guard covers it, or the guard's anchor widened
- assertions that survive the four mutations above — the multi-line raw-fetch case in particular

The corrected count for #1188 (22, not 34) in the issue comment stands and is independent of this.

PR #1232, opened 2026-08-14
fix(1223): a transient /object_info timeout must not refuse a safe widget edit

### LINKED ISSUE #1223: panel_set_widget refuses live edit when backend schema probe times out
A live canvas edit was partially completed, but panel_set_widget refused to set a widget on an existing H3Keyframes node because both schema probes timed out. The graph was reachable and panel_disconnect mutations succeeded immediately before this call.

### Steps to reproduce
1. Have a connected ComfyUI canvas containing a custom H3Keyframes node.
2. Call panel_set_widget on that node.
3. Observe refusal after api.getNodeDefs and GET /object_info time out.

### Exact error (scrubbed)
"cannot verify the node type against the ComfyUI backend — no usable /object_info schema was obtained ... api.getNodeDefs() did not answer ... GET /object_info did not answer ... Refusing to write ... (#458)"

### Fix
Applied locally: no. Upstream-only: yes. Precise change needed: when the live graph has already identified the node type and graph mutations are otherwise working, use a bounded cached schema or retry/backoff path rather than refusing a safe widget edit solely because a transient object_info probe timed out. Preserve stale-cache safety for genuinely unknown node types.

### Environment
Windows 11 / ComfyUI 0.33.0 / NVIDIA RTX 5090 32 GB / comfyui-mcp 0.51.45 / comfyui-mcp-panel 0.14.34.

Related (same root cause): #982

artokun:
Fixes #1223 (related: #982, #458, #716, #1161).

## The report

`panel_set_widget` refused a live edit on an existing `H3Keyframes` node because *both* `/object_info` probes timed out. The graph was reachable — `panel_disconnect` mutations had succeeded moments before — and the node type was already identified on the live canvas.

## Why it times out while the backend is fine

ComfyUI serves HTTP from the same process that runs the graph, so a heavy step (a VAE decode, a model load) blocks the event loop and `/object_info` — megabytes on a large install — waits behind it. A **busy** backend, not an absent one, and it is the ordinary condition *during a render*, which is exactly when an agent is editing widgets.

## The trust root

A ComfyUI process **cannot change the set of types it serves without restarting** — `NODE_CLASS_MAPPINGS` is built at import time. A restart drops this tab's websocket. So a `/object_info` map observed on the **current backend connection** still describes the process answering now; that provenance is something the LiteGraph registry lacks, which is why #458 refuses to trust it.

`web/js/lib/object-info-snapshot.js` holds the type names of the last whole map observed, and authorizes from it only when **all four** hold:

1. **A whole map was observed, and its observer said so** (`whole: true`). A per-class `/object_info/<Type>` payload would make every other type read as absent and the ever-seen gate diagnose the install as removed packs — and that payload *can* reach the refresh path.
2. **The socket is up** (`comfyBackendSocketDown`).
3. **No reconnect since the schema was FETCHED** — the epoch is captured at issuance and re-checked when the answer lands.
4. **The backend never answered.** A timeout, a client that returned nothing, or a 504. Anything that answered — threw, a non-gateway status, an empty schema, a stalled body — keeps the existing refusal.

## Three review rounds, and what each one caught

This was rebuilt twice. Recording it because each defect looked correct when written.

**Round 1 — the fix did not fire, and could reopen #458.**
- On a normal startup `seedObjectInfoHistory()` performs the ONLY whole read; `registerComfyNodeDefs()` never runs. So the reported case found no snapshot and was refused exactly as before. **The fix was inert.**
- The epoch was read *after* the fetch, and the clear meant to catch that never ran: the `reconnected` handler's payload-less `refreshComfyNodeDefs()` is coalesced into any in-flight run (`refresh-coalesce.js:50-58`), so it never reaches `registerComfyNodeDefs`. A pre-restart schema could be filed under post-restart provenance.
- The payload was retained by reference, and `registerNodesFromDefs` runs `beforeRegisterNodeDef` hooks that mutate defs in place.
- Snapshot defs reached `refreshCombos`, which does `w.options.values = first.slice()` — rewriting a live dropdown *backwards*, persistently.
- The ineligibility reason was spliced into `failures`, so a two-transport failure reported "Tried **3** routes" (#982's own defect).
- A comment cited `recordsWholeSchemaOnly`, a symbol that does not exist.

**Round 2 — two more P1s.**
- **A 502 is not silence.** nginx/Caddy emit it *immediately* when they cannot CONNECT to a stopped ComfyUI — evidence the process is **gone**. Only 504 (connected, upstream quiet) now counts. Including 502 would have let a pre-restart schema authorize a write to a removed type.
- **A cache hit is not evidence.** The issuance epoch was captured before `objectInfoCache.read()`, but a hit returns a payload fetched up to a TTL earlier, and the reconnect in between never invalidates the cache. The epoch is now captured *inside* the loader, which runs only on a miss.

**Round 3 — two P2s.**
- The `!preloadedDefs` rule was too blunt: when `graph_add_node` hits an unregistered type it fetches a whole map, and `registerComfyNodeDefs` cleared the snapshot then declined to re-record it — so an add that registers a new type ended with **no snapshot**, reproducing the original refusal. The resolver now files its own payload, gated on `freshDefsAreSingleClass`.
- The HTTP status was read twice; a getter-backed response can answer differently each read, so 504 could classify as silence while 502 was reported.

## Storage is type NAMES, not the schema

Copying the key names is both the detachment and the whole of what this needs — the fence asks "does the backend define this type". It also keeps a ~5.4MB schema from being pinned for the life of a connection, and it means the values cannot supply stale combo lists to anything downstream. The map is null-prototype so a type named `__proto__` cannot reach `Object.prototype`.

## Disclosure

A write authorized this way returns `schema_source: "last-observed"` and a `schema_note`. On **its own fields, never `warning`** — that channel is single-slot and priority-ordered, so appending there would displace a warning about what the write actually *does*.

## What this does not close

Conditions 2 and 3 rest on the tab *noticing* the socket went away. A half-open TCP connection to an already-dead process does not report itself. Exploiting it needs the restart **and** a pack removal across it **and** a write to that exact type inside the window. An **install** is harmless (a type missing from the snapshot is still refused); only a **removal** can authorize something it should not. This widens the race `object-info-cache.js` already documents for its own 1.5s window rather than opening a new class of hole, and it is written into the module header.

## Verification

- **4363 unit tests pass** (44 new).
- **34 mutations applied, 32 killed.** The two survivors are equivalent mutants, documented in the source: one writes a provably-equal value, the other is defence-in-depth over a check `record` already performs.
- Mutation found two weak tests of mine. A source regex asserted `/objectInfoSnapshot\.record\(defs, \{/` — which `if (false) record(...)` still matches, so disabling the startup snapshot (the defect that made v1 inert) left it green. And a fixed-width slice from the `reconnecting` handler ran into `status`, which also clears. Both replaced with tests that execute the shipped code.
- The shipped `getFreshObjectInfo` and `seedObjectInfoHistory` bodies are extracted from the panel and driven directly, so the reporter's scenario runs the production path.
- The `node-def-refresh.test.mjs` diff is 17 lines. It was 1990 — that file is mixed CRLF/LF on `main` with no `.gitattributes`, so any editor normalises it and buries the real change; the additions were made as byte-preserving splices instead.

### DIFF
--- a/browser_tests/unit/add-node-socket-proof-scope.test.mjs
+++ b/browser_tests/unit/add-node-socket-proof-scope.test.mjs
@@ -204,6 +204,8 @@ function realGraphAddNode(comfy, overrides = {}) {
   const { app, LG, graph } = comfy;
   const context = { app, LG, graph, rootGraph: graph, workflow: { uuid: "wf" } };
   const calls = { single: [], whole: 0 };
+  // #1223 — what the add filed into the last-observed-schema snapshot, and with what epoch.
+  const snapshotFilings = [];
 
   const api = {
     // The whole-schema route. Counting it is how these tests prove #780's saving survives.
@@ -230,7 +232,11 @@ function realGraphAddNode(comfy, overrides = {}) {
     // #1223 — the last-observed-schema snapshot the add path files its WHOLE payload into,
     // and the connection epoch it stamps that with. Both are module scope in the real file,
     // so this rebuilt scope has to name them or the executor throws ReferenceError.
-    objectInfoSnapshot: { record: () => true, clear: () => {} },
+    //
+    // A RECORDING spy, not a stub: the add is one of the readers that must feed the
+    // fallback, and the epoch it stamps has to be the one at the WHOLE request — not the
+    // one before the optional per-class probe that may precede it.
+    objectInfoSnapshot: { record: (defs, opts) => (snapshotFilings.push({ defs, opts }), true), clear: () => {} },
     backendReconnectEpoch: 0,
     api,
     refreshComfyNodeDefs: async (defs) => app.registerNodesFromDefs(defs ?? backendObjectInfo()),
@@ -294,7 +300,7 @@ function realGraphAddNode(comfy, overrides = {}) {
      const executors = {${addNodeMatch[0]}};
      return executors.graph_add_node;`,
   );
-  return { graph_add_node: factory(...names.map((n) => deps[n])), calls };
+  return { graph_add_node: factory(...names.map((n) => deps[n])), calls, snapshotFilings };
 }
 
 // ---------------------------------------------------------------------------
@@ -681,4 +687,38 @@ test("#1180: the registration wait measures itself on the monotonic clock", () =
   assert.match(fn, /const startedAt = monotonicNow\(\);/, "the deadline must be taken on the monotonic clock");
   assert.match(fn, /while \(monotonicNow\(\) < deadline\)/, "…and the poll must read the SAME clock it was set on");
   assert.doesNotMatch(fn, /Date\.now\(\)/, "no wall-clock reading may survive in this function");
+});
+
+test("#1223: the add files a WHOLE schema, and never a single-class payload", async () => {
+  // graph_add_node is one of the readers that must feed the last-observed-schema fallback:
+  // an add that registers a new type is often the last thing to have read a whole map, and
+  // dropping it leaves the next render-time widget edit refused for want of it.
+  //
+  // UNREGISTERED type ⇒ the #780 fast path is skipped and the WHOLE schema is fetched.
+  const comfy = makeComfy();
+  delete comfy.LG.registered_node_types.SeedVR2VideoUpscaler;
+  const { graph_add_node, snapshotFilings } = realGraphAddNode(comfy);
+  await graph_add_node({ class_type: 'SeedVR2VideoUpscaler', workflow_uuid: 'wf' });
+
+  assert.equal(snapshotFilings.length, 1, 'one whole schema was fetched, so one is filed');
+  const [{ defs, opts }] = snapshotFilings;
+  assert.ok(Object.keys(defs).length > 1, 'the WHOLE map, not a single-class payload');
+  assert.equal(opts.whole, true, 'and the wholeness claim is stated');
+  assert.equal(
+    opts.observedAtEpoch,
+    opts.currentEpoch,
+    'the epoch must be read where the WHOLE request goes out — a reconnect during the ' +
+      'optional per-class probe before it would leave this stale and the record rejected',
+  );
+});
+
+test("#1223: an add that takes the single-class fast path files NOTHING", async () => {
+  // A one-entry /object_info/<Type> payload reaching the snapshot would make every other
+  // type read as absent, and the #458 ever-seen gate would diagnose the whole install as
+  // removed packs. The fast path only runs for an ALREADY-REGISTERED type.
+  const comfy = makeComfy();
+  const { graph_add_node, snapshotFilings, calls } = realGraphAddNode(comfy);
+  await graph_add_node({ class_type: 'SeedVR2VideoUpscaler', workflow_uuid: 'wf' });
+  assert.ok(calls.single.length > 0, 'the fast path ran');
+  assert.deepEqual(snapshotFilings, [], 'and nothing was filed from it');
 });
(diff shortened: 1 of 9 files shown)


artokun:
## Review outcome: NOT mergeable as written. Two independent reviews, twelve findings, two of them fatal.

Recording this before rework so the reasoning is on the record rather than buried in a force-push.

### The fix does not fire in the reported scenario

On a normal fresh tab, `seedObjectInfoHistory()` performs the only whole `/object_info` read and records only `objectInfoHistory` — `registerComfyNodeDefs()` never runs. So when the first widget edit happens while ComfyUI is rendering and both probes time out (exactly #1223), `authorize()` reports *no snapshot* and the refusal this PR exists to remove is preserved. The PR claimed to fix a case it does not reach.

### It can reopen #458 — the hole its own header promises not to reopen

`objectInfoSnapshot.record(defs, backendReconnectEpoch)` reads the epoch **after** the await, so a payload fetched before a restart gets stamped with post-restart provenance.

The clear that was supposed to prevent this never runs. The `reconnected` handler calls `refreshComfyNodeDefs()` with no payload and no `force`, and `makeRefreshCoalescer` handles that case with `await current; return;` — `runRegister` is never invoked, so neither `objectInfoCache.invalidate()` nor `objectInfoSnapshot.clear()` executes for that reconnect. Verified in `web/js/lib/refresh-coalesce.js:50-58`.

Sequence: a forced refresh starts → user uninstalls a pack and restarts ComfyUI → `reconnected` bumps the epoch and joins the in-flight run without clearing → the in-flight fetch resolves with the **pre-restart** schema and is recorded under the **new** epoch → the next silent-probe write authorizes a type the running backend no longer defines.

The same mechanism defeats the burst cache's generation guard: `object-info-cache.js` deliberately refuses to *store* a result issued before an `invalidate()`, but still returns it to its waiter. The set_widget record site takes that retired answer and promotes it into a store with no TTL at all.

### The rest

- **Stored by reference.** `registerNodesFromDefs` runs `beforeRegisterNodeDef` hooks that mutate defs in place (Comfy's upload hook adds an input the backend never declared), so the snapshot would later return frontend-mutated data as backend evidence.
- **A client resolving `null`/`undefined` is tagged `ANSWERED_UNUSABLE`**, contradicting the oracle's own stated semantics five lines above ("Only a client that returned NOTHING … leaves the question unanswered"). On any frontend whose `getNodeDefs` swallows its failure, the fallback is dead.
- **A proxy 504 disqualifies silence.** ComfyUI behind nginx/Caddy (the standard remote/RunPod shape) answers a hung backend with a gateway timeout — the fix would never fire there, and the refusal would blame the backend for the proxy's synthetic error.
- **The snapshot's ineligibility reason is spliced into `failures`**, so a two-transport failure reports "Tried **3** routes". That is the #982 defect the surrounding comments exist to prevent, and my own test locked it in.
- **Snapshot defs reach `refreshCombos`**, which does `w.options.values = first.slice()` — rewriting the live dropdown *backwards* to the older list, persistently.
- **`recordsWholeSchemaOnly` does not exist.** I wrote a backticked identifier into two comments claiming a test pins it; the actual pin is an untitled source-regex test. The next reader greps for it and concludes the guard was deleted.
- **The single-class `/object_info/<Type>` payload can reach the whole-schema path** across an await in `assertAddNodeResolvableRefreshing`, and `record()` does not reject it.
- **1958 lines of line-ending churn** in `node-def-refresh.test.mjs` hide the 32 real added lines (`--ignore-cr-at-eol` shows 32/6). Direction is the opposite of what one review said — `origin/main` is LF, this commit stored CRLF, and there is no `.gitattributes`.

### Direction for v2

Store a **detached** representation built from the observed key names rather than retaining the payload, stamp it with the epoch captured **at fetch issuance** and reject the record if the epoch moved, clear it from the socket handlers directly instead of depending on the coalescer reaching `registerComfyNodeDefs`, seed it from the startup read, and stop letting gateway-class statuses and a nothing-returning client count as "the backend answered".

PR #1236, opened 2026-08-14
fix(1569): a widget write that the queue will not honour is not a success

artokun:
**Draft — claiming artokun/comfyui-mcp#1569.** A red test is in; the fix follows.

## What is NOT the cause

The obvious guess is that `panel_set_widget` assigns `widget.value` and skips the widget's
callback, leaving a custom node's derived state untouched. **That is not what happens.**
`applyWidgetWrite` assigns the value *and* fires the widget's own callback — the same single
invocation a manual UI edit performs. I checked before writing anything.

## What is

The **verification** step. After writing, it compares against `w.value` — which is what the
canvas shows. ComfyUI does not queue `w.value`: `graphToPrompt` asks the widget for
`serializeValue()` when it defines one, and a custom node that keeps its authoritative state
elsewhere derives that from the state rather than from the assignment.

So the two diverge, and a check that only ever reads one of them cannot see it. That is
exactly the reported shape:

> Tool reports success and graph query shows new strings, but queued workflow uses stale
> custom-widget state.

## What this can and cannot fix

It cannot make `Ideogram4PromptBuilderKJ` honour the assignment — that is the node's own
design. What it can stop is **reporting success for a write that will not take effect**,
which is the difference between re-setting the widget three times and knowing immediately to
wire a `CLIPTextEncode` instead (which is what the reporter eventually did).

## The constraints, already pinned

Three of the five tests pass today, and they are what keeps the fix from being worse than the
bug:

- a widget with **no** `serializeValue` must be untouched — that is nearly all of them;
- a `serializeValue` that **agrees** must stay an ordinary success;
- a **throwing** `serializeValue` must not fail a write that otherwise worked. The check is a
  diagnostic, not a new gate.

## Status

- [x] Red test reproducing the divergence, plus the three constraints above
- [ ] Compare the written value against what will actually be serialized
- [ ] A message naming both values and saying the canvas moved while the queue did not
- [ ] Mutation pass, codex review, live verification on the rig

Refs artokun/comfyui-mcp#1569


artokun:
**Closing without a code change.** Three designs, three review rounds, three kills — each on
evidence from a real upstream node. Recording the outcome rather than shipping the last one
standing, because every version was worse than the silence it replaced.

## What is actually wrong (this part is solid)

`panel_set_widget` reported success while the queued prompt used stale state. The obvious
cause — that the panel assigns `widget.value` and skips the widget's callback — **is not it**.
`applyWidgetWrite` already fires the callback, exactly as a manual UI edit does.

ComfyUI does not queue `w.value`. `graphToPrompt` asks the widget for `serializeValue()` when
it defines one, and a node that keeps its authoritative state elsewhere builds that from the
state. So the write lands on the widget, `panel_query_graph` shows the new strings, and the
run sends the old ones.

## Why nothing shipped

**Attempt 1 — fail when the serialized value did not move between before and after.**
Killed: calling `serializeValue()` has **side effects**. Upstream's `RecordAudio` stops an
active recording and starts an upload from inside its serializer; a before/after comparison
does that twice, to satisfy a diagnostic. And it is `async` there, so every call returns a
fresh Promise — the two samples are never comparable. A check that can stop a user's
recording, or fail a write that worked, is worse than the bug.

**Attempt 2 — disclose on success when the widget defines its own serializer.**
Killed: ordinary `CLIPTextEncode` prompt widgets install `serializeValue` via the
dynamic-prompts extension. The note would fire on the most common write in the product — every
prompt edit — and drown out `write_warning`, which means something actually went wrong.

**What that leaves.** Detecting the divergence requires invoking the serializer, and invoking
the serializer is precisely the thing that must not happen. I could not find a discriminator
that separates the reported prompt-builder from an ordinary prompt field without calling it.

## What I am doing instead

The diagnosis goes on the issue, which stays open, so this is not re-investigated from
scratch. The reporter's workaround — wiring a `CLIPTextEncode` into the sampler's conditioning
— is correct and remains the answer for that node.

If a discriminator turns up (a marker on the widget, an upstream convention for
state-derived serializers, or a post-queue comparison in `panel_run`, where the submitted
prompt is already known), this becomes shippable. None of those exist today.

Refs artokun/comfyui-mcp#1569

Issue #805, opened 2026-08-08
panel_set_widget errors after integer widget normalizes requested value by one


artokun:
### What happened / root cause
panel_set_widget mutates an integer widget to a frontend-normalized value, then reports failure because the retained value differs from the request by one. On OllamaTextDescriber.max_tokens, requesting 4096 retained 4097; requesting 8192 retained 8193. The workflow mutation did occur, so the error is misleading and leaves outcome handling awkward.

### Steps to reproduce
1. Open a live workflow containing OllamaTextDescriber.
2. Call panel_set_widget for widget "max_tokens" with value 4096 (or 8192).
3. Observe the widget becomes 4097 (or 8193) and the tool returns an error instead of a normalized-success result.

### Exact error (scrubbed)
Widget "max_tokens" on node 1 (OllamaTextDescriber) did not retain the requested value: wrote 4096 but it became 4097.
Widget "max_tokens" on node 4 (OllamaTextDescriber) did not retain the requested value: wrote 8192 but it became 8193.

### Fix
applied locally: no
upstream-only: yes
Precise change needed: panel_set_widget should distinguish normal frontend/node widget normalization (clamp/step/canonicalization) from a failed write. When the widget reports a deterministic normalized value after the set, return success with requested_value, actual_value, and normalized:true, or pre-normalize against widget config before comparing. It should not return an outcome-ambiguous failure after the mutation has already applied.

Temporary workaround: accept the retained odd value; it is functionally equivalent for this workflow.

### Environment
Windows 11; ComfyUI 0.31.0; RTX 5080 16 GB; comfyui-mcp 0.50.36; panel 0.11.44.

artokun:
Fixed on main in #806 (`3c49b85`) — thank you for the two data points, they were enough to pin the rule exactly.

Your diagnosis was right: the write applied, and the verification step was a strict `actual === expected`, so ordinary widget quantization was reported as a failed write.

Worth naming why this direction is the nastier one. The surrounding code exists to stop the tool claiming success for work that did not happen; this was the mirror — claiming failure for work that **did**. And "did not retain" reads as transient, so the natural response is a retry, which normalizes identically every time. Both your nodes looked broken for the same reason.

**The rule.** Only a value the widget's own declared config explains *exactly*:

```
snap(v) = clamp(min + round((v - min) / step) * step, min, max)
```

[...]

PR #1241, opened 2026-08-14
feat(757): panel_set_widget can create an rgthree lora_N slot

### LINKED ISSUE #757: panel_set_widget cannot create dynamic widget slots (Power Lora Loader rgthree lora_N)

artokun:
Closes #757.

> **Status: draft, with a two-round cap.** After round 6 the user decided this PR gets **two more review rounds** (7 and 8). If round 8 still finds a creating-path-only defect, it is parked as a draft. Round 7 is done; **this is the last round before that call.**
>
> **A prediction in this PR was falsified, and the record should say so.** After round 5 I predicted that further findings would be about the *shared* write path rather than creation-only. Round 6 found a creating-path-only defect, and round 7 found another. The correct model is the one the reviewer proposed: **a newly created row has a settling period an existing row does not**, and both defects lived inside it. See round 7 for what that means and what it costs.

A freshly added `Power Lora Loader (rgthree)` carries only `divider, PowerLoraLoaderHeaderWidget, divider, ➕ Add Lora`. The `lora_1`, `lora_2`, … rows do not exist until the user clicks **➕ Add Lora**, a DOM-only control an agent cannot activate — so every write to `lora_1` on a new node was refused for a widget that no tool could bring into existence. Writing an *existing* row already worked, so this is the creation half only.

## Read this first: four rounds of review, one root cause

Every serious defect this PR has had came from the **same seam**: the row was created *before* `runSetWidget`, which yields at `await getFreshObjectInfo()` on a cache miss. A live graph mutation therefore sat in the graph across a network request, and each round found a different thing that could happen in that window:

| Round | What the window allowed |
| --- | --- |
| 1 | a refused write left the row behind |
| 2 | an undo transaction held across it could **wedge ComfyUI's ChangeTracker for the whole session** |
| 2 | a concurrent frame's write could be rolled back under it |
| 3 | a **user's own hand-edit** during the window could be rolled back by an unrelated failure |

Rounds 1 and 2 each added a guard. **Round 3 removed the window instead**, and the guards with it. That is the shape change worth reviewing: creation now happens at `runSetWidget`'s *synchronous write boundary*, so no `await` separates the creation, the write and the rollback, and nothing — no other command frame, no user gesture, no undo capture — can run in between.

**Round 4 found no P1s** and confirmed the mutation-ordering class closed. **Round 5 found one P1 — caused by round 4's own fix** — and it settles a question worth stating up front:

> A refused write is **not a no-op in the undo history**, and cannot be made one. `applyWidgetWrite` captures its write and then captures its rollback in a second envelope of its own, so an *ordinary* refused write already leaves two snapshots a Ctrl+Z can step back into. Round 4 asked the creating path to do better; the only way to do better is to hold one envelope across the verification, and that moves the capture — and every pack `serializeValue` the capture runs — past the last check.

Round 4 chose the better history and lost the verification. Round 5 restores the verification and accepts the same history every other write has. **Symmetry with the ordinary write is the invariant**; being better than the shared write path was never on offer.

## How the route is kept away from the typo guard

The panel deliberately refuses to auto-press a pressable control (`pressable-widget.js`), because a generic "this node has one button, press it" rule would mutate the graph on an ordinary **typo** — overwhelmingly the most common reason a widget name misses. So creation is keyed on three independent facts, all required:

1. the node **type** is `Power Lora Loader (rgthree)`;
2. the requested name matches `lora_<n>` and is **absent** from the node;
3. the **value** is a lora-slot the existing writer already accepts.

A typo cannot satisfy all three, and `pressableWidgetHint` stays the answer for every other node and every other missing name.

`addNewLoraWidget` is pack-private and version-dependent, so it is feature-detected (a loud, actionable refusal when absent) and its **effect** is verified by comparing the widget list afterwards — the maintainer's live probe found generic callbacks (`callback`, `onMouseClick`, `mouseClickCallback`) that accept the call and create nothing.

---

### 1. (P1) Create the row at the synchronous post-authorization write boundary

**The defect.** Creating before `runSetWidget` put the mutation on the far side of an awaited `/object_info` fetch. A user edit landing in that window is captured by ChangeTracker together with the transient row (splitting create-and-assign across undo steps), and if the request then fails, the rollback can delete a row **the user edited by hand**. Losing a manual edit is worse than the bug this feature fixes.

**The fix.** `runSetWidget` takes an injected `prepareWriteTarget` hook and calls it inside `write()` — after the workflow fence, immediately before `applyWidgetWrite`, with the rollback in the same `try/catch`:

```js
assertTargetStillCurrentNow();
const prepared = typeof prepareWriteTarget === "function" ? prepareWriteTarget() : null;
try {
  return applyWidgetWrite(node, widgetName, value, { … });
} catch (err) {
  try { prepared?.undo?.(); } catch { /* never replace the refusal */ }
  throw err;
}
```

`applyWidgetWrite` is synchronous, so the create → write → rollback triple is uninterruptible. The stale-combo and upload-asset retries re-enter `write()`, and each attempt rolls its own preparation back **before** the await that follows — so no half-made target spans `refreshCombos` or the asset probe either.

**What this deletes.** The `rowClaims` `WeakMap`, `noteRgthreeLoraRowWritten`, the panel's `assertActiveWorkflowCommandTarget` call on the creation path (the fence is now `runSetWidget`'s own, re-checked at the boundary), and the panel's `try/catch/undo` around the call. Net **−131 lines** in the panel.

### 2. (P2) An unknown row counter is an *incomplete* rollback, not a clean one

When a pack exposes no readable `loraWidgetsCounter`, `counterBefore === null` was treated as "nothing changed". But the shipped `addNewLoraWidget` **increments first and constructs second**, so a throw has probably already spent a row name — the caller was told a retry was safe when a name had in fact been consumed. Only a counter we can *read* and that now *reads back unchanged* licenses "nothing was changed". Both refusal branches made the unconditional claim (the pack-throw path and the added-no-widget path below it); both now disclose it.

---

### 1. (P2) The cleanup has to happen inside the undo envelope

**The defect.** `applyWidgetWrite` closes its own envelope *before* the #240 read-back verifies, and rolls a rejected value back inside a second envelope of its own. Every close that reaches zero makes ChangeTracker capture. So a **post-assignment** refusal — a widget setter that normalizes the value away, so the read-back rejects what was written — captured a snapshot holding the created row and the rejected value, and the cleanup then removed the row with nothing open and nothing captured. The tracker's newest snapshot was a graph that no longer existed, and the next Ctrl+Z restored the command that had just been refused. A refusal is supposed to be a no-op in the undo history too.

> **Superseded in round 5.** The mechanism below — one outer envelope around the whole triple — is what produced round 5's P1. Only the *cleanup* is bracketed now. The finding was half right, and the half that survives is "the newest capture must describe the graph that exists"; see round 5.

**The fix as it stood in round 4.** The write and its cleanup share ONE outer envelope, so the inner closes never reach zero and exactly one capture happens — at the end, over the graph as it really finished:

```js
const prepared = typeof prepareWriteTarget === "function" ? prepareWriteTarget() : null;
if (!prepared) return doWrite();          // nothing grown: bookkeeping unchanged
beforeChange?.();
try {
  try { return doWrite(); }
  catch (err) { /* undo + disclose */ throw err; }
} finally { afterChange?.(); }
```

**This is not the round-2 defect returning.** That one held a transaction across an `await`, where a workflow switch could detach the canvas before the close and wedge the tracker at `changeCount === 1` for the session. Everything inside this bracket is synchronous and cannot yield — the property moving the creation here established in the first place. A mutation that reopens the envelope across a deferred cleanup is in the suite.

**One deliberate deviation from the review's wording.** The review asked for preparation, application and cleanup in one envelope; the envelope here opens once there *is* something to protect, leaving the preparation just outside it. Opening first would bracket a graph that never changed whenever the preparation **refuses** (a pack with no `addNewLoraWidget`), and whether an unchanged capture becomes a no-op undo entry is ComfyUI's business — not something this file can verify from here. So a refusal that changed nothing keeps doing no bookkeeping at all, which is an already-tested property. Nothing is lost: the capture is a whole-graph snapshot taken at the close, so it contains the created row regardless of which side of the open it was minted on, and no capture can occur in between because only a close that reaches zero captures.

### 2. (P2) A refusal must report a row name it could not give back

This is the residual the previous round flagged and deliberately did not fix; the reviewer found it independently, which is the mandate to do it properly.

A pack that exposes `addNewLoraWidget` but **hides or freezes** `loraWidgetsCounter` let a slot-shaped request with an invalid field create a row, be rejected by `applyWidgetWrite`, and have the row taken back out — while the **name stayed spent**. `remove()` still reported plain success, so the caller heard only that its value was invalid. Retrying the corrected value under the same `lora_N` then minted a *later* row and was refused a second time on the name — the same unfollowable advice the mismatch refusal already had to stop giving.

`remove()` now returns `{ removed, incomplete }`. Whether the counter was restored is decided by **reading it back off the node**, not by trusting `restoreRowCounter`'s return — that function reports `false` both when it could not write and when there was nothing to undo, so its return cannot tell those apart, and a frozen counter it "restored" reports `true` while nothing moved. `runSetWidget` appends the disclosure to the refusal **in place**, so the error keeps its type and its `combo` / `emptyOptions` flags and a recoverable stale-combo miss still retries.

---

### 1. (P1) Verify the creating write the same way an ordinary write is verified

**The defect.** `applyWidgetWrite` verifies **after** its own `afterChange` has fired — deliberately, and the source says why: *"an afterChange hook can itself re-stale a widget or change the promotion topology, and that must be caught too"*. That close is also when ComfyUI serializes the graph for the undo capture, and serialization runs every node's own `serializeValue`.

Round 4's outer envelope stopped `applyWidgetWrite`'s close from reaching zero. The capture — and every `serializeValue` it runs — therefore moved *past* the last verification. An rgthree loader in **Separate Model & Clip** mode rewrites `strengthTwo: null` to `1` exactly there, so the creating path reported plain success for a value an existing-row write would have reported as normalized (#805).

**The fix.** `applyWidgetWrite` is called with nothing wrapped around it, exactly as every other write path calls it. Only the cleanup is bracketed:

```js
const prepared = typeof prepareWriteTarget === "function" ? prepareWriteTarget() : null;
try {
  return applyWidgetWrite(node, widgetName, value, { … });   // verifies after its own capture
} catch (err) {
  if (prepared) {
    safeHistoryHook(beforeChange);                            // the cleanup gets its own envelope
    try { /* undo + disclose */ } finally { safeHistoryHook(afterChange); }
  }
  throw err;
}
```

The test for it is pack-independent: a history probe that rewrites the widget value **at capture depth**, which is where a `serializeValue` runs. Re-nesting the write in an outer envelope is in the mutation set and is killed by it.

### 2. (P2) Exception-safe history hooks around the cleanup

The cleanup's envelope is opened by the graph's own hook, and a pack or extension can make that throw. An escaping **open** skipped the cleanup entirely — the row *and* the row name it spent left behind while the caller got an error, i.e. mutate-then-refuse caused by the bookkeeping rather than the write. An escaping **close** could replace the refusal the caller actually needs to read. Both are now routed through `safeHistoryHook`, mirroring `widget-write.js`'s own `safeBefore`/`safeAfter`.

### 3. (P2) Confirm the mismatched-row counter actually rewound

`restoreRowCounter` reports whether it ran an assignment, not whether the assignment **took**. A counter that accepts `++` but silently ignores writes back — a getter-only property, a clamping pack, a proxy — was advertised as rewound, and the remedy the refusal printed (*"set `lora_2` instead"*) only works if the number really came back. The mismatch branch now reads the counter back off the node, the same rule the two round-3 branches and `remove()` already use.

---

## Review round 7 — the settling period

Rounds 6 and 7 both found **creating-path-only** defects, which is what falsified the symmetry prediction. The model that explains both:

> A row created by the pack's own **➕ Add Lora** button is unsettled for exactly as long as it takes the canvas to repaint — microseconds, and always before a human can touch it. A row created by an agent is written to **in the same synchronous command**, with no frame in between. Anything the pack defers to first-draw is therefore still unset when our write is verified.

### 1. (P1) Initialize the new row's strength mode before the capture

Read from the shipped pack (`rgthree-comfy/web/comfyui/power_lora_loader.js`, located on the dev machine rather than inferred):

```js
constructor:    this.showModelAndClip = null;
draw:           this.showModelAndClip =
                  node.properties["Show Strengths"] === "Separate Model & Clip";
serializeValue: const v = { ...this.value };
                if (!this.showModelAndClip) { delete v.strengthTwo; }
                else { this.value.strengthTwo = this.value.strengthTwo ?? 1; … }
```

`null` is **falsy**, so an undrawn row serializes down the *first* branch and silently drops `strengthTwo` instead of normalizing it. The write therefore verified clean — nothing had touched the value — and the first draw afterwards flipped `showModelAndClip` to `true`, at which point `serializeValue` rewrote `strengthTwo: null` to `1`. A verified success that changes at the next queue or save.

Creation now performs the same synchronisation the first draw would, and **only** that: the value adjustment sitting beside it in `draw()` is skipped, exactly as the pack skips it when `oldShowModelAndClip` is `null`. A created row is byte-identical to the same row created by the button and drawn once.

The test is the invariant, not the mechanism: **a Separate Model & Clip creation must report what an existing-row write reports.** Both now refuse with "did not retain the requested value", because in that mode the pack overrides `strengthTwo` for either path. Symmetry is the pass condition — not success.

### 2. (P2) Roll back when the tail throws

The helper's tail mutates a row that is already on the node while the caller still holds no handle to undo it: `remove` is only returned at the very end, and `runSetWidget` evaluates the whole preparation *before* entering the try that would clean up. A throw there — a hostile `computeSize`, a raising `setDirty`, a widget rejecting the mode assignment — stranded the row, the name it spent and the height it grew, while the command reported a failure that had changed the graph. The tail now undoes itself before rethrowing, and `setDirty` is best-effort, matching the resize beside it.

### What this costs, stated plainly

The settle is **one named field, mirroring one line of one pack's `draw`**. It is correct and verified against the shipped source, but it is not a general mechanism: `settleCreatedRow` is the place where the *next* draw-deferred field would go, not a guarantee that none exists. The general form of this problem — "make the row fully settled before anything reads it" — has no synchronous solution, because settling is defined by a canvas frame that a single command cannot wait for without reintroducing the mutation-across-await class that rounds 1–3 closed.

The shape that removes the settling period entirely is **creation as its own committed step**, with the value write issued afterwards as an ordinary existing-row write. That is a tool-surface decision, not a bug fix, and it is the honest alternative if round 8 finds a third field.

---

## Rounds 1 and 2 (for history)

<details>
<summary>The nine earlier findings, and which of them round 3 superseded</summary>

**1. (P1) The route was dead code — values arrive as JSON strings.** `coerceWidgetValue` turns the string into an object at `widget-write.js:508-512`, *after* the classifier ran, so `isLoraSlotObject(rawValue)` answered "not a slot" for every real request. `slotValue()` now parses a JSON string before the shape test; `isLoraSlotObject` remains the only judge, so nothing about what counts as a slot is widened.

**2. (P2) The monotonic counter made the refusal's remedy unfollowable.** `addNewLoraWidget()` increments before it names, and removing the row does not undo the increment — so `set "lora_7" instead` had already moved the next name to `lora_8`. Undoing means undoing both halves; when the counter cannot be rewound the message names the state instead of promising a retry that would not work.

**3. (P1) A refused write left the row behind.** `createRgthreeLoraRow` returns `{ created, remove }`. *Round 3 moved the caller of `remove()` from the panel into the write boundary.*

**4/5. (P1) The transaction must not span the await across canvas detachment.** Round 1's fix — bracketing creation and write in one `graph.beforeChange()`/`afterChange()` pair — was worse than the bug it fixed. LiteGraph delivers those hooks through *attached* canvases (`canvasAction(cb) { … for (const c of this.list_of_graphcanvas) cb(c) }`), so a tab switch mid-await leaves the close reaching nobody and ChangeTracker stuck at `changeCount === 1` for the rest of the session. The creation is not bracketed at all; one undo step still results, because ChangeTracker captures whole-graph *snapshots* and the pair `runSetWidget` opens for the write closes with the row already present.

**6. (P1) A rollback must not delete a row another request has written.** Round 2 added a claim (`WeakMap` keyed on the widget object) so a rollback only removed a row nobody else had written. **Round 3 deleted this entirely** — with create and write in one synchronous stretch, the interleaving it defended against is unreachable, so exactly one of two overlapping requests can create. Do not look for `rowClaims`; it is gone.

**7. (P2) Resize the loader after adding a row.** rgthree's own button resizes right after calling `addNewLoraWidget`; marking the canvas dirty only repaints. Height only — widening a node the user sized is not ours to do — and the prior size is restored on rollback.

**8. (P2) Do not rewind a counter advanced by later edits.** Create `lora_1`, add `lora_2`, roll back, and an unconditional rewind mints `lora_1` and a **duplicate** `lora_2`. The rewind requires the counter to still read exactly this call's increment.

**9. (P2) Roll back partial effects when the pack method throws.** The exception path removes whatever landed, rewinds the counter and the size, and says plainly when it could not undo it all.

</details>

---

## Verification

Both mechanisms were read out of the shipped sources rather than reasoned about: `rgthree-comfy/web/comfyui/power_lora_loader.js` for the increment-then-append order and the resize step, and the ComfyUI frontend bundle for `LGraph.beforeChange` → `canvasAction` → `ChangeTracker`.

Testing is at three levels, deliberately:

- the **shipped `graph_set_widget`** is extracted from the panel and run against doubles (the `graph-edit-node.test.mjs` pattern), because source-text assertions cannot answer "does a refused write leave the row behind" or "is this one undo step";
- the **real `runSetWidget`** is driven directly (the `set-widget-refresh.test.mjs` pattern) for the new hook — the executor tests stub `runSetWidget`, so on their own they verify the panel's half of the contract against a *model* of the other half that could drift from it;
- source-text assertions pin the two structural rules (no `await` in the write stretch; the panel reaches `createRgthreeLoraRow` only from inside the hook).

Results:

- `npm run test:unit`: **4489 tests, 4488 pass, 0 fail, 1 todo** (i18n, i18n-render, tool-vocabulary and `check-panel-scope` gates all green).
- **59 mutations, every one killed.** Round 7 added the P1 itself (*the new row is never settled* — killed by the asymmetry test, which reports `creation ok=true where an existing row reported ok=false`), *the settle reads the mode backwards*, *the settle reads the wrong node property*, *the settle invents the field on a build that never had it*, *an unreadable node fails the creation*, *a throw in the tail strands the row*, and *setDirty stops being best-effort*. Round 5 added the P1 itself (*the write is nested in an outer envelope again*), both P2s, and the two unguarded-history-hook variants.
- **One mutation was deliberately removed rather than killed.** Closing and reopening the cleanup's envelope — adding an extra intermediate capture — *survives*, and correctly so: the only property claimed is that the newest capture describes the graph that exists. Killing it would mean asserting the minimal-history property that round 5 established is unavailable to any write path. A mutation kept alive by a pretend assertion is worse than no mutation.
- Earlier rounds: Round 3 added: *the target is minted before the `/object_info` await again* (killed behaviourally — nothing may be minted on a stale canvas, and nothing may survive across the retry await); *minted before the workflow fence*; *the boundary drops its own fence re-check*; *the rollback is deferred out of the synchronous stretch* (`setTimeout`, invisible to the source-text assertions); *an unknown counter is called a clean rollback again*, once per branch; and *a genuinely clean rollback is never reported as clean*, so the honest message stays reachable. Round 4 added both P2s themselves plus *the envelope closes before the cleanup*, *the envelope never closes* (the round-2 defect, as a mutation), *the disclosure is thrown as a NEW error, losing the combo retry*, and *the panel swallows the disclosure*.
- **One round-4 mutation LIVED on its first run** — *the panel swallows the row-creation's disclosure* — exposing a real gap: nothing connected the panel's hook to the text the caller actually reads, because the executor tests stubbed the layer that does the joining. Closed with an end-to-end executor test over a counter-hiding pack, and the stub now models the annotation as well.
- The harness restores from an in-memory snapshot rather than `git checkout` (the work was uncommitted), and treats a needle matching anything other than exactly once as a **survivor** — on this CRLF checkout an LF-authored multi-line anchor that silently missed would otherwise read exactly like a kill.

## Notes for the reviewer

- **Is the P1 class closed?** For this feature, yes, and by construction rather than by guard: there is no yield point between the fence, the creation, the write and the rollback, and the retry paths roll back before their awaits. The mutation that puts the creation back before the fetch is killed by two *runtime* tests, not only by the source-text rule. Round 4 reviewed this path and found no P1s.
- **The residual flagged in round 3 is now fixed.** A rollback that cannot put everything back — a surviving row, or a spent row name — is disclosed on the refusal itself. The cross-layer signal it needed is a single optional string returned by the hook's `undo()`, which `set-widget.js` appends without knowing anything about lora rows.
- **The undo-history question is settled, and the answer is "same as every other write".** A refused creating write leaves intermediate captures, exactly as a refused ordinary write does; what is guaranteed is that the *newest* capture describes the graph that exists. Making it a true no-op requires holding an envelope across the verification, which is what round 5's P1 was.
- **Where the remaining risk actually sits — corrected.** After round 5 this section predicted that further findings would be about the *shared* write path. That was wrong: rounds 6 and 7 were both creating-path-only, and both sat in the newly created row's **settling period** — the window between `addNewLoraWidget()` and the first canvas draw, in which pack state the draw is responsible for is still unset. Round 7 closes the one field that affects serialization (`showModelAndClip`, verified against the shipped pack source). It does not prove there is no second field, and no synchronous mechanism can, because settling is defined by a frame this command cannot wait for.
- **The alternative shape, if round 8 finds a third.** Make creation its own committed step and let the caller issue the value write as an ordinary existing-row write. That removes the settling period by construction rather than fixing its symptoms, at the cost of two calls and a refused write leaving an empty row behind. It is a tool-surface decision, and the evidence for it is now two rounds of findings rather than a hunch.
- **Disclosure.** The reply carries `created_widget` on a field of its own, not in `warning`: that channel is single-slot and priority-ordered, so appending would displace a warning about what the write actually does.
- **Shape.** `prepareWriteTarget` is a generic "the target may not exist yet" hook, and the only implementation is rgthree's. That is deliberate: it is the same shape `reconcileUnknownWidgetNames` already has in this file (mutate the node so the caller's name resolves, post-authorization, behind the fence). If a *second* growable target ever wants this hook, that is the moment to promote it to an explicit `panel_add_widget_slot` the caller invokes deliberately rather than a side effect of a write.
- **Scope.** Keyed strictly to the one node type. `panel_remove_widget` (comfyui-mcp#938) already removes these same rows by the mirror-image mechanism.

### DIFF
--- a/browser_tests/unit/rgthree-lora-row.test.mjs
+++ b/browser_tests/unit/rgthree-lora-row.test.mjs
@@ -1,0 +1,217 @@
+/**
+ * #757 — `panel_set_widget` could not CREATE an rgthree Power Lora Loader row.
+ *
+ * The rows exist only after the user clicks "➕ Add Lora", a DOM-only control an agent
+ * cannot activate, so every write to `lora_1` on a fresh node was refused for a widget no
+ * tool could bring into existence.
+ *
+ * The load-bearing constraint is what this must NOT become. The panel deliberately refuses
+ * to auto-press a pressable control, because a generic "press this node's button" rule
+ * would mutate the graph on an ordinary TYPO — the overwhelmingly common reason a widget
+ * name misses. Most of these tests exist to pin that the route cannot be reached by
+ * anything but the real case.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import {
+  isRgthreeLoraRowCreation,
+  createRgthreeLoraRow,
+  POWER_LORA_LOADER_TYPE,
+} from "../../web/js/lib/rgthree-lora-row.js";
+
+const SLOT = { on: true, lora: "x.safetensors", strength: 0.5, strengthTwo: null };
+
+/** A Power Lora Loader as it looks fresh from panel_add_node: no rows yet. */
+function loader({ counter = 1, addNew = true, widgets = null } = {}) {
+  const node = {
+    id: 153,
+    type: POWER_LORA_LOADER_TYPE,
+    widgets: widgets ?? [
+      { name: "divider" },
+      { name: "PowerLoraLoaderHeaderWidget" },
+      { name: "divider" },
+      { name: "➕ Add Lora" },
+    ],
+    removeWidget(w) {
+      const i = node.widgets.indexOf(w);
+      if (i >= 0) node.widgets.splice(i, 1);
+    },
+  };
+  if (addNew) {
+    // rgthree's real behaviour: a MONOTONIC counter, so names are not positional.
+    node.addNewLoraWidget = () => {
+      node.widgets.push({ name: `lora_${counter++}`, value: { on: true, lora: null, strength: 1, strengthTwo: null } });
+    };
+  }
+  return node;
+}
+
+// ---------------------------------------------------------------------------
+// The classifier — three independent facts, all required
+// ---------------------------------------------------------------------------
+
+test("#757 the reported case classifies: right type, lora_N name, slot-shaped value", () => {
+  assert.equal(isRgthreeLoraRowCreation(loader(), "lora_1", SLOT), true);
+});
+
+test("#757 a TYPO does not reach the creation route", () => {
+  // The whole reason the panel refuses to auto-press a button. `strenght` or `lora1` or
+  // `seed` must all get the ordinary refusal (and its pressable hint), never a new row.
+  for (const name of ["lora1", "loras_1", "lora_", "seed", "strenght", "LORA_1", ""]) {
+    assert.equal(isRgthreeLoraRowCreation(loader(), name, SLOT), false, `name ${JSON.stringify(name)}`);
+  }
+});
+
+test("#757 another node type never reaches it, even with a lora_N name and a slot value", () => {
+  const other = { ...loader(), type: "LoraLoader" };
+  assert.equal(isRgthreeLoraRowCreation(other, "lora_1", SLOT), false);
+  const noType = { ...loader(), type: undefined, comfyClass: undefined };
+  assert.equal(isRgthreeLoraRowCreation(noType, "lora_1", SLOT), false);
+});
+
+test("#757 a value that is not a lora slot never reaches it", () => {
+  // A slot is minted to receive a row. Growing the node for a value the writer would then
+  // refuse leaves a stray row behind and reports a failure — worse than refusing up front.
+  for (const value of [null, undefined, 5, "x.safetensors", [], { on: true }, { on: true, lora: "a", strength: 1, extra: 1 }]) {
+    assert.equal(isRgthreeLoraRowCreation(loader(), "lora_1", value), false, `value ${JSON.stringify(value)}`);
+  }
+});
+
+test("#757 an EXISTING row is left to the ordinary write path", () => {
+  const n = loader();
+  n.widgets.push({ name: "lora_1", value: { on: false, lora: null, strength: 1, strengthTwo: null } });
+  assert.equal(isRgthreeLoraRowCreation(n, "lora_1", SLOT), false, "minting over it would duplicate the row");
+});
+
+test("#757 the classifier is total — a hostile node answers false, never throws", () => {
+  const hostile = {
+    get type() {
+      throw new TypeError("disposed");
+    },
+  };
+  assert.doesNotThrow(() => isRgthreeLoraRowCreation(hostile, "lora_1", SLOT));
+  assert.equal(isRgthreeLoraRowCreation(hostile, "lora_1", SLOT), false);
+});
+
+// ---------------------------------------------------------------------------
+// Creation, and the post-verify that makes a pack-private call safe
+// ---------------------------------------------------------------------------
+
+test("#757 the reported case: the row is created and named", () => {
+  const n = loader();
+  const events = [];
+  const r = createRgthreeLoraRow(n, "lora_1", {
+    beforeChange: () => events.push("before"),
+    afterChange: () => events.push("after"),
+    setDirty: () => events.push("dirty"),
+  });
+  assert.deepEqual(r, { created: "lora_1" });
+  assert.ok(n.widgets.some((w) => w.name === "lora_1"), "the row now exists for the write that follows");
+  assert.deepEqual(events, ["before", "after", "dirty"], "the mutation is bracketed for undo");
+});
+
+test("#757 a pack without addNewLoraWidget refuses LOUDLY, and changes nothing", () => {
+  // Feature detection, as ltx-director.js does for its own pack-private entry point. A
+  // renamed or dropped method must produce an actionable refusal, never a silent no-op.
+  const n = loader({ addNew: false });
+  const before = n.widgets.length;
+  assert.throws(() => createRgthreeLoraRow(n, "lora_1", {}), /does not expose addNewLoraWidget/);
+  assert.equal(n.widgets.length, before);
+});
+
+test("#757 a call that adds NOTHING is caught — the effect is verified, not the call", () => {
+  // The probe that motivated this file found pack callbacks that accept a call and create
+  // nothing. Only comparing the widget list catches that.
+  const n = loader();
+  n.addNewLoraWidget = () => {};
+  assert.throws(() => createRgthreeLoraRow(n, "lora_1", {}), /ran but added no widget/);
+});
+
+test("#757 a pack method that THROWS is attributed to the pack", () => {
+  const n = loader();
+  n.addNewLoraWidget = () => {
+    throw new Error("rgthree exploded");
+  };
+  assert.throws(() => createRgthreeLoraRow(n, "lora_1", {}), /rgthree pack's own addNewLoraWidget\(\) threw \(rgthree exploded\)/);
+});
+
+test("#757 afterChange still runs when the pack method throws", () => {
+  const n = loader();
+  n.addNewLoraWidget = () => {
+    throw new Error("boom");
+  };
+  const events = [];
+  assert.throws(() =>
+    createRgthreeLoraRow(n, "lora_1", {
+      beforeChange: () => events.push("before"),
+      afterChange: () => events.push("after"),
+    }),
+  );
+  assert.deepEqual(events, ["before", "after"], "an unclosed beforeChange would corrupt the undo stack");
+});
+
+test("#757 a MONOTONIC counter: the wrong row is taken back out and the real name is named", () => {
+  // rgthree's loraWidgetsCounter only ever increases, so after a row is removed the next
+  // created row is NOT the removed name. A refusal that left the stray row behind could not
+  // be safely retried.
+  const n = loader({ counter: 7 });
+  const before = n.widgets.length;
+  assert.throws(
+    () => createRgthreeLoraRow(n, "lora_1", {}),
+    /this node's next row is "lora_7", not "lora_1"[\s\S]*has been removed again/,
+  );
+  assert.equal(n.widgets.length, before, "nothing is left behind");
+  assert.ok(!n.widgets.some((w) => w.name === "lora_7"), "including the row it minted");
+});
+
+test("#757 the stray row is removed even on a node with no removeWidget method", () => {
+  const n = loader({ counter: 9 });
+  delete n.removeWidget;
+  const before = n.widgets.length;
+  assert.throws(() => createRgthreeLoraRow(n, "lora_2", {}), /next row is "lora_9"/);
+  assert.equal(n.widgets.length, before, "the splice fallback still cleans up");
+});
+
+test("#757 consecutive creates work: lora_1 then lora_2", () => {
+  const n = loader();
+  assert.deepEqual(createRgthreeLoraRow(n, "lora_1", {}), { created: "lora_1" });
+  assert.deepEqual(createRgthreeLoraRow(n, "lora_2", {}), { created: "lora_2" });
+  assert.deepEqual(
+    n.widgets.map((w) => w.name).filter((x) => x.startsWith("lora_")),
+    ["lora_1", "lora_2"],
+  );
+});
+
+// ---------------------------------------------------------------------------
+// The panel wiring
+// ---------------------------------------------------------------------------
+
+import { readFileSync } from "node:fs";
+const PANEL_SRC = readFileSync(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url), "utf8");
+
+test("#757 creation runs AFTER the history seed and BEFORE runSetWidget", () => {
+  // The seed is what makes the #458 authorization meaningful, and growing a node is a
+  // mutation: it must not happen on a call authorization is about to refuse.
+  const seed = PANEL_SRC.indexOf("await awaitObjectInfoHistorySeed();", PANEL_SRC.indexOf("async graph_set_widget("));
+  const create = PANEL_SRC.indexOf("createRgthreeLoraRow(node, widget, {");
+  const run = PANEL_SRC.indexOf("const result = await runSetWidget(");
+  assert.ok(seed > 0 && create > seed, "creation is after the history seed");
+  assert.ok(run > create, "…and before the write");
+});
+
+test("#757 the uuid fence brackets the creation", () => {
+  // The user can switch workflows during the awaits above; a row must never be grown on a
+  // canvas the caller did not address (#570/#718).
+  const at = PANEL_SRC.indexOf("if (isRgthreeLoraRowCreation(node, widget, value)) {");
+  assert.notEqual(at, -1);
+  const block = PANEL_SRC.slice(at, PANEL_SRC.indexOf("createRgthreeLoraRow(node, widget, {", at));
+  assert.match(block, /assertActiveWorkflowCommandTarget\(\{/, "the fence runs before the mutation");
+});
+
+test("#757 the created row is disclosed on its own field, not in `warning`", () => {
+  assert.match(PANEL_SRC, /created_widget: createdLoraRow/);
+  const sw = PANEL_SRC.slice(PANEL_SRC.indexOf("async graph_set_widget("));
+  const body = sw.slice(0, sw.indexOf("async graph_remove_widget("));
+  assert.ok(!/warning:[^\n]*createdLoraRow/.test(body), "it must not displace a warning about the write");
+});
--- a/browser_tests/unit/rgthree-lora-row.test.mjs
+++ b/browser_tests/unit/rgthree-lora-row.test.mjs
@@ -477,6 +477,28 @@ test("#757 remove() never throws, whatever the node does", () => {
   assert.doesNotThrow(() => made.remove(), "an undo on an error path must not replace the refusal");
 });
 
+test("#757 the mismatch refusal CONFIRMS the rewind instead of assuming it", () => {
+  // ROUND-5 P2. `restoreRowCounter` reports whether it ran an assignment, not whether the
+  // assignment took. A counter that accepts `++` but silently ignores writes back — a
+  // getter-only property, a clamping pack, a proxy — made it return true while the number
+  // stayed spent, and the refusal then printed the remedy that only works if it came back:
+  // "Set lora_2 instead", when lora_2 is already gone too.
+  const n = loader({ nextRow: 1 });
+  let frozen = 0;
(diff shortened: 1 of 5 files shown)


artokun:
**PARKED at the agreed two-round cap.** Not merged, left as a draft on purpose.

Round 7 fixed the `showModelAndClip` settle (verified against rgthree's shipped source on this machine, not inferred). Round 8's review was **cancelled at 30 minutes without converging**, so there is no clean verdict — and absence of a finding is not a pass. Given rounds 1-7 each surfaced a case where the command reports success while state is wrong, merging on `59/59 mutations + 4488 tests` alone would be claiming a verification this PR has not had.

**State for whoever picks this up:**

- 8 review rounds; a P1 in six consecutive ones. Severity fell steadily and the last two were creating-path-only, which falsified the round-5 prediction that remaining findings would be shared-path.
- The mechanism is understood, not mysterious: **a row created by the ➕ button is unsettled only until the next repaint — always before a human can touch it. A row created by an agent is written to in the same synchronous command, with no frame in between.** Anything the pack defers to first draw is still unset when our write is verified. `settleCreatedRow` handles the one such field `serializeValue` reads (`showModelAndClip`), confirmed against the pack source; it is not proof that no other exists.
- **The recommended next shape, if this is resumed:** creation as its own committed step, with the value written afterwards as an ordinary existing-row write. That removes the settling period by construction, at the cost of two calls and a refused write leaving an empty row — i.e. an explicit `panel_add_widget_slot` tool surface rather than growing a node as a side effect of a write. That recommendation is backed by two rounds of evidence rather than speculation.

Everything else here is sound and reusable: the classifier requires node type + `lora_N` shape + a slot-shaped value so a typo cannot reach it, the pack method is feature-detected and its effect verified, the monotonic counter is rewound on every refusal path, and the deliberate refusal to auto-press a control (`pressable-widget.js`) is untouched.

PR #1243, opened 2026-08-15
Add panel-open MCP autostart and provider chooser

artokun:
## Summary

- add an Autostart MCP setting, enabled by default, in both the Connect menu and Settings
- request MCP startup only when the Agent panel opens and no MCP endpoint is already available
- proxy launcher status/start calls through fixed same-origin ComfyUI routes
- adapt connection guidance to whether autostart is enabled and whether the companion is installed
- discover supported local/CLI providers, select a sole match automatically, and show a chooser when several are available
- preserve the existing setup instructions when no provider is detected
- add localized UI copy plus unit, proxy, and browser coverage

## Security and behavior

The Registry-scanned panel never spawns a process or accepts arbitrary commands. It talks only to fixed ComfyUI proxy routes; the separately installed MCP launcher owns OS integration and authenticates its loopback broker.

## Validation

- `npm run test:unit` — 4,405 passed, 1 todo
- `npm run typecheck`
- `python -m unittest browser_tests.unit.test_launcher_proxy` — 3 passed
- focused Playwright autostart spec passed before the final rebase
- the post-rebase Playwright rerun reached a local-fixture limitation: the running ComfyUI instance served a different custom_nodes checkout, so the worktree panel could not mount; no product assertion failed

## Companion change

- MCP launcher and live-provider discovery: https://github.com/artokun/comfyui-mcp/pull/1590


artokun:
## The launcher proxy was a confused deputy — pushed a fix (52c4d46e), still not ready to merge on my say-so

### The hole

`POST /comfyui_mcp_panel/launcher/start` was `async def _launcher_start(_request)` — the request was ignored entirely. No Origin, Referer, or CSRF check.

The companion launcher (comfyui-mcp #1590) is properly secured on its own terms: 127.0.0.1 bind, 32-byte random token, constant-time `safeTokenEqual`, one fixed command. This pack proxies it so the token stays server-side and the browser never sees it — right instinct, and *precisely* what created the hole. Holding the token here means the token stops authenticating the **caller** and only authenticates the **proxy**. So any page in any tab the user has open could send

```html
<form method="POST" action="http://127.0.0.1:8188/comfyui_mcp_panel/launcher/start">
```

A cross-origin form POST is a "simple request": no preflight, nothing for the user to approve. The pack then attached the bearer token on that page's behalf and the launcher opened a process. The attacker never needing to *read* the reply is the whole point — the side effect has already happened. `--enable-cors-header` widens it further (a plain `fetch()` POST works too, not just a form).

Same applied to `GET /comfyui_mcp_panel/launcher/status` and, unlisted in the review but the same class, `POST /comfyui_mcp_panel/launcher/handshake` — all three proxy the token, so all three are guarded.

### The allow-rule, and why this mechanism

**Refuse a request whose declared `Origin` (or `Referer`, when `Origin` is absent) has an authority different from the `Host` the browser addressed.**

Why an origin check rather than a CSRF token or a required custom header: both of those rest on an assumption this deployment breaks. With `--enable-cors-header`, ComfyUI answers `Access-Control-Allow-Origin: *` and permissive preflights — a foreign page can read a minted CSRF token straight out of a GET response, and can get a preflight for `X-Requested-With` approved. Both defenses evaporate. The `Origin` *request* header does not: the browser sets it, page script cannot forge it, and CORS response headers govern who may READ a response, never who may CLAIM an origin.

Clause by clause:

- **Compare against the request's own `Host`**, not `COMFYUI_URL` / `_detect_comfyui_url()`. One ComfyUI answers on many authorities (127.0.0.1:8188, localhost:8188, a LAN IP, a tunnel hostname) and each serves a legitimate panel. `Host` is the authority the browser actually addressed, so `Origin`-authority == `Host` states exactly "the page that called me was served by me".
- **Compare authority (host+port), not the full origin string.** A TLS-terminating proxy leaves the page on `https://` while this server still sees plain HTTP; a scheme-strict compare would 403 every remote deployment. Claiming `http://` for an `https://` page needs an active network attacker who could already inject script into that page.
- **`X-Forwarded-Host` is deliberately not consulted** — it is a request header, and under `--enable-cors-header` a foreign page could get its preflight approved and then name itself. Trusting it hands over the answer key.
- **Non-http(s) origins are refused**, including the literal `null` a sandboxed iframe or `data:` document sends.
- **Absent `Origin` *and* absent `Referer` is allowed**, load-bearing in both directions. The same-origin GET status probe the panel polls sends neither, so demanding a header would break the panel; and browsers attach `Origin` to every POST they issue (cross-origin form POSTs included), so a header-less caller is not a browser page — it is curl or a local script, which can already read `~/.comfyui-mcp/launcher.json` and call the launcher directly. This guard defends the browser's ambient authority; it is not local-process authentication and cannot be made into it.

Known limits are written into the source comment rather than left for someone to rediscover: a page served *by* this ComfyUI origin passes (same-origin content is indistinguishable from the panel by construction); DNS rebinding makes Origin and Host agree, but it already exposes every other ComfyUI route, so it is a host-level concern; and a frontend dev server on another port is cross-origin by definition and is refused — that is the rule working.

### Two smaller items in the same file

- `payload.pop("token", None)` was a **denylist over a launcher response reflected into the browser**. Replaced with an allowlist (`_LAUNCHER_RESULT_KEYS`: `ok`, `protocol`, `orchestrator_running`, `already_running`, `started`, `start_in_progress`, `minimized`, plus a code-shaped `error`). A denylist is only as current as the last person who remembered to update it — the day the launcher grows a `config_path` or a second secret, a denylist ships it. The launcher's 500 path returns `error: <Error.message>`, free text that can carry a filesystem path, so a non-code-shaped `error` is collapsed to `launcher_error`.
- `"message": str(error)` reflected aiohttp's error text — which quotes the loopback host and the launcher's ephemeral port — into the page. Now a fixed non-identifying reason. The panel branches on `error` and never renders this text.

### Verification

`browser_tests/unit/test_launcher_proxy.py` grew from 3 to 18 tests. They drive the **real handlers** (`_register_launcher_routes` is now module-level and takes a route collector + a `web` double) and the **real proxy call** (fake `aiohttp` in `sys.modules`) — nothing greps the source, since `if False: guard()` matches a regex just fine. The cross-origin cases assert the launcher was never called, not merely that the status was 403: for a form POST the side effect *is* the attack.

- `npm run test:unit` — **4406 tests, 4405 pass, 0 fail, 1 todo** (i18n, render, tool-vocabulary and check-panel-scope gates all green; `npm install` was run first so check-panel-scope actually executed).
- `python -m unittest browser_tests.unit.test_launcher_proxy` — **18 tests, OK**.
- `python -m py_compile __init__.py`, the CI AST one-statement/eval check, and `bandit -r . -s B101,B112,B311` (Registry parity) — all clean.

**Mutation test** (committed first; each mutation applied to bytes and restored from an in-memory pristine copy, never `git checkout`):

| # | Mutation | Result | Killed by |
|---|---|---|---|
| M1 | Remove the origin check entirely (guard always returns "allowed") | RED — killed | 8 tests |
| M2 | Allow a foreign origin (drop the authority comparison, keep the scheme check) | RED — killed | 7 tests |
| M3 | Accept the literal `null` / non-http(s) origins | RED — killed | `test_opaque_and_look_alike_origins_are_refused` |
| M4 | Restore the denylist (`pop("token")`, reflect everything else) | RED — killed | `test_only_expected_keys_reach_the_browser`, `test_a_free_text_launcher_error_is_collapsed_to_a_code` |
| M5 | Reflect aiohttp's error text into the page again | RED — killed | `test_an_unreachable_launcher_does_not_name_its_host_or_port` |

Suite green again after restore, working tree byte-identical to the commit.

### Not done, flagging instead

- **`POST /comfyui_mcp_panel/advertise_bridge` looks like the same class of hole** and is out of scope for this fix. A cross-origin POST can set `_ADVERTISED_BRIDGE_URL`, and the panel then dials that `wss://` on Connect — i.e. a foreign page could point the panel at an attacker-controlled bridge. The `wss://` check restricts the scheme, not the destination. `_cross_origin_denial()` is now available and would apply cleanly (the local orchestrator that legitimately POSTs here sends no `Origin`, so it keeps working), but I did not want to widen this PR unasked.
- **The Python tests do not run in CI.** `ci.yml` runs `py_compile` and bandit but no `unittest`. This new suite is stdlib-only and would be a one-line step; two of the *existing* python test modules need `aiohttp`, so full discovery would need a `pip install` first.
- **No e2e run** — Playwright needs a live ComfyUI. The panel calls all three routes through `api.fetchApi` with relative paths, which is same-origin by construction, and the existing specs intercept `**/comfyui_mcp_panel/launcher/**` before the server sees them.

Not merged, not marked ready, no version/CHANGELOG edits.

PR #1253, opened 2026-08-15
fix(1606): a Manager that CANNOT report a failure stops reading as one reporting none

### LINKED ISSUE #1606: panel_install_node legacy Manager can drain queue without surfacing install failure
On Manager 3.x legacy mode, panel_install_node queued comfyui-reactor and returned pending with no failure. panel_node_queue_status immediately reported an empty idle queue, and panel_list_nodes confirmed the pack was absent. The wrapper cannot surface why the legacy Manager discarded or failed the install, forcing a manual source-install workaround.

### Steps to reproduce
1. On ComfyUI-Manager 3.x legacy, call panel_install_node with id comfyui-reactor.
2. Poll panel_node_queue_status.
3. Verify with panel_list_nodes.

### Exact error (scrubbed)
No error was reported. Result said queued:true, installed:false, verified:false, pending:true; the queue then became idle and the pack was absent.

### Fix
Capture and return the legacy Manager task failure/log detail when a queued install drains without producing a verified pack, or retain a failed task record long enough for panel_node_queue_status to report it.

### Environment
Windows 11; ComfyUI 0.33.0; GPU RTX 3090 24GB; comfyui-mcp 0.51.54; panel 0.14.41; ComfyUI-Manager 3.x legacy.

artokun:
Fixes artokun/comfyui-mcp#1606.

## What the reporter saw

On ComfyUI-Manager 3.x legacy, `panel_install_node comfyui-reactor` returned
`queued:true, installed:false, verified:false, pending:true` with **no failure**.
`panel_node_queue_status` immediately showed an **empty idle queue**, and
`panel_list_nodes` confirmed the pack was absent. Nothing anywhere said why, so
they fell back to a manual source install.

## Why nothing said why

#1539 gave install the Manager's own terminal verdict by reading
`/v2/manager/queue/history?ui_id=`. Released 3.x registers no such route — and
the record is not merely unexposed, it is **deleted**. From ComfyUI-Manager
**3.41**'s `glob/manager_server.py` (read off this machine's own install, the
exact build the report names), `task_worker`:

```python
nodepack_result[ui_id] = msg          # do_install returns 'success', or the error text
...
PromptServer.instance.send_sync("cm-queue-status",
    {'status': 'done', 'nodepack_result': nodepack_result, ...})
nodepack_result = {}                  # <-- and it is gone
task_queue = queue.Queue()
```

`/manager/queue/status` derives `done_count` from `len(nodepack_result)`, so the
line **after** the broadcast is what makes total/done/in_progress all read 0.
The reporter's "empty idle queue" is produced *by the task finishing*, and it
looks identical whether the task succeeded or failed. No later HTTP read can
recover the outcome.

So the broadcast is the only statement of the outcome that ever exists — and a
panel living in the ComfyUI page is already a client of it. `send_sync` with no
`sid` goes to every connected client, and ComfyUI-Manager's own UI reads it the
same way (`api.addEventListener("cm-queue-status", …)` in
`js/custom-nodes-manager.js`). ComfyUI's frontend dispatches an unknown socket
type only to listeners that registered for it (`_registered.has(t.type)` guards
the `default:` arm of its socket handler), which is what makes the subscription
load-bearing rather than decorative.

## What this does

- **Subscribe at panel load** (`setupListeners`, in its own try so a diagnostic
  capture can never preempt the backend-down/reconnect listeners it sits above)
  and keep what the frame says in a bounded, insertion-ordered log.
- **`verifyInstalled` passes it as `capturedFailure`**, correlated by the `ui_id`
  *this* command submitted. It feeds `classifyInstallOutcome`'s existing
  `taskFailure` branch, so no new verdict logic: the reported install now fails
  with the Manager's own words — `Cannot resolve install target:
  'comfyui-reactor@latest'` — instead of reporting pending.
- **`nodes_queue_status` lists captured failures**, and stops returning a bare
  `{status}` that a build with no history cannot be distinguished from a v4 with
  nothing to report. `failure_reporting` is now `complete` / `partial` /
  `unavailable`, and an unreadable history says so rather than reading as a
  clean queue.
- A later **success for the same pack retires** the earlier failure, and captures
  **age out** (30 min) of the poll surface, so a defeat that has been undone is
  not re-reported.

Additive in both directions: a build that never emits the frame reads exactly as
it does on main today, and the capture can only ever *add* a positive failure —
`queueEventFailureReason` returns null for anything it cannot read.

## Proof

`npm run test:unit` — **4552 pass, 0 fail** (27 new, in
`browser_tests/unit/manager-task-broadcast.test.mjs`). The install/verify tests
drive the **real extracted `waitForQueueDrain` + `verifyInstalled` +
`nodes_queue_status`** from the panel source, not helper copies.

Three mutations, each killed:

| deleted | fails |
|---|---|
| `capturedFailure:` in `verifyInstalled` | 3 tests |
| `subscribeManagerTaskResults()` in `setupListeners` | 1 test |
| `managerTaskResults.note(ui_id, …)` in `nodes_install` | 1 test |

There is also a permanent negative control in the suite: *"WITHOUT the capture
the same install is only 'unverified' (the bug)"* runs the same legacy scenario
against an empty log and asserts the pre-fix verdict.

## Not verified here

The chain is read from source at every link — Manager 3.41's worker, ComfyUI's
`send_sync` broadcast, the frontend's `_registered` dispatch guard — but **not
run end-to-end on a live legacy ComfyUI**: no instance was running for this
session and this run was not allocated one. If the frame never arrives, the
behaviour degrades to main's, plus the honest `failure_reporting: unavailable`
note. A live legacy install of a bogus id is the check that would close that gap.

> Codex gate: **INDETERMINATE** — the codex account's usage limit is exhausted
> until Aug 19. Per autopilot rules an indeterminate gate is not a pass, so this
> is **held, not merged**.

### DIFF
--- a/browser_tests/unit/install-terminal-failure.test.mjs
+++ b/browser_tests/unit/install-terminal-failure.test.mjs
@@ -116,6 +116,10 @@ function buildVerifyInstalled({ routes, calls = [], budgetMs }) {
     INSTALL_VERIFY_BUDGET_MS: budgetMs ?? 4000,
     AbortSignal,
     setTimeout,
+    // comfyui-mcp#1606 — the capture the verifier now also consults. EMPTY here
+    // on purpose: these tests pin the HTTP history path, and an empty log proves
+    // the capture cannot short-circuit or alter it.
+    managerTaskResults: ManagerInstall.createManagerTaskResultLog(),
   };
   const factory = new Function(
     ...Object.keys(deps),
--- a/browser_tests/unit/manager-dialect.test.mjs
+++ b/browser_tests/unit/manager-dialect.test.mjs
@@ -971,6 +971,10 @@ function nodesInstallDeps(overrides) {
     reProbeManagerDialect: async () => "v2",
     managerQueueControl: async () => {},
     verifyInstalled: async () => ({ state: "installed", status: QUEUE_STATUS }),
+    // comfyui-mcp#1606 — the ui_id ↔ pack correlation the handler records so a
+    // captured Manager failure can name what it was installing. A REAL log, so
+    // a regression in that call surfaces here instead of hitting a no-op stub.
+    managerTaskResults: ManagerInstall.createManagerTaskResultLog(),
     ...overrides,
   };
 }
--- a/browser_tests/unit/manager-install.test.mjs
+++ b/browser_tests/unit/manager-install.test.mjs
@@ -860,6 +860,10 @@ function loadNodesInstall({ budgetMs, detect, reProbe, managerV2, managerCall, m
     "MANAGER_FETCH_TIMEOUT_MS",
     "NODES_INSTALL_COMMAND_BUDGET_MS",
     "AbortSignal",
+    // comfyui-mcp#1606 — the handler correlates its ui_id with the pack, so the
+    // Manager's completion broadcast can name what failed. A real log, not a
+    // stub: an injected no-op would let the correlation regress unnoticed here.
+    "managerTaskResults",
     `${stallMatch[0]}\nconst handler = { ${body} };\nreturn handler.nodes_install;`,
   );
   return factory(
@@ -878,6 +882,7 @@ function loadNodesInstall({ budgetMs, detect, reProbe, managerV2, managerCall, m
     15000,
     budgetMs,
     AbortSignal,
+    ManagerInstall.createManagerTaskResultLog(),
   );
 }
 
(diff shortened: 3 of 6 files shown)


artokun:
**Gate: INDETERMINATE — held, not merged.**

```
$ node .claude/autopilot/codex-gate.mjs --worktree .../1606-failure-visibility --base origin/main --name 1606
[gate] INDETERMINATE — codex account quota exhausted. Merge HOLDS.   (exit 2)
```

Confirmed directly against `codex exec`, not just through the wrapper:

> ERROR: You've hit your usage limit. Visit https://chatgpt.com/codex/settings/usage to purchase more credits or try again at Aug 19th, 2026 8:43 PM.

An indeterminate gate is not a pass, so this stays a draft. **What unblocks it:** codex credits (or the Aug 19 reset) → re-run the gate command above → merge if it returns SHIP.

Meanwhile: CI green (`Lint pack` success, run 31906623935) and `npm run test:unit` locally green at 4552 pass / 0 fail. No self-review is offered in place of the gate — a self-review reads its own guard as obviously right, which is the defect class this gate exists to catch.

PR #1258, opened 2026-08-15
fix(1571): a subgraph conversion that broke the graph stops reporting success

### LINKED ISSUE #1571: panel_subgraph_group leaves bypass-node boundary link broken and nested run-to-node cannot fingerprint
### What happened
Converting the Krea2 T2I group with panel_subgraph_group preserved a bypassed RBG_Smart_Seed_Variance node incorrectly. A full panel_run then failed with `No link found in parent graph for id [302:192] slot [0] conditioning`. Also panel_run(to_node_id:183) for the nested SaveImage refused with `run-to-node scope for node 183 cannot be dispatched safely: the prompt could not be fingerprinted (graphToPrompt failed)`.

### Steps to reproduce
1. Load krea2-combo and wrap the T2I group via panel_subgraph_group.
2. The group includes bypassed RBG_Smart_Seed_Variance feeding two KSamplers.
3. Try nested run-to-node or full panel_run.

### Fix / workaround
Applied locally to workflow: entered the subgraph, connected CLIPTextEncode conditioning directly to both samplers, removed the bypassed RBG node, then full panel_run succeeded. Upstream should avoid exposing a disconnected boundary input for bypassed nodes during group-to-subgraph conversion and support fingerprinting nested output targets after conversion.

### Environment
Windows 11; ComfyUI 0.33.1; RTX 5090 32GB; comfyui-mcp 0.51.45; panel 0.14.35.

Related (same root cause): #1294

artokun:
Fixes artokun/comfyui-mcp#1571

## What was reported

`panel_subgraph_group` wrapped a group containing a **bypassed** `RBG_Smart_Seed_Variance`, reported `{subgraph: {node_id: 302, …}}`, and the workflow could not be run afterwards. One broken graph, three different answers:

| call | answer |
| --- | --- |
| `panel_subgraph_group` | `{subgraph: {node_id: 302, …}}` — success |
| `panel_run {}` | `No link found in parent graph for id [302:192] slot [0] conditioning` |
| `panel_run {to_node_id: 183}` | `the prompt could not be fingerprinted (graphToPrompt failed)` |

Only one named anything, and what it named (`[302:192]`) is a flattened id the caller had never seen. The reporter repaired the graph by hand and filed both halves as one issue — concluding, from the third answer, that run-to-node "cannot fingerprint **nested** output targets". Nesting was never involved.

## Where the corruption comes from

`ExecutableNodeDTO.resolveInput`, read out of the shipped ComfyUI_frontend bundle:

```js
if (i.linkId == null) return                    // unconnected: fine
const link = this.graph.getLink(i.linkId)
if (!link) throw new InvalidLinkError(
  `No link found in parent graph for id [${this.id}] slot [${slot}] ${i.name}`)
```

`this.graph` there is the node's **own** graph — for a node that now lives inside a subgraph, that is the Subgraph. So the error states one precise fact: an input references a link id that does not exist in the graph the node is now in. Worth recording against the report: the reporter described this as "a disconnected boundary input", and it is not. A boundary input with no link is handled gracefully by that same function (`if (linkId == null) … return`). What throws is a **dangling link id** — a reference to a link that was never written into the new subgraph's link table.

`LGraph.convertToSubgraph` can produce exactly that and does not check. `getBoundaryLinks` skips any input link it cannot resolve (`console.warn('Failed to resolve link ID […]')`, `continue`), and `mapSubgraphInputsAndLinks` skips a whole connection group whose first resolved connection has no `input` — while the reconnect loop that follows indexes `subgraphNode.inputs[i - 1]` off a counter incremented for **every** group.

That is the frontend's code, not ours. We cannot fix it here. We can stop reporting it as a success.

## What changed

**1. Both conversion tools verify their own output.** New `web/js/lib/subgraph-conversion-integrity.js`, reporting in two tiers:

- **dangling input on a node the serializer does not skip → refuses.** Provably unserializable with no modelling at all: `graphToPrompt` skips virtual / muted / bypassed nodes and then resolves **every** input of the ones it keeps, unconditionally — no type matching, no consumer analysis. The refusal names the node, slot and link, prints the `[302:192]`-shaped error the run would otherwise have shown first, and says the subgraph **is still on the canvas** — the opposite of the sibling `assertSubgraphNodeLanded` message, so the caller does not retry and wrap the same nodes twice.
- **dangling input on a skipped node, and unfed boundary slots → reported, never fatal.** Surfaced as a `warning` string plus structured `dangling_inputs` / `unfed_boundary_inputs`.

**#1571's own node is in the second tier, and that is the point.** It is bypassed, so the reported conversion now *succeeds*, carrying the finding it used to hide. Three rounds of the review gate killed three successive attempts to decide reachability *through* a bypassed node, each a closer approximation of `_getBypassSlotIndex` and each still wrong:

1. it is driven by the **consumer's** type, not the output's, so a wildcard consumer short-circuits the type check entirely;
2. a connected output whose only consumer is itself muted reaches nothing;
3. `LiteGraph.isValidConnection` accepts comma-delimited union types (`IMAGE,MASK`), which a string comparison does not.

An approximation of a resolver is not the resolver. That modelling is gone. The links are provably missing — that part is measured and reported; whether serialization trips over them is stated as ComfyUI's verdict to give, not the panel's, and the message says so and names what gives it.

The obvious shortcut — call `app.graphToPrompt()` and see whether it throws — is deliberately **not** taken: it runs `applyToGraph()` on every virtual node, so probing with it mutates the graph as a side effect of asking a question about it.

Everything in the module fails toward **silence** on an input it cannot read: it gates the report of a mutation that already happened, so an unfamiliar graph shape must produce no findings rather than a graph-wide accusation.

**2. The scoped-run fingerprint stops discarding the cause.** `dispatchScopedRun` wrapped `app.graphToPrompt()` in `try { … } catch { }` with an *empty binding*, so ComfyUI's `InvalidLinkError` — which names the offending node outright — was caught and thrown away one line before the message that needed it. The cause is now quoted verbatim, attributed to the frontend ("that is the frontend's own error, not the panel's diagnosis of it"), bounded at 400 chars, whitespace-flattened, and **omitted entirely when nothing threw** — this path also fires with no `graphToPrompt` at all, and inventing a cause there would be the same defect pointed the other way.

## Evidence

The fixture is not invented. Node 192 / `RBG_Smart_Seed_Variance` / `mode: 4` / input `conditioning` / `link: 505`, feeding KSamplers 265 and 273 with SaveImage 183 downstream, are read verbatim from `packs/krea2-combo/workflow.json` in the mcp repo — the workflow the report names. The corruption shape is read from `ExecutableNodeDTO`.

- **41 new tests** (32 conversion-integrity + 9 scope-fingerprint-cause) and 4 added to `subgraph-stale-outline`, including wiring assertions on both call sites and behaviour tests driving the real `graph_subgraph_group` / `graph_create_subgraph` executors — a helper the tool never consults is the shape that shipped this bug in the first place.
- Full panel unit suite: **4540 passed, 0 failed** (includes `check-panel-scope`, `check-tool-vocabulary`, both i18n checks).
- **19/19 mutations killed**, including: delete either call site; drop either advisory key; refuse on everything (the rounds 1–3 over-refusal); refuse on nothing; reclassify muted / bypassed / virtual / ordinary nodes; compare link ids by identity; treat an unreadable link table as empty; walk into nested subgraph definitions; make the warning assert the workflow cannot run; and — the original defect — restore the empty `catch {}` in the fingerprint path.

## Deliberately not done

- **No auto-repair.** The dangling reference could be cleared to make the graph serializable again, but that silently drops a connection the user still wants, and re-deriving the lost boundary link from the parent's link table is graph surgery this cannot verify live. Report, name both recoveries, mutate nothing.
- **The frontend bug itself is untouched** — it belongs upstream in ComfyUI_frontend.
- **No live-rig verification.** No ComfyUI was running for this run, so the behaviour tests drive the real executors against fixtures rather than a live canvas.

### DIFF
--- a/browser_tests/unit/scope-fingerprint-cause.test.mjs
+++ b/browser_tests/unit/scope-fingerprint-cause.test.mjs
@@ -1,0 +1,114 @@
+// comfyui-mcp#1571, second half — `panel_run { to_node_id: 183 }` refused with:
+//
+//   "run-to-node scope for node 183 cannot be dispatched safely: the prompt could not be
+//    fingerprinted (graphToPrompt failed)"
+//
+// The refusal is correct: no fingerprint means the panel cannot tell its own dispatch from
+// unrelated queue traffic, so it fails closed (#556). What was wrong is that it threw the
+// reason away. `dispatchScopedRun` wrapped `app.graphToPrompt()` in `try { … } catch { }`
+// with an empty binding, so ComfyUI's own `InvalidLinkError: No link found in parent graph
+// for id [302:192] slot [0] conditioning` — which names the offending node outright — was
+// caught and discarded one line before the message that needed it.
+//
+// The cost is visible in the report itself: the reporter titled it "nested run-to-node
+// cannot fingerprint" and asked for "fingerprinting nested output targets" to be
+// supported. Nesting was never involved. Their graph had been left unserializable by a
+// subgraph conversion, and a FULL `panel_run` on the same graph failed the same way.
+
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+
+import {
+  scopeUnattributableError,
+  describeFingerprintFailure,
+} from "../../web/js/lib/run-scope-guard.js";
+
+/** The error ComfyUI_frontend 1.48.7 actually throws for the reported graph. */
+const INVALID_LINK = new Error(
+  "No link found in parent graph for id [302:192] slot [0] conditioning",
+);
+INVALID_LINK.name = "InvalidLinkError";
+
+test("#1571 the serializer's own message reaches the caller", () => {
+  const msg = scopeUnattributableError({ toNodeId: 183, cause: INVALID_LINK });
+  assert.match(msg, /No link found in parent graph for id \[302:192\] slot \[0\] conditioning/);
+  // …and the refusal it is embedded in is unchanged: still fail-closed, still #556.
+  assert.match(msg, /run-to-node scope for node 183 cannot be dispatched safely/);
+  assert.match(msg, /Nothing was queued/);
+  assert.match(msg, /#556/);
+});
+
+test("#1571 the quote is attributed, so it does not read as the panel's own diagnosis", () => {
+  const msg = scopeUnattributableError({ toNodeId: 183, cause: INVALID_LINK });
+  assert.match(msg, /ComfyUI's serializer failed with/);
+  assert.match(msg, /the frontend's own error, not the panel's diagnosis/);
+});
+
+test("#1571 the refusal contradicts the theory the silence created", () => {
+  // The reporter's conclusion was that run-to-node cannot handle NESTED targets. Whatever
+  // else this message says, it has to say that much is not the cause.
+  const msg = scopeUnattributableError({ toNodeId: 183, cause: INVALID_LINK });
+  assert.match(msg, /not specific to run-to-node or to nesting/);
+});
+
+test("#1571 with nothing thrown, no cause is invented", () => {
+  // This path also fires when nothing threw at all — a frontend with no `graphToPrompt`,
+  // or a prompt that canonicalizes to nothing. Attaching a cause there would be the same
+  // defect pointed the other way: a reporter sent to fix a graph that is fine.
+  for (const nothing of [undefined, null, "", "   ", new Error(""), 42, {}]) {
+    const msg = scopeUnattributableError({ toNodeId: 7, cause: nothing });
+    assert.match(msg, /could not be fingerprinted/, String(nothing));
+    assert.doesNotMatch(msg, /ComfyUI's serializer failed with/, String(nothing));
+    assert.doesNotMatch(msg, /No link found/, String(nothing));
+  }
+  // The historical call shape — no argument at all — must still produce the old message.
+  assert.match(scopeUnattributableError({ toNodeId: 7 }), /cannot be dispatched safely/);
+  assert.doesNotMatch(scopeUnattributableError({ toNodeId: 7 }), /serializer failed with/);
+});
+
+test("#1571 a string cause is accepted as well as an Error", () => {
+  // Not every layer that can fail here throws an Error object.
+  assert.match(describeFingerprintFailure("boom from an extension"), /boom from an extension/);
+});
+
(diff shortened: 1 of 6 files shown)


artokun:
**HOLDING — the review gate could not run. Not merging.**

The codex gate returned **INDETERMINATE** (exit 2), not a pass:

```
```

Confirmed directly rather than trusting the gate's own classification — `codex exec` on a
one-line probe answers:

```
ERROR: You've hit your usage limit. … try again at Aug 19th, 2026 8:43 PM.
```

So this is a real quota exhaustion with a known reset, not a sandbox or empty-verdict
misfire. **An indeterminate gate is not a pass**, and a self-review is not a substitute for
one — a self-review reads its own guard as obviously right, which is exactly the defect
class this change has already been wrong about three times. The PR stays a draft.

**What unblocks it:** re-run

```
node .claude/autopilot/codex-gate.mjs \
  --worktree C:/Users/Artokun/code/comfyui-mcp-panel/.claude/worktrees/1571 \
  --base origin/main --name 1571
```

after **Aug 19 2026 20:43** (or once credits are topped up). Exit 0 → squash-merge; exit 1 →
one more revision round, but note the rounds below.

**State at the hold**, so a resume does not redo any of it:

- Head `c58cda0b`, CI **green** (Lint pack, 2m3s).
- Full panel unit suite: **4540 passed / 0 failed**, including `check-panel-scope`,
  `check-tool-vocabulary` and both i18n checks.
- **19/19 mutations killed** — both call sites, both advisory keys, all four node
  classifications, the link-id comparison, the unreadable-link-table guard, the
  one-level walk, the warning's strength, and the original empty `catch {}`.
- The PR body above is current as of `c58cda0b`. The earlier body described the
  round-1..3 design where a dangling input on a *bypassed* node refused; that is no
  longer what this does, and #1571's own node now **warns and succeeds**.

**Gate rounds already spent: 3.** Rounds 1–3 each killed a closer approximation of
`_getBypassSlotIndex` and each replacement was still wrong; the response was to stop
modelling the resolver rather than approximate it a fourth time. If the next round asks for
reachability *through* a bypassed node to be decided here, that is the ask this change
already refused on purpose — the argument is in the module header, and the counter-example
for each attempt is in the commit message of `c58cda0b`.

Issue #1111, opened 2026-08-12
Fix workflow tab switch leaving previous graph on canvas


artokun:
### What happened / root cause
panel_open_workflow switched the active workflow identity/title to an LTX workflow, but panel_graph_outline still returned the previous H3 graph. The open response detected a graph mismatch and reported every loaded node existed, yet the subsequent active-tab read showed the wrong 40-node H3 canvas rather than the saved 49-node LTX graph. Loading the saved workflow explicitly with panel_load_workflow(path) recovered the correct graph.

### Steps to reproduce
1. Keep two saved workflows open with distinct graphs.
2. Modify/render the first workflow.
3. Call panel_open_workflow for the second saved workflow after a reconnect.
4. Call panel_list_workflows, then panel_graph_outline.
5. Observe the second workflow identity is active while the first workflow graph remains on canvas.

### Exact error (scrubbed)
workflow_open RAN and the canvas IS bound to the requested workflow, but the graph on it does not match the state that was loaded. The graph differed on nodes and links. A later graph outline showed the previous workflow's graph.

### Fix
Applied locally: no
Upstream-only: yes
Suggested change: make workflow_open completion wait for and verify the target workflow's graph repaint/activation transaction before publishing the new workflow identity. If verification sees the previous graph, retry the frontend activation once or fail without changing the active binding. panel_load_workflow(path) was a successful workaround.

### Environment
Windows 11; ComfyUI 0.30.2; RTX 3090 24 GB; comfyui-mcp 0.50.44; panel 0.11.43.

artokun:
Related: #1089 is the same wrong end state — identity moves to the requested workflow while the canvas keeps the previous graph — but reported a **clean success** (`modified: false`, correct uuid) rather than the mismatch you saw. Their next planned calls were `panel_remove_node`, and Save-As preserves ids, so those deletions would have landed on the wrong workflow.

Your report supplies the discriminator that one was missing: `panel_graph_outline` returning the PREVIOUS graph proves the panel was reading the stale graph, not merely rendering it. Taking this one, since it carries the reproduction.

artokun:
Shipped in **panel 0.14.11** — the disclosure half. The cause is not fixed and this stays open for it.

[...]

Issue #1089, opened 2026-08-11
panel_open_workflow reported success with the correct uuid, but the canvas kept the PREVIOUS workflow's graph (near-miss: edits would have hit the wrong workflow)


artokun:
### What happened

After a Save-As, `panel_open_workflow` reported a clean success for the ORIGINAL workflow — right path, right filename, right `workflow_uuid`, `modified: false` — but the live canvas still held the graph of the workflow I had just saved-as to. There was no warning of any kind on this call.

I only caught it because the next `panel_connect` failed with "No node with id 341", which made no sense for that workflow. Reading both files off disk showed the disk state was correct and it was the CANVAS that was stale.

This is a near-miss for silent data loss: my next steps were `panel_remove_node` calls. Had those node ids happened to exist in the stale graph (they largely did — the two workflows share most ids, since Save-As preserves them), I would have deleted nodes from the WRONG workflow and then saved over it, with every tool call reporting success.

### Steps to reproduce

1. Open workflow A (`Qwen3VL_Flux_Text_Styler`, 31 nodes).
2. `panel_save_workflow({name: "Text na pruhlednem pozadi (alfa)"})` → Save-As. Response correctly flags `workflow_instance_changed: true` and B becomes active.
3. Edit B (remove nodes), `panel_save_workflow()` → B saved correctly.
4. `panel_open_workflow({path: "Qwen3VL_Flux_Text_Styler"})` → reports success:

```json
{
  "opened": { "path": "workflows/Qwen3VL_Flux_Text_Styler.json", "filename": "Qwen3VL_Flux_Text_Styler" },
  "routing_key": "wf:workflows/Qwen3VL_Flux_Text_Styler.json",
  "active_routing_key": "wf:workflows/Qwen3VL_Flux_Text_Styler.json",
  "active_matches_target": true,
  "workflow_uuid": "75e36464-...",
  "modified": false,
  "open_seq": 2
}
```

5. `panel_query_graph({fields:'ids'})` → returns **B's** 27 node ids (no `4`/`LoadImage`, no `341`, but `342`/`344`/`345` present). A on disk has 31 nodes including `4` and `341`.

### Why the success report is the bug

[...]

PR #1274, opened 2026-08-15
fix(1623): an open where only presentation moved is not a failure

artokun:
Reported in artokun/comfyui-mcp#1623 — `panel_open_workflow` reports a false error when the frontend normalizes per-node fields.

## What the reporter got

Two consecutive workflow switches, on a library with DOM/custom nodes (ComfySketch, MarkdownNote), each returning `isError: true` with a message that ends:

> workflow_open RAN, the canvas IS bound to X, and every node that was loaded is on it with the same id and type … nothing extra appeared … the content is reported **UNCONFIRMED rather than failed** … You are on the right workflow and **there is no missing work to redo**.

A reply cannot say "no missing work to redo" and fail. They took the failure at face value and spent a `panel_graph_outline` recovery read on a graph that was already correct.

## The mechanism — two lists answering different questions

| set | question it answers | contents |
|---|---|---|
| `RECOMPUTED_NODE_FIELDS` | is this difference explained by a rewrite this panel **measured**? | `size` (height-only), `inputs` |
| `COSMETIC_NODE_FIELDS` | could anything **authored** have been lost? | `size`, `pos`, `order`, `color`, `bgcolor` |

The **disclosure** asks the second. The **pass/fail** asked the first. So `pos`, `order`, `color`, `bgcolor` — and a `size` whose **width** moved, which is exactly what a DOM widget's box does — failed the open while the sentence attached to that failure said nothing was lost. The two agreed by accident when the only cosmetic field anyone hit was a height, and stopped agreeing the moment one wasn't.

## The change

`openContentDifferenceIsPresentationOnly` is promoted to a single exported predicate — nodes is the only differing surface, the node set is intact, every differing field is in `COSMETIC_NODE_FIELDS` — and **both** the sentence and the verdict now read it. The open reports applied on that weaker ground, publishes `workflow_uuid`, and discloses the fields that moved under its own key (`presentation_rewritten`) with its own note.

It is asked off the **same frozen snapshot** the strict proof reads. Answering it from a second serialization would reopen the hole codex r3 closed: a synchronous serialization hook could show a cosmetic-only difference to one question and a changed widget to the other.

## What is NOT widened — this is the part to review hardest

`widgets_values` stays outside **both** sets, so the reporter's prescribed fix ("treat size/order/**widgets_values** as success") is **not** granted. `RECOMPUTED_NODE_FIELDS` is not touched at all.

`resolveOpenRebindVerdict` records why `content` blocks success in the first place: `loadGraphData` catches a mid-`configure()` throw and returns, leaving the complete node id/type set, the links and the panel's marker over nodes that silently **lost their widget values and properties** — byte-identical to "the loader normalized them", with no discriminator to separate the two.

That failure cannot present as presentation-only. `configure()` writes `widgets_values`, `properties`, `title`, `flags` and `mode` in the same pass as the cosmetic five, and every one of them is outside `COSMETIC_NODE_FIELDS` — so a load that died mid-configure answers false on the first node it reached, and the nodes it never got to answer false on everything. The discriminator that comment says does not exist **is which fields differ**, and the panel already computes it. The guard #1111 and #1089 exist for is untouched.

`COSMETIC_NODE_FIELDS` remains a **denylist**: an unknown field from any node pack reads as non-cosmetic and still fails the open. `color`/`bgcolor` are authored and a difference there IS an authoring difference — they ride along because the claim being licensed is only "no node and no value was lost", and the reply **names every field that moved** so a caller who stores colours finds out.

## Verified

- `npm run test:unit`: **4543 tests, 0 failures**, including the i18n, tool-vocabulary and panel-scope gates.
- **9 mutations applied, all 9 killed** — occurrence-verified (each `from` string must appear exactly once or the mutation is reported skipped, not silently passed). Including M1, which restores the reported behaviour by deleting `|| contentProof.presentationOnly`, and M6, which discards the answer at the strict proof's own refusal so the fix goes inert for the reporter's case.
- `sameNodeSet` in the predicate **survived** deletion at first — `classifyNodeDifference` computes `fields` only after the sets match, so its `cosmeticOnly:true` already implies it. The check is kept (the predicate is exported and takes a caller-supplied shape) and now has a test that kills the mutation, with the overlap written down rather than left as a coincidence.

- **I did not reproduce the reporter's session.** If one of their two switches differed in `widgets_values` — the issue body lists it — that call still fails, correctly, and I say so on the issue. Admitting it needs the rewrite characterised the way `size` and `inputs` were, and I had no live frontend to measure against; guessing at it is what `RECOMPUTED_NODE_FIELDS`'s own growth rule forbids.
- I did not touch `describeOpenRebindOutcome`'s reassurance branch. It is still reachable — a live canvas can move between the verdict and the message — and it now asks the same question the verdict does, so it can no longer contradict it.
- I did not admit a `definitions` surface alongside `nodes` here. #1252 (open) is doing that accounting for the message; this stays on the narrow `["nodes"]` gate and will merge on top of whichever lands first.

### DIFF
--- a/browser_tests/unit/definitions-renumber.test.mjs
+++ b/browser_tests/unit/definitions-renumber.test.mjs
@@ -126,11 +126,17 @@ test("#886 WIRING: the content proof consults it for a definitions surface", () 
   assert.match(src, /import \{ definitionsDifferOnlyByLinkRenumber \} from "\.\/definitions-renumber\.js";/);
   assert.match(src, /definitionsDifferOnlyByLinkRenumber\(state\?\.definitions, actualState\?\.definitions\)/);
   // The surface set must be a SUBSET of { nodes, definitions }; anything else refuses.
-  assert.match(src, /s !== "nodes" && s !== "definitions"\)\) return NOT_PROVEN;/);
+  // #1623 renamed the refusal value: every refusal AFTER the diff is computed now
+  // carries the separate presentation-only answer out with it instead of discarding
+  // it. What is pinned is unchanged — this surface set still refuses.
+  assert.match(src, /s !== "nodes" && s !== "definitions"\)\) return notProven;/);
+  // ...and `notProven` must be a REFUSAL of the content proof, not a second success
+  // door: a rename that made it `proven: true` would pass the line above.
+  assert.match(src, /const notProven = \{\r?\n\s*proven: false,/);
   // `nodes` must NOT be mandatory. Requiring it refused the reported case outright:
   // #886 is a graph where `definitions` is the ONLY differing surface, so the earlier
   // gate wired this fix into a branch its own bug report could never reach (review).
-  assert.doesNotMatch(src, /if \(!surfaces\.includes\("nodes"\)\) return NOT_PROVEN;/);
+  assert.doesNotMatch(src, /if \(!surfaces\.includes\("nodes"\)\) return (?:NOT_PROVEN|notProven);/);
   assert.match(src, /if \(!unique\.includes\("nodes"\)\)/);
 });
 
--- a/browser_tests/unit/open-content-proof.test.mjs
+++ b/browser_tests/unit/open-content-proof.test.mjs
@@ -96,6 +96,9 @@ test("#1001 a byte-identical repaint is still reported as exact", () => {
     proven: true,
     exact: true,
     fields: [],
+    // #1623 — a graph that matched has no difference to classify, so the weaker
+    // "nothing authored was lost" ground is not what carried it.
+    presentationOnly: false,
   });
 });
 
@@ -227,6 +230,10 @@ test("#1001 an unreadable root proves nothing — absence of comparison is not e
       proven: false,
       exact: false,
       fields: [],
+      // #1623 — and it must not answer the WEAKER question either. A root nobody
+      // could read supports "nothing authored was lost" exactly as little as it
+      // supports "the content was reproduced".
+      presentationOnly: false,
     });
   }
   assert.equal(graphRootReproducesStateContent({ rootGraph: rootOf(state), state: null }).proven, false);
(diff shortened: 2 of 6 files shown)


artokun:
## HOLD — the codex gate is INDETERMINATE, not passed

```
node .claude/autopilot/codex-gate.mjs --worktree .../1623-cosmetic-open --base origin/main --name 1623
exit 2
```

Confirmed directly rather than taken from the wrapper:

```
$ codex exec "Reply with exactly: GATE_PROBE_OK"
ERROR: You've hit your usage limit. Visit https://chatgpt.com/codex/settings/usage
       to purchase more credits or try again at Aug 19th, 2026 8:43 PM.
```

An indeterminate gate is not a pass, so this PR stays a draft and is **not** merged.
Nothing about the change is blocked — CI is green and the suite is at 4543/0 — only the
review is.

**What unblocks it:** codex credits, or the quota resetting on **Aug 19 2026 20:43**.
Re-run the gate command above; on exit 0 this is mergeable as-is.

This diff is worth a real adversarial read rather than a rubber stamp — it makes an open
that previously FAILED report success, which is the error direction that costs. The two
questions I would put to a reviewer first:

1. Can a load that died mid-`configure()` produce a difference confined to
   `{size, pos, order, color, bgcolor}`? The claim is no, because `configure()` writes
   `widgets_values`, `properties`, `title`, `flags` and `mode` in the same pass and every
   one of them is outside that set — and a load that aborts leaves every later node
   wholly unconfigured. If there is a frontend path that writes the cosmetic five and
   then throws before the rest, this gate is wrong.
2. `color`/`bgcolor` are **authored**, not recomputed. They ride along because the claim
   licensed is only "no node and no value was lost", and every differing field is named
   on the reply. If that is too generous, the narrowing is to split them out of the
   pass/fail set while leaving them cosmetic for the disclosure.

PR #1277, opened 2026-08-15
fix(1168): a wedged orchestrator's death is recorded where the panel concludes it

### LINKED ISSUE #1168: A WEDGED orchestrator's death is never recorded, so a turn it was running is never resumed
Split out of #1163 after two attempts to fix it from the teardown side each introduced a worse bug. Nothing currently ships for this — the gap is real but **unfixed by design**, pending a proper one.

## The gap

The panel decides a turn needs resuming from how long the bridge was down, and the outage clock is started by the socket `close` listener. A **wedged** orchestrator never closes its socket: it holds the connection open and answers nothing. So when the panel gives up and force-respawns it, no outage was ever recorded, the fresh handshake measures nothing, the `ready` ack reads no drop, and the turn the respawn killed is never resumed. The user watches the indicator stop and the agent never picks the work back up.

Same user-visible outcome as #1145, reached through the recovery path instead of the backoff.

## Why the obvious fixes are wrong — both already tried

**Attempt 1: record the outage in the shared teardown helper** (`stop()`/`setUrl()`/`destroy()`). Those carry roughly seventeen call sites: Disconnect, provider switch, soft reload, hard restart, the secure-tunnel re-point, every autoconnect retarget, and the user editing the bridge URL. On most the orchestrator is alive, and on several a turn is still armed — `connectBackend()` clears the mid-task marker only when `switching`, `hardRestart()` only on its success branch (#1166). Review found this would nudge a working agent after a refused soft reload (the 10s POST timeout is measured as downtime) and after a failed hard restart. That converts a MISSED nudge into a FALSE one, which is strictly worse.

**Attempt 2: record it in the two recovery functions.** `tryAutoReclaim` is defensible, but `tryHandshakeRedial` is not a death at all — its own doc says it is the *agent-not-ready-**yet*** case: "the WS opened before the agent spawned, so no `models` arrived on this socket." Nothing dropped and nothing was killed. Stamping there records a phantom outage on an ordinary slow first connect.

## What a real fix has to establish

- **Distinguish "the orchestrator DIED" from "it has not booted yet" and from "the user re-pointed the bridge."** The teardown functions cannot tell these apart, which is why both attempts failed there. The decision has to be made where the death is actually concluded — and `tryHandshakeRedial` proves not every handshake timeout is one.
- **Confirm the reclaim really kills anything.** Review flagged that this pack's `/comfyui_mcp_panel/connect` may ignore `force`, so "the reclaim killed the wedged orchestrator" may be false for the published pack. Worth checking before building on it.
- **Probably a different contract than `noteBridgeClosed()`.** That name means "a socket closed"; this needs "the panel concluded the agent behind this socket is gone". Overloading the socket-event name is what made both attempts read as plausible while being wrong.
- **Interaction with #1166.** A stale mid-task marker on the hard-restart failure branches would turn any outage recorded across a restart into a false nudge, so that likely wants fixing first.

## Not urgent

The failure is a missed nudge in a comparatively rare path (a wedged orchestrator the panel force-respawns mid-turn). Every attempt so far has traded it for a false nudge in common paths, which is the worse direction — an idle agent is recoverable by the user, a working agent told to start over is not.

artokun:
Closes #1168.

## The gap, restated in terms of what this build can actually do

The panel decides a turn needs resuming from **how long the bridge was down**, and that
clock is started by the socket `close` listener. A **wedged** orchestrator never fires
one: it holds the connection open and answers nothing. So the outage its death caused is
not merely mismeasured — **it is never opened at all**. The clock starts only when the
wedged process eventually exits, so a user who restarts the agent promptly measures the
**restart**: three seconds, under the 6s threshold, no nudge, turn stranded.

The issue's own triage comment established the two facts that change the shape of the fix,
and I re-verified both:

- `externalOrchestratorMode()` returns a hardcoded `true`, so `onHandshakeTimeout` always
  takes the external branch and `tryAutoReclaim` is unreachable (it re-guards on the same
  predicate at its first line).
- `force` appears **0 times** in `__init__.py` — the pack's `/connect` ignores it entirely.

So the panel **cannot respawn the orchestrator**, and no fix can be built on "record the
death when the panel kills it". The death must be concluded from the panel's own
observation or not at all.

## The fix

`noteAgentGone()` — a new tracker event — **opens an outage exactly as a close does**, at
the one place the panel concludes the agent behind the socket is gone.

That place is `showExternalHintOnce()`'s latch, and it is chosen rather than convenient.
Everything above it is the ladder of reasons the conclusion is *not earned yet*: the
patient cold-start window, the bounded handshake redial (the agent-not-ready-***yet***
case), and #1136's fallback dial for a configured URL that outlived its process. #1136
moved that latch there precisely to stop the panel over-claiming "no agent is listening",
and concluding a death is the identical claim. Stamping earlier is what sank both previous
attempts.

It is a separate method from `noteBridgeClosed()` for the **decision**, not the effect: a
close is *observed*, this is *concluded*, and only the caller can be trusted to have earned
it. Keeping the names apart is what stops this from being spelled the way the rejected
attempts were.

| | stamped at | why it was wrong |
|---|---|---|
| Attempt 1 | shared teardown helper | ~17 call sites (Disconnect, provider switch, soft reload, hard restart, tunnel re-point, autoconnect retarget). Most are a **live** orchestrator being re-pointed. |
| Attempt 2 | `tryHandshakeRedial` | Its own doc calls it the agent-not-ready-***yet*** case. Nothing dropped, nothing was killed. |

## What it deliberately does NOT do — and the review round that established that

My first draft passed the conclusion into `shouldNudgeAfterMidTaskReconnect` as a second
input that **outranked the clock**. The gate returned NO-SHIP, correctly.

The panel cannot distinguish the two states that latch collapses. *"The orchestrator died"*
and *"my own socket went stale while it kept working"* — a half-open TCP after a laptop
sleep, a NAT idle-kill, a VPN flap — both present as an open socket answering nothing, and
both reach the same give-up. With the threshold bypassed, the second sends *"your
connection dropped mid-task … continue exactly what you were doing"* into a turn that never
stopped: the exact harm both earlier attempts were rejected for, and a behaviour change
from `main`.

The draft's safety argument was that a surviving turn's `turn:working` re-announce would
stand the conclusion down first. **This repo already records that no such ordering is
guaranteed** — the #1163 test carries a comment saying it was rewritten specifically to
stop pinning that assumption, and `session-rebind.js` notes ten lines below that `ready`
goes out synchronously while the models frame comes from an async continuation.

The gap this issue reports is an outage the predicate **never got to see**, not one it
judges wrongly. So the reader is unchanged and the interval remains the sole gate:

- a wedge is now measured from where the panel concluded it (**~123s** in the reported
  sequence) instead of from the wedged process's exit (**~3s**);
- a conclusion the panel got **wrong** is measured too — a live orchestrator that answers
  again moments later measures short and is left alone;
- a genuine wedge requires the user to read the hint and restart the agent by hand, which
  never fits inside the 6s window.

A conclusion is evidence that an outage **began**, never that it was long enough to have
killed the turn.

#1166, which the issue named as a prerequisite (a stale mid-task marker on hard-restart
failure branches), is **closed** on main.

## Verification

Mutation, not a green suite. Each of these turns the suite **red**:

| mutation | result |
|---|---|
| delete the write site | ✗ red |
| method kept but emptied (records nothing) | ✗ red |
| drop the first-opener guard (a conclusion re-stamps a running outage) | ✗ red |
| turn start no longer closes an outage a conclusion opened | ✗ red |
| record moved **above** #1136's fallback (attempt-2's over-claim) | ✗ red |

One equivalent mutant survived and is not a gap: removing the `t !== null` clock guard is a
no-op because `readClock()` already normalises a non-finite reading to `null`.

The regression test for the gate's P1 pins the reviewer's exact sequence and **relies on no
frame ordering at all**. The wiring test now also asserts the predicate is called with the
outage *alone*, so the rejected shape cannot return silently. The call-site pin counts
occurrences (`=== 1`) rather than scanning teardown bodies — the previous negative pin
searched `stop()/setUrl()/destroy()` for a literal `bridgeOutage.` and would have **passed**
against the very design it named.

- `npm run test:unit` — **4532 pass, 0 fail**
- `check-panel-scope` — OK (a fresh worktree has no `node_modules`, and this gate silently
  skips without one; I linked one in. It is the only thing that catches a name resolving in
  a sibling function)
- `tsc --noEmit` — clean

## Out of scope, stated plainly

An **already-handshaken** socket that wedges still has no detector — the handshake timeout
only arms on a socket that opened and never handshook. Closing that needs a heartbeat,
which is a new mechanism under the freeze, not this bug.

---

adversarial review instead. That review returned NO-SHIP on the first draft; the finding is
addressed above and re-gated.

### DIFF
--- a/browser_tests/unit/session-rebind.test.mjs
+++ b/browser_tests/unit/session-rebind.test.mjs
@@ -608,7 +608,7 @@ test("#1138 wiring: the call site NEGATES the predicate and returns", () => {
   const src = readFileSync(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url), "utf8");
   assert.ok(
     src.includes(
-      "if (!shouldNudgeAfterMidTaskReconnect({ outageMs: bridgeOutage.outageMs() })) return;",
+      "if (!shouldNudgeAfterMidTaskReconnect({ outageMs: bridgeOutage.outageMs()",
     ),
     "the ready-ack mid-task branch must NEGATE the predicate and read the measured outage",
   );
@@ -638,13 +638,15 @@ test("#1138 wiring: the call site NEGATES the predicate and returns", () => {
   ]) {
     assert.ok(src.includes(snippet), site);
   }
-  // NOTE: two attempts to also pin the WEDGE-death paths were removed with the code
-  // they described. A wedged orchestrator holds its socket open and never fires a
-  // close, so nothing records that death — a real gap, tracked separately. It is not
-  // pinned here because the assertions that tried to were themselves wrong: the
-  // negative half scanned stop()/setUrl()/destroy() for a literal `bridgeOutage.`,
-  // while the design it claimed to reject put the stamp in a shared helper CALLED
-  // from those bodies — so it would have passed against the very regression it named.
+  // The WEDGE-death path is a FOURTH write site and is pinned separately, at the end of
+  // this file (#1168). It is not listed here because it is not a close: a wedged
+  // orchestrator holds its socket open and never fires one, so the event it records is
+  // the panel's own conclusion rather than a socket transition. Two earlier attempts to
+  // pin it here were removed with the code they described — the negative half scanned
+  // stop()/setUrl()/destroy() for a literal `bridgeOutage.`, while the design it claimed
+  // to reject put the stamp in a shared helper CALLED from those bodies, so it would
+  // have passed against the very regression it named. #1168's pin counts the call
+  // instead, which a wrapper cannot evade.
 });
 
 // ── #1145: the interval must be the OUTAGE, not one backoff step ─────────────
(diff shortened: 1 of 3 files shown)


artokun:
## Verification pass — merge HOLDS on the gate, not on the change

Picked this up to drive it to a terminal state. Everything that does not depend on codex is done;
the gate itself cannot run, so it is not merged.

### 1. The fix is reachable from production — traced, not assumed

The one thing the unit tests cannot see is whether `noteAgentGone()` is ever the **first** opener of
an outage. It is, and the reason is in the close listener's own note: a panel-driven teardown nulls
`sock` synchronously, so its late `close` fails the `isActive()` guard and records nothing.

- `tryHandshakeRedial` recovers via `client.setUrl(currentUrl())` (`comfyui-mcp-panel.js:30466`),
  and `setUrl` nulls `sock` before the close event lands (`:19912`–`:19930`). **The redial ladder
  opens no outage.**
- So the handshake-timeout caller reaches `showExternalHintOnce()` with the clock still unstarted:
  WS open, port held by a wedged orchestrator, `MID_TASK_KEY` armed. On `main` the eventual
  handshake calls `noteHandshake()` with `startedAt === null` and records nothing — `outageMs() === 0`,
  no nudge. With this change the outage runs from the conclusion. That is the reported bug.
- The other caller is safe for the stated reason, and it is structural rather than incidental:
  `emitStatus("disconnected")` is only reached from `scheduleReconnect` (`:19680`–`:19682`), which
  only runs from the `!closed` branch of a **real** close — so an outage is always already open and
  `noteAgentGone()` is a no-op. A user Disconnect sets `closed = true` and never gets there, so this
  does **not** open the door attempt 1 was rejected for.

### 2. Mutation-verified independently (5/5 killed)

Re-ran the PR's table from scratch against `browser_tests/unit/session-rebind.test.mjs`:

| mutation | result |
|---|---|
| delete the write site | ✗ red (2 fail) |
| method kept but emptied | ✗ red (2 fail) |
| drop the first-opener guard | ✗ red (1 fail) |
| turn start no longer closes a concluded outage | ✗ red (4 fail) |
| record moved above #1136's fallback | ✗ red (1 fail) |

### 3. Suite and CI

- `npm run test:unit` — **4533 tests, 4532 pass, 0 fail** (`check-panel-scope` included, and this
  worktree has `node_modules`, so it did not silently skip)
- `tsc --noEmit` — clean
- CI on `fix/1168-wedge-death-recorded` — **success**

### 4. Why it is not merged

```
node .claude/autopilot/codex-gate.mjs --worktree …/wt-1168 --base origin/main --name 1277
[gate] INDETERMINATE — codex account quota exhausted. Merge HOLDS.   (exit 2)
```

Confirmed directly, not just from the gate's pattern match — a bare `codex exec` returns
`You've hit your usage limit … try again at Aug 19th, 2026 8:43 PM`, and neither phrase appears
anywhere in the diff, so this is not a false positive. Exit 2 is **not** a pass.

Left open and unmerged. Re-run the gate on or after **2026-08-19**; nothing else is outstanding.

Issue #1184, opened 2026-08-13
connectBackend commits the new backend before it can fail, so an aborted switch leaves the panel claiming a provider it never connected to


artokun:
## Summary

`connectBackend()` commits the new backend to memory, to `localStorage` and to the UI **before** it checks whether the old provider's session could be durably invalidated. When that check fails it returns — leaving the panel persistently claiming a backend it never connected to.

Found while reviewing #1171 (PR #1183) and confirmed independently by review. **Pre-existing**, and deliberately not fixed there: #1171 is a one-flag fix to the reload re-entrancy guard, and rolling back a half-committed backend switch is a different change to a different state machine.

## The ordering

In `connectBackend()` (`web/js/comfyui-mcp-panel.js`), before the guard:

| step | effect |
|---|---|
| `seedPrefsForBackendSwitch(id)` | prefs reseeded for the new backend |
| `selectedBackend = id` | committed in memory |
| `localStorage.setItem(STORAGE_KEY_BACKEND, id)` | **persisted across reloads** |
| `refreshModelChip()` | chip repainted to the new provider |
| `endTurnLocally()` | working indicator cleared |
| `client.armContext(replay)` | fresh-session replay armed |
| `if (!await invalidateDurableAgentSession()) { … return; }` | **aborts here** |

## Consequences

- The chip and model chip show the new provider; the panel is not connected to it.
- `STORAGE_KEY_BACKEND` persists the new choice, so a reload adopts a backend the user never successfully switched to.
- The armed one-shot replay — the *"Conversation so far — continued in a fresh AI session"* preamble — stays armed against the **old** provider's still-live session, so the next message is prefixed with a fresh-session block for a session that never changed.
- `endTurnLocally()` has already cleared the working indicator for a turn on the old provider that may still be running.

Until PR #1183 this return was also completely **silent** — no transcript line at all. That PR adds the same disclosure `hardRestart` uses, which makes the pause visible but does not make the state coherent, because the state was already committed.

## What a fix has to decide

Not obvious, which is why it is filed rather than folded in:

1. **Roll back** — restore `selectedBackend`, `STORAGE_KEY_BACKEND`, the prefs seed and the chip to the previous backend, and drop the armed replay. Most correct, and the most state to get right.
[...]

PR #1278, opened 2026-08-15
fix(1198): the Settings path must not leave the saved default naming a backend it never reached

### LINKED ISSUE #1198: The Settings backend path commits the durable default before #1184's guard can abort
## Summary

#1184 stops `connectBackend()` committing the new backend before the old provider's session is durably invalidated. It cannot cover the **Settings** path, because that path commits *before* `connectBackend` is ever called — and the commit it makes is the durable one.

Found while reviewing PR #1196 (#1184) and confirmed from source. Filed separately because the fix is not a reordering and the obvious repair re-enters a loop the code explicitly warns about.

## The ordering

The Settings dropdown is a ComfyUI setting. ComfyUI **persists the value, then notifies**:

```js
onChange: (v) => {
  if (suppressSettingOnChange || !settingsArmed) return;
  panelHooks.applyBackend?.(v);
}
```

`applyBackend` writes a transcript line and calls `connectBackend(id)`. So by the time #1184's guard can abort, `SETTING_BACKEND` — the *saved default*, which outlives the tab and wins on a fresh profile — already names a backend the panel may never connect to.

The chip and model-popover paths do not have this problem: they call `connectBackend` directly and #1184 covers them.

## Why the obvious fix is wrong

Restoring the setting means calling `setSetting(SETTING_BACKEND, previous)`, which fires `onChange` again. The panel already has a scar from exactly this:

> the old `setSetting(SETTING_BACKEND, id)` re-entered through `SETTING_BACKEND.onChange` → `applyBackend` → `connectBackend` → `setSetting` → … (ComfyUI fires onChange async, AFTER `setSetting`'s `suppressSettingOnChange` has already reset), so each switch overlapped multiple connects and the bridge's close-old-on-new-hello looped

That is the 9181 "ready"/"waiting" storm. A rollback here walks straight back into it, and `suppressSettingOnChange` is documented as **best-effort** precisely because the notification is async.

## Options

1. **Validate before ComfyUI persists.** Needs a pre-change hook; ComfyUI's settings API notifies after the fact, so this may not be expressible.
2. **Roll back through a re-entrancy-safe path** — a switch generation counter rather than the best-effort suppress flag, so the echo is *identified* instead of *suppressed*. Closest to correct; the most state to get right.
3. **Accept and disclose.** The panel already says "Default backend → X" up front; if the switch then aborts, say so. Cheapest, and honest, but leaves the saved default wrong.
4. **Decide legality earlier.** `applyBackend` could run the invalidate itself before touching anything, and simply not call `connectBackend` on failure — the setting is still wrong, but nothing else is.

## Reproducing

Needs `invalidateDurableAgentSession()` to answer false, which needs the history store's write to genuinely fail. Note a merely *slow* store does not do it: a capped IndexedDB open yields `null` and `flush()` maps null to true. The real trigger is history past the local shadow's limits — `LOCAL_SHADOW_THREADS = 20` / `LOCAL_SHADOW_MESSAGES = 200` — plus a disk hiccup inside the 2s open cap.

(Those are the limits; an earlier comment of mine in the panel cited 400/30, which were probe *points* from a measurement table, not the boundaries. Corrected in #1186.)

artokun:
Fixes #1198.

## What was wrong

ComfyUI writes `SETTING_BACKEND` and only *then* notifies the panel, so #1184's guard runs far
too late to stop it: by the time `connectBackend` can abort, the **saved default** —
`comfy.settings.json`, not browser-scoped, so it wins on a fresh profile where no runtime pick
shadows it — already names a backend the panel may never connect to.

## What ComfyUI actually does

Read out of the shipped frontend (`src/platform/settings/settingStore.ts`, via the sourcemap
beside the bundle) rather than assumed, because all three facts are load-bearing and one of them
contradicts the issue:

* **(A)** `onChange` is dispatched **synchronously** from inside `setSettingValue`, and is handed
  the **previous** value as its second argument — which this handler had always dropped, and
  which is the only record of what to restore to.
* **(B)** Writing an unchanged value is a complete no-op — no notification, no server call.
* **(C)** The server write happens **after** the notification, not before. The issue says
  "persists, then notifies"; it is the other way round. That does not rescue anything —
  `onChange` cannot veto and its return value is discarded — but it is why option 1 (validate
  before ComfyUI persists) is not expressible: there is no pre-change hook, only a notification
  already too late to refuse.

## The fix — the issue's option 2

The reason option 2 was called "the most state to get right" is that writing this setting is what
caused the 9181 "ready"/"waiting" connect storm. Two things keep it out of that loop:

* a **compare-and-swap**, atomic for free because the read and the write sit in one synchronous
  block. A rollback that finds the setting has moved on does nothing rather than clobbering a
  newer pick;
* the echo is **identified, not suppressed**. The module marks the value it is about to write and
  consumes the matching notification. A value marker is correct whether dispatch is synchronous
  (today) or deferred (any future frontend), which is exactly what `suppressSettingOnChange` — a
  time window, documented as best-effort — cannot promise. `isSelfWrite` is therefore asked
  **before** that flag, or the flag swallows the echo and strands the marker.

Two things the tests found that reading did not:

* restoring to `previous` is wrong when two Settings switches overlap and both abort (a wedged
  history store fails every switch, so trying twice hits it every time): the second rollback's
  `previous` is the first switch's un-reached backend, landing the default on this very defect one
  step down. A superseded rollback now hands its resting place to the switch that superseded it,
  so a chain of any depth collapses to the last backend actually reached;
* "Default backend → X" was printed before the switch was even attempted, so an abort left the
  transcript asserting a durable change the panel then undid. It is announced once it is true. No
  new catalog key — the abort already discloses, and English is frozen (#1135).

`connectBackend` returns the whole result now: `switched` is false for **both** an abort and an
ordinary first connect, and only `reason` separates them — rolling back on the boolean would wipe
the default a first-connect user just chose.

Restoring to `selectedBackend` was considered and rejected: a chip pick moves the runtime backend
*without* touching the saved default (#1184's FIX 1), so the two legitimately diverge and that
would write a default the user never chose.

## Verification

**The three ComfyUI facts were re-checked against the frontend installed on this machine**, not
just cited: `onChange(setting, newValue, oldValue)` dispatches `setting.onChange(newValue,
oldValue)` synchronously; `if (newValue === oldValue) return undefined` precedes any notification;
`await api.storeSetting(...)` follows it; and `get()` falls back through `getDefaultValue()`, so
`oldValue` is always a non-empty string for a setting declaring `defaultValue: "claude"` — the
`NO_PREVIOUS` branch really is belt-and-braces for a different frontend, as the comment claims.

**Mutation-tested — the suite is red for each of the five ways this fix could be undone:**

| Mutation | Tests that go red |
| --- | --- |
| drop the `rollback` call in `applyBackend` | 5, incl. the primary defect test |
| branch on `switched` instead of `reason` | first-connect keeps its default; wiring |
| ask `suppressSettingOnChange` before `isSelfWrite` | the 9181 storm test; wiring |
| remove the compare-and-swap | 4, incl. "never clobbers a newer pick" |
| remove the superseded-chain resolution | two overlapping aborted switches |

24 new tests drive the **real** `onChange` and **real** `applyBackend`, extracted from the panel
source, over a model of the real ComfyUI settings store — in both synchronous and deferred
notification modes — and count connects, so a rollback that re-entered would show up as a second
connect.

Full unit suite green (4564 pass / 0 fail) after rebasing onto current `main`, which had itself
landed +172 lines in the panel monolith these tests regex over. CI Lint pack green.

### DIFF
--- a/browser_tests/unit/settings-backend-default.test.mjs
+++ b/browser_tests/unit/settings-backend-default.test.mjs
@@ -1,0 +1,546 @@
+// #1198 — the Settings path commits the durable default before #1184's guard can abort.
+//
+// #1184 stopped `connectBackend()` committing anything until the old provider's session was
+// durably invalidated. The Settings dropdown slips past that entirely, because ComfyUI has
+// already written `SETTING_BACKEND` by the time it notifies the panel at all:
+//
+//   combo → ComfyUI writes the SAVED default → onChange → applyBackend → connectBackend →
+//   #1184's guard aborts → the saved default still names a backend never connected to.
+//
+// `STORAGE_KEY_BACKEND` (which #1184 leaves alone on an abort) shadows it on this machine,
+// so the damage only surfaces where there is no runtime pick to shadow it: a fresh profile,
+// another browser, a cleared site. That is what makes it worth a test rather than a look.
+//
+// THE HARD PART IS NOT THE ROLLBACK, IT IS ROLLING BACK WITHOUT THE STORM. Writing this
+// setting is what caused the 9181 "ready"/"waiting" connect storm, so the tests below drive
+// the panel's REAL onChange and REAL applyBackend over a model of the REAL ComfyUI settings
+// store, and count connects. A rollback that re-entered would show up as more than one.
+
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+
+import { BACKEND_SWITCH } from "../../web/js/lib/backend-switch.js";
+import {
+  SETTINGS_BACKEND_ROLLBACK,
+  createSettingsBackendDefault,
+  planSettingsBackendRollback,
+} from "../../web/js/lib/settings-backend-default.js";
+
+const PANEL_LF = readFileSync(
+  fileURLToPath(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url)),
+  "utf8",
+).replace(/\r\n/g, "\n");
+
+const SETTING_BACKEND = "comfyui-mcp.defaultBackend";
+
+// ---------------------------------------------------------------------------
+// A model of ComfyUI's settings store, transcribed from the shipped frontend.
+//
+// From `src/platform/settings/settingStore.ts` (ComfyUI_frontend 1.50.3, read out of the
+// sourcemap shipped beside the bundle). All three behaviours below are load-bearing for
+// the fix, so they are modelled rather than assumed:
+//
+//   (A) `onChange(newValue, oldValue)` is dispatched SYNCHRONOUSLY from inside
+//       `setSettingValue`, and it is handed the PREVIOUS value;
+//   (B) `if (newValue === oldValue) return undefined` — an unchanged write is a complete
+//       no-op: no notification, no server call;
+//   (C) `settingValues[key] = newValue` and `await api.storeSetting(...)` both happen
+//       AFTER the notification, so a handler still reads the OLD value while it runs.
+//
+// `deferNotify` models a frontend that delivers the notification later instead. Nothing in
+// the settings API promises (A), and `suppressSettingOnChange` is documented in the panel as
+// best-effort for exactly that reason — so the fix has to hold in both modes, and is tested
+// in both.
+// ---------------------------------------------------------------------------
+function createComfySettingStore({ initial = {}, deferNotify = false } = {}) {
+  const values = { ...initial };
+  const handlers = new Map();
+  const serverWrites = [];
+  const queue = [];
+  return {
+    register: (key, onChange) => handlers.set(key, onChange),
+    get: (key) => values[key],
+    serverWrites,
+    /** ComfyUI's `settingStore.set`, in the order the real one does it. */
+    set(key, value) {
+      const oldValue = values[key];
+      if (value === oldValue) return; // (B)
+      const notify = () => handlers.get(key)?.(value, oldValue); // (A)
+      if (deferNotify) queue.push(notify);
+      else notify();
+      values[key] = value; // (C) — after the notification, not before
+      serverWrites.push({ key, value });
+    },
+    /** Deliver whatever a deferred frontend would have delivered later. */
+    drain() {
+      while (queue.length) queue.shift()();
+    },
+  };
+}
+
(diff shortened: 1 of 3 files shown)


artokun:
## Status: verified and CI-green, but the merge is **held by the codex gate**

Picking this up where it was left. Summary of what changed and what the gate said.

**Rebased onto current `main`** (`9a19d06`). `main` had landed +172 lines in
`web/js/comfyui-mcp-panel.js` since this branch was cut, and this PR's tests extract the real
`onChange` / `applyBackend` bodies by regex over that file — so the rebase was done first to make
sure the suite was verified against post-merge content rather than a stale base. Full unit suite
green afterwards: **4564 pass / 0 fail**. CI Lint pack green on `aae5ba3`.

**Verified the fix, not just the suite.** Details in the PR body; the two things worth repeating:

* the three load-bearing claims about ComfyUI's settings store were re-read from the frontend
  installed on this machine, and all three hold as transcribed — including the one that
  contradicts the issue (the server write happens *after* the notification, not before);
* five separate mutations, one per load-bearing decision, each turn the suite red. The fix is not
  passing by construction.

**Gate: `INDETERMINATE` (exit 2). Not merged.**

```
```

Exit 2 is not a pass and not a fail — the gate could form no opinion, because the codex account is
out of quota. Confirmed directly rather than trusting the gate's pattern match:

```
$ codex exec --skip-git-repo-check "Reply with exactly: PROBE_OK"
ERROR: You've hit your usage limit. … try again at Aug 19th, 2026 8:43 PM.
```

So this is the weekly limit, not the short session window — a retry before **Aug 19 20:43 PT**
will hit the same wall. (The gate's own log also contains `session limit · resets 4pm` lines from
the surrounding run; the direct probe above is the authoritative one.)

**Left open and left as a draft on purpose** — draft status is what keeps it from being merged
while the gate is holding it. The work itself is complete; nothing here is waiting on more code.

**To finish:** re-run the gate after the quota resets, and merge on exit 0.

```
node C:/Users/Artokun/code/comfyui-mcp/.claude/autopilot/codex-gate.mjs \
  --worktree C:/Users/Artokun/code/comfyui-mcp-panel/.claude/worktrees/1198-settings-backend \
  --base origin/main --name 1278
```

Note for whoever does: #1198 is a same-repo owner-qualified reference, so squash-merging this will
**not** auto-close it. Close it by hand.

PR #1279, opened 2026-08-15
fix(1209): refresh a drifted graph fence before the next call, not after it refuses

### LINKED ISSUE #1209: Refresh graph fence after active workflow change event
After the panel injected an ACTIVE WORKFLOW CHANGED event, the next read-only panel_graph_outline call was still stamped with the previous workflow-instance fence and failed closed. The session already knew the new active workflow, so requiring an explicit panel_set_workflow_target(mode: current) retry is avoidable friction.

### Steps to reproduce
1. Keep a panel agent session open.
2. Switch to another workflow tab.
3. Immediately call panel_graph_outline from the resumed turn.
4. Observe workflow instance mismatch; call panel_set_workflow_target(mode: current), then retry successfully.

### Exact error
workflow instance mismatch: command issued for the previous workflow instance while the active canvas reports the new instance. Nothing was applied.

### Fix
applied locally: no
upstream-only: yes
Precise change needed: when the orchestrator injects an active-workflow-changed event into the turn, refresh that session's graph-command fence to the reported live workflow UUID before the next read-only graph call. Alternatively, auto-retry read-only panel graph operations once after a confirmed active-tab event. Mutations should remain fail-closed.

### Environment
Windows 11; ComfyUI 0.30.2; NVIDIA RTX 5090 32GB; comfyui-mcp 0.50.1; panel package reported installed 0.13.7.

artokun:
Fixes #1209

## What was wrong

The orchestrator stamps every `panel_*` graph command with the workflow instance uuid the panel last advertised in a hello, and the panel refuses a command whose stamp no longer names the live canvas (`workflow instance mismatch`).

**#607 made the recovery reachable, not unnecessary.** Its re-hello is EDGE-triggered on an actual refusal, so the first graph call after the panel's identity moved still fails — and both reporters spent a `panel_set_workflow_target({mode:"current"})` round-trip on a fence the panel could already see was stale.

## Why the existing switch-path re-hello does not cover it

`onWorkflowMaybeChanged` re-hellos only when `workflowTabId()` changes. That id is the saved **handle** (`wf:<path>`), and it is UNCHANGED when a workflow is replaced *in place* under the same tab — loaded into the open canvas, reopened at the same path, restored by ComfyUI at startup. `workflowStableUuid()` is per-**instance**, and the `loadGraphData` wrapper re-mints it on exactly those replacements. So the fence identity moves while the route the poll watches does not, and nothing re-advertises.

That is precisely the second report in the thread: the **first** `panel_graph_outline` of a brand-new session, refused (`58ec734e…` vs live `ed48682a…`) with no tab switch and no prior mutation.

## The change

`web/js/comfyui-mcp-panel.js` — three edits:

1. `noteWorkflowIdentityDrift()` + hook, alongside the existing `noteWorkflowInstanceMismatch()`.
2. The hook itself, registered in the bridge client: compares the live `workflowStableUuid()` against `lastAdvertisedWorkflowUuid` — the panel's own record of what actually reached the orchestrator, published only by a hello that made it to the wire. Those differing IS the stale fence, observed from the only side that holds both halves. The send is the bounded #607 re-hello, unchanged.
3. The 600 ms workflow poll calls it on the path where the route did **not** change.

**The fence is not weakened.** Mutations stay fail-closed against whatever stamp is current; this only replaces a stamp the panel can *prove* is out of date. The issue asked for read-only auto-retry as an alternative — this is strictly narrower: nothing is retried, one advertisement is corrected.

### The proactive path needs a bound of its own

This is not just the #607 hook wired to the poll, and finding that out is what the first round of tests caught. #607's budget is keyed on the LIVE identity and resets whenever it changes; under a churning identity that is every tick, so a level-triggered caller re-hellos forever — a greeting storm, the same failure the churning `tmp:` id produced before it was keyed on the workflow object. The drift path therefore counts consecutive attempts that did not **converge**, lets only convergence replenish them, and charges only a re-hello that was actually **dispatched** (`workflowInstanceMismatchRehello()` now returns whether it sent). An identity that never settles goes quiet after three tries, leaving today's #607 recovery untouched.

## Evidence

`browser_tests/unit/binding-recovery.test.mjs` — 9 new tests in the #607 cluster, extracting the shipping hook verbatim.

**Verified by mutation** — each of these turns the suite red, and reverting turns it green:

| mutation | result |
|---|---|
| remove the poll call site | ✖ 1 fail (the wiring test) |
| no-op the drift hook | ✖ 5 fail |
| remove the drift bound | ✖ 3 fail |
| charge the bound for undispatched attempts | ✖ 7 fail |

- `npm run test:unit` — **4535 tests, 0 fail**.
- The 12 pre-existing #606/#607 tests in the same file still pass — `workflowInstanceMismatchRehello`'s behaviour is unchanged, only its return value is new.

## Not done

- **No browser/e2e run — it could not be run.** The suite does not start ComfyUI (`playwright.config.ts`: *"PREREQUISITES (this suite does NOT start ComfyUI for you)"*), nothing was listening on `localhost:8188`, and no ComfyUI was allocated to this run. Even with one up, the browser loads the pack junctioned into `custom_nodes` — the main checkout, not this worktree — so the run would not have exercised this change; and the tier-1 MockBridge fixtures model no in-place workflow replacement with a real hello round-trip. Stating this rather than implying coverage I do not have.
- What *does* cover the browser-side risk: CI's JS-syntax lint, and `check-panel-scope` — the gate that catches a new module-level binding resolving in a sibling scope, which is a `ReferenceError` the moment the line runs. It needed `npm install` in the worktree to run at all (a fresh worktree inherits no `node_modules`, and it prints CANNOT RUN rather than a false OK); it reports OK. This change renders nothing — it sends one hello frame.
- The `Date.now()` calls in the #607 backoff are pre-existing and untouched; the drift path uses no clock.

---

**Gate disclosure: NOT GATED. This PR is held.**

The codex gate was re-run on head `c043916` on 2026-08-15 and returned **exit 2 (INDETERMINATE)** — the codex account is over its usage limit and does not reset until **2026-08-19 8:43 PM**. An indeterminate gate is not a pass, and no substitute is one either: the `claude-gate.mjs` run recorded here is an independent adversarial review, not the gate this repo merges on. Do not merge until the codex gate has actually returned exit 0.

Everything else is finished — see the verification comment below.

### DIFF
--- a/browser_tests/unit/binding-recovery.test.mjs
+++ b/browser_tests/unit/binding-recovery.test.mjs
@@ -705,3 +705,241 @@ test("#606 the binding refusal marks the reload remedy as REQUESTED, not confirm
     );
   }
 });
+
+// ---------------------------------------------------------------------------
+// #1209 — the fence is refreshed BEFORE the refusal, not only after it
+//
+// #607 made the recovery reachable. It is edge-triggered on an actual refusal, so
+// the first graph call after the panel's live identity moved still fails and the
+// reporter still spends a panel_set_workflow_target({mode:"current"}) round-trip on
+// a stale fence the panel could already see.
+//
+// It moves without a tab switch. The switch-path re-hello is gated on
+// workflowTabId() changing, and that id is the saved HANDLE (`wf:<path>`): a workflow
+// replaced IN PLACE under the same tab keeps it while workflowStableUuid() — the
+// per-instance fence identity — is re-minted. Two reporters hit exactly that, one
+// after a tab switch and one on the FIRST call of a brand-new session.
+// ---------------------------------------------------------------------------
+
+/**
+ * Extract the budget block, the #607 hook AND the #1209 drift hook together, so the
+ * delegation between them is the shipping one. Deleting the drift hook, or making it
+ * fire unconditionally, fails a test below.
+ */
+function buildDriftHook({ sendHello, stableUuid }) {
+  const declStart = SRC.indexOf("const MISMATCH_REHELLO_MAX_PER_IDENTITY");
+  assert.notEqual(declStart, -1, "the re-hello budget constants are missing");
+  const marker = "workflowIdentityDriftRehello = () => {";
+  const bodyStart = SRC.indexOf(marker);
+  assert.notEqual(bodyStart, -1, "the bridge client must register the drift re-hello hook");
+  // Same brace arithmetic as buildRehelloHook: the marker ENDS with the body's own
+  // "{", and the slice must end at that index plus the body length (using bodyStart
+  // silently lops the tail off and every assertion below becomes meaningless).
+  const bodyOpen = bodyStart + marker.length - 1;
+  const body = balancedFrom(SRC, marker, bodyOpen);
+  const slice = SRC.slice(declStart, bodyOpen + body.length);
+  assert.ok(
+    slice.includes("workflowInstanceMismatchRehello() === true"),
+    "the drift hook must reuse the bounded #607 send, and charge only a DISPATCHED one",
+  );
+  assert.ok(
+    slice.includes("driftRehelloUnconverged = 0;"),
+    "the extracted region is truncated — the drift bound's replenishment is missing",
+  );
+  assert.ok(
+    slice.includes("mismatchRehelloLandedCount += 1;"),
+    "the extracted region is truncated — the #607 landed bookkeeping is missing",
+  );
+  const factory = new Function(
+    "sendHello",
+    "workflowStableUuid",
+    `let workflowInstanceMismatchRehello = null;\n` +
+      `let workflowIdentityDriftRehello = null;\n${slice};\n` +
+      `return {\n` +
+      `  drift: workflowIdentityDriftRehello,\n` +
+      `  mismatch: workflowInstanceMismatchRehello,\n` +
+      `  advertise: (uuid) => { lastAdvertisedWorkflowUuid = uuid; },\n` +
+      `};`,
+  );
+  return factory(sendHello, stableUuid);
+}
+
+test("#1209 a fence that drifted under an unchanged route is re-advertised, unprompted", async () => {
+  // The reported shape: the orchestrator was told A at connect, the canvas is now B,
+  // and NOTHING has been refused yet. Before this fix the panel sat on that until a
+  // command failed; the reporter's next read-only panel_graph_outline was that command.
+  let live = "uuid-A";
+  const dbl = helloDouble(() => live);
+  const h = buildDriftHook({ sendHello: () => dbl.send(), stableUuid: () => live });
+  dbl.harness = h;
+  h.advertise("uuid-A");
+
+  h.drift();
+  await settle();
+  assert.equal(dbl.calls.length, 0, "a fence that still names the live canvas sends nothing");
+
+  live = "uuid-B"; // in-place replace: same wf:<path> route, fresh instance uuid
+  h.drift();
+  await settle();
+  assert.deepEqual(dbl.calls, ["uuid-B"], "the drifted fence is re-advertised without a refusal");
+
+  // …and it QUIESCES: once the advertisement lands, the panel and the orchestrator
+  // agree again, so the 600ms poll must stop sending. A re-hello per tick would be a
+  // greeting storm — the exact failure the tmp:-id churn produced.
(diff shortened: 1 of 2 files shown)


artokun:
## Verification pass — independently re-run, not taken from the description above

Picked this up as unfinished work and re-derived every claim rather than trusting the table in the body.

### The diff does what it says

The premise holds in the source, not just in the prose:

- `workflowTabId()` (`web/js/comfyui-mcp-panel.js:3221`) returns the saved **handle** `wf:<path>` for a persisted workflow — unchanged by an in-place replacement.
- `workflowStableUuid()` (`:2874`) is per-**instance**, and its own comment records that the `loadGraphData` wrapper "INVALIDATES a stale object-cache and re-mints whenever the content loaded into an existing instance changes (in-place replace)".

So the fence identity moves while the route the poll watches does not, and before this change nothing re-advertised. That is the gap, and it is the one both reporters describe.

The wiring is real, not theoretical: `setInterval(() => onWorkflowMaybeChanged(), 600)` at `:29046`, and the new call sits on the `wfid === currentWorkflowId` early-return path at `:27104`.

### Verified by mutation — my own harness, six mutations

Backed up by file copy and restored by copy (never `git checkout`, which would have discarded the working tree). First run reported "ANCHOR NOT FOUND" for every multi-line mutation: the checkout is CRLF (33820/33820 lines) and the anchors were authored with `\n`. Lifting the anchors to the file's own EOL is what made the harness real — worth stating, because a mutation that silently fails to apply reports as "not found", and a sloppier harness would have reported it as *green* and read as proof of the opposite.

| mutation | result |
|---|---|
| M1 remove the poll call site | ✖ 1 fail |
| M2 no-op the drift hook | ✖ 5 fail |
| M3 remove the drift bound | ✖ 3 fail |
| M4 charge the bound for undispatched attempts | ✖ 7 fail |
| **M5 drop the equality/convergence check (fire unconditionally)** | ✖ 3 fail |
| **M6 revert `workflowInstanceMismatchRehello`'s dispatch return** | ✖ 3 fail |

M1–M4 reproduce the body's table. **M5 and M6 are new** — they were not tested before and they close the two gaps that mattered most: that the hook is genuinely conditional (not a re-hello every tick that happens to look right), and that the `return true` added to the #607 hook is load-bearing rather than cosmetic. Both go red. The tree was confirmed clean afterwards and the suite green again.

### Suite and CI

- `npm run test:unit` in the worktree: **4535 tests, 0 fail, 1 todo**.
- CI on head `c043916`: **Lint pack — pass** (2m1s). That single job is the whole of CI here, and it runs `npm ci`, `npm run typecheck` and `npm run test:unit`, so the suite is genuinely gated, not just run locally.
- `check-panel-scope` passes, which is the check that catches the specific hazard this diff creates — a new module-level binding (`workflowIdentityDriftRehello`) resolving in a sibling scope, a `ReferenceError` the moment the line runs.

### Two risks I chased down and am satisfied by

**A proactive full hello could re-greet the user** ("agent ready" — the storm the title-update path exists to avoid). It does not: the drift path delegates to the *same* `sendHello` the tab-switch re-hello and the #607 refusal path already call, and `5f0d1ac` (#1138, "a live-socket re-advertise must not read as a fresh connect") is already in `main`. This adds no new greeting surface.

**`workflowStableUuid()` is now called every 600 ms**, and it is not a pure read — it mints, calls `setWorkflowObjectUuid`/`rememberWorkflowUuidOwner`, and can write into `extra`. It is idempotent once the object cache is seeded (`objectUuid` short-circuits both the mint and the whole fork branch, which is gated on `!objectUuid`), so steady-state cost is a WeakMap lookup. It is also gated behind `lastAdvertisedWorkflowUuid` being set, so it does not run at all before a hello has landed.

### Why this is not merged

The codex gate was re-run on this head and returned **exit 2 — INDETERMINATE**:

```
```

Confirmed directly against the CLI rather than inferred from the gate's own classifier:

```
ERROR: You've hit your usage limit. ... try again at Aug 19th, 2026 8:43 PM.
```

Indeterminate is not a pass. The work is complete and I would otherwise have merged it; it is held solely on the gate and should be re-gated after **2026-08-19 8:43 PM**, then squash-merged if it returns exit 0. Left as a draft, matching every other open PR here.

`#1209` is left **open** for the same reason — the fix is not on `main`, so there is nothing to close it on.

Issue #342, opened 2026-07-30
graph_set_widget accepts out-of-enum combo values and silently drops links on dynamic-combo slots


artokun:
### What happened / root cause

`graph_set_widget` (panel handler, `web/js/comfyui-mcp-panel.js`) assigns `w.value = value` with **no validation against the widget's own option list**, then fires `w.callback?.(...)`.

For a `COMFY_DYNAMICCOMBO_V3` widget, that callback rebuilds the node's input slots based on the selected key. If the value is not a valid key, the combo collapses: the slots it owned are **removed entirely**, and any links attached to those slots are **silently dropped**. No error, no warning — `graph_set_widget` returns success.

The resulting graph *looks* wired in `panel_graph_outline` (the outline renders the orphaned link against whatever slot now occupies that index), so the breakage is only visible by querying the node's detail and reading `connected_from: null`.

### Steps to reproduce

1. Add `easy saveVideo` (ComfyUI-Easy-Media). Its `input_mode` has exactly two valid options: `images+audio`, `video`.
2. Connect `VAEDecode.IMAGE` → `input_mode.images`. Succeeds.
3. Call `panel_set_widget(node_id=<id>, widget="input_mode", value="images")` — note `"images"` is **not** a valid key.
4. Returns success: `{ set: { previous: "images+audio", value: "images" } }`.
5. `panel_query_graph({ids:[<id>], fields:"detail"})` now shows slots reduced to `input_mode`, `output_mode.save_metadata`, `output_mode`, `filename_prefix` — `input_mode.fps` / `input_mode.images` / `input_mode.audio` are gone and the IMAGE link is lost.
6. `panel_graph_outline` misleadingly renders `VAEDecode → <node>.output_mode.save_metadata` (a BOOLEAN slot).

### Exact error (scrubbed)

None — that is the bug. Silent success on an invalid value.

### Fix

`applied locally: yes` (needs a hard browser refresh to take effect; not yet verified at time of filing)
`upstream-only: no`

Guard the assignment by validating against `w.options.values` first, only rejecting when a finite option list can be positively enumerated (so non-combo and dynamically-sourced widgets are unaffected):

```diff
@@ graph_set_widget({ node_id, widget, value }) {
+    // Validate against the widget's own option list before assigning. An
+    // out-of-enum value silently collapses dynamic combos (COMFY_DYNAMICCOMBO_V3)
+    // and removes the input slots they own, dropping any links on those slots
[...]

PR #1340, opened 2026-08-19
fix(342): the outline resolves a link's live target slot through the backlink — a stale target_slot renders nothing instead of fabricating connectivity

artokun:
## Mechanism

`panel_graph_outline` rendered each outgoing link's target as `tgt.inputs[l.target_slot]?.name` — the `target_slot` recorded on the link record at connect time. That index is captured before a transformation that invalidates it: when the target node's inputs are **compacted** — a `COMFY_DYNAMICCOMBO_V3` selection rebuilding the node's slots, or the reporter's removed dynamic `ref_video_0` shifting the tail — the recorded index survives while the slot it named does not. The outline then reported the link against whatever slot now *occupies* that index: `VAEDecode → easy saveVideo.output_mode.save_metadata`, a BOOLEAN, over a graph whose IMAGE link no longer executes. `panel_query_graph`, which reads the live `inputs[].link` backlink, correctly showed it gone — so the two readers disagreed and the cheaper one lied.

The original #342 complaint — `graph_set_widget` accepting an out-of-enum combo value — was fixed in 0.11.7 (#233/#240). What the 2026-08-18 reproduction on panel 0.14.43 / frontend 1.49.6 describes is this outline staleness, which that fix never touched.

## Fix

- `liveLinkTargetInput()` in `web/js/lib/graph-read.js` resolves the input a link actually lands on by scanning the **target node's own `inputs[].link` backlinks** — the same live truth the prompt serializer walks — never the recorded `target_slot`. Returns `{ index, name }`, or `null` when nothing backlinks the link (an orphaned record).
- The outline executor uses it: an **orphaned link renders nothing**, matching `panel_query_graph`. A link whose live slot merely *shifted* renders the slot it actually feeds.
- `Number(null) === 0`, so a real link id of `0` is guarded from matching an *unlinked* input.
- The fallback is honest about being fail-**open**, not fail-closed: when the target is not in this graph's node set (or carries no `inputs` array) the recorded slot is still rendered as a bare index, exactly as before. That case is self-disclosing — the id it names appears nowhere else in the outline — and was deliberately left alone rather than broadened.

## What the first pass was missing, and the refutation

The fix landed with eight unit tests, all of them exercising the **helper**. Measured on a clean worktree: reverting the entire wiring hunk in `web/js/comfyui-mcp-panel.js` — the outline back to rendering the recorded `target_slot`, the bug fully restored — left the suite at **5038 tests, 5036 pass, 1 fail (the known #671 wall-clock flake), 1 todo**, with all eight #342 tests green. The fix was unpinned; a later edit could have removed it silently.

Two commits close that:

1. **`browser_tests/unit/outline-coverage.test.mjs`** — four tests that lift the outgoing-link render block *verbatim* out of the `graph_outline` handler and execute it, so the assertion runs the production source rather than a copy. With the wiring reverted they report the exact strings #342 was filed about:

   | case | actual (fix reverted) | expected |
   | --- | --- | --- |
   | orphaned link | `['12.output_mode.save_metadata']` | `[]` |
   | slot shifted by compaction | `['5.fps']` | `['5.ref_video_1']` |

2. **`browser_tests/outline-orphaned-link-renders-nothing.spec.ts`** — the same state built in a **real graph in a real ComfyUI**: connect `EmptyLatentImage.LATENT → VAEDecode.samples`, then remove the target's input slot without disconnecting (what the dynamic-combo rebuild leaves behind — record alive in `graph.links` and on the origin's `outputs[].links`, no input backlinking it). `vae` shifts into the freed index. Pre-fix panel: **FAIL**, `Expected substring: not "2.vae"` — the live handler attributing the dropped link to the slot that took its index. With the fix: **PASS**. Preconditions are asserted before the outline is read, so a future frontend that cleans the record up on splice fails the spec as "no longer reproducing #342" instead of passing vacuously.

## Verification

Run from the worktree against a ComfyUI 0.33.1 / frontend 1.48.7 started on an isolated `--base-directory` with this branch junctioned into `custom_nodes`.

- `npm run test:unit` — **5042 tests, 5040 pass, 1 fail, 1 todo**. The single failure is `#671 verifyInstalled … fast-draining install`, the known wall-clock flake; re-run alone it is **125/125 green**.
- `npm run typecheck` — clean. `npm run check:scope` — every name resolves.
- `npx playwright test --workers=1` — **52 passed / 21 failed** with this branch, against a **19 failed / 54 passed** baseline measured back-to-back on the same rig with only `web/js/` reverted to `origin/main`. The failure set is the same pre-existing `MockBridge.waitForUserMessage timed out` cluster (chat history, conversation persistence, streaming, workflow-chat identity) — none of those specs call `graph_outline`. The two that differ between the runs, `agent-drive.spec.ts:113` ("highlight issued BEFORE the first page lands") and `conversation-persistence.spec.ts:261` ("a stale tab republishes the removed thread"), are both ordering races and both are absent from one run and present in the other; neither touches the changed path.
- The specs that *do* exercise `graph_outline` — `outline-orphaned-link-renders-nothing`, `outline-reports-renamed-labels`, `outline-tags-not-forgeable` — are **8/8 green** with the fix.

## Gate

Not self-gated. `codex exec` is account-exhausted until 2026-08-19 20:43 and the substitute `claude-gate.mjs` is blocked on the weekly account limit, so no gate script was run here — the orchestrator gates and merges this.

This addresses the underlying cause of #342; closing that issue is the orchestrator's call, so this PR does not carry an auto-close reference.

### DIFF
--- a/browser_tests/outline-orphaned-link-renders-nothing.spec.ts
+++ b/browser_tests/outline-orphaned-link-renders-nothing.spec.ts
@@ -1,0 +1,109 @@
+/**
+ * #342 — the outline must not fabricate connectivity out of a stale `target_slot`.
+ *
+ * A link record's `target_slot` is captured at connect time. When the target node's
+ * inputs are COMPACTED afterwards — a `COMFY_DYNAMICCOMBO_V3` rebuilding its slots on a
+ * selection change, a removed dynamic `ref_video_N` input shifting the tail — the index
+ * survives the transformation that invalidated it. `panel_graph_outline` rendered
+ * `tgt.inputs[l.target_slot].name`, so it reported the link against whatever slot now
+ * OCCUPIES that index. The reporter's outline said
+ * `VAEDecode → easy saveVideo.output_mode.save_metadata` (a BOOLEAN) over a graph whose
+ * IMAGE link no longer existed, while `panel_query_graph` — which reads the live
+ * `inputs[].link` backlink — correctly showed it gone.
+ *
+ * This spec builds that state in a REAL graph in a REAL ComfyUI: connect two nodes, then
+ * remove the target's input slot WITHOUT disconnecting, which is exactly what the
+ * dynamic-combo rebuild leaves behind (link record alive in `graph.links` and in the
+ * origin's `outputs[].links`; no input backlinks it any more). The unit tests pin the
+ * render block; this pins that a live graph, a live outline handler and a live frontend
+ * still agree.
+ *
+ * Measured here against the pre-fix panel: the outline row read `→ <id>.vae`.
+ */
+import { test, expect } from './fixtures/panelTest'
+import { claimFreshCanvas, settleCanvas } from './fixtures/canvasIdentity'
+
+test('an outgoing link whose target slot was removed renders NOTHING in the outline', async ({
+  page,
+  panel,
+  mockBridge,
+}) => {
+  await panel.goto()
+  await panel.setBridgeUrl(mockBridge.url)
+  await panel.openSidebar()
+  await panel.connect()
+  await claimFreshCanvas(page, mockBridge)
+
+  const made = await page.evaluate(() => {
+    const w = window as any
+    const app = w.comfyAPI?.app?.app || w.app
+    const LG = w.LiteGraph || w.comfyAPI?.litegraph?.LiteGraph
+    const graph = app?.canvas?.graph ?? app?.graph
+    const latent = LG.createNode('EmptyLatentImage')
+    const decode = LG.createNode('VAEDecode')
+    graph.add(latent)
+    graph.add(decode)
+    // VAEDecode inputs are [samples (LATENT), vae (VAE)]. Wire the LATENT into slot 0.
+    latent.connect(0, decode, 0)
+    const linkId = decode.inputs?.[0]?.link ?? null
+    const before = {
+      slot0: decode.inputs?.[0]?.name ?? null,
+      linkId,
+      // The origin still lists the link on its output — the outline walks THIS.
+      onOutput: (latent.outputs?.[0]?.links ?? []).includes(linkId),
+    }
+    // The compaction: the slot the link landed on is REMOVED and the tail shifts up,
+    // with no disconnect. `vae` now sits at index 0 — the index the link still records.
+    decode.inputs.splice(0, 1)
+    const after = {
+      slot0: decode.inputs?.[0]?.name ?? null,
+      backlinked: (decode.inputs ?? []).some((i: any) => i?.link === linkId),
+      recordedTargetSlot: graph.links?.[linkId]?.target_slot ?? null,
+      stillOnOutput: (latent.outputs?.[0]?.links ?? []).includes(linkId),
+    }
+    return { decodeId: String(decode.id), latentId: String(latent.id), before, after }
+  })
+
+  // Preconditions — if any of these drift, the spec is no longer reproducing #342 and
+  // must be repaired rather than relaxed.
+  expect(made.before.slot0, 'precondition: the LATENT lands on `samples`').toBe('samples')
+  expect(made.before.linkId, 'precondition: the connect produced a link').not.toBeNull()
+  expect(made.before.onOutput, 'precondition: the origin output lists the link').toBe(true)
+  expect(made.after.slot0, 'precondition: compaction shifted `vae` into the freed index').toBe('vae')
+  expect(made.after.backlinked, 'precondition: no input backlinks the link any more').toBe(false)
+  expect(
+    made.after.recordedTargetSlot,
+    'precondition: the link record still points at the freed index',
+  ).toBe(0)
+  expect(
+    made.after.stillOnOutput,
+    'precondition: the orphaned record survives on the origin output — this is what the outline walks',
+  ).toBe(true)
+
+  await settleCanvas(page)
+
+  const outline = await mockBridge.command('graph_outline', {})
+  expect(outline.ok).toBe(true)
+  const text = String(outline.result?.outline ?? '')
+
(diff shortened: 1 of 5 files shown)


artokun:
Reconciliation note — a second session was dispatched on #342 before this PR appeared in the survey, and reached the same root cause independently. Rather than open a competing PR I reviewed this one and pushed the delta onto this branch (`e2b7ef7`, `b232f3a6`, fast-forward — the original `9307624c` is untouched). The body has been rewritten to carry the evidence.

**The mechanism here is right.** Independently checked: the incoming (`←`) side of the outline already walks `n.inputs[].link`, so it was never affected; the outgoing (`→`) side was the only reader of its kind (`grep target_slot` over `web/js/` leaves only the disconnect diagnostics and the subgraph-rail plumbing, neither of which renders connectivity). `byId` is built from `graph._nodes`, so the `!tgt` branch means a dangling record, and the fallback there is fail-**open**, not fail-closed as the first body claimed — it still renders `target_id.target_slot`. Left as-is (that case is self-disclosing: the id it names appears nowhere else in the outline), but the code comment now says so instead of promising a guarantee it does not make.

**What was missing was the pin, not the fix.** All eight original tests exercise `liveLinkTargetInput` itself. Reverting the entire wiring hunk in `web/js/comfyui-mcp-panel.js` — the bug fully restored — left the suite at 5038 tests / 5036 pass / 1 fail (the known #671 flake) / 1 todo, every #342 test green. Two commits close that:

- four tests that lift the outgoing render block verbatim out of the `graph_outline` handler and **run** it, so the assertion is on the production source. Fix reverted, they report `['12.output_mode.save_metadata']` where `[]` is expected, and `['5.fps']` where `['5.ref_video_1']` is — the reporter's exact symptom.
- a Playwright spec that builds the state in a **real** graph (connect, then remove the target's input slot without disconnecting, so `vae` shifts into the freed index). Pre-fix panel: FAIL, `Expected substring: not "2.vae"`. With the fix: PASS.

Browser gate run on ComfyUI 0.33.1 / frontend 1.48.7, `--workers=1`: 52 passed / 21 failed on this branch against a back-to-back 54/19 baseline with only `web/js/` reverted — same pre-existing `MockBridge.waitForUserMessage timed out` cluster either way, and the three specs that actually call `graph_outline` are 8/8 green. Unit gate 5042/5040, typecheck and panel-scope clean.

No gate script was run (codex account-exhausted until 2026-08-19 20:43; the `claude-gate.mjs` substitute is blocked on the weekly limit) and neither session merges — this is for the orchestrator. The `Closes #342` line was dropped from the body for the same reason: closing the issue is the orchestrator's call.

artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and the `claude-gate.mjs` fallback spawns a headless `claude` blocked on the account weekly limit. Under the owner's standing authorisation, this PR was gated by an **independent adversarial reviewer that had never seen either author's reasoning**, run against `.claude/autopilot/review-taxonomy.md`. Verdict: **SHIP**, head `b232f3a`.

This PR had two authors from two concurrent sessions, so the gate re-ran the second author's claims rather than adopting them.

- **The unpinning reproduced.** Reverting `web/js/comfyui-mcp-panel.js` to `origin/main` in a scratch worktree gives 62/64, and exactly the two new **call-site** tests fail, with the strings #342 was filed about (`['12.output_mode.save_metadata']`, `['5.fps']`). The eight helper tests stayed green — confirming that before those call-site tests existed, the entire wiring hunk could be deleted with the suite green. Restored: 64/64. Extra mutation: deleting `inp.link != null &&` at `web/js/lib/graph-read.js:96` is killed by `graph-read.test.mjs:614`.
- **Reachability proven, not inferred.** One production call site, `comfyui-mcp-panel.js:10496`, inside `GRAPH_TOOL_EXECUTORS.graph_outline`. There is no build step — `web/js` is served directly.
- **The reopened reproduction was reproduced LIVE and closed live.** On an isolated ComfyUI (frontend 1.48.7) with this branch junctioned in, `origin/main`'s `web/js` printed `1 EmptyLatentImage … → 2.vae` for a link no input backlinks. With the fix, it passes.
- **The scoping does not leave #342 half-fixed.** `panel_query_graph`'s adjacency (`:10676-10684`) is built purely from `inp.link`, matching the reporter's own observation that query_graph was correct — the outline was the sole stale reader on that path. The remaining stale `target_slot` readers (`web/js/lib/use-everywhere-links.js:124,149`, `web/js/lib/safe-remove-node.js:120`) are same-class latent bugs on UE-broadcast and subgraph-rail paths the repro never touches. Follow-up material.
- **Fail-open fallback accepted.** `:10490-10493` fires only where nothing can be verified, and `liveLinkTargetInput` never throws (null, `{}`, and hole-arrays tested); an orphan renders nothing, which is what the reporter asked for. Minor pre-existing note: its bare-index `id.N` output is byte-identical to the *verified* unnamed-input output at `:10498`, so "unverified recorded slot" and "live slot, unnamed input" collapse — ambiguous before this change too.
- **Playwright spot-check: the attribution holds, but this PR body undercounts.** It says 3 specs call `graph_outline`; 8 do. All 8 at `--workers=1`: 12 passed / 2 failed. `blank-canvas-not-wedged.spec.ts:44` passes on solo re-run (load flake). `open-unverified-names-recovery.spec.ts:24` fails **identically with `web/js` reverted** — pre-existing. `agent-drive.spec.ts` is fully green on the branch, and `conversation-persistence.spec.ts` fails the same 7 including `:261` on baseline.
- Gates: typecheck, `check-panel-scope`, `check-tool-vocabulary`, i18n all clean; 5042 unit tests, 1 failure = the known #671 flake (125/125 alone).

The only defect-shaped finding is the spec undercount in the PR body, which is prose rather than code.

PR #1341, opened 2026-08-19
fix(1264): a starved post-open frame wait no longer latches the workflow-switch fence

### LINKED ISSUE #1264: panel_open_workflow timeout leaves workflow-switch fence stuck

artokun:
## Root cause

`workflow_open`'s re-baseline step (`clearSpuriousOpenModified`) awaits ONE animation frame via `nextFrame()` while holding a reload step of the switch/reload critical section (`beginWorkflowReloadStep`). `requestAnimationFrame` does not fire in a hidden or occluded tab, and a step in flight is deliberately immune to the section's 30s age-out (`activeWorkflowReloadGuard` never expires while `pending > 0`) — that immunity exists so a genuine load cannot lose the fence mid-write. But a render frame is not a load: on a backgrounded ComfyUI tab the unbounded rAF await never settled, `pending` stayed > 0 forever, and the fence latched — every graph/workflow command was refused ("still switching/reloading") and the open's reply/receipt was undeliverable until a manual F5, exactly the #1264 reports (fence held 2m+; `panel_open_workflow` timing out at 15s with outcome undetermined).

## Fix

Bound the frame wait with the repo's one shared bounded-step primitive, `withTimeout` from `web/js/lib/bounded-step.js` (already imported and used by six other paths in this bundle). After `NEXT_FRAME_FALLBACK_MS = 2000` — well under the bridge's 15s reply window — a starved frame degrades to "the capture may not have landed yet": the tab keeps a cosmetic modified flag (the LOUD direction) instead of wedging the bridge. `withTimeout` never rejects and never cancels the bounded work, so a late-firing frame still runs the re-baseline when it can.

This is the only awaited rAF in the open path — the remaining `requestAnimationFrame` uses are UI render loops that hold no fence.

## Tests

`browser_tests/unit/open-frame-bound.test.mjs` pins both halves:

1. the primitive's guarantee the fix relies on — a never-settling step resolves through its fallback instead of pending forever (and a late-settling step is dropped once the bound has fired);
2. the SHIPPING `nextFrame` in the bundle routes its rAF wait through `withTimeout` with a named, positive bound that stays under the bridge's 15s reply window.

## Verification (worktree, origin/main @ 25cef01e)

- `npm ci` — clean, 0 vulnerabilities
- `node --test browser_tests/unit/open-frame-bound.test.mjs` — 3/3 pass
- `npm run test:unit` — 5034 tests, 5033 pass, 0 fail, 1 todo (i18n checks + tool-vocabulary + panel-scope gates all included and green)
- `npm run typecheck` — clean

Closes #1264

### DIFF
--- a/browser_tests/unit/open-frame-bound.test.mjs
+++ b/browser_tests/unit/open-frame-bound.test.mjs
@@ -1,0 +1,75 @@
+// comfyui-mcp-panel#1264 — a timed-out workflow_open left the switch fence stuck.
+//
+// workflow_open's re-baseline step (clearSpuriousOpenModified) awaits ONE
+// animation frame while holding a reload step of the switch/reload section.
+// rAF does not fire in a hidden or occluded tab, and a step in flight is
+// deliberately immune to the section's 30s age-out (that immunity exists so a
+// genuine load cannot lose the fence mid-write) — so on a backgrounded tab the
+// frame wait never settled, the fence latched forever, every graph/workflow
+// command was refused, and neither the open's reply nor its open receipt could
+// be delivered until a manual page reload. The reporters' only workaround was
+// F5.
+//
+// The fix bounds the frame wait with the repo's one bounded-step primitive: a
+// starved frame degrades to a timer fallback (the tab keeps a cosmetic
+// modified flag — the LOUD direction) instead of wedging the bridge. These
+// tests pin both halves:
+//
+//   1. the primitive's guarantee the fix relies on — a never-settling step
+//      resolves through its fallback instead of pending forever;
+//   2. the SHIPPING nextFrame in the bundle routing its rAF wait through that
+//      primitive with a positive bound that stays well under the bridge's
+//      15s reply window, so the open's reply/receipt can still land.
+
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+
+import { withTimeout } from "../../web/js/lib/bounded-step.js";
+
+const here = dirname(fileURLToPath(import.meta.url));
+const source = readFileSync(join(here, "../../web/js/comfyui-mcp-panel.js"), "utf8").replace(/\r\n/g, "\n");
+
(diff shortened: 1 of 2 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and the `claude-gate.mjs` fallback spawns a headless `claude` blocked on the account weekly limit. Under the owner's standing authorisation, this PR was gated by an **independent adversarial reviewer that had never seen the author's reasoning**, run against `.claude/autopilot/review-taxonomy.md`. Disclosure is the condition that substitution was accepted under. Verdict: **SHIP**, on head `4ac0c28`.

- **The latch is real and is now bounded.** Starved case constructed: hidden tab, so rAF never fires, so `pending` stays above 0 and `activeWorkflowReloadGuard()` never ages out. Bounded at 2s per site, at most 4s per open, inside the 15s reply window.
- **Fence release is safe on every exit.** Both `clearSpuriousOpenModified` call sites (`comfyui-mcp-panel.js:15663-15670`, `:15832-15839`) wrap in `try/finally { endWorkflowReloadStep }`, and the helper's own try/catch (`:6072-6100`) means it cannot throw at all. `withTimeout` never rejects and never cancels, so "the frame never came" degrades to a timer while a late frame still runs the re-baseline.
- **Reachability:** `nextFrame` has exactly one caller (`:6077`), inside `workflow_open`'s re-baseline. It is the only awaited rAF; the other eight `requestAnimationFrame` uses are render loops holding no fence.
- **Mutation:** reverting `nextFrame` to the unbounded promise fails `open-frame-bound.test.mjs` test 3. The test asserts the shipping source, not just the primitive.
- 5034 unit tests, one failure = the known #671 `verifyInstalled` load flake (125/125 in isolation, fails identically on all three branches). i18n, i18n-render, tool-vocabulary and panel-scope gates green (panel-scope run with a real TypeScript, not its silent cannot-run path); `tsc --noEmit` clean.
- Noted, not a defect: Chrome clamps `setTimeout` to at least 1s in hidden tabs, so the bound is "at least 2s", not exactly 2s. "Forever" is still gone.
- Playwright not run: the diff bounds a promise and changes nothing the sidebar renders. No browser-level coverage is claimed.

Issue #1192, opened 2026-08-13
graph_add_node's bounds do not compose: serialised, they sum past the 30s command budget


artokun:
## Summary

`graph_add_node`'s bounds are each defensible alone and do not compose. Serialised on one add they sum past the bridge's 30s command budget, so the worst case is a bare "tab did not reply" instead of the worded refusal the bounds exist to preserve.

Found by review on #1180 (PR #1186) and confirmed there. **Deliberately not fixed in that PR**: #1180's scope is bounding the individual `getNodeDefs` call sites that could hang, and it does that. This is the same class of defect one level up — a budget for a whole COMMAND rather than for one refresh run — and it needs its own design.

**This is not a regression.** Before #1180 these steps were unbounded, so the same path could hang forever. #1180 makes it finite; this issue makes it fit.

## The composition

On the scenario the panel is most likely to meet it — a ComfyUI restart, which is exactly when a reconnect-triggered refresh is already in flight — adding a newly-installed class costs:

| step | bound |
|---|---|
| `boundedGetNodeDefs()` (fast path, or the whole-schema fallback) | 10,000 ms |
| `refresh(freshDefs)` → coalescer `await current` on the in-flight run | that run's budget (9,000 ms) + its unbounded local work |
| …then `startRun(freshDefs)`, a second full run | 9,000 ms + unbounded local work |
| `awaitRequiredCustomWidgetRegistration` (the 2,500 ms widen runs inside it) | 5,000 ms |

**33,000 ms of bounded waiting**, before the unbounded local work (`registerNodesFromDefs`, `reapplyDefsToLiveNodes`) that legitimately runs inside those runs. The command budget is 30,000 ms and the orchestrator's ui-bridge read default is 20,000 ms.

## Why the pieces cannot just be shrunk

The coalescer's `await current` is a wait on a run that ALREADY STARTED, under its own deadline. A caller cannot retroactively bound work someone else began, so shrinking the add's own numbers does not cap that term. Any real fix has to decide what the add does when the join alone would exhaust its budget — most likely refuse in words ("a node-def refresh is still running, retry"), which is recoverable, rather than proceed unregistered or block.

## Related, and probably part of the same fix

[...]

Issue #289, opened 2026-07-30
panel_add_node fails with "Unknown node type" for freshly installed packs — frontend node-def cache is stale after install+restart, and no tool can refresh it


artokun:
### What happened / root cause

The standard autonomous flow documented for the panel — `panel_search_nodes` → `panel_install_node` → `panel_node_queue_status` → `panel_restart_comfyui` → resume and build with `panel_add_node` — **cannot complete without manual user intervention.**

After installing `seedvr2_videoupscaler` and restarting ComfyUI, the *server* had the new nodes (confirmed via `get_node_info`, which returned full definitions for `SeedVR2VideoUpscaler`, `SeedVR2LoadDiTModel`, `SeedVR2LoadVAEModel`, `SeedVR2TorchCompileSettings`), but every `panel_add_node` for those class_types failed:

```
Error: Unknown node type "SeedVR2LoadDiTModel" — check the exact class_type via panel_get_graph or panel_search_nodes
```

Root cause: `panel_add_node` validates the requested `class_type` against the **ComfyUI frontend's** LiteGraph node registry, which is populated from `/object_info` at page load. A ComfyUI server restart does **not** cause the already-open frontend to re-fetch `/object_info` — the panel reconnects over the websocket and the agent resumes, but the browser's node-definition map is still the pre-install one. So the agent resumes into a state where it believes it can build, and every add silently 404s against a stale cache.

Notably `panel_add_node` for *pre-existing* types (`VHS_LoadVideoPath`, `VHS_VideoCombine`) worked fine in the same session, which is what makes this confusing to diagnose — the graph tools appear healthy, only the new types are missing, and the error message points you at `panel_search_nodes` (which happily confirms the pack *is* installed), sending you in a circle.

There is **no panel tool to refresh node definitions** — no equivalent of the frontend's "Refresh Node Definitions" command. The only recovery is to ask the user to press F5 in the ComfyUI tab. That converts a fully autonomous install→build→run task into one that hard-blocks on a human keystroke, mid-task, for a reason the user has no way to anticipate.

The error message is also actively misleading: it attributes the failure to a wrong `class_type` when the name was correct and the real problem is a stale client-side cache.

### Steps to reproduce

1. Open ComfyUI with the agent panel; leave the tab open for the whole test.
2. Install a pack providing new node types: `panel_install_node({id: "seedvr2_videoupscaler"})`.
[...]

Issue #1242, opened 2026-08-14
Auto-refresh stale node schema before panel_add_node retry


artokun:
panel_add_node refused to add LoadImage because the registered frontend node schema was stale. Reproduction: open a fresh workflow, call panel_add_node for LoadImage while its frontend definition is stale, observe the refusal, then call panel_refresh_nodes and retry; the identical add succeeds. Exact error: LoadImage required input image was added or retyped since this page loaded its node schema. Suggested fix: automatically refresh node definitions and retry once on this exact stale-schema condition. Applied locally: no; workaround verified. Environment: Windows 11; ComfyUI 0.33.1; frontend 1.48.7; RTX 4060 Laptop 8GB; comfyui-mcp 0.50.97; panel 0.11.90.

Related (same root cause): #852, #1085

artokun:
Additional report on Windows 11 / ComfyUI 0.28.0 / frontend 1.45.21 / RTX 3060: comfyui-mcp 0.51.56 and panel 0.14.43. During sequential live-canvas adds, `panel_add_node(LoadImage)` returned a successful added-node payload (id 196) while the sidebar simultaneously reported `graph_add_node failed`: `"LoadImage" required input "image" was added or retyped since this page loaded its node schema`. After `panel_refresh_nodes`, the node was present and usable. This adds an inconsistent-outcome case: the tool reports success despite the visible stale-schema refusal, so retrying based on the error can be ambiguous.

artokun:
Additional report: on macOS / ComfyUI 0.33.3 / Apple MPS, the first `panel_add_node(LoraLoader)` in a blank workflow reported the schema-refresh disclosure but created the node correctly; the second add used the refreshed schema. The reporter finds that disclosure confusing. Metadata lists comfyui-mcp 0.52.91 (the body also says 0.52.67) and panel 0.7.3, but the installed panel version could not be mechanically verified.

PR #1342, opened 2026-08-19
fix(1192): graph_add_node gets ONE budget, so its bounds stop adding up

artokun:
## Root cause

`graph_add_node`'s bounds were each defensible alone and did not compose. Serialised on one add — the scenario the panel is most likely to meet on a ComfyUI restart, when a reconnect-triggered refresh is already in flight — they summed to ~41,000 ms of bounded waiting (seed wait 8,000 + fast/whole-schema fetch 10,000 + coalescer join ~9,000 + unbounded local + own run ~9,000 + widget registration 5,000) against the 30,000 ms window `panel_add_node` relays this command in. Past that window the reply the bounds exist to produce never leaves the tab, and the user gets a bare `did not reply to "graph_add_node" within 30000 ms`, which names nothing and is not actionable.

## Fix

One deadline for the whole command, so the bounds stop ADDING:

- **`web/js/lib/command-budget.js` (new)** — `makeCommandBudget(totalMs, now)`: one monotonic deadline taken at the top of the command; every step draws its bound from what is left. It computes ALLOWANCES only — `withTimeout` still applies every one, so no second timeout primitive. Never returns a non-positive bound (`withTimeout` reads `ms <= 0` as NO BOUND — the #1188 trap).
- **`web/js/lib/refresh-coalesce.js`** — `opts.joinMs` bounds the caller's WAIT on a run someone else started, never the run (a caller cannot retroactively shorten work already begun). An abandoned join drops the payload and returns `REFRESH_JOIN_ABANDONED` rather than starting a competing run — two concurrent `registerNodesFromDefs` passes are the stampede the coalescer exists to prevent. Dropping a payload while *refusing* is safe; #289 P2 forbids dropping one while claiming success.
- **`web/js/comfyui-mcp-panel.js`** — `graph_add_node` takes one 25,000 ms budget (mirroring `NODES_INSTALL_COMMAND_BUDGET_MS` against the same 30 s relay window: 5 s of reply/serialise/websocket slack). The seed wait, the single-type fast-path fetch, and the whole-schema fetch each take `budget.bounded(theirOwnConstant)`; the whole-schema fetch is allowed the same 10,000 ms `graph_set_widget`'s oracle allows, settling the two paths' disagreement the issue flagged. The widget-registration wait takes the *smaller* of its own 5,000 ms and what the command has left, and the socket-proof widen is derived from the wait's *effective* deadline. An abandoned join refuses in words — "a node-def refresh started by something else was still running… NOTHING WAS ADDED… RETRY in a few seconds" — recovered through a structured flag plus the live registry, never by parsing the resolver's error text. A registration wait cut short by the budget discloses that in its refusal instead of sending the caller to a tab reload.

## Verification

- `npm ci` — clean, 0 vulnerabilities
- `npm run test:unit` — **5059 tests, 5058 pass, 0 fail, 1 todo** (includes i18n, tool-vocabulary, and panel-scope gates). One earlier full-suite run flaked in `manager-install.test.mjs` #671 `verifyInstalled` (real-clock timing, untouched by this change — history of wall-clock fixes in that file, e.g. 72bbe2b5); passes in isolation and on rerun.
- `npm run typecheck` — clean
- New/extended tests: `add-node-command-budget.test.mjs` (drives the shipped `graph_add_node` body extracted from the panel source against the real coalescer), `command-budget.test.mjs`, `refresh-coalesce.test.mjs`, `single-node-def.test.mjs` (source-level pins, incl. the `withTimeout` wiring of the panel's coalescer).
- Mutation harness (15 targeted mutations of the fix): each caught by the suite — several as hangs reported red, per the harness's own hang≠pass rule.

the underlying cause of #1192

### DIFF
--- a/browser_tests/unit/_panel-constants.mjs
+++ b/browser_tests/unit/_panel-constants.mjs
@@ -13,6 +13,9 @@
 import { readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
 
+import { makeCommandBudget } from "../../web/js/lib/command-budget.js";
+import { REFRESH_JOIN_ABANDONED } from "../../web/js/lib/refresh-coalesce.js";
+
 export const PANEL_SRC = readFileSync(
   fileURLToPath(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url)),
   "utf8",
@@ -35,13 +38,26 @@ export const CUSTOM_WIDGET_REGISTRATION_TIMEOUT_MS = readPanelNumber(
   "the registration deadline",
 );
 
+/**
+ * #1192 — the divisor is a NAMED constant now, because the widen's bound is no longer a
+ * fixed number: it is that fraction of whatever deadline the registration wait actually
+ * received, which under a command budget can be far less than the 5000ms standalone one.
+ * Still read from source rather than restated, for the reason above.
+ */
+export const WIDEN_SOCKET_PROOF_DIVISOR = readPanelNumber(
+  /const WIDEN_SOCKET_PROOF_DIVISOR = (\d+);/,
+  "the widen divisor",
+);
+
 /** Derived exactly as the panel derives it, divisor included, so the two cannot drift. */
 export const WIDEN_SOCKET_PROOF_TIMEOUT_MS = Math.floor(
-  CUSTOM_WIDGET_REGISTRATION_TIMEOUT_MS /
-    readPanelNumber(
-      /WIDEN_SOCKET_PROOF_TIMEOUT_MS = Math\.floor\(CUSTOM_WIDGET_REGISTRATION_TIMEOUT_MS \/ (\d+)\)/,
-      "the widen divisor",
-    ),
+  CUSTOM_WIDGET_REGISTRATION_TIMEOUT_MS / WIDEN_SOCKET_PROOF_DIVISOR,
+);
+
+/** #1192 — the whole-command deadline `graph_add_node` takes on its first line. */
+export const ADD_NODE_COMMAND_BUDGET_MS = readPanelNumber(
+  /const ADD_NODE_COMMAND_BUDGET_MS = (\d+);/,
+  "the add_node command budget",
 );
 
 export const NODE_DEFS_FETCH_TIMEOUT_MS = readPanelNumber(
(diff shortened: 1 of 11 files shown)


artokun:
Independent review of this PR from a parallel drain session that was dispatched on #1192
before this PR existed. **I am not opening a competing PR** — this one is correct and is the
substance of the fix. One incompleteness and one residual, both with numbers, plus the browser
gate this PR does not claim.

### What I checked, and it holds up

This is **not** a bigger constant. It makes "this one step was slow" distinguishable from "the
command has spent its allowance": `REFRESH_JOIN_ABANDONED` is a `Symbol`, distinct from the
`undefined` a successful join resolves to, and the add recovers the reason through **that flag
plus the live registry** rather than by parsing the resolver's error text — which matters,
because `assertAddNodeResolvableRefreshing` swallows the refresh throw by design
(`web/js/lib/node-resolve.js:103`), so the reason never reaches its wording anyway.

Verified independently rather than taken from the body:

- The relay window is right. `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS = 30_000` at comfyui-mcp
  `src/orchestrator/panel-tools.ts:821`, applied to `graph_add_node` at `:9650`. The issue's
  20,000 ms ui-bridge default is not the binding number for this command; the PR is right to
  correct it.
- Every other `refreshComfyNodeDefs` caller passes no `joinMs`, so none can receive the new
  Symbol and none changes behaviour.
- Four mutations, all killed — including the two that matter most here:

| mutation | result |
|---|---|
| delete the one-line `withTimeout` wiring on the coalescer | ✖ red (the call site is pinned, not just the helper) |
| restore the unbounded resolver join | **hang** — the add never answers |
| drop `budget.bounded(...)` from the widget-registration wait | ✖ red |
| remove `bounded()`'s 1 ms floor (the #1188 "0 means NO BOUND" trap) | **hang** |

### The incompleteness: there is a FIFTH wait

The budget is threaded through four awaits. `graph_add_node` has five. `web/js/comfyui-mcp-panel.js:11786`:

```js
```

No `joinMs`. The coalescer's forced branch awaits any run **already in flight** before queueing
its own, so unbounded this is the same term the issue is about, at the one await inside this
command that names no bound — and it sits after the resolver, reached with the window already
partly spent. On the scenario the PR targets (a ComfyUI restart with a reconnect refresh still
running) the add parks there until the relay gives up.

Refuted on the rig, not argued: with a real in-flight run held open and the drift path driven,
**the add never answers at all**. Bounded, it gives up at 279.8 ms of a 400 ms budget and
refuses in words. A source-level count of `budget.` call sites cannot see this — only a test
that drives the drift path does.

Fix + test are stacked on this branch so the delta is readable on its own: **PR #1349**.

### The residual: `joinMs` bounds the JOIN, never the run after it

Driven against the real coalescer: join bound 280 ms, in-flight run 250 ms (settles inside the
bound), the caller's own run 2,000 ms → `refresh()` resolved in **2,251 ms, 8.0× its own join
bound**. With the shipped numbers a join may end as late as **20,000 ms** into the command
(25,000 budget − 5,000 reserve), and the own run that follows draws nothing from the budget —
`NODE_DEFS_RUN_BUDGET_MS = 9_000` bounds only the *waiting* inside a run, and
`registerNodesFromDefs` is deliberately unbounded (3,972 ms, per this PR's own comment). So a
worst case near **33,000 ms** survives on the path where the join succeeds close to its bound.

Still far better than the ~41,000 ms it replaces, and every other path composes — but the body's
"the bounds stop ADDING" is stronger than what one path delivers, and closing that last term
means bounding a *run* from its caller, which is exactly #1193. Worth saying here rather than
discovering it later.

### Browser gate

This PR claims unit tests and typecheck, not the browser suite. I ran it, `--workers=1`, against
a real ComfyUI 0.33.1 serving **this branch's** panel from its own instance and base-directory
(so the loaded module is the changed one, not the shared checkout's): **52 passed / 21 failed,
10.1 min**. I then re-ran the same 21 files against **`origin/main`** on the same rig: **21
failed, the exact same tests, name for name.** The e2e delta attributable to this change is
**zero**; the failures are pre-existing here and sit entirely in the chat /
conversation-persistence / MockBridge specs. Neither spec that exercises `add_node`
(`auto-layout`, `blank-canvas-not-wedged`) is among them.

The machine was loaded during both runs, and since this issue is about timing I would not read
the absolute numbers as anything — it is the main-vs-branch comparison that carries the claim.

One nit unrelated to correctness: the body ends `Closes #1192`, which auto-closes the issue on
merge. Closing is the orchestrator's call here.

artokun:
Orchestrator note: I edited one line of this PR body, replacing `Closes #1192` with "the underlying cause of #1192". No code or claim was changed.

Reason: this PR is the base of a stacked pair. **#1349** branches off it and closes the fifth wait — `graph_add_node` has five waits and this PR budgets four; the fifth is #1242's drift recovery at `comfyui-mcp-panel.js:11786`, calling `refreshComfyNodeDefs(undefined, { force: true })` with no `joinMs`, where the coalescer's forced branch first awaits any run already in flight. With `Closes #1192` in place, merging this PR would close the issue while that path was still unbounded.

There is also a disclosed residual that outlives both PRs: `joinMs` bounds the join, not the run after it (measured at 8.0x its bound), so the worst case is ~33,000 ms against a 30,000 ms relay window (`OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS`, `src/orchestrator/panel-tools.ts:821`). Closing that means bounding a *run* from its caller, which is panel#1193 — separate, and correctly left alone here.

So #1192's close decision stays with the merge sequence rather than firing on this PR alone. Nothing here reflects on the work: an independent review confirmed this is a mechanism and not a bigger constant — it raises no existing timeout, adding only a cap and a named divisor while `NODE_DEFS_FETCH_TIMEOUT_MS`, `NODE_DEFS_RUN_BUDGET_MS`, `OBJECT_INFO_SEED_WAIT_MS` and `CUSTOM_WIDGET_REGISTRATION_TIMEOUT_MS` all keep their values, and four mutations of it were killed (two red, two hangs).

PR #1343, opened 2026-08-19
fix(1261): a rename is announced as a rename, not a workflow switch

### LINKED ISSUE #1261: Workflow rename is reported as a switch with conflicting identity
Renaming the active workflow from "Untitled 2026-08-15 14-11-07" to "Krea2" emitted an ACTIVE WORKFLOW CHANGED event claiming a different workflow identity, while panel_set_workflow_target(mode=current) remained bound to the old instance and panel_graph_outline returned the same graph. The ComfyUI log confirms this was a file move/rename, not a tab switch. This is misleading and made a safe edit look like it targeted the wrong workflow.

### Steps to reproduce
1. Keep a saved workflow active.
2. Rename it in the frontend.
3. Observe the panel event reporting a workflow switch/new wf identity.
4. Call panel_set_workflow_target(mode=current); it reports already_current on the prior instance.
5. panel_graph_outline returns the same graph.

### Exact error
No thrown error. Conflicting identity/status output:
- panel event says active workflow changed to Krea2 with a different wf id
- panel_set_workflow_target says already_current on the prior workflow instance
- server log records moving the old workflow JSON to Krea2.json

### Fix
applied locally: no
upstream-only: yes
On workflow rename, preserve the active workflow instance identity, update its routing path/name atomically, and emit a rename event rather than a workflow-switch event. panel_set_workflow_target(mode=current) should report the updated path/name for the same instance.

### Environment
Windows 11
ComfyUI 0.31.1
Frontend 1.48.7
GPU NVIDIA GeForce RTX 4070 Ti SUPER 16 GB
comfyui-mcp 0.51.51
panel 0.14.37

artokun:
## Root cause

Renaming (or save-as-ing) the active saved workflow makes ComfyUI mutate the SAME workflow object's path in place: the instance identity survives, and only the derived route id moves `wf:<old>` → `wf:<new>`. The panel-owned-session branch of `onWorkflowMaybeChanged` (`web/js/comfyui-mcp-panel.js`) announced **every** route-id change with the switch wording — `The user switched the open workflow on the canvas … re-read it (panel_graph_outline) before assuming or editing anything, since earlier turns may refer to a different workflow.` So a rename read as a switch to a *different* workflow identity, contradicting the instance binding (the workflow uuid follows the object across a rename, so `panel_set_workflow_target(mode:"current")` correctly reported `already_current` on the same instance) and making a safe edit look like it targeted the wrong workflow.

The per-workflow-history branch already discriminated this shape (case 2b); the panel-owned branch did not.

## Fix

- New pure discriminator `isSameInstanceRename` in `web/js/lib/workflow-route-change.js`: **same live object + `wf:`→`wf:` route change can only be a rename** — a genuine switch always arrives on a different object. The discriminator is the object (via the proxy-safe `sameWorkflowObject`), never the path text, and it fails closed on anything unproven.
- The panel-owned branch now announces a rename as a rename: `The open workflow was RENAMED … "old" is now "new". Same canvas, same conversation, same node ids — only its file path changed.` The re-hello still happens (the route id genuinely moved); only the claim changes.
- The existing case-2b rename check now uses the same helper, so both branches share one discriminator.

## Verification

- `npm ci` — clean, 0 vulnerabilities.
- `npm run test:unit` (i18n-check + i18n-render-check + check-tool-vocabulary + check-panel-scope + node --test) — **5036 pass, 0 fail**, including 6 new tests in `browser_tests/unit/workflow-route-change.test.mjs` covering: the reporter's rename case, a genuine switch on a different object, unproven comparisons failing closed, first-save (`tmp:`→`wf:`) not being a rename, unsaved `tmp:`→`tmp:`, and non-string ids.

Closes #1261

### DIFF
--- a/browser_tests/unit/workflow-route-change.test.mjs
+++ b/browser_tests/unit/workflow-route-change.test.mjs
@@ -1,0 +1,85 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { isSameInstanceRename } from "../../web/js/lib/workflow-route-change.js";
+
+/**
+ * #1261 — a rename must not read as a workflow SWITCH.
+ *
+ * Renaming the active saved workflow moves its route id wf:<old> → wf:<new> on the
+ * SAME live workflow object. The panel's poll used to announce every route-id change
+ * as "the user switched the open workflow", so a rename read as a switch to a
+ * different workflow identity and sent the agent re-reading the canvas it was still
+ * bound to. The discriminator is the object, never the path text.
+ */
+
+test("the reporter's case: same live object, wf:→wf: id change — a rename, not a switch", () => {
+  assert.equal(
+    isSameInstanceRename({
+      sameWorkflowObject: true,
+      previousRouteId: "wf:Untitled 2026-08-15 14-11-07",
+      newRouteId: "wf:Krea2",
+    }),
+    true,
+  );
+});
+
+test("a genuine tab switch arrives on a DIFFERENT object — never a rename", () => {
+  // Two tabs can legitimately show the same file, so the path text cannot decide
+  // this. Only the object comparison can, and here it says: different workflow.
+  assert.equal(
+    isSameInstanceRename({
+      sameWorkflowObject: false,
+      previousRouteId: "wf:old",
+      newRouteId: "wf:new",
+    }),
+    false,
+  );
+});
+
+test("an unproven object comparison fails closed — not a rename", () => {
+  assert.equal(
+    isSameInstanceRename({
+      sameWorkflowObject: undefined,
+      previousRouteId: "wf:old",
+      newRouteId: "wf:new",
+    }),
+    false,
+  );
+  assert.equal(
+    isSameInstanceRename({ previousRouteId: "wf:old", newRouteId: "wf:new" }),
+    false,
+  );
+});
+
+test("a first save (tmp:→wf:) is NOT a rename — the adopt case owns that shape", () => {
+  assert.equal(
+    isSameInstanceRename({
+      sameWorkflowObject: true,
+      previousRouteId: "tmp:6f2c1d4e-9b7a-4c1e-8d3f-2a5b7c9d1e2f",
+      newRouteId: "wf:Krea2",
+    }),
+    false,
+  );
+});
+
+test("an UNSAVED canvas never renames (tmp:→tmp:)", () => {
+  assert.equal(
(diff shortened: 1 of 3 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and the `claude-gate.mjs` fallback spawns a headless `claude` blocked on the account weekly limit. Under the owner's standing authorisation, this PR was gated by an **independent adversarial reviewer that had never seen the author's reasoning**, run against `.claude/autopilot/review-taxonomy.md`. Disclosure is the condition that substitution was accepted under. Verdict: **SHIP**. Head reviewed: `8b54662`.

- **The discriminator is identity, not name.** `isSameInstanceRename` requires `sameWorkflowObject === true`, and `sameWorkflowObject` (`web/js/lib/workflow-chat-identity.js:44`) layers `===`, then Vue-raw `===`, then per-instance `changeTracker`, with a header stating that path is deliberately NOT evidence. So "rename A to B's name" and "two tabs on one file" both resolve correctly; no name comparison happens anywhere.
- **Ordering verified:** `renaming` and `previousName` are computed before `currentWorkflowId`/`currentWorkflowRef` are reassigned (`comfyui-mcp-panel.js:28681-28691`), and the branch is unreachable when the id is unchanged (`:28551` early-returns).
- The other candidate emitter of the reported symptom, `manualChangeBanner`'s ACTIVE WORKFLOW CHANGED (`:8557`), keys on `workflowStableUuid`, which is anchored on the live object's uuid and explicitly preserved across a rename (`:3330-3340`). It does not fire here — the `armContext` line this PR changes is the one that did.
- **Finding, and it is being fixed in a follow-up rather than blocking: the shipped behaviour change is unpinned.** Inverting the helper kills 2 of 6 tests, but replacing the `renaming ? ...RENAMED... :` ternary at `:28699` with the old unconditional switch wording kills **zero**. `isSameInstanceRename` appears in one test file only, and no test anywhere contains the old string. This is the repo's own "install line invisible to tests" shape — the fourth independent instance found this session. #1344 got it right with source assertions and this should copy that.
- Nit: `previousName` splits on forward slash only, so a Windows-style path renders whole rather than as a basename. Cosmetic.
- 5037 tests, one failure = the known #671 load flake. Four gates green, `tsc --noEmit` clean. Playwright not run: the human-facing `tr()` line is untouched and only an agent-facing string changes.

Issue #604, opened 2026-08-04
Canvas/tab binding desync after ComfyUI restart: reads blocked but MUTATIONS still route to the wrong workflow tab (nodes deleted from wrong graph)


artokun:
### What happened / root cause
After a ComfyUI restart + panel reconnect (and again after `panel_reload scope:frontend`), the agent's graph binding repeatedly pointed at the WRONG canvas while the user had multiple tabs open (7+ tabs, several unsaved `tmp:` tabs):

1. `panel_graph_outline` correctly refused with `The live graph is out of sync with the active workflow: the workflow reports N node(s), but the canvas is bound to a different graph` — but MUTATING commands did NOT get the same guard: `panel_add_node`, `panel_connect`, `panel_set_widget`, `panel_edit_node`, and `panel_remove_node` all EXECUTED against whatever canvas happened to be bound, which was a different tab than the one just confirmed by `panel_graph_outline` moments earlier.
2. Concretely: 10 `panel_add_node` calls landed on tab A; the immediately following `panel_connect` batch executed against tab B (errors referenced node types that only exist on tab B); a later `panel_remove_node` batch executed against tab C and DELETED 4 correctly-configured loader nodes from a graph the agent had just verified on a different canvas. The agent had to manually re-add and re-wire them.
3. `panel_open_workflow` returned `workflow_open could not prove that the active canvas was rebound to the requested workflow` on EVERY call (for both `tmp:` keys and persisted filenames), yet sometimes DID switch the view (applied:"unknown" receipts) — so the agent cannot deterministically navigate tabs.
4. `panel_new_workflow` succeeded (`created:true`) but the fresh empty tab was immediately out-of-sync for `panel_graph_outline`.
5. One `panel_list_workflows` call failed transiently with `panel tab wf:workf was replaced by a different workflow — in-flight "workflow_list" cancelled`.
6. `tmp:` routing keys regenerate on every reconnect, so keys captured before a reconnect are useless afterward (`no connected tab with id ...`).

Root cause (from behavior): the out-of-sync guard that protects `graph_outline` is not applied consistently to the mutation path, and the canvas→workflow binding is re-established lazily/racily after reconnects and tab switches, so a stale binding can serve an entire mutation batch.

### Steps to reproduce
1. Open 6+ workflow tabs (mix of persisted and unsaved) in ComfyUI with the agent panel connected.
2. Restart ComfyUI (or trigger a panel frontend reload) mid-session.
[...]

PR #1344, opened 2026-08-19
fix(1215): an untagged canvas is no longer captured into the target tab's state on a switch

### LINKED ISSUE #1215: After panel_open_workflow tab switch, graph reads (panel_graph_outline/query) silently return the PREVIOUS tab's graph while the fence/active both name the new workflow
Sequence: active tab was `video_upscale_seedvr2` → agent called `panel_open_workflow(path:"workflows/image_krea2_turbo_t2i.json")` → result reported applied:true, active_matches_target:true, workflow_uuid 7e7ab2d6 (the krea2 tab). `panel_set_workflow_target(mode:'current')` then reported graph_binding:'bound', already_current on instance 7e7ab2d6. `panel_list_workflows` confirms active = image_krea2_turbo_t2i (active_confirmed:true).

Yet `panel_graph_outline` (called twice, with a rebind in between, ~45s after the open) returns the FULL GRAPH OF THE PREVIOUS TAB (`video_upscale_seedvr2`: LoadVideo/SeedVR2VideoUpscaler/FrameInterpolate nodes, ids 1-13, including live randomized seed values) — not the krea2 t2i graph that is on disk for the active tab (SaveImage + ResolutionSelector + subgraph b0e5ca93…). No mismatch error is raised: the read succeeds under the NEW instance fence while serving the OLD tab's canvas object.

This is the dangerous inverse of the instance-mismatch protection: the fence, list, and open receipt all agree on the new workflow, so the agent has no signal that reads (and presumably WRITES — panel_add_node/panel_connect) would target the stale canvas. The agent only caught it because the returned node types obviously belonged to the other workflow; a subtler graph would have been silently mis-edited. Agent stopped before editing.

Likely cause: the frontend graph accessor used by graph commands still points at the previous tab's graph object after a programmatic workflow_open (canvas repaint/handoff not awaited), while the identity/fence bookkeeping updates immediately.

### Steps to reproduce
1. Open two workflows in tabs (A active).
2. `panel_open_workflow` to B (applied:true, active_matches_target:true).
3. Immediately (or within ~1 min) call `panel_graph_outline` — observe it returns tab A's nodes while `panel_list_workflows` says B is active (active_confirmed:true).

### Exact error (scrubbed)
No error — silent wrong-canvas read. Open receipt also carried geometry_rewritten:['size'] (#1001 note), possibly related to the open not fully settling.

### Fix
applied locally: no  ·  upstream-only: yes
Precise change needed: after a programmatic workflow_open, graph commands must resolve the canvas object THROUGH the same identity they stamp (workflow instance uuid) — or fail closed with an instance mismatch when the live graph object's identity differs from the fence, instead of serving whatever graph the accessor currently holds. A repaint-settled barrier (await the frontend's activeWorkflow → graph object swap) before ack'ing workflow_open would also close the window.

### Environment
OS: Windows 11 Pro 10.0.26200 · ComfyUI: Desktop, port 8000 · comfyui-mcp: 0.51.38 (npx cache 28eae33d) · panel: 0.13.7 · frontend: Desktop-bundled (version not read — graph reads untrusted in this state)

artokun:
## Root cause

`workflow_open`'s repaint poisons the target tab's in-memory state and then faithfully reproduces it. After `s.openWorkflow(target)` moves the active pointer to tab B but before anything repaints, the executor calls `target.changeTracker.checkState()` — ComfyUI's capture, which serializes the STILL-MOUNTED root (tab A's canvas) into the now-active tracker, i.e. into B's `activeState`. The repaint then loads that same state, so all four rebind proof parts are true statements about a poisoned source, and the open returns PROVEN, publishing B's `workflow_uuid` while A's graph sits on the canvas under B's stamp. Every later graph read honestly answers from the wrong graph — the silent wrong-canvas success in #1215's reports (reproduced as late as mcp 0.51.57 / panel 0.14.44, where `panel_graph_outline` returned an unrelated default canvas after a proven open).

The #968 guard already refused the capture on a POSITIVE identity conflict (`describeLiveCanvasBinding` → `foreign`), but recorded its own residual: an UNTAGGED root answers `unknown`, and `unknown` captured. After a restart with `Comfy.Workflow.Persist=false` — exactly the reported configuration — nothing has stamped the roots yet, so every root is untagged and the poisoning path was wide open. #968/#1089/#1111 are the tagged ancestors of this same defect; this closes the untagged residual they left.

## Fix

Panel-side only, in `web/js/comfyui-mcp-panel.js`:

1. **Snapshot the pre-switch active workflow** (`activeBefore = activeWorkflowRef()`) before `openWorkflow` moves the pointer.
2. **Tighten the capture gate.** `bound` still captures outright (that is the #874 already-current case and the #604/#708 divergence, where node-written values only exist on the mounted canvas), `foreign` still skips. `unknown` now captures only when the pointer did NOT move — on a switch, `openWorkflow` moved the pointer and repainted nothing, so the mounted canvas is the previous tab's and an untagged capture can only import the other tab's graph. The codex note that previously blocked this ("never guesses from 'did the pointer move'") is REWRITTEN in place rather than silently inverted: that objection holds for refusals, where pointer movement proves nothing; here the snapshot gates a WRITE whose own precondition is "the mounted canvas is the target's".
3. **Disclose what remains.** A state poisoned before this gate existed is untagged, so `describeRepaintSourceBinding` can only answer `unknown` for it and the existing `foreign_source_state` disclosure never fires. A new, deliberately weaker `unproven_source_state` key is emitted when the open switched tabs and the repaint source carried no resolvable identity — it says VERIFY THE GRAPH BEFORE EDITING, names `panel_graph_outline`/`panel_load_workflow`, and is explicit that an untagged state is the NORMAL case for a workflow the panel never stamped, so it fires on good opens too ("MAY be, not IS").

No refusal is added anywhere: the #1089 analysis stands — refusing after the pointer moved strands the session with every `graph_*` command fenced, including the `panel_load_workflow` the refusal would recommend.

Accepted cost, recorded in the comment: in the #604/#708 divergence (target's canvas mounted while the pointer sat on another tab) a skipped capture forfeits preserving node-written values that only ever existed on that mounted canvas; the repaint reverts them. The alternative is the silent wrong-graph serve this fixes.

## Verification

- `npm ci` — clean, 0 vulnerabilities.
- `npm run test:unit` — **5034 tests, 5033 pass, 0 fail, 1 todo (pre-existing)**.
- `npm run typecheck` (`tsc --noEmit`) — clean.
- New tests in `browser_tests/unit/open-rebind-proof.test.mjs` pin: the pre-switch snapshot precedes `openWorkflow`; the gate requires `bound` or a non-moved pointer (and `if (captureBinding !== "foreign")` alone must not come back); `sourceUnproven` requires pointer-moved + no foreign finding + no `workflow_uuid` on the source; the `unproven_source_state` reply carries the verify/read/recover wording with the MAY-not-IS hedge.

Not verified: a live-browser reproduction of the original sequence (no ComfyUI instance driven here). The fix follows the mechanism the in-repo triage (`docs/triage/2026-08-14-open-issues.md`, #1215 section) established from the captured receipts, and the behavior change is pinned by the unit tests above.

Closes #1215

### DIFF
--- a/browser_tests/unit/open-rebind-proof.test.mjs
+++ b/browser_tests/unit/open-rebind-proof.test.mjs
@@ -782,11 +782,13 @@ test("#702: the fence note is ONLY on the content-unverified outcome", () => {
 //
 // Nothing was fooled. All four parts of resolveOpenRebindVerdict are taken
 // against the root the loader produced, so each was a true statement about a
-// poisoned SOURCE state. #968 closed the writer this repo owns and recorded
-// that an untagged root is still captured; a state already poisoned still
-// reaches the load, and the load is what makes it durable — it stamps the
-// target's identity onto the foreign graph, so a later save writes it to the
-// target's file.
+// poisoned SOURCE state. #968 closed the writer this repo owns, and #1215 then
+// closed the untagged-root residual it had recorded — the capture now runs on
+// an untagged root only when the pointer never moved. A state poisoned BEFORE
+// that gate still reaches the load, and the load is what makes it durable — it
+// stamps the target's identity onto the foreign graph, so a later save writes
+// it to the target's file — which is why the open reply carries the
+// `unproven_source_state` disclosure for exactly that combination.
 // ---------------------------------------------------------------------------
 
 /** A serialized tab state carrying the panel's durable workflow tag. */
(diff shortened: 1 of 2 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and the `claude-gate.mjs` fallback spawns a headless `claude` blocked on the account weekly limit. Under the owner's standing authorisation, this PR was gated by an **independent adversarial reviewer that had never seen the author's reasoning**, run against `.claude/autopilot/review-taxonomy.md`. Disclosure is the condition that substitution was accepted under. Verdict: **SHIP**, on head `1e67f26`.

- **Absent and mismatched are now distinguishable** at the capture gate (`comfyui-mcp-panel.js:15394-15400`): `bound` captures, `foreign` skips, `unknown` captures only if the pointer did not move. `activeBefore` is snapshotted at `:15171`, genuinely before `await s.openWorkflow(target)` at `:15245`, and compared proxy-safely. The #874 already-current case still captures.
- **Mutations die on the call site, not the helper:** restoring `if (captureBinding !== "foreign")` fails one test; setting `sourceUnproven = false` fails another. 39/39 green unmutated.
- **Finding, non-blocking but worth fixing: the guard over-warns on the most common open.** `sourceUnproven` (`:15479-15483`) ignores `wasOpen`, which is in scope from `:15164` and which this file's own comment at `:15161` says proves the case. Concretely: `panel_open_workflow("never-opened.json")` from a different active tab, on a workflow the panel never stamped, gives `wasOpen === false` with no tracker to poison and the disk copy as the repaint source — yet the reply still carries the ~1000-char `unproven_source_state` block. It hedges MAY rather than asserting IS, so it is not wrong data, but it is a false positive on the normal shape and costs the caller an extra `panel_graph_outline` each time. A follow-up will gate it on `wasOpen`.
- Nit: the empty-string tag is treated as tagged here, while the adjacent `describeRepaintSourceBinding` (`web/js/lib/graph-binding.js:1596`) also rejects falsy. Unreachable from panel-written state; consistency only.
- The #604/#708 cost is real, documented in the diff, and the correct trade against the silent wrong-graph serve #1215 reproduced on 0.51.57 / 0.14.44.
- Playwright not run (the diff adds a reply key); no live-ComfyUI reproduction of the original sequence, as the PR itself states.

Issue #389, opened 2026-07-31
Panel graph read tools return an empty graph while the session reports missing nodes


artokun:
### What happened / root cause
The session context for the active workflow `workflows/Unsaved Workflow.json` reported red missing-model nodes 345 and 346. In the same agent turn, `panel_list_workflows` reported that workflow as active and modified, but both `panel_graph_outline` and `panel_get_errors` returned `node_count: 0` with no errors. This makes it impossible to inspect or repair the reported nodes.

Root cause is not yet isolated. The active-workflow registry and graph-read bridge appear out of sync.

### Steps to reproduce
1. Open a modified `Unsaved Workflow` containing red missing-model nodes.
2. Start the attached sidebar workflow conversation.
3. Call `panel_list_workflows`, then `panel_graph_outline` and `panel_get_errors`.
4. Observe the workflow listed as active while graph-read results contain zero nodes.

### Exact results
- `panel_list_workflows`: active `workflows/Unsaved Workflow.json`, modified true, persisted false.
- `panel_graph_outline`: `{ "node_count": 0, "group_count": 0, "viewing": { "scope": "root" } }`
- `panel_get_errors`: `{ "node_count": 0, "errored_count": 0, "nodes": [] }`

### Fix
Applied locally: no. Upstream-only: yes pending a reproducible active-tab/graph-state bridge trace. No user workflow data was changed.

### Environment
Windows 11; ComfyUI 0.29.0; RTX 5090 32 GB plus RTX 3090; comfyui-mcp 0.48.21; panel 0.11.20.

artokun:
Regression report on current versions: comfyui-mcp 0.52.167 and panel 0.15.150 (ComfyUI 0.33.0, Linux). Starting from an empty active `workflows/Untitled 2026-09-01 16-45-15.json`, the user loaded the MiniMax H3 Reference to Video template into the visible canvas. Despite `mutations_ready=true`, `canvas_binding=bound`, and `graph_binding=bound`, a subsequent `panel_graph_outline` still returned a silent `node_count=0`, `group_count=0`; `panel_list_workflows` continued to name the same active tab. This is the false-clean empty graph read previously addressed here, now recurring on the latest panel release.

artokun:
[...]

PR #1345, opened 2026-08-19
fix(1233): a proven read keeps its stale-tag bypass through the executor's own fence re-assert

### LINKED ISSUE #1233: Tab switch can leave stale workflow UUID and reject live graph reads
After the user switched to an unsaved workflow tab, panel_graph_outline rejected a read with [root-workflow-uuid-mismatch]. The active tab routing key was known, and reopening it with panel_open_workflow repaired the binding. This looks like a panel/orchestrator tab-switch race leaving a stale workflow identity tag.

### Steps to reproduce
1. Keep an unsaved workflow tab open.
2. Switch the visible workflow tab.
3. Immediately call panel_graph_outline.
4. Observe root-workflow-uuid-mismatch despite the panel event naming the active workflow.

### Exact error (scrubbed)
[root-workflow-uuid-mismatch] The live canvas carries a different workflow's identity tag than the active workflow, so this command was NOT applied.

### Fix
Applied locally: no
Upstream-only: yes
Precise change needed: rebind/update the live canvas workflow UUID atomically when the active tab changes (including unsaved tabs), before graph tools validate root identity. The successful temporary recovery was panel_open_workflow with the tab routing key.

### Environment
Windows 11; ComfyUI 0.30.0; frontend version not available in the current environment summary; RTX 4070 SUPER 12 GB; comfyui-mcp 0.51.49; panel 0.14.2.

artokun:
## Root cause

The binding fence runs **twice** for `graph_outline` and `graph_get_errors`:

1. Bridge dispatch asserts with `graphCommandBindingBar(msg.cmd)` — the read bar, which carries the **#995 stale-tag bypass** (a read on a canvas proven content-equal to the active tab, exclusive among open tabs, proceeds without touching the tag).
2. The executor then **re-asserts** for the #389 baseline read guard — but passed **no options**, and the bypass defaults OFF.

So the re-assert recomputed the same `root-workflow-uuid-mismatch` and refused the canvas the dispatch fence had just proven. That is the reported recurrence: a tab switch onto an **unsaved/modified** tab leaves the previous tab's identity tag on the reused `app.graph` (the #817 mechanism — clear/configure does not reset `graph.extra`); the #995 bypass admitted the read at dispatch; the executor's bare re-assert revoked it. Every other read executor relies on the dispatch fence alone, which is why the surviving symptom named `panel_graph_outline`, and why `panel_open_workflow` (which re-stamps the root) was the only recovery.

This is the residual the 2026-08-14 triage doc predicted for #1233: not the #817 clean-tab rebind, not the #995 dispatch bypass, not the #1209/#1279 orchestrator fence refresh — all shipped — but the gap between them.

## Fix

Both re-asserts now pass the command's **own classified read bar** plus the executor-only `includeBaselineReadGuard: true`:

- `web/js/comfyui-mcp-panel.js` — `graph_outline` and `graph_get_errors` executors call the fence with `{ ...graphCommandBindingBar(<cmd>), includeBaselineReadGuard: true }`.
- The bypass stays opt-in via the one classification site (`graphCommandBindingBar`); nothing new can acquire it by omission.

## What still refuses (pinned by tests)

- a foreign canvas (content differs from the active tab's own state)
- a dirty or identical **twin** tab (exclusivity unprovable)
- every **mutation** on the same evidence (the bypass is read-only by classification)

## Verification

- New `browser_tests/unit/stale-tag-read-executor.test.mjs` drives the **shipping fence** (extracted from the monolith, the graph-binding.test.mjs idiom) through the reported state: dispatch bar admits it, bare executor defaults refuse it with the report's exact error, the fixed executor bar passes it and writes nothing — and the refuse-cases above still refuse. 7/7 pass.
- `npm run test:unit`: 5036 pass / 1 fail — the failure is `#671 verifyInstalled ... fast-draining install` in manager-install.test.mjs, a timing-sensitive flake that **also fails on pristine origin/main** (verified by stashing this change: 5029 pass / same 1 fail) and passes in isolation (125/125).
- `npm run typecheck` (`tsc --noEmit`): clean.

Not verified: a live ComfyUI rig (none available in this environment) — the fix is proven at the extracted-shipping-code level only.

Closes #1233

### DIFF
--- a/browser_tests/unit/read-only-graph-commands.test.mjs
+++ b/browser_tests/unit/read-only-graph-commands.test.mjs
@@ -107,9 +107,10 @@ test("the two other unlisted reads are deliberately still refused", () => {
 // reporter's refusal came from, and it is a whole-file search, so it is exact.
 //
 // NOT TESTED, verified by reading: `graph_get_errors`'s executor writes nothing —
-// it calls `assertGraphBoundToActiveWorkflow(graph, rootGraph)` with no options
-// (so it already took the read-shaped defaults; only the DISPATCH bar was wrong)
-// and otherwise only reads `graph._nodes` and builds a report.
+// it re-asserts the fence on this command's own read bar plus the baseline guard
+// (`graphCommandBindingBar("graph_get_errors")` with `includeBaselineReadGuard`,
+// since #1233 — bare defaults had silently dropped the #995 stale-tag bypass) and
+// otherwise only reads `graph._nodes` and builds a report.
 //
 // A source-scanning version of that check was written and REMOVED. Bounding a
 // method body by counting braces needs real tokenization: masking comments first
--- a/browser_tests/unit/stale-tag-read-executor.test.mjs
+++ b/browser_tests/unit/stale-tag-read-executor.test.mjs
@@ -1,0 +1,260 @@
+/**
+ * #1233 — after a tab switch onto an UNSAVED (never-saved, modified) workflow tab,
+ * panel_graph_outline was refused `[root-workflow-uuid-mismatch]` even though the
+ * #995 dispatch-fence bypass had already PROVEN the canvas: content-equal to the
+ * active tab's own tracker state, exclusive among open tabs, nothing written.
+ *
+ * The fence runs TWICE for those reads. Bridge dispatch asserts with
+ * `graphCommandBindingBar(msg.cmd)` — the read bar, which carries the #995
+ * stale-tag bypass — and the graph_outline / graph_get_errors executors then
+ * RE-assert for the #389 baseline guard. The executors passed NO options, and the
+ * bypass defaults OFF, so the re-assert recomputed the same mismatch and refused
+ * the canvas the dispatch fence had just admitted. Every other read executor
+ * relies on the dispatch fence alone, which is why the surviving symptom named
+ * graph_outline. Re-opening the workflow re-stamps the root tag, which is why
+ * panel_open_workflow "fixed" it.
+ *
+ * Proven here against the SHIPPING fence, extracted from the monolith and driven
+ * with doubles (the graph-binding.test.mjs idiom): the dispatch bar passes the
+ * reported state, bare executor defaults refuse it, and the read bar the
+ * executors now re-assert with passes it — while a foreign canvas, a twin, and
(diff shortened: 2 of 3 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and the `claude-gate.mjs` fallback spawns a headless `claude` blocked on the account weekly limit. Under the owner's standing authorisation, this PR was gated by an **independent adversarial reviewer that had never seen the author's reasoning**, run against `.claude/autopilot/review-taxonomy.md`. Disclosure is the condition that substitution was accepted under. Verdict: **SHIP**. Head reviewed: `1470c96`.

- **The "claim outlives its window" hypothesis was refuted: nothing is cached.** `assertGraphBoundToActiveWorkflow` (`comfyui-mcp-panel.js:7297`) re-derives `rootContentProvesActiveWorkflowDespiteEdits` plus open-tab exclusivity from live state on **every** call, so the executor re-asserts at `:10287`/`:14294` are fresh proofs, synchronous, in the same dispatch — not a carried verdict. No cross-async identity read is introduced, so `tabIncarnation` is not in play.
- The delta is exactly one flag (`staleTagReadBypass` false to true); everything else matches bare defaults, and it is the behaviour `:20163` already documents.
- **Scope claim verified:** the other seven assert sites are MUTATION_BINDING_BAR (`:7790`, `:7996`, `:9895`, `:11852`, `:13630`) or post-nav proof callbacks (`:17870`, `:17990`). No other bare-default read re-assert exists.
- `graph_outline`'s executor does write via `syncGraphNodeAreas(graph)` after the assert, but only derived bounding cache from live pos/size, undo-recorded, and already on every passing read; since the bypass proves content equality, the output is identical whichever tab owns the canvas. Not a new hazard.
- **Mutation:** pointing both asserts at bare defaults fails the source guard "both re-asserting read executors pass their OWN classified bar". The six behavioural tests drive the extracted fence rather than the executor and survive — thin, but it dies, and the source guard is the mitigation this repo prescribes for exactly that shape.
- 14/14 on both touched files; full suite 5036 pass, one failure = the known #671 17.7s flake the PR itself named. Playwright not run — the diff changes nothing the sidebar renders.

PR #1346, opened 2026-08-19
fix(1193): a slow-but-healthy fetch no longer starves the combo refresh of its budget

### LINKED ISSUE #1193: The refresh run budget is thin: refreshComboInNodes alone measures ~4.8s of a 9s run
## Summary

`NODE_DEFS_RUN_BUDGET_MS` is 9000ms for a whole `registerComfyNodeDefs` run. Measured on a real rig, the combo phase alone needs about half of that — so a slow-but-healthy fetch can leave it too little and it is abandoned, reporting `combo_refresh_failed` for a refresh that would have completed.

Follow-on from #1180 (PR #1186). Not a regression from a hang — before #1180 the combo phase was unbounded and always completed — but it is a real narrowing that #1180 introduced, and the numbers say the margin is thinner than the budget's own comment implies.

## Measured

In-page, ComfyUI 0.32.0, frontend 1.48.7, 4320 node types, warm:

```
api.getNodeDefs()             456 ms
registerNodesFromDefs()      3972 ms   (unbounded local work, excluded from the budget)
refreshComboInNodes()        4846 ms
```

`refreshComboInNodes` is expensive because it does its own `getNodeDefs()` and then walks every node in the graph rewriting each combo widget's options (`reloadNodeDefs` in the frontend's `app.ts`).

## The arithmetic

The fetch phase is allowed `2/3 × 9000 = 6000ms`; the combo phase gets whatever is left of the run deadline.

| fetch actually takes | combo is left | combo needs | result |
|---|---|---|---|
| 456 ms (measured, warm) | 8544 ms | 4846 ms | fits |
| 1062 ms (measured, cold) | 7938 ms | 4846 ms | fits |
| 4000 ms (slow link) | 5000 ms | 4846 ms | barely |
| 6000 ms (its full allowance) | 3000 ms | 4846 ms | **abandoned** |

So the failure needs a fetch slow enough to eat most of its allowance while still succeeding. That is exactly a congested or remote backend — the case the bound was written for.

The consequence is not just a wrong verdict: `nodeDefsRefreshConfirmed` stays false, which puts the missing-asset scan back into over-report-safe mode and is what reopens #610's false "model still missing".

## Why it is not simply "raise the budget"

The budget is sized so two serialised runs fit the orchestrator's 20s ui-bridge read default (`2 × 9000 = 18,000`). Raising it to 12,000 makes that 24,000 and breaks the property it was chosen for. Any real fix has to either

1. give the combo phase a guaranteed floor rather than only a remainder, and accept that a run can exceed the budget by that floor; or
2. size the whole thing from measured phase costs rather than from the read default alone; or
3. stop treating the combo refresh as something worth waiting for at all — it mutates the graph correctly whether or not the panel waits (see the note at the combo bound), so the panel could report it as "started, not confirmed" instead of failing it.

(3) is probably right and is the smallest change, but it moves what `nodeDefsRefreshConfirmed` means, which several callers depend on.

## Related

- #1192 — the same class one level up: `graph_add_node`'s bounds sum past the 30s command budget.
- #1188 — 34 more unbounded awaits of the same family.

artokun:
Closes #1193

## Root cause

`registerComfyNodeDefs` runs under one 9000ms deadline (`NODE_DEFS_RUN_BUDGET_MS`, #1180). The fetch phase may consume its full 2/3 share (6000ms) and still succeed — leaving the combo phase a 3000ms remainder against a measured 4846ms cost (`refreshComboInNodes` re-fetches `/object_info` and rewrites every combo widget in the graph). In exactly the congested/remote-backend case the budget was written for, a **healthy** combo refresh was abandoned: `panel_refresh_nodes` reported `combo_refresh_failed` over a refresh that would have completed, `nodeDefsRefreshConfirmed` stayed false (reopening #610's false "model still missing"), and the loader combo stayed stale so the next `panel_set_widget` rejected the newly installed value — the reported repro.

## Fix

The issue's option (1): a **floor** under the combo phase's remainder — `NODE_DEFS_COMBO_FLOOR_MS = 6000` (the measured 4846ms plus a quarter), applied as `Math.max(nodeDefsBudgetLeft(runDeadline), NODE_DEFS_COMBO_FLOOR_MS)` at the combo bound. A fast fetch is untouched (its remainder exceeds the floor); a slow-but-successful fetch no longer starves the combo phase.

Option (3) — stop waiting for the combo refresh — was rejected and the reason is recorded at the constant: it fixes only the verdict's wording while leaving every symptom in place (combos still stale when the reply lands, the follow-up widget write still refused).

**Stated cost:** a run can now exceed its budget by up to the floor — worst-case bounded waiting 15,000ms per run, 30,000ms for two serialized runs against the 20s bridge read default, only in the case the floor exists for. The #1192 ratchet test carries the new number (33000 → 45000), so the sum is tracked rather than silently grown.

## What changed

- `web/js/comfyui-mcp-panel.js` — the floor constant (with the measurement and the cost stated), the combo bound, the timeout message ("…remaining budget or its floor"), and the budget comments updated to match.
- `browser_tests/unit/node-def-refresh.test.mjs` — three new tests driving the **shipped** executor on an injected clock and virtual timers: the issue's failing row now completes `refreshed: true`; the floor arms at exactly 6000ms when the remainder runs out and still bounds a never-answering combo; a fast fetch keeps the whole remainder (the floor never shrinks it). The harness's `nodeDefsBudgetLeft` now shares the injected clock when one is given — it previously read the real clock, which mismeasured a fake-clock deadline.
- `browser_tests/unit/_panel-constants.mjs`, `refresh-graph-guard.test.mjs` — the new constant read from source / threaded into the rebuilt executor.
- `browser_tests/unit/single-node-def.test.mjs` — the combo-call structural pin updated to the floor shape; floor arithmetic pinned (must beat the worst-case remainder, must not exceed the run budget); the #1192 ratchet counts the floor.

## Verification

- `npm ci` — clean.
- `npm run typecheck` — exit 0.
- `npm run test:unit` (i18n-check + i18n-render-check + check-tool-vocabulary + check-panel-scope + `node --test browser_tests/unit/*.test.mjs`) — **exit 0, 5033 pass / 0 fail.**
- Mutation check: with the `Math.max(…, NODE_DEFS_COMBO_FLOOR_MS)` reverted to a bare remainder (constant kept), the two key #1193 tests fail and pass again with the fix — they detect the pre-fix behavior, not just the text.
- Note: `#671 verifyInstalled … fast-draining install` (a real-wall-clock, 3s-budget test) flaked twice on this loaded machine during full-suite runs — including once on **clean origin/main with these changes stashed** — and passes in isolation both ways. Pre-existing environmental flake, unrelated to this change.

### DIFF
--- a/browser_tests/unit/_panel-constants.mjs
+++ b/browser_tests/unit/_panel-constants.mjs
@@ -61,6 +61,11 @@ export const NODE_DEFS_FETCH_SHARE = (() => {
   return Number(m[1]) / Number(m[2]);
 })();
 
+export const NODE_DEFS_COMBO_FLOOR_MS = readPanelNumber(
+  /const NODE_DEFS_COMBO_FLOOR_MS = (\d+);/,
+  "the combo phase's guaranteed floor",
+);
+
 /**
  * The panel's sentinel for "this call did not answer", mirrored so a rebuilt executor
  * compares against the same value its injected `boundedGetNodeDefs` resolves.
--- a/browser_tests/unit/node-def-refresh.test.mjs
+++ b/browser_tests/unit/node-def-refresh.test.mjs
@@ -22,6 +22,7 @@ import { withTimeout } from "../../web/js/lib/bounded-step.js";
 import {
   COMBO_NO_ANSWER,
   COMBO_OK,
+  NODE_DEFS_COMBO_FLOOR_MS,
   NODE_DEFS_FETCH_SHARE,
   NODE_DEFS_FETCH_TIMEOUT_MS,
   NODE_DEFS_NO_ANSWER,
@@ -389,6 +390,7 @@ function buildRegisterComfyNodeDefs({
   appValue,
   apiValue,
   timers,
+  now,
   recordObjectInfoTypes = () => ({}),
   reapplyDefsToLiveNodes = () => {},
 }) {
@@ -410,6 +412,7 @@ function buildRegisterComfyNodeDefs({
     "NODE_DEFS_FETCH_TIMEOUT_MS",
     "NODE_DEFS_RUN_BUDGET_MS",
     "NODE_DEFS_FETCH_SHARE",
+    "NODE_DEFS_COMBO_FLOOR_MS",
     "nodeDefsBudgetLeft",
     "monotonicNow",
     "NODE_DEFS_RETRY_DELAYS_MS",
@@ -456,8 +459,13 @@ function buildRegisterComfyNodeDefs({
     NODE_DEFS_FETCH_TIMEOUT_MS,
     NODE_DEFS_RUN_BUDGET_MS,
     NODE_DEFS_FETCH_SHARE,
-    nodeDefsBudgetLeft,
-    monotonicNow,
+    NODE_DEFS_COMBO_FLOOR_MS,
+    // #1193 — ONE clock for the deadline and the arithmetic that reads it. The shared
+    // helper reads the REAL monotonicNow; a fake-clock deadline measured against it is a
+    // different bound than the one the panel arms (its own nodeDefsBudgetLeft shares the
+    // panel's clock). Without the override the shared helper is passed through unchanged.
+    now ? (deadline, share = 1) => Math.max(1, Math.floor((deadline - now()) * share)) : nodeDefsBudgetLeft,
+    now ?? monotonicNow,
     NODE_DEFS_RETRY_DELAYS_MS,
     // #716 — the shipping function drops the widget-write burst cache after a successful
     // fetch. A spy, so the harness can prove that happens rather than merely tolerate it.
@@ -872,6 +880,15 @@ function virtualTimers() {
         t.fn();
       }
     },
+    // Advance the clock to `limitMs`: fire every armed timer whose delay is within it,
+    // so a test can let a 4846ms step COMPLETE without firing a 6000ms bound.
+    fireUpTo(limitMs) {
+      for (const [id, t] of [...pending]) {
+        if (t.ms > limitMs) continue;
+        pending.delete(id);
+        t.fn();
+      }
+    },
   };
 }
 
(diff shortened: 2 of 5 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**, against `.claude/autopilot/review-taxonomy.md`. Verdict: **SHIP**, head `9451fd2`.

The specific thing this issue invites is "a bigger constant wearing a mechanism's clothes", so that was the hunt.

- **It is a mechanism.** `Math.max(nodeDefsBudgetLeft(runDeadline), NODE_DEFS_COMBO_FLOOR_MS)` (`comfyui-mcp-panel.js:1461`) decouples the combo allowance from what the fetch spent — raising `NODE_DEFS_RUN_BUDGET_MS` could not do that. It implements #1193's own stated option (1), cost included.
- **Nothing stops running.** Combo is still awaited and can only get MORE time; a third test pins that a fast fetch keeps the whole remainder. A failed fetch throws out of `fetchNodeDefsWithRetry` (`:1270`) before the combo phase, so the floor cannot pay out on a run with no defs.
- **Mutation:** reverting `Math.max(...)` to the bare remainder fails 3 tests, including the source-text call-site pin at `single-node-def.test.mjs:318`. The call site itself is asserted — which matters, because five PRs this session shipped with helper-only tests that stayed green when the wiring was deleted.
- **Starvation moves, and the gate quantified where.** Worst case per run 9000 to 15000; two serialized runs 18000 to 30000. Mitigating: `refresh_nodes` is dispatched with `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS = 30_000` (`src/orchestrator/panel-tools.ts:821`), not the 20s default the issue cites, so the worst case lands *at* that ack timeout rather than past it. #1192's ratchet carries it (33000 to 45000, with 45000 exactly the new worst case — zero headroom).
- **Named gaps, neither disqualifying.** The floor is one rig's number (6000 = a measured 4846 times 1.24 at 4320 types), so a larger install can still starve — but when it misses, the behaviour is the old pessimistic verdict, not a new wrong output. And there are **no new before/after measurements**: the PR re-uses the issue's figure and verifies on a virtual clock, so the sizing rests on that single-rig number.
- 5032 unit tests pass, 1 failure = the known #671 wall-clock flake (fails on the other branch and on this loaded box too). `check-panel-scope` OK. Playwright not run — it needs ComfyUI at :8188, which is the owner's rig and currently down, and its numbers would be near-meaningless for a timing change under a dozen concurrent agents anyway.

PR #1347, opened 2026-08-19
fix(654): a ComfyUI restart the bridge SURVIVES re-advertises the tab route

artokun:
Refs the underlying cause of artokun/comfyui-mcp-panel#654.

## The mechanism

The panel re-registers its bridge route by exactly one act: sending a `hello`.
On the ComfyUI-restart path nothing sends one any more.

The pack is pure-frontend and can no longer spawn the orchestrator, so
`externalOrchestratorMode()` is hardcoded `true` (`comfyui-mcp-panel.js:4574`).
The orchestrator therefore runs out-of-band and **always survives** a ComfyUI
restart. Follow what that leaves:

| step | what happens |
|---|---|
| the bridge socket | never closes — the orchestrator did not die |
| the socket `open` handler | never fires, so its `sendHello()` never runs |
| `onComfyReconnected` | `shouldResumeAfterComfyReconnect` returns `false` on `bridgeConnected` (#278, correctly) and the handler returns having done nothing |
| `connectAgent()` (had it been reached) | `connect()` early-returns on an already-OPEN socket, so it cannot produce a hello either |

The tab sits on a live socket it never re-announced. Orchestrator-side,
`awaitPostRestartReachable` waits for a **strictly newer** hello generation
(`current.generation > before.generation`), which can never arrive — so
`panel_restart_comfyui` reports a confirmed server cycle with the tab-reconnected
and graph-tools-ready flags both false, on **every** restart, and a manual
browser refresh is the only recovery. That is this report verbatim, including
the 2026-08-14 and 2026-08-15 recurrences.

This is the remaining half of Cluster A in `docs/triage/2026-08-14-open-issues.md`
(committed to main in #1234). The other half, #584's version re-probe, landed in
0.15.1 as #1317 and hooks the same `reconnected` signal.

## The fix

The repair this panel already ships for a `free_vram` bounce (#310):
**re-advertise the route**. Triggered by the `reconnected` event rather than a
command reply, because at `comfy_reboot` reply time ComfyUI is on its way down
and a hello then reaches nothing.

**#278's guard is not relaxed.** `if (bridgeConnected) return false` is what
stops a benign WS blip bouncing a live agent session, and it stands untouched
(pinned by a test). This is a second, cheaper action on the branch that guard
declines, and the two are mutually exclusive by construction — resume needs the
bridge DOWN, this needs it UP. A cross-product test asserts they can never both
fire rather than reasoning that they cannot.

A landed hello is not a cheap frame — it bumps the agent-session epoch, draws a
fresh `ready` ack and runs the orchestrator's panel sync — which is why it is
gated on four things rather than fired on every `reconnected`:

- **the bridge stayed UP.** If it dropped, its own socket-open hello already
  registered the tab; a second here is two greetings and two ready acks for one
  outage (#1138's harm).
- **no hello LANDED during the outage** — counted from hellos that provably
  reached the wire, never from ones the panel meant to send. This covers the
  deployment where the bridge *does* drop and re-hellos before `reconnected`
  arrives.
- **one-shot per outage.** `reconnected` repeats.
- **a MEASURED outage** at or past the same 6s threshold
  `shouldNudgeAfterMidTaskReconnect` already uses — deliberately one number, not
  a second to keep in sync. The elapsed interval is the only thing separating a
  restart from a blip (asset view, image check, tab refocus), and absent, zero,
  negative or unreadable all mean "not measured", never "long".

The outage accounting is a tracker over injected time, like #1145's bridge one
and for the same reason: **the defect it prevents is a sequence.** ComfyUI
announces a drop as `reconnecting` *and* as a null `status` payload and emits
several across one restart, so the first opens the outage and later ones must not
move its start — re-stamping reads a 26s restart as a 3s blip and withholds the
recovery. It measures on the monotonic clock, holds `null` (not `0`) as its
no-outage sentinel because `performance.now()` legitimately reads ~0 near page
load, and reads **live** while an outage is open so the answer does not depend on
which of the two `reconnected` listeners happens to run first.

## The refutation

Not asserted — run. **Thirteen mutations, thirteen caught.**

| mutation | caught by |
|---|---|
| delete the re-advertise call (`rehello` → no-op) | e2e + wiring |
| swap it for `connectAgent()` (dead on an open socket) | wiring |
| delete the `reconnecting` down-stamp | wiring |
| delete the null-`status` down-stamp | wiring |
| delete the `reconnected` close | wiring |
| delete `landedHelloCount++` | wiring |
| wall clock instead of monotonic | wiring |
| move the branch after the resume gate's early return | wiring |
| invert the bridge-up guard | behaviour |
| drop the blip threshold | behaviour |
| drop the landed-hello suppression | behaviour |
| drop the one-shot claim | behaviour |
| tracker re-stamps on every down signal (#1145 shape) | tracker sequence |
| falsy-0 sentinel instead of `=== null` (#1138 shape) | tracker sequence |

**One survived at first and is worth naming.** The wiring assertion for the
re-advertise was a presence check on `client?.rehello?.();` — and this file holds
three other `rehello` call sites (#607, #310, #508), so deleting *this* one left
the assertion green. That is "wiring tests must check PATHS" exactly. It is now
positional: the call must appear inside the #654 branch, and `connectAgent(` must
not.

## Browser verification

`npx playwright test --workers=1` from this worktree — **result reported below.**

The new e2e (`bridge-reregisters-after-restart.spec.ts`) drives ComfyUI's own
`reconnecting` → `reconnected` api events under a **live** MockBridge and
requires:

- a fresh hello, on the **same** socket (`connectionsOpened` unchanged — no drop,
  no redial, no page reload), carrying a real route;
- a `graph_outline` routing through afterwards — announcing the registration is
  not the same as being usable, which is what `Connected: none` was about;
- **silence** for a sub-second blip (the control half — a fix that re-greeted on
  those would be #1138 wearing this issue's clothes);
- **silence** for a repeated `reconnected`.

Two harness facts, both found by running it:

1. The api singleton is `window.comfyAPI?.api?.api`, not `app.api`. Dispatching
   on the latter reaches an object the panel never listened on, and the control
   half would then pass for the wrong reason. The helper now asserts the dispatch
   was accepted.
2. This spec did not route the worktree's source, so it was measuring the dev
   ComfyUI's installed checkout. Run once in that state, the new test produced
   **no re-advertise at all** — this issue reproducing in the browser against the
   installed build. `routeWorktreeSource` is now applied in the new test, with
   that finding recorded beside it.

`MockBridge` gains a `connectionsOpened` counter: the status pill cannot tell a
re-advertise on a live socket from a socket that dropped and dialled back, and
only the first is what this scenario produces.

Unit: 5045/5047. The single failure is the known `#671 verifyInstalled` wall-clock
flake under full-suite contention — 125/125 in isolation, and it touches nothing
in this change.

- **Did not widen `shouldRehelloAfterCommand`.** Its narrowness is pinned on
  purpose, and a hello at `comfy_reboot` reply time reaches a server on its way
  down.
- **Did not touch #1096** ("bridge survived, tab route unrestored"), whose
  trigger is a plain socket drop rather than a restart, nor mcp#1670/#1671 or
  panel#1262.
- **Did not change the orchestrator side.** `graph_tools_ready` and `ready` still
  fail closed on an unproven reconnect, which is right; this makes the proof
  arrive instead of weakening what counts as proof.
- **Did not verify the #585 resume interaction end-to-end.** The re-advertise's
  `ready` ack reaches `handleRebootResumeAck`, which is the post-restart resume
  the panel's own `comfy_reboot` message promises and which is currently
  unreachable in this deployment — I believe that is a second thing this fixes,
  but I did not drive a real agent turn across a restart to confirm it, so I am
  claiming the route registration only. Reviewers should look hardest there.

---

Codex was unavailable (account exhausted until 2026-08-19), so this was **not**
self-gated; the orchestrator owns gating and merging.

### DIFF
--- a/browser_tests/bridge-reregisters-after-restart.spec.ts
+++ b/browser_tests/bridge-reregisters-after-restart.spec.ts
@@ -164,8 +164,6 @@ test('a ComfyUI restart under a LIVE bridge re-advertises the tab route', async 
   })
   const socketsAtStart = mockBridge.connectionsOpened
 
-  /** ComfyUI's backend socket goes away for `downMs` and comes back. The BRIDGE
-   *  is deliberately untouched — that is the whole point of this shape. */
   const fireComfyApiEvent = async (name: string) => {
     // The SAME resolution the panel itself uses to find the api singleton
     // (`window.comfyAPI?.api?.api || window.api`). Reaching for `app.api`
@@ -183,6 +181,8 @@ test('a ComfyUI restart under a LIVE bridge re-advertises the tab route', async 
     expect(delivered, `ComfyUI's api singleton must accept the ${name} event`).toBe(true)
   }
 
+  /** ComfyUI's backend socket goes away for `downMs` and comes back. The BRIDGE
+   *  is deliberately untouched — that is the whole point of this shape. */
   const bounceComfyBackend = async (downMs: number) => {
     await fireComfyApiEvent('reconnecting')
     await page.waitForTimeout(downMs)
--- a/browser_tests/bridge-reregisters-after-restart.spec.ts
+++ b/browser_tests/bridge-reregisters-after-restart.spec.ts
@@ -31,6 +31,7 @@
  */
 import { test, expect, deleteSavedWorkflow } from './fixtures/panelTest'
 import { MockBridge } from './fixtures/MockBridge'
+import { routeWorktreeSource } from './fixtures/worktreeSource'
 
 test.setTimeout(180_000)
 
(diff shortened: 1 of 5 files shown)


artokun:
**Correction to the coverage claim in this PR, recorded before the gate decision.**

The full Playwright `--workers=1` suite did **not** complete for this branch. Two runs were abandoned: the first was invalidated when the owner's ComfyUI on `:8188` died mid-run, and the second was stopped because it had been re-pointed at `http://localhost:8194`, an instance this work did not start and should not have borrowed. Any blanket "browser suite green" reading of this PR is therefore unsupported.

What *is* evidenced, and is not nothing:

- the targeted spec passes **2/2**, and fails with the fix deleted;
- a genuine browser reproduction — run against the installed build without `routeWorktreeSource`, the new test yields **no re-advertise at all**;
- **13 mutations, 13 caught.** Worth naming: one survived on the first attempt. The rehello wiring assertion was a *presence* check, and this file has three other `client?.rehello?.()` call sites, so deleting the new one left the suite green. It is now positional. That is this project's recurring "an install line is invisible to tests" failure, caught rather than shipped.

The full-suite gate is outstanding and the merge decision should be made with that known, not around it.

Unrelated to the fix, one piece of test litter is likely left in the owner's workflow library from a run whose cleanup could not execute after `:8188` vanished: `Untitled 2026-08-18 21-30-56.json`.

artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**. Verdict: **SHIP**, head `7c49653`, merged with the browser gate knowingly outstanding.

- **The audit gap this closes is genuinely still open.** The 2026-08-18T00:33Z comment named "bridge survived, tab route unrestored" as the only remaining row and routed it to #1096 / draft PR #1134. #1134 is CLOSED, and #1096 is a different trigger (a plain socket drop) which this fix declines by construction. It does diverge from the audit's orchestrator-side recommendation by fixing it panel-side.
- **The load-bearing premise was verified end-to-end rather than taken as given:** `externalOrchestratorMode()` is an unconditional `return true` (`:4596-4604`, six callers); `awaitPostRestartReachable` requires `generation > before.generation && tabSessionId === before.tabSessionId` (`src/orchestrator/panel-tools.ts:7189`); and `src/services/ui-bridge.ts:109` documents that every hello gets a new value even on the same socket/tab id — so a rehello on the surviving socket does flip it.
- **Identity handled correctly (taxonomy class 2), with no `tabId` comparison anywhere.** `landedHelloCount` is captured at `noteDown`, BEFORE the outage, and read at `reconnected`; `outageSeq` is stamped at outage open and `outageMs()` reads live while open, so both `reconnected` listener orders agree. The #1327 ordering trap is defended, not assumed.
- **The positional assertion holds.** Deleting `client?.rehello?.();` (`:31771`) fails the wiring test; swapping it for `connectAgent()` also fails. The asserted slice is ~10 lines and cannot be satisfied by the file's three other rehello sites — which is the whole point, since the original presence-check version stayed green when the new call was deleted.
- **11 mutations run, 11 caught.** Including inverting the bridge-up guard (4 fails), dropping the threshold, dropping the landed-hello suppression, dropping the one-shot, deleting `landedHelloCount++`, deleting `noteUp()`, wall clock, tracker re-stamp, and the falsy-0 sentinel.
- **Two residuals, both disclosed rather than papered over.** (1) A restart shorter than 6s is indistinguishable from a blip and will not recover — safe direction, leaves today's behaviour; an exact signal (`ssGet(REBOOT_KEY)`, read 55 lines below at `:31800`) exists but only covers panel-initiated restarts, which is why the duration heuristic is defensible. (2) **#654 may not fully close on the reporter's timings**: their recoveries were ~26-28s while `reconnectWaitTiming()` budgets 20s (`panel-tools.ts:1115`), so the restart reply's `panel_tab_reconnected` will still read false on that rig without raising `COMFYUI_PANEL_RECONNECT_WAIT_S`. This fixes the WEDGE — later graph calls route again without a manual reload — which is what the PR claims.
- **Browser gate outstanding, and the PR body overstates it.** The line promising a `--workers=1` result is unsupported and should be read as struck: two runs were abandoned (one invalidated when :8188 died mid-run, one stopped because it had been re-pointed at an instance the author did not start), and the gate could not run it either for the same reason. What stands is the targeted spec at 2/2 failing with the fix deleted, plus a browser reproduction where the installed build yields no re-advertise at all.
- No overlap with #1341/#1343/#1344/#1345 — same file, unrelated regions, no shared hunks.

PR #1349, opened 2026-08-19
fix(1192): the FIFTH wait in graph_add_node draws from the command budget too

artokun:
**Stacks on #1342** (`fix/1192-add-node-command-budget`), which is the substance of the #1192
fix and is correct. The base branch here is that PR, so the diff below is only my delta: the
one call site its budget did not reach, plus what I measured reviewing it.

## The defect

#1342 threads the command budget through **four** of `graph_add_node`'s waits — the baseline
seed, the single-type fetch, the whole-schema fetch, and the widget-registration wait — and
bounds the resolver's join on an in-flight refresh.

`graph_add_node` has **five**. The fifth is #1242's schema-drift recovery, at
`web/js/comfyui-mcp-panel.js:11786` on that branch:

```js
```

No `joinMs`. The coalescer's FORCED branch awaits any run **already in flight** before it queues
its own (`web/js/lib/refresh-coalesce.js`, the `if (current)` block), so unbounded this is
precisely the term #1192 is about — a wait on a run started under someone else's deadline —
arriving through the one `await` inside this command that named no bound. It sits *after* the
resolver, so it is reached with the window already partly spent, and on the reported scenario
(a ComfyUI restart, whose reconnect refresh is still running) the add parks there until the
30,000 ms relay gives up and the user gets the bare `did not reply to "graph_add_node"` that
this whole change exists to replace.

A source-level count of `budget.` call sites cannot see this: the budget was threaded through
four awaits and the fifth simply was not one of them.

For the record, since "a bigger constant is not a fix" is the obvious worry with any change of
this shape: **#1342 raises no existing timeout.** `git diff origin/main...` over `web/js` adds
exactly two numeric constants — `ADD_NODE_COMMAND_BUDGET_MS = 25000` (a CAP every step draws
from, not an enlargement) and `WIDEN_SOCKET_PROOF_DIVISOR = 2` (naming an existing divisor).
`NODE_DEFS_FETCH_TIMEOUT_MS` (10000), `NODE_DEFS_RUN_BUDGET_MS` (9000),
`OBJECT_INFO_SEED_WAIT_MS` (8000) and `CUSTOM_WIDGET_REGISTRATION_TIMEOUT_MS` (5000) all keep
their values. This delta adds no constant at all.

## The fix

Same allowance as the resolver's join, and the same reserve held back, so a drift recovery
cannot eat the window the widget-registration wait still needs. An abandoned recovery gets its
**own** reason rather than falling into the generic branch: a Symbol is not `typeof "object"`,
so it would have been reported as reason `"unknown"` — i.e. as a refresh that FAILED, which
sends the caller to the wrong recovery. Nothing failed; the refresh is still running and is
still registering the schema the add wants, so the reason says retry.

Healthy adds are unchanged: `joinMs` is only consulted inside the coalescer's `if (current)`
branch, so with no refresh in flight this call takes the path it always did.

The second commit updates #1242's two existing pins, which pinned the OLD literal
(`refreshComfyNodeDefs(undefined, { force: true })` — one as a source substring, one as
`deepEqual(opts, { force: true })`). **Updated, not loosened:** each now pins *more* than it
did — the source pin matches the call's shape *and* requires the budget-derived `joinMs`; the
options pin asserts `force` is set, that `joinMs` is finite and positive, and that nothing else
is passed. Verified in both directions: remove the bound from the panel again and both files go
red.

## Refutation (run, not reasoned)

Removed the `joinMs` again and re-ran: the new test does not merely fail — **it hangs, and
`graph_add_node` never answers at all.** That is the reported failure, reproduced at the call
site the budget missed. With the bound in place the add gives up at **279.8 ms** of a 400 ms
budget with a 120 ms reserve and refuses in words about the drift.

I also ran four mutations against #1342's own work. All killed:

| mutation | result |
|---|---|
| delete the `withTimeout` wiring on the coalescer (a one-line install) | ✖ red — the call site is pinned, not just the helper |
| restore the unbounded resolver join | **hang** — the add never answers |
| drop `budget.bounded(...)` from the widget-registration wait | ✖ red |
| remove `bounded()`'s 1 ms floor (the #1188 "0 means NO BOUND" trap) | **hang** |

## Measured

- The relay window #1342 cites is **correct, and I verified it independently**:
  `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS = 30_000` at comfyui-mcp
  `src/orchestrator/panel-tools.ts:821`, applied to `graph_add_node` at `:9650`. The issue's
  20,000 ms ui-bridge default is *not* the binding number for this command.
- **`joinMs` bounds the JOIN only, never the run that follows it.** Driven against the real
  coalescer: join bound 280 ms, in-flight run 250 ms (settles inside the bound), the caller's
  own run 2,000 ms → `refresh()` resolved in **2,251 ms — 8.0× its own join bound**.
- The drift recovery's live bound, read off the shipped constants by #1242's own harness:
  **`joinMs: 19999.3`** — i.e. 25,000 − 5,000, as designed.

## What still adds up (stated, not fixed here)

With `ADD_NODE_COMMAND_BUDGET_MS = 25_000` and `ADD_NODE_POST_REFRESH_RESERVE_MS = 5_000`, a
join is allowed to end as late as **20,000 ms** into the command. The caller's own run that
follows draws nothing from the budget: `NODE_DEFS_RUN_BUDGET_MS = 9_000` bounds only the
*waiting* inside a run, and `registerNodesFromDefs` is deliberately unbounded (3,972 ms
measured, per #1342's own comment). So a worst case of roughly **33,000 ms** survives on the
path where the join succeeds near its bound — still past the 30,000 ms window, though far
better than the ~41,000 ms it replaces, and every other path composes.

Closing that last term means bounding a *run* from its caller, which is exactly the subject of
the separate open issue #1193 (`refreshComboInNodes` alone at ~4.8 s of a 9 s run budget). I
deliberately did not touch it, and this PR does not resolve it.

## Verification

**Browser suite — `npx playwright test --workers=1`**, against a real ComfyUI 0.33.1 running on
its own port and base-directory with this branch junctioned into its `custom_nodes`, so the
module the browser loads is the changed one and not the shared checkout's (confirmed by
fetching the served bundle and finding this commit's code in it):

| tree | result |
|---|---|
| this branch (`a11eb0fa`) | **53 passed / 20 failed**, 10.6 min |
| #1342's head (`fc0e4b09`) | 52 passed / 21 failed, 10.1 min |
| `origin/main`, same 21 spec files, its own instance | **21 failed — the same tests, name for name** |

The e2e delta attributable to this change is **zero**: the 20 failures here are a strict subset
of the 21 that `origin/main` fails on this rig, and they sit entirely in the chat /
conversation-persistence / MockBridge specs. Neither spec that exercises `add_node`
(`auto-layout`, `blank-canvas-not-wedged`) is among them.

This issue is about timing, so plainly: the machine was loaded during these runs and the
wall-clock numbers are noisy. It is the branch-vs-`origin/main` comparison, not the absolute
count, that carries the claim.

**CI does not run on this PR, and that is structural, not a failure.**
`.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`, and this PR's base is
#1342's branch. It will fire automatically the moment #1342 merges and GitHub retargets this to
`main`. I ran that job's own steps locally instead:

- `node --check` over `web/js` — **167 files, 0 errors** (the CI step verbatim)
- `npm run typecheck` — clean
- `npm run test:unit` — see below; includes the tool-vocabulary, panel-scope and i18n gates
- Python steps are untouched by this diff (JS and test files only)

`npm run test:unit` — **5,061 tests, 5,059 pass, 1 fail, 1 todo.** The single failure is the
known #671 `verifyInstalled` wall-clock flake in `manager-install.test.mjs` (26.1 s), untouched
by this diff and **125/125 green when that file is run alone**.

Gate: **not run.** `codex exec` is account-exhausted and I was instructed not to run a gate
script — the orchestrator gates and merges.

This addresses the underlying cause of #1192 together with #1342. Neither closing the issue nor
merging is a decision of mine.

### DIFF
--- a/browser_tests/unit/add-node-command-budget.test.mjs
+++ b/browser_tests/unit/add-node-command-budget.test.mjs
@@ -540,3 +540,71 @@ test("#1192: every step stalling at once still replies inside the command budget
   gate.resolve();
   await built.inFlightStarted?.catch(() => {});
 });
+
+// ---------------------------------------------------------------------------
+// 4. THE FIFTH WAIT: #1242's drift recovery is inside this command too.
+//
+// `graph_add_node` has FIVE waits, not four. The fifth is #1242's `force: true` refresh,
+// run when the registered schema has drifted from the backend's. The coalescer's forced
+// branch AWAITS ANY IN-FLIGHT RUN before queueing its own, so unbounded it is the same
+// term this issue is about — reached later, with the window already partly spent.
+//
+// Counting `budget.` call sites in the source cannot see this: the budget was threaded
+// through four awaits and the fifth simply was not one of them. Only a test that DRIVES
+// this path finds it, which is why this one holds a real in-flight run open and asserts
+// the add still answers at all.
+// ---------------------------------------------------------------------------
+
+test("#1192: the #1242 drift-recovery refresh is bounded by the command budget too", async () => {
+  // The class is ALREADY registered, so the resolver returns on its fast path and never
+  // calls refresh — the only refresh this add performs is #1242's. `driftedRequiredInput-
+  // Names` is forced non-empty so that recovery actually runs; that is the path under test.
+  const comfy = makeComfy();
+  comfy.app.registerNodesFromDefs({ ExistingNode: backendObjectInfo().ExistingNode });
+  const gate = deferred(); // a reconnect refresh that never lands
+  const built = realGraphAddNode({
+    comfy,
+    holdInFlight: gate.promise,
+    budgetMs: 400,
+    reserveMs: 120,
+    overrides: { driftedRequiredInputNames: () => ["seed"] },
+  });
+
+  const started = Date.now();
+  // UNBOUNDED, this call parks on the held-open run and the add never answers at all — the
+  // reported failure, reproduced. Bounded, it gives up and refuses in words about the drift.
+  await assert.rejects(
+    () => built.graph_add_node({ class_type: "ExistingNode" }),
+    (err) => {
+      assert.match(err.message, /required input/i, "the refusal is the drift one, worded");
+      return true;
+    },
+  );
+  const elapsed = Date.now() - started;
+  assert.ok(
+    elapsed < 3000,
+    `the drift recovery took ${elapsed}ms — it must give up at the command's bound, not ` +
+      "wait out a run started by something else",
+  );
+  gate.resolve();
+});
+
+test("#1192: an abandoned drift recovery is reported as a RETRY, not as 'unknown'", async () => {
+  // The generic branch turns any non-true verdict into reason "unknown". A Symbol lands
+  // there unless it is handled FIRST, and "unknown" is the shape of answer this repo keeps
+  // getting burned by: it cannot tell a refresh that FAILED from one this command simply
+  // stopped waiting for. Only the second is cleared by retrying.
+  const addBody = addNodeMatch[0];
+  assert.match(
+    addBody,
+    /force: true,\s*\n\s*joinMs: budget\.remaining\(\) - ADD_NODE_POST_REFRESH_RESERVE_MS,/,
+    "the #1242 drift refresh must draw its join bound from the command budget",
+  );
+  const at = addBody.indexOf("if (verdict === REFRESH_JOIN_ABANDONED)");
+  assert.ok(at > 0, "an abandoned drift recovery must be distinguished from a failed one");
+  assert.match(
+    addBody.slice(at, at + 700),
+    /retry/i,
+    "…and its reason must name the remedy that actually works",
+  );
+});
--- a/browser_tests/unit/add-node-schema-auto-refresh.test.mjs
+++ b/browser_tests/unit/add-node-schema-auto-refresh.test.mjs
@@ -279,7 +279,20 @@ test("#1242: the reporter's add succeeds in one call — the panel refreshes fir
   assert.equal(comfy.graph._nodes.length, 1);
   assert.equal(refreshCalls.length, 1, "exactly one refresh — the recovery, not a retry storm");
   assert.equal(refreshCalls[0].defs, undefined, "the refresh is the whole-schema forced one");
-  assert.deepEqual(refreshCalls[0].opts, { force: true });
+  assert.equal(refreshCalls[0].opts.force, true, "the recovery is the forced whole-schema one");
+  // #1192 — and it is BOUNDED now. This recovery awaits any refresh already in flight before
+  // queueing its own; unbounded that wait is the composition defect #1192 is about, arriving
+  // at the one await inside this command that named no bound. Asserting the bound EXISTS is
+  // stronger than the literal `{ force: true }` this replaces, which could not see it.
+  assert.ok(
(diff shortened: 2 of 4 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had seen neither author's reasoning**, as a stacked pair with #1342. Verdict: **SHIP**.

**The verdict was issued against `a11eb0fa`; this merges `33cd702e` after a rebase.** Rather than assume the rebase was inert, the delta was compared by `git patch-id --stable`: `1f30dd2c…` before and after, i.e. **byte-identical** (4 files, 124 insertions / 4 deletions). The rebase used `git rebase --onto origin/main fc0e4b09`, because a plain rebase wrongly re-applied #1342's commits — it landed as a merge commit.

**The core refutation is a hang, not a red test, and it re-reproduced after the rebase.** `graph_add_node` has five waits; #1342 budgets four. The fifth is #1242's drift recovery, which called `refreshComfyNodeDefs(undefined, { force: true })` with no `joinMs`, and the coalescer's forced branch first awaits any run already in flight. Deleting the `joinMs:` line at `:11984` makes `#1192: the #1242 drift-recovery refresh is bounded by the command budget too` run **150,000 ms** — killed only by the harness cap — against a whole-file baseline of **2,992 ms**. A second test fails fast because `REFRESH_JOIN_ABANDONED` never arrives.

**The merge order was structurally forced, not stylistic:** `REFRESH_JOIN_ABANDONED`, `ADD_NODE_POST_REFRESH_RESERVE_MS` and `makeCommandBudget` had zero occurrences on `main` and `web/js/lib/command-budget.js` did not exist there, so this could not compile before #1342.

**Independently verified after #1342 merged: no pin was lost in that conflict resolution.** The conflict was in the pins file itself, which is exactly where a guard gets silently dropped. On current `main`, each mutation proven applied by content-anchored replacement with a throw-on-miss:
- deleting the `withTimeout,` install (`comfyui-mcp-panel.js:1710`) → RED, killed by the source-level **call-site** pin in `single-node-def.test.mjs`
- removing `bounded()`'s 1 ms floor (`web/js/lib/command-budget.js:109`) → RED, 3 kills, one of which *hung* and died on the 45 s cap
- reverting #1346's `Math.max(nodeDefsBudgetLeft(runDeadline), NODE_DEFS_COMBO_FLOOR_MS)` (`:1507`) → RED, 2 behavioural kills plus #1346's own source pin

**CI note worth recording:** this PR's checks had never run, because `ci.yml` uses the default `pull_request` types, which exclude `edited` — retargeting the base alone does not trigger it. A close-and-reopen produced the `reopened` event and CI passed (run `32224707548`, 2m17s) with commits untouched.

**#1192 stays OPEN.** A disclosed residual survives both PRs: `joinMs` bounds the *join*, never the run after it, so a join ending near its 20,000 ms bound plus a ~13,000 ms own run still exceeds the 30,000 ms relay window. That is now tracked as #1351 — note the authors had pointed at #1193, which is already closed and whose fix redistributes budget *inside* a run rather than bounding one.

Issue #1339, opened 2026-08-19
panel_set_widget transiently rejects matching re-derived workflow identity after download event


artokun:
### What happened / root cause
After a background model download completed, panel_set_widget on the still-active live workflow failed once because the workflow had "no trusted identity", even though the error itself said the live canvas identity and session fence had already been re-derived to the same workflow UUID. The identical retry immediately succeeded. This appears to be a transient workflow-fence refresh race after canvas mutations/background events.

### Steps to reproduce
1. Load a workflow onto the live canvas.
2. Add/connect/move/group nodes through panel tools.
3. Start and complete a background model download.
4. Call panel_set_widget on an existing node immediately after completion.
5. First call may fail with the identity/fence error; exact retry succeeds.

### Exact error (scrubbed)
Error: "graph_set_widget" cannot be safely targeted to the active workflow: this workflow has no trusted identity for the panel to fence the command against.
CHECKED: the live canvas DOES carry an identity (...) and this session's fence has been re-derived onto it. RETRY THIS EXACT CALL ONCE — nothing was applied.

### Fix
Applied locally: no. Upstream-only: yes.
Temporary workaround verified: retry the exact idempotent panel_set_widget call once, as instructed by the error. It succeeded without rebinding or reloading.
Suggested upstream change: when the live canvas identity matches the newly re-derived session fence within the same request, continue the mutation instead of returning a retry-only failure, or refresh the fence before the safety decision.

### Environment
Windows 11
ComfyUI 0.33.1
ComfyUI frontend 1.48.7
GPU NVIDIA RTX 5080 16 GB
comfyui-mcp 0.51.12
panel 0.14.24

Related (same root cause): #1014

artokun:
Claimed by the backlog-drain session. Picking up abandoned in-progress work found in the worktree `panel-1339` (branch `fix/1339-fixed-seed-nodes-contradiction`, three modified files including a test, never committed). Driving it to completion; if no PR appears within ~8h, consider the claim lapsed.

artokun:
## Two findings. Neither of them is in this repo.

### 1. The abandoned work in `panel-1339` is a different bug

The worktree I picked up (branch `fix/1339-fixed-seed-nodes-contradiction`, three modified
files, never committed) does not touch workflow identity at all. Its `1339` is
`artokun/comfyui-mcp#1339` — the *MCP* repo's rgthree-seed issue, fixed here in #1082 and
[...]

PR #1350, opened 2026-08-19
fix(mcp#1339 r2): the fixed-seed ARRAY and its SENTENCE come from one predicate

artokun:
## What this is

Recovery of abandoned uncommitted work, plus the finding that it is **not** the issue it was filed under.

I was dispatched to drive panel issue 1339 (`panel_set_widget` transiently rejects a matching
re-derived workflow identity after a download event) and handed an abandoned worktree —
branch `fix/1339-fixed-seed-nodes-contradiction`, three modified files, never committed,
0 commits ahead of its base.

**The inherited diff addresses a different bug than the panel issue it was branched for.**
The `1339` in that branch name is `artokun/comfyui-mcp#1339` — the *MCP* repo's rgthree-seed
issue, fixed here in PR #1082 and referenced throughout `web/js/lib/scoped-batch-seed.js`.
The panel repo's own issue 1339 is unrelated. Two repos, one issue number.

This PR ships the inherited work on its own merits. The panel-issue-1339 finding is in a
comment on that issue.

## The mechanism this fixes

A batch run against an rgthree Seed node computed two answers about the same question from
two different predicates:

- the prose warning (`rgthreeFixedSeedNote`) selected on `varies === false`
- the structured `fixed_seed_nodes` field, assigned at the call site in
  `comfyui-mcp-panel.js`, selected on `armed === false`

Those agree for a concrete seed value and **disagree for an armed node whose
`randomMin`/`randomMax` (or `step`) admit exactly one value** — `armed: true, varies: false`.
That is the case the note's own text calls the confusing one, because the user *did* press
the button and the seed still repeats.

Because the field is only attached when the note is non-empty, the disagreement is not
hypothetical: a run against such a node returned

```
fixed_seed_note: "ALL 10 ITEMS OF THIS BATCH WILL USE THE SAME SEED from node 649"
fixed_seed_nodes: []
```

A program branches on the array. Only a human reads the sentence. So the one case the
warning exists for is the one case a caller reading the structured field concluded was clean.

## The fix

`varies === false` is the correct predicate, so it now lives once in
`repeatingRgthreeSeeds()` in `web/js/lib/scoped-batch-seed.js`, and both consumers call it.
`armed` stays on each returned entry alongside `degenerate_range`, so a reader can still tell
the two ways a seed repeats apart — what is removed is the ability for the two answers to
disagree.

Deliberately **not** shared with `rgthreeQueueTimeSeedInput` (mcp#1124): that one gates on
`armed` because a degenerate-range node still *substitutes* at queue time and so is still
volatile for the drift hash. Same nodes, opposite question.

## What I did with the inherited work, and why

**Adopted unchanged.** Judged on its merits:

- It applies cleanly to current `main` (v0.15.3), 40+ commits past the base it was written on,
  and the defect it names is still live there — the call site on `main` still reads
  `rgthreeSeeds.filter((s) => s && s.armed === false)`.
- Run as it stood in the abandoned worktree, its test file passed **33/33**. It was a
  completed fix with its test, not a test-shaped statement of a bug with nothing behind it.
- It carries a source guard asserting on **every** `fixed_seed_nodes` assignment in the
  monolith, not just the first, and on the import — the two things a helper-level test is
  blind to for a one-line call-site change.

The three-file diff was copied out to a patch before anything was touched; nothing in the
abandoned worktree was checked out, stashed, or reverted, and no work was done inside it.

## Refutation

Green before and after proves nothing, so I deleted the fix twice and watched it fail:

| mutation | result |
|---|---|
| call site reverted to `rgthreeSeeds.filter((s) => s && s.armed === false)` | **1 failed** — `#1339 r2 source guard: the CALL SITE uses the shared predicate, not its own filter` |
| `repeatingRgthreeSeeds` predicate reverted to `armed === false` | **2 failed** — `an ARMED node with a degenerate random range still repeats`, and `an ARMED, degenerate node is in the ARRAY, not only in the prose` |
| restored | 33/33 pass |

Both halves are pinned, and the call-site half is pinned by something that can actually see it.

## Verification

- `npm run test:unit` — **5033/5035 pass, 1 fail**. The single failure is
  `manager-install.test.mjs:836` (`expected 'installed', actual 'unverified'`), the known
  wall-clock flake that fires under machine load; eleven agents are running on this box.
  Re-run alone: **125/125 pass**. It does not touch this change's files.
- Panel scope gate (`check-panel-scope.mjs`), i18n check, i18n render check and the tool
  vocabulary gate all pass.
- Playwright browser suite, `--workers=1`, from this worktree — result appended below.

## Gating

Not gated by me. `codex exec` is account-exhausted until 2026-08-19 20:43 and the substitute
`claude-gate.mjs` is blocked on the weekly account limit, so **no gate has been run on this
PR** — I am not substituting my own review for one. The orchestrator gates and merges.

### DIFF
--- a/browser_tests/unit/scoped-batch-seed.test.mjs
+++ b/browser_tests/unit/scoped-batch-seed.test.mjs
@@ -21,6 +21,7 @@ import {
   scopedBatchSeedNote,
   findRgthreeSeedNodes,
   rgthreeFixedSeedNote,
+  repeatingRgthreeSeeds,
   rgthreeQueueTimeSeedInput,
 } from "../../web/js/lib/scoped-batch-seed.js";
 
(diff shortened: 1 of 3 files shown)


artokun:
### Playwright: NO VERDICT — the browser gate on this PR is outstanding

The PR body says the `--workers=1` result would be appended here. It cannot be, and I would
rather say so than leave the sentence implying coverage.

The run was launched from this worktree (`73 tests in 31 files`, `--workers=1`) and was killed
at a 10-minute harness cap before it finished. Its output had been piped through `tail`, so:

- the pipeline exited **0 — that is `tail`'s status, not Playwright's**, and cannot distinguish
  a pass from a fail or from a kill;
- the captured tail is 62 lines, all bare test-name lines with the `───` padding Playwright uses
  for its *did not run* listing. There is **no `N passed` summary, no `✓`/`✘`, no failure block**.

So: **the browser suite has not been run to completion against this change.** I have not re-run it
(machine is under load from eleven agents, and two runs have already been invalidated by that).

What IS verified from this worktree:

- `npm run test:unit` — 5033/5035 pass. The single failure is `manager-install.test.mjs:836`
  (`expected 'installed', actual 'unverified'`), the known wall-clock flake under load; that file
  alone passes **125/125**. It shares no code with this change.
- `check-panel-scope.mjs` (the compiler-based gate) — OK, every name in the monolith resolves,
  which is what confirms the new `repeatingRgthreeSeeds` import is actually wired.
- i18n check, i18n render check, tool-vocabulary gate — all pass.
- The two-mutation refutation described in the PR body.

The change is a pure data-shape fix in the run-result payload (`fixed_seed_nodes`), with no DOM or
sidebar rendering path, so I would expect the browser suite to be indifferent to it — but that is
an expectation, not an observation, and it should not be treated as one.

artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**. Verdict: **SHIP**, head `aa3e205`.

**Provenance verified first, because this PR came out of a recovery job.** The `1339` in the originating branch name is `artokun/comfyui-mcp#1339` (the MCP repo's rgthree-seed issue), **not** panel#1339 — two repos, one issue number. That MCP issue is CLOSED/COMPLETED, so the gate's first question was whether anything remains to fix.

**It does, and the defect was reproduced against `origin/main`'s own module.** The owner's closure covered detector *blindness* (panel #1082, v0.14.3), not the array/prose split. An armed node with `randomMin === randomMax === 5` — and independently `min=0, max=5, step=100` — scans as `armed: true, varies: false`, so `rgthreeFixedSeedNote` names node 649/650 in prose while `comfyui-mcp-panel.js:14135`'s `filter(s => s.armed === false)` returns `[]`. A live contradiction, and pointedly: the owner's own closing instruction on that issue tells the reporter to check `fixed_seed_nodes` — the exact field that answers "clean" for the exact case he calls confusing.

**The surviving predicate is the correct one, not merely the agreeing one.** Collapsing two predicates is only right if one was wrong. `varies === false` is right because the field is attached *only when the note is non-empty*, and the note's own gate is already `varies === false` — so the population gate and the membership predicate must be the same, or `[]`-beside-a-sentence is guaranteed. Since `armed === false` is a strict subset of `varies === false`, this is a pure **widening**: nothing previously reported is dropped, and `armed` plus `degenerate_range` remain on each entry. No code consumer of `fixed_seed_nodes` exists in either repo, so the widening breaks nothing downstream.

**Reachability:** the call site is inside `async graph_run(...)` (`:13615`), dispatched at `:20052` and `:32879`, with the orchestrator mapping `panel_run → graph_run` (`comfyui-mcp/src/orchestrator/panel-tools.ts:9076`, `:10442`).

**Four mutations, each confirmed applied with `git diff` before running** — deliberately via Python byte-replace rather than perl, because this repo's CRLF files silently swallow `$`-anchored substitutions and a mutation that never applied is indistinguishable from one that survived. Call site to `armed===false`: 1 fail. Predicate to `armed===false`: 2 fail. Predicate to `varies !== true`: 1 fail. Deleting the import: 1 fail. Restored: 33/33.

Unit suite 5035 tests, 1 failure = the known #671 `verifyInstalled` wall-clock flake (125/125 alone). i18n, i18n-render, tool-vocabulary and `check-panel-scope.mjs` all OK.

**One P3 finding, test-only, worth fixing later:** `browser_tests/unit/scoped-batch-seed.test.mjs:465` asserts that every node the note names is in the array "and vice versa" — but the prose truncates at 5 ("and N more") while the array is complete. With 7 repeating nodes the array holds 1-7 and the prose holds 1-5, so nodes 6 and 7 violate the asserted equality. Product behaviour (array ⊇ prose) is correct; the *test's* invariant is overstated and only passes because no fixture exceeds 5. The landmine is that someone adding a 6-node case sees red and "fixes" it by narrowing the array. It should be loosened to a superset assertion.

No browser coverage claimed and none required — the diff changes which entries populate a JSON field in a tool reply.

Issue #1284, opened 2026-08-16
Do not flag KJNodes virtual SetNode/GetNode as missing backend node types


artokun:
### What happened / root cause
panel_get_errors reports KJNodes' SetNode/GetNode virtual wiring nodes as missing_node_type because they are frontend-registered virtual nodes rather than backend /object_info nodes. This produces a misleading 12-node error report on a valid Ideogram workflow. panel_strip_workflow correctly resolves these buses into real links and returns a runnable graph, so the graph itself is valid.

### Steps to reproduce
1. Install/load the bundled ideogram-txt2img workflow, which uses KJNodes SetNode/GetNode buses.
2. Call panel_get_errors.
3. Observe missing_node_type errors for GetNode and SetNode.
4. Call panel_strip_workflow; it successfully flattens the buses and produces a valid 29-node graph with no SetNode/GetNode types.

### Exact error (scrubbed)
missing_node_types: ["GetNode", "SetNode"]
unchecked_nodes includes the SetNode/GetNode ids with reason "node type not found in /object_info".

### Fix
No local source patch applied. The precise fix is for panel_get_errors' node-type validation to recognize KJNodes' frontend virtual SetNode/GetNode registration (or defer to panel_strip_workflow's virtual-node resolver) instead of treating absence from backend /object_info as a missing custom node. Current workaround: use panel_strip_workflow for authoritative wiring validation.

### Environment
OS: darwin
ComfyUI: 0.33.1
ComfyUI frontend: 1.48.7
GPU: mps, 48 GB VRAM
Python: 3.13.12
PyTorch: 2.10.0
comfyui-mcp: 0.51.56
panel: 0.14.41

Related (same root cause): #475

artokun:
Drive-by note from the family working comfyui-mcp#1657 / #1648 — **not claiming this issue**, just leaving evidence so the next pass does not implement the prescribed fix as written, because half of it would introduce a bug.

**The `unchecked_nodes` half is addressed** by #1353. Those entries come from the live combo scan, where a frontend-only node can never resolve a class and so always landed in `unchecked_nodes` with `"node type not found in /object_info"`. Virtual nodes are now skipped before the lookup — nothing is claimed about them, because there was never a judgement available to withhold.

[...]

PR #1353, opened 2026-08-19
fix(1657,1648,1284): a frontend VIRTUAL node is not a missing one — the derivable signal, at three call sites

artokun:
A backlog audit grouped `artokun/comfyui-mcp#1400`, `artokun/comfyui-mcp#1648` and
`artokun/comfyui-mcp#1657` as one family across three call sites. Two of the three are
panel-side and are fixed here; the third stays where its owner parked it. The audit's
prescription — "make the call sites consult `FRONTEND_ONLY_NODE_TYPES`" — is refuted
below, with the case that refutes it.

## The shared mechanism

Three surfaces answer "is this node type real?" from ONE observation — the server's
`/object_info` does not list it — and that observation cannot tell apart:

* **A.** a node the FRONTEND registers as VIRTUAL, absent from `/object_info` **by
  design** (KJNodes' Get/Set bus, rgthree's Label and Fast Groups toggles, litegraph's
  own Note / MarkdownNote / Reroute / PrimitiveNode); and
* **B.** a node whose class this page never registered, so litegraph left a defless
  PLACEHOLDER.

Both are absent from `/object_info`. Only **B** is a fault. Collapsed, they report a
working 422-node canvas as a wall of missing node types.

## The derivable signal — not a list of names

`node.isVirtualNode === true`, read off the live node instance. Not a heuristic and not
this repo's invention: it is the exact flag ComfyUI's own serializer reads to decide what
never reaches the backend. From the shipped frontend 1.48.7 bundle:

```js
for (let e of i.values()) {
  if (e.isVirtualNode || e.mode === NEVER || e.mode === BYPASS) continue;
  ...
  a[String(e.id)] = { inputs: t, class_type: e.comfyClass, _meta: { title: e.title } }
}
```

and its `ExecutableNodeDTO` proxies it straight through
(`get isVirtualNode() { return this.node.isVirtualNode }`). So this flag is the only thing
standing between a virtual node and a `class_type: undefined` prompt entry — exactly the
property these surfaces need, asserted by the authority that acts on it.

Confirmed in each pack's own source on this machine's install, rather than inferred:

| source | sets it |
|---|---|
| `comfyui-kjnodes/web/js/setgetnodes.js` | `this.isVirtualNode = true` on `SetNode` and `GetNode` — their own comment: "purely frontend and does not impact the resulting prompt" |
| `rgthree-comfy/web/comfyui/base_node.js` | `RgthreeBaseVirtualNode`, extended by Label, Fast Groups Bypasser/Muter, Fast Bypasser/Muter, Node Collector, Reroute (rgthree) |
| the ComfyUI frontend itself | Note, MarkdownNote, PrimitiveNode, Reroute, `SubgraphNode` |

A pack nobody has heard of yet is covered on the same terms: a virtual node that does not
set this flag is serialized into the prompt and rejected by the server, so a pack cannot
both omit the flag and work.

## Refutation of the audit's prescription

**`FRONTEND_ONLY_NODE_TYPES` is two different sets in two repos, and the audit read the
wrong one.** The panel's (`web/js/lib/node-resolve.js`) has listed Label, Fast Groups
Bypasser/Muter, Fast Bypasser/Muter, Node Collector, Reroute (rgthree), SetNode and
GetNode since #475/#496. Only the orchestrator's
(`comfyui-mcp/src/services/workflow-converter.ts`) is limited to
Reroute/Note/MarkdownNote/PrimitiveNode + the Get/Set bus. So "rgthree Label and Fast
Groups are absent" holds for `check_runtime` only.

**And adopting that list at these call sites would have shipped a bug.** `panel#1284` and
`artokun/comfyui-mcp#1648` come from the SAME rig — darwin, ComfyUI 0.33.1, frontend
1.48.7, comfyui-mcp 0.51.56, panel 0.14.41 — and on it ComfyUI's OWN missing-nodes store
named `GetNode`/`SetNode`, because that tab had never loaded KJNodes' JS. Those nodes
really were dead placeholders. A name allowlist containing "GetNode"/"SetNode" would have
declared them fine and sent a run to the queue that could not execute.

The same predicate gives the RIGHT answer on both rigs, which is the whole argument for
it: on `#1657`'s rig the packs are loaded, the instances report `isVirtualNode`, and they
are dropped; on `#1284`'s rig they are placeholders, report nothing, and stay reported.

The audit's second clause — "`panel_run` and `panel_get_errors` never consult it" — is
literally true and, for `panel_run`, deliberately so: that pre-flight reads the SERIALIZED
prompt precisely so it does not need a name list (#1460). It is not changed here.

## Per-call-site changes

**1. `panel_get_errors`' live combo scan** (`live-combo-availability.js`) — the source of
the verbatim `"node type not found in /object_info"` reason both reporters quote. A
virtual node has no backend definition and therefore no combo the server could judge, so
it is now skipped BEFORE the per-class lookup. Nothing is claimed about it, which is
honest rather than lenient — there was never a judgement available to withhold — and it
saves a wasted round trip plus a slot in the 80-class cap.

**2. `collectMissingAssets`** (`comfyui-mcp-panel.js`) — the shared collector behind
`missing_node_types`. ComfyUI's `missingNodesError` store is a LOAD-TIME snapshot the
frontend never re-evaluates (`stale-placeholders.js` measured exactly that), so it keeps
naming a type after the page learned to resolve it. A type is dropped only when EVERY live
instance of it declares `isVirtualNode`. Filtered in the collector rather than in
`graph_get_errors`, so the tool, the turn-start validation banner and the stale-red-flag
adjudication cannot disagree — this family reached three call sites because each site
decided separately.

**3. `panel_run`'s pre-flight refusal** (`missing-node-preflight.js`) — **the refusal was
right and the remedy was wrong.** `#1648`'s reporter installed KJNodes, restarted ComfyUI,
and was told by this message to install the pack and restart ComfyUI. The message named
ONE cause for a state that has two, and named the one they had already ruled out. It now
reads this page's litegraph registry (DIAGNOSIS ONLY — the go/no-go is still
`unrunnableNodeIds(built)`, unchanged) and distinguishes:

* type not registered here → names both causes and prescribes **RELOAD the ComfyUI tab**,
  because a pack's frontend JS is fetched at page load and a server restart does not add
  its classes to an already-open tab. For a frontend-only type this is the only
  possibility, since nothing the server reports can confirm or deny it.
* type registered here → says so, and explicitly does NOT send the user to reload.
* registry unreadable → the message is **byte-identical to before**, pinned verbatim in a
  test. An unobserved registry must not become a claim about one.

## What stays refused / reported

* `panel_run` refuses exactly the same runs as before — only the text changed. Pinned by a
  wiring test asserting the gate is still the serialized prompt and that
  `registered_node_types` appears exactly once, inside the message construction.
* A type with ANY live placeholder instance → still reported (the unanimity rule).
* A type with NO live instance → still reported; a stale store entry is never cleared this
  way.
* A genuinely uninstalled pack → placeholders, no flag, still reported, run still refused.
* The combo scan's real `unavailable` findings are untouched (pinned).
* `check_runtime` in the orchestrator is not touched at all.

## Refutation (mutation runs)

Each fix was DELETED and the suite re-run. The first attempt reported all-green because
the perl patterns silently failed to match CRLF files — the harness now proves each
mutation landed (byte delta + a post-condition grep) before believing a verdict.

| mutation | result |
|---|---|
| delete the skip in `scanComboAvailability` | RED — killed by the CALL SITE 1 behaviour test **and** its wiring test |
| delete the filter in `collectMissingAssets` | RED — killed by both CALL SITE 2 wiring tests |
| stop passing the registry from `graph_run` | RED — killed by both CALL SITE 3 wiring tests |
| loosen `isVirtualNode === true` to a truthy check | RED |
| drop the unanimity rule | RED |
| make an unreadable registry assert "not registered" (`?? {}`) | RED |

20 new tests, split per call site — a helper-only suite here would have left two of three
unwired and stayed green.

## Verification

* `npm run test:unit` — 5084 tests, 5082 pass, 1 fail: the known `#671`
  `verifyInstalled` wall-clock flake under load. Re-run alone: `manager-install.test.mjs`
  125/125 green. Nothing in this change touches it.
* `npm run typecheck` clean; `npm run test:e2e:list` compiles and discovers 74 specs.
* **Browser e2e NOT run.** It requires a real ComfyUI on `localhost:8188`; that rig is the
  owner's Desktop and is down, and I was instructed not to touch it or borrow another
  instance. No spec in the 32-file suite exercises `graph_get_errors` or `graph_run`'s
  pre-flight, so the gate would have added no coverage for this change — but I did not run
  it and am not implying coverage I do not have.
* Not gated: `codex exec` is exhausted until 2026-08-19 and `claude-gate.mjs` is blocked on
  the weekly limit. The orchestrator gates and merges.

## Issue verdicts

* **`artokun/comfyui-mcp#1657`** — the underlying cause is addressed here: sites 1 and 2.
* **`artokun/comfyui-mcp#1648`** — the underlying cause is addressed here: site 3. Note
  the correction to the report — the refusal was correct on that rig; the misdiagnosis was
  the remedy.
* **`artokun/comfyui-mcp#1400`** — NOT addressed here and deliberately so. Its remaining
  half is `check_runtime`, which is headless and has no node instances to read, and the
  panel command that would carry this signal to it is parked under the freeze by the
  owner's own comment of 2026-08-14. Parking confirmed on the issue, with the signal now
  identified and proven so the parked work has a known shape.
* **`panel#1284`** — the underlying cause is addressed here, and its own `unchecked_nodes`
  complaint goes away. Its `missing_node_types: ["GetNode","SetNode"]` does NOT, and must
  not: on that rig those were placeholders. That issue's thread is where that correction
  belongs.
* **`panel#1296`** (`panel_add_node` rejects Fast Groups Bypasser) — NOT covered. Same
  family, different call site: `assertAddNodeResolvableRefreshing` requires live-registry
  membership, and on a tab that never loaded rgthree's JS there is none. Its refusal text
  has the same defect site 3 fixes here — it sends the user to install a pack they have.
* **`panel#1300`** (`panel_move_group` refuses groups containing Label) — NOT covered.
  Different mechanism: its pre-flight verifies a child WRITE landed by reading position
  back, which is a different collapsed pair and not a node-type question.

### DIFF
--- a/browser_tests/unit/frontend-virtual-nodes.test.mjs
+++ b/browser_tests/unit/frontend-virtual-nodes.test.mjs
@@ -254,9 +254,24 @@ test("#1648 CALL SITE 3: a REGISTERED type is not sent to reload the tab", () =>
   // page, so "reload" would send the user round the same loop from the other side.
   const msg = missingNodeRunRefusal([{ id: "7", type: "SomeDisplayNode", registered: true }]);
   assert.match(msg, /DOES have a node class registered for SomeDisplayNode/);
-  assert.match(msg, /reloading will not change it/);
+  assert.match(msg, /reloading\s+will not change those/);
   assert.doesNotMatch(msg, /RELOAD the ComfyUI tab/);
   assert.doesNotMatch(msg, /install the pack that provides these types/);
+  // Nothing was left unnamed, so no leftover clause.
+  assert.doesNotMatch(msg, /could not be named from the/);
+});
+
+test("#1648 CALL SITE 3: an offender the canvas could not name is never swept into the claim", () => {
+  // `registered: null` is a node whose type the graph could not read. The "registered here"
+  // sentence must stay scoped to what was actually looked at — otherwise it asserts
+  // "nothing is missing" off a node nobody examined.
+  const msg = missingNodeRunRefusal([
+    { id: "7", type: "SomeDisplayNode", registered: true },
+    { id: "8", type: null, registered: null },
+  ]);
+  assert.match(msg, /DOES have a node class registered for SomeDisplayNode/);
+  assert.doesNotMatch(msg, /nothing is missing here/);
+  assert.match(msg, /The other 1 node could not be named from the canvas, so nothing is established about it/);
 });
 
 test("#1648 CALL SITE 3: one unregistered offender in a mixed set still gets the reload remedy", () => {
--- a/browser_tests/unit/frontend-virtual-nodes.test.mjs
+++ b/browser_tests/unit/frontend-virtual-nodes.test.mjs
@@ -1,0 +1,300 @@
+// comfyui-mcp#1657 / #1648 / panel#1284 — "absent from /object_info" is not "missing".
+//
+// Three surfaces collapsed two states into one answer:
+//   A. a node the FRONTEND registers as virtual, which never reaches the backend and is
+//      absent from /object_info BY DESIGN; and
+//   B. a node whose class this page never registered, so litegraph left a placeholder.
+//
+// The discriminator is `node.isVirtualNode === true` — the flag ComfyUI's own serializer
+// reads to skip a node (frontend 1.48.7: `if (e.isVirtualNode || …) continue`), set in
+// their own source by KJNodes' SetNode/GetNode and by rgthree's RgthreeBaseVirtualNode
+// (Label, Fast Groups Bypasser/Muter, Fast Bypasser/Muter, Node Collector, Reroute).
+//
+// EVERY behavioural test below is paired with its case-B twin, because that is the pair a
+// name allowlist gets wrong: panel#1284 and comfyui-mcp#1648 come from the same rig, where
+// GetNode/SetNode were REAL placeholders (the tab had never loaded KJNodes' JS) and an
+// allowlist naming "GetNode"/"SetNode" would have declared a broken run fine.
+//
+// Each of the three call sites is asserted separately. A helper-only suite here is blind
+// by construction — it would leave two of the three unwired and stay green.
+
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+
+import {
+  isFrontendVirtualNode,
+  frontendVirtualTypesAmong,
+  withoutFrontendVirtualTypes,
+} from "../../web/js/lib/frontend-virtual-nodes.js";
+import { scanComboAvailability } from "../../web/js/lib/live-combo-availability.js";
+import {
+  describeUnrunnable,
+  missingNodeRunRefusal,
+} from "../../web/js/lib/missing-node-preflight.js";
+
+const ROOT = join(dirname(fileURLToPath(import.meta.url)), "../..");
+const PANEL = readFileSync(join(ROOT, "web/js/comfyui-mcp-panel.js"), "utf8");
+
+/** A live KJNodes/rgthree-style virtual node: its class set the flag. */
+const virtualNode = (id, type) => ({ id, type, isVirtualNode: true, widgets: [{ name: "Constant", value: "MODEL" }] });
+/** The SAME type as a defless placeholder — the pack's JS never loaded in this tab. */
+const placeholderNode = (id, type) => ({ id, type, widgets: [{ name: "Constant", value: "MODEL" }] });
+
+// ── the predicate ───────────────────────────────────────────────────────────────
+
+test("#1657 only an explicit isVirtualNode===true is proof; everything else stays reported", () => {
+  assert.equal(isFrontendVirtualNode({ isVirtualNode: true }), true);
+  // The whole point is that a placeholder cannot forge this by accident. Truthy-but-not-true
+  // shapes are the ones a loose check would swallow.
+  for (const bad of [
+    {},
+    { isVirtualNode: false },
+    { isVirtualNode: undefined },
+    { isVirtualNode: "true" },
+    { isVirtualNode: 1 },
+    null,
+    undefined,
+    "GetNode",
+    42,
+  ]) {
+    assert.equal(isFrontendVirtualNode(bad), false, JSON.stringify(bad));
+  }
+});
+
+test("#1657 a type is virtual only when EVERY live instance says so", () => {
+  const nodes = [
+    virtualNode(1, "GetNode"),
+    virtualNode(2, "GetNode"),
+    virtualNode(3, "Label (rgthree)"),
+    { id: 4, type: "KSampler" },
+  ];
+  assert.deepEqual([...frontendVirtualTypesAmong(nodes)].sort(), ["GetNode", "Label (rgthree)"]);
+  // One placeholder of the same type refutes the whole type: that type has a real problem.
+  assert.deepEqual(
+    [...frontendVirtualTypesAmong([...nodes, placeholderNode(5, "GetNode")])],
+    ["Label (rgthree)"],
+  );
+});
+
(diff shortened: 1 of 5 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**. Verdict: **SHIP**, head `f43b71f5`.

This PR refused a prescription from the backlog audit that dispatched it, so the gate's first job was deciding who was right. **The author was right on both counts, verified independently rather than taken on trust.**

- **The audit read the wrong repo's set.** Two same-named sets exist: mcp `src/services/workflow-converter.ts:2305` is `NON_EXECUTING` ∪ `WIRING_VIRTUAL` with no rgthree entries, while panel `web/js/lib/node-resolve.js:63` **already** lists Label, Fast Groups Bypasser/Muter, Fast Bypasser/Muter, rgthree Reroute, Node Collector and Get/Set — added by `a9cf8712` for #475. The audit described the mcp set and prescribed it for panel surfaces.
- **The allowlist would have shipped a bug, reproduced live.** On a clean rig with KJNodes absent (panel#1284's exact state), a loaded `GetNode` is a bare `LGraphNode` placeholder with `isVirtualNode` undefined and `graphToPrompt()` emits `{id:"1", class_type: null}`. The PR's filter still reports it, so it stays refused. **A literal list containing "GetNode" would have removed it and declared a genuinely dead run fine.** The refusal in mcp#1648 was correct; only its remedy was wrong.

**`isVirtualNode` is authoritative.** Exactly 11 occurrences in the shipped 1.48.7 bundle: the serializer's `if(e.isVirtualNode||e.mode===NEVER||e.mode===BYPASS)continue`, the DTO proxy, and `SubgraphNode … isVirtualNode=!0`. Critically, **`undefined` fails toward reporting at all three sites** — site 1 is not skipped and stays scanned, site 2 lands the type in `refuted`, site 3 does not use the flag at all and reproduces the old message verbatim.

**What stays refused, verified live:** an uninstalled pack is still reported and still refused; a real `unavailable_widget_values` finding still fires; the run gate remains `unrunnableNodeIds(built)`. A virtual node could never have produced an `unavailable` finding pre-fix (no `/object_info` entry → `combos === null` → `unknown`), so the skip costs no coverage.

**8/8 mutations applied and killed**, each proven by byte delta plus post-condition re-read — including two the gate added itself. Both author and gate hit the CRLF trap: MSYS `grep`/`cat -A` strip the CR and report "LF" while the tree is CRLF, so multi-line patterns silently no-op. Worth remembering repo-wide.

**Two non-blocking follow-ups recorded:**
1. **A false reassurance in the header comment.** It says a real placeholder cannot carry the flag because "no saved workflow can put it on a node whose class was never registered". It can: `LGraphNode.configure` in 1.48.7 is an arbitrary-key copy (`for(let t in e){…this[t]=e[t]}`), so a workflow JSON carrying `"isVirtualNode": true` on an unregistered type produces a defless placeholder with the flag as an own property, and `graph.serialize()` round-trips it. **Harm is nil** — ComfyUI's own serializer skips that node too, so it never reaches the backend, and the guard is exactly as trustworthy as the serializer it borrows from. The sentence should still be corrected rather than left as a claim that is false as written.
2. **A reachable inconsistency (P2), demonstrated live:** `collectMissingAssets` filters `nodeTypes` but leaves `nodeCount`, so the state `hasMissingNodes:true, missingNodeCount:2, nodeTypes:[]` is reachable — `graph_get_errors` then reports `missing_node_count: 2` with no names, and the turn-start banner says "2 node type(s) not installed". The alarm survives with the names stripped, which is narrower than before. It does not affect mcp#1657's rig, whose symptom comes from `live-combo-availability.js` (call site 1, fully fixed).

Issue #1172, opened 2026-08-13
panel_refresh_nodes reports refreshed but new CheckpointLoaderSimple retains empty ckpt validation list


artokun:
### What happened / root cause
On a live canvas, a newly added CheckpointLoaderSimple had an empty checkpoint combo/validation list even though the connected ComfyUI /models endpoint listed five checkpoints. panel_refresh_nodes returned {ok:true, refreshed:true}, and panel_query_graph showed the requested ckpt_name widget, but panel_run validation still rejected it because the server prompt contained ckpt_name options []. The refresh result is misleading and did not repopulate the loader's combo choices.

### Steps to reproduce
1. Add CheckpointLoaderSimple to the live canvas with panel_add_node.
2. list_local_models action:list model_type:checkpoints confirms juggernautXL_ragnarokBy.safetensors exists.
3. panel_set_widget ckpt_name to that file.
4. panel_run: validation fails: Value not in list ... not in [].
5. Call panel_refresh_nodes; it returns ok:true refreshed:true.
6. Re-set ckpt_name and panel_run again; same rejection.

### Exact error
CheckpointLoaderSimple: Value not in list (ckpt_name: 'juggernautXL_ragnarokBy.safetensors' not in [])

### Workaround
None through live-canvas tools; muted the experimental branch so the user's valid workflow remains runnable.

### Fix
Applied locally: no. Upstream-only: yes. panel_refresh_nodes should rebuild combo option arrays used by nodes added in the current tab, and should not report refreshed:true if those definitions remain empty.

### Environment
Windows 11; RTX 4090 24GB; ComfyUI 0.32.0; comfyui-mcp 0.51.17; panel 0.14.15.

artokun:
Closing as fixed — verified on current main: #1218 (shipped v0.14.34) rebuilds combo options during the reapply sweep and discloses authoritatively-empty lists scoped to types on the graph (empty_combo_lists note), and #1316 closed the report-refreshed-over-shrink hole. The eight #1172 pinning tests pass on v0.15.1. The remaining add-time empty-list disclosure is scoped out per the triage doc — worth its own small issue if wanted.

PR #1354, opened 2026-08-19
fix(1193): an abandoned combo refresh is a missing confirmation, not stale dropdowns

artokun:
Follow-up to #1346 (merged as ee47a27a while this was being measured). It keeps that PR's
analysis of the underlying cause of #1193 and replaces its remedy, because the remedy is
measurably sized for a condition it never fires in — and pays for it with the one property
`NODE_DEFS_RUN_BUDGET_MS` exists to hold.

**Round 2** (gate NO-SHIP): the replacement verdict failed UNSAFE on V2 combo specs and
could suppress a real missing-asset warning. Fixed, measured, and pinned — see "Round 2"
below.

## What I measured

In a real ComfyUI page driven by Playwright (frontend 1.48.7, 848 node types, warm):

```
                              getNodeDefs()      refreshComboInNodes()
unthrottled                        211 ms                 555 ms
every /object_info delayed 3000ms  3156 ms                3318 ms   (+2763 ms)
```

That `refreshComboInNodes()` re-fetches is documented in the base already
(`comfyui-mcp-panel.js:874/914/1457`); what is new here is the **quantified second
payment** — delaying the backend by 3 s moved the combo phase by 2.76 s, so the phase's
cost tracks the backend's latency rather than sitting near its warm figure. Also measured:
the cost scales with node-TYPE count, not graph size — adding 200 nodes left it unchanged
(223 ms vs 231 ms). I could not reproduce the reporter's 4846 ms (no 4320-type install
here); that number is quoted as theirs, never as mine.

## Why the floor could not have worked

`NODE_DEFS_COMBO_FLOOR_MS = 6000` pays out only when the remainder has dropped below it,
which requires the FETCH to have spent >3000 ms of bounded waiting (local work is excluded
by the give-back). That means `/object_info` was slow. The combo phase then issues its own
`/object_info` **against that same slow backend**, so it needs ≈ that latency again plus its
local work — on the reporter's rig ≈ 3000 + (4846 − 456) ≈ 7.4 s and up, past a 6000 ms
floor sized from a WARM measurement. Its helpful window on the rig that filed #1193 is
empty.

And the cost was real: the merged version put one run's worst-case bounded waiting at
15,000 ms and two serialized runs at 30,000 ms against the bridge's 20 s read default — its
own comment says so. Past that read the user gets a bare "did not reply within 20000 ms",
which names nothing: the #1161 symptom this budget was written to prevent.

## The fix

The budget is untouched (2 × 9000 < 20,000 again). The combo phase's **verdict** moves
instead:

- **`web/js/lib/asset-staleness.js`** — `reapplyDefsToLiveNodes(root, defs, stats)` reports
  what the sweep observably did: `nodesSwept`, `combosRebuilt`, `combosSkipped`, and
  `completed`, set as the **last statement of the walk**. `comboRebuildCovered(stats)` is
  the single predicate both readers ask.
- **`web/js/comfyui-mcp-panel.js`** — an **ABANDONED** combo call over a **covered** local
  rebuild is disclosed, not failed: `refreshed: true` with `combo_refresh_confirmed: false`
  and a note. A combo call that **THREW** still fails — a throw is the frontend reporting
  something wrong, a timeout is only the absence of an answer.
- The shared trust flag becomes `comboRan || comboAbandonedAfterRebuild`. What that flag has
  always stood for is "an authoritative payload was obtained AND the live lists were rebuilt
  from it"; a covered sweep satisfies exactly that, and it is work the panel *observes and
  counts*, where `refreshComboInNodes()` resolving is a call answering whose per-node effect
  the panel never sees.
- The remaining failure names the bound it was given and the budget it came from, so
  "starved down to a sliver by a slow fetch" and "given the whole budget and hung" stop
  being the same sentence.

**What breaks if the panel stops waiting** — checked, not assumed. The call still RUNS; only
the wait ends. Consumers of `refreshed`/the trust flag are `refresh_nodes`' reply,
`withRefreshTimeout` → `get_errors`' `trustCombo`, and `graph_add_node`'s drift re-check —
the last re-reads the registry rather than trusting the verdict, so it is unaffected.
Function-valued option sources are unaffected: they derive their own list and
`assetCandidateResolvesLive` invokes them.

## Round 2 — the V2 blocker the gate found

The gate was right, and the failure was in the unsafe direction.
`refreshComboOptionsFromDefs` handled only the V1 `[[opt,...], config]` spec. A V2
`["COMBO", {options:[…]}]` was walked past in silence while the sweep still set
`completed`, so `refreshed: true` and `nodeDefsRefreshConfirmed` were granted over lists
that had never been rebuilt — and a deleted model still sitting in a stale list is then
resolved as present and its warning suppressed.

**Which fix I chose: both, because neither half is sufficient.**

1. **Rebuild V2** (`authoritativeComboValues`). Not optional — measured against a live
   ComfyUI 0.33 `/object_info` (848 types): **V1 61 inputs, V2 467, V2-remote 1,
   `COMFY_DYNAMICCOMBO_V3` 120.** V2 is the majority shape; counting-only would have left
   the disclosure permanently off on any modern install.
2. **Count what could not be rebuilt** (`combosSkipped`), and require zero.
   `comboRebuildCovered` needs `completed === true` AND zero skips. Remote V2 (the list is a
   separate fetch — the frontend shows "Loading…") and V3 dynamic (options are `{key,
   inputs}` objects selecting SUB-INPUTS) are refused rather than guessed at, and a shape
   published after this was written is caught by the live widget already presenting an
   option array.

Verified against the real widgets the frontend builds, not fixtures: `LatentConcat.dim`
(V2) materialises as `options.values = ["x","-x","y","-y","t","-t"]`, so writing
`spec[1].options` is the right field; `LoadImageOutput.image` (remote V2) has no array at
all and value `"Loading..."`; `SaveImageAdvanced.format` (V3) shows `["png","exr"]` — the
keys — which is exactly why a naive rebuild there would look right.

**Before/after, executed against the shipped modules** (`isStaleAssetCandidate` driven with
`trustCombo` = what the panel would grant):

```
BEFORE  V1  combosRebuilt=1 skipped=0 covered=true  list=[sd15]        SUPPRESSED=false
BEFORE  V2  combosRebuilt=0 skipped=0 covered=true  list=[anime,sd15]  SUPPRESSED=true   <-- the bug
AFTER   V1  combosRebuilt=1 skipped=0 covered=true  list=[sd15]        SUPPRESSED=false
AFTER   V2  combosRebuilt=1 skipped=0 covered=true  list=[sd15]        SUPPRESSED=false
```

The end-to-end test is `#1193 END TO END: a V2 combo is rebuilt, so a deleted model is still
REPORTED missing`, and it drives the whole path to the suppression. Reverting the two halves
of the fix:

| code state | asset-staleness.test.mjs |
|---|---|
| as shipped here | 0 failing |
| BEFORE — V1-only, no skip counting (what the gate executed) | **3 failing**, incl. "a V2 combo must be rebuilt, not walked past" |
| V2 handled, skips NOT counted | 1 failing |
| skips counted, V2 NOT handled | 2 failing |

**Does the safety net turn the fix off?** Measured on the live install, over every
combo-bearing type instantiated in a real page (373 types): **528 combo widgets rebuilt,
121 skipped, 0 dynamic**. Every skip is an API/dynamic node (Bria, HeyGen, Kling, Meshy,
Qwen, Reve, Topaz — all `COMFY_DYNAMICCOMBO_V3`) or `LoadImageOutput` (remote). A graph of
ordinary local nodes — the ones missing-model candidates come from — is fully covered and
the disclosure fires; a graph containing one of those nodes withdraws the claim and falls
back to today's `combo_refresh_failed`, which is the safe direction.

## Refutation I ran

**19 mutations, 19 killed** (was 18/19; the survivor is now pinned):

| mutation | caught by |
|---|---|
| **`completed` alone gates the claim** (the gate's finding) | `#1193: a sweep that FINISHED but skipped a combo does NOT license the disclosure` |
| **completion claimed before the walk** (the gate's survivor) | `#1193 the sweep REPORTS what it rebuilt…` |
| V2 specs unhandled again | `#1193 END TO END…`, `authoritativeComboValues…` |
| skips uncounted / `covered` ignores the counter / accepts truthy `completed` | `a combo the sweep CANNOT rebuild is counted…`, `comboRebuildCovered needs BOTH facts` |
| skip test asks only the spec / only the live widget | `a combo the SPEC does not announce…`, `a combo the sweep CANNOT rebuild…` |
| dynamic (function) sources counted as skips | `a function-valued option source is NOT a skip` |
| trust flag ignores the local rebuild | `#635 … verdict through the panel wiring`, `#1193 … disclosed, not failed` |
| disclosure branch removed / broadened to credit a THROWN combo | `#1193 … disclosed, not failed`, `… THREW is still a failure` |
| reply drops the disclosure (the #981 whitelist hole) | `#1172 WIRING …`, `#1193 … survives the field whitelist` |
| refusal stops naming the bound | `#1193 … names the bound it was given` |
| armed bound recomputed / sweep not asked for its stats | `#1180 a whole refresh RUN …` |
| verdict collapses back to `refreshed:false` / claims CONFIRMED | `#1193 … disclosed, not failed` |

On the gate's survivor: `reapplyDefsToLiveNodes` has exactly one production caller
(`comfyui-mcp-panel.js:1466`), inside `if (defs)` and `if (rootGraph)`, so the `!defsByType`
early return is unreachable from production today. That is a property of the CALL SITE, one
refactor from changing, and the failure it would cause is the silent one — so it is pinned
directly rather than argued away.

The `THREW` mutation is not hypothetical either: the **first** version of this change
credited a rejected combo refresh, because the phase reifies its outcome and `didThrow`
stays false for it. The test found that before review did.

## Rebase

Rebased onto current `origin/main` (8193b226), which carries both #1342 (82f330e1) and
#1349. One conflict, in `single-node-def.test.mjs`, where #1192's merged work replaced the
tracked-ceiling ratchet with the `naiveSum > budget` FITS assertion.

**Resolution: main's side, minus the floor term.** Everything #1192 added survives — the
`RELAY_WINDOW_MS` slack check, `naiveSum > budget`, the per-step "no single step exceeds the
command budget" loop, and the five `budget.bounded(...)` call-site pins. `naiveSum` is now
`33000 > 25000` (was `41000 > 25000`), so the property #1192 asserts still holds with the
floor gone. The only thing dropped is my own earlier widen block, which duplicated a
stronger pin main already has in the `#580` test (it reads the divisor from source and
asserts `widen > 0 && widen < registration`).

## Stale prose corrected

Three sites said trust requires the frontend call to have run, which this change makes
false — `registerComfyNodeDefs`' opening contract, `withRefreshTimeout`'s doc, and
`refresh_nodes`' doc. All three now say the lists must have been REBUILT and name both ways
that can happen. The `NODE_DEFS_RUN_BUDGET_MS` block and the trust-flag comment carry the
measurement and the V2 reason.

## Verification

- `npm run typecheck` — exit 0.
- `npm run test:unit` (i18n-check, i18n-render-check, check-tool-vocabulary,
  **check-panel-scope OK**, `node --test browser_tests/unit/*.test.mjs`) — **5205 tests,
  5203 pass, 1 fail**, the fail being `#671 verifyInstalled … fast-draining install`, the
  documented real-wall-clock flake: it passes in isolation (125/125, three times) and this
  box is running eleven agents.
- **Playwright, run twice back to back on the same ComfyUI (:8211), `--workers=1`:**

  | panel under test | passed | failed |
  |---|---|---|
  | this branch | 55 | 20 |
  | `origin/main` (baseline, same instance, immediately after) | 54 | 21 |

  **Regressions: NONE** — the branch's 20 failures are a strict subset of main's 21, and the
  parser that compared them asserts it parsed exactly as many specs as each run declared
  (a first version silently matched nothing and would have reported "no regressions" from
  an empty set). The one extra baseline failure, `autostart-mcp.spec.ts:26`, is a
  spawn-timing flake. The 20 shared failures are the suite's pre-existing state on this box
  (chat-history / conversation-persistence / streaming, most timing out in
  `MockBridge.waitForUserMessage`) — none touch node-def refresh, combos, or the missing-asset
  scan. The branch's panel was verified live in the served bundle before the run
  (`comboRebuildCovered` present), and the borrowed slot was restored afterwards.
  Baseline caveat: it ran against main@1e3580d0, two commits ahead of this branch's base.

## Out of scope, seen and left alone

`isEmptyComboSpec` — the #1172 empty-combo disclosure — is V1-only too, so on a current
install it is blind to most empty lists. Filed separately rather than widened into this PR:
it is a disclosure, not the trust path, and the interaction is safe in the meantime (an
empty V2 list is now rebuilt to `[]`, so nothing resolves against it and candidates are
over-reported rather than suppressed). Likewise the `COMBO_API_ABSENT` remedy, which still
says combo lists "cannot be rebuilt in place" — false since #1172.

## Relationship to #1342 / #1192

Disjoint, and #1342 did not address the underlying cause of #1193: it gives `graph_add_node`
ONE command budget so its bounds stop summing, where this is about how a single
`registerComfyNodeDefs` run splits its 9000 ms between the fetch and the combo phase. The
one place they meet is the budget test above, and this change *lowers* that sum rather than
raising it.


artokun:
## Round 2 — gate NO-SHIP addressed

**The blocker (V2 fails unsafe): fixed. I chose BOTH remedies**, because neither is sufficient alone. Measured on a live ComfyUI 0.33 `/object_info` (848 types): **V1 61 combo inputs, V2 467, V2-remote 1, `COMFY_DYNAMICCOMBO_V3` 120** — V2 is the majority shape, so counting-only would have left the disclosure permanently off on any modern install, and rebuilding-only would still claim coverage over remote/dynamic shapes.

- `authoritativeComboValues(spec)` reads V1 and V2 and returns null for anything it cannot derive.
- a null return over a widget that IS a combo increments `stats.combosSkipped`; `comboRebuildCovered(stats)` requires `completed === true` **and** zero skips, and both the disclosure and the trust flag ask that one predicate.

**Before/after, executed against the shipped modules** (`isStaleAssetCandidate` with `trustCombo` = what the panel grants):

```
BEFORE  V2  combosRebuilt=0 covered=true  list=[anime,sd15]  MISSING-ASSET SUPPRESSED=true   <-- the finding
AFTER   V2  combosRebuilt=1 covered=true  list=[sd15]        MISSING-ASSET SUPPRESSED=false
```

New test `#1193 END TO END: a V2 combo is rebuilt, so a deleted model is still REPORTED missing` drives that path. Against the exact pre-gate code (V1-only, no counting) `asset-staleness.test.mjs` goes **0 → 3 failing**; V2-without-counting fails 1, counting-without-V2 fails 2.

Real shapes confirmed in a live page rather than from fixtures: `LatentConcat.dim` (V2) materialises as `options.values = ["x","-x","y","-y","t","-t"]`; `LoadImageOutput.image` (remote) has no array and reads `"Loading..."`; `SaveImageAdvanced.format` (V3) shows the option KEYS, which is why a naive rebuild there would look right.

**Does the safety net switch the fix off?** Over every combo-bearing type instantiated in a real page (373 types): **528 rebuilt, 121 skipped**. Every skip is an API/dynamic node (Bria, HeyGen, Kling, Meshy, Qwen, Reve, Topaz — all V3) or `LoadImageOutput`. Ordinary local graphs are covered; a graph with one of those falls back to today's `combo_refresh_failed`, the safe direction.

**Also addressed:**

1. **Rebase** onto `origin/main` (8193b226, carrying #1342 and #1349). One conflict, in `single-node-def.test.mjs`. Resolution: main's side minus the floor term. Every #1192 pin survives — `RELAY_WINDOW_MS` slack, `naiveSum > budget`, the per-step loop, and the five `budget.bounded(...)` call-site pins. `naiveSum` is now **33000 > 25000** (was 41000 > 25000), so #1192's property still holds with the floor gone. The only thing dropped is my own earlier widen block, which duplicated a stronger pin main already has in the `#580` test.
2. **Stale prose** corrected at all three sites (`registerComfyNodeDefs`' contract, `withRefreshTimeout`, `refresh_nodes`) — each now says the lists must have been REBUILT and names both ways that can happen.
3. **Your survivor** (`stats.completed = true` above the `!defsByType` early return): confirmed unreachable — `reapplyDefsToLiveNodes` has exactly one production caller, `comfyui-mcp-panel.js:1466`, inside `if (defs)` and `if (rootGraph)` — **and pinned anyway**, because that is a property of the call site and the failure it would cause is silent. **19/19 mutations killed** (a second survivor I found, the live-widget arm of the skip test, is pinned too).
4. **PR body** no longer claims the re-fetch is invisible from the source; it claims the quantified second payment, which is what is actually new.

**Playwright**, run twice back to back on :8211 with `--workers=1`: branch **55 passed / 20 failed**, `origin/main` baseline on the same instance **54 passed / 21 failed**. **Regressions: NONE** — the branch's failures are a strict subset of main's. (The comparison's first parser matched nothing and would have reported "no regressions" from an empty set; the one used asserts it parsed exactly as many specs as each run declared.) The extra baseline failure is `autostart-mcp.spec.ts:26`, a spawn-timing flake. Branch code was verified present in the served bundle before the run, and the borrowed `ap-1` slot was restored afterwards.

`npm run test:unit`: 5205 tests, 5203 pass, 1 fail — `#671 verifyInstalled`, the documented wall-clock flake, 125/125 in isolation. `check-panel-scope` OK. `typecheck` clean. Not merged, no gate script run.

artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by **independent adversarial reviewers that had never seen the author's reasoning**. First round: **NO-SHIP**. After rework, second round: **SHIP**, head `9e26dd4f` (matches the reviewed SHA).

**The first round found this failing UNSAFE**, which is the direction that matters: `refreshComboOptionsFromDefs` handled only V1 specs, silently skipping V2, while `stats.completed = true` fired regardless and the call site read only `.completed`. So `refreshed: true` was granted over combo lists never rebuilt, **suppressing a real missing-asset warning** — where the merged #1346 fails safe.

**The second round found the hole was wider than that.** Reproduced against the pre-gate commit with shipped modules: `V2 combosRebuilt=0 covered=true MISSING-ASSET SUPPRESSED=true`, and also `V3dyn covered=true SUPPRESSED=true`. After the fix: `V2 combosRebuilt=1 SUPPRESSED=false`, and `V3dyn combosSkipped=1 covered=false SUPPRESSED=false` — failing safe.

**The design's justification reproduced to the unit** against a live `/object_info` on a real 848-type install: **V1 61, V2 467, V2-remote 1, `COMFY_DYNAMICCOMBO_V3` 120, 373 combo-bearing types**. V2 is the majority shape, so counting-only would leave the disclosure permanently off on a modern install, while rebuilding-only would still claim coverage over remote and dynamic shapes. Hence both halves: `authoritativeComboValues()` reads V1 and V2 and returns null for what it cannot derive, a null over a real combo increments `stats.combosSkipped`, and `comboRebuildCovered()` requires `completed === true` **and** zero skips — asked by both the disclosure and the trust flag.

**The safety net does not switch the fix off.** Running the shipped sweep over the real payload: `swept=845 rebuilt=528 skipped=121`, independently reproducing the author's figures. The 121 land on 111 types — **91 `api_node:true`** plus 17 truly-local (`LoadImageOutput`, `SaveVideo`, `SaveImageAdvanced`, `SaveAudioAdvanced`, niche `comfy_extras`). No mainstream loader, sampler or encode node, so the fix engages on exactly the missing-checkpoint case this issue is about. `comboRebuildCovered` is graph-wide, so one `SaveVideo` switches the disclosure off for that run — but that degrades to main's existing `combo_refresh_failed`, never to a false grant. A utility limit, not a safety one.

**On the merged floor this removes:** its helpful window on the rig that filed #1193 is **empty**. It fires only when fetch F > 3000 ms while combo need ≈ F + 4390 ms, so 6000 ms never covers the case it exists for; two runs ≤ 20 s force floor ≤ 10000−F, and the best retune buys a ~500 ms band with zero slack. Removal is right; retune is not.

**Mutations.** Library: V2 branch removed → 2 killed (including the end-to-end); skip increment removed → 2; `combosSkipped` term → 4; `completed` term → 4; and the previously-*surviving* `stats.completed = true` on the `!defsByType` early return is now killed. Panel wiring, proving reachability rather than the helper alone: disclosure branch disarmed → 2; `covered()` weakened to `.completed` → 2; trust flag drops the new term → 2; stats arg dropped from the sweep → 2; whitelist forward removed → 1.

**Rebase integrity checked pin by pin**, since the conflict was in the pins file — where a guard gets silently dropped. Seed wait, fast path, whole-schema fetch and registration wait all die when mutated. **P4, the coalescer join, SURVIVES — but it is pre-existing, not dropped here**: the assertion is a single `assert.match` over `graph_add_node`'s body and there are *two* `joinMs: budget.remaining() - ADD_NODE_POST_REFRESH_RESERVE_MS` sites (`:11993`, `:12084`), so unwrapping one leaves the other satisfying the regex. The identical mutation survives on the pre-rebase base `8193b226` too. Filed as a follow-up rather than held against this PR. `RELAY_WINDOW_MS` slack, `naiveSum > budget` (now 33000 > 25000) and the per-step loop all still assert.

**Playwright is not evidence here and is not claimed as such:** no spec under `browser_tests/` references `refresh_nodes`, `refreshComboInNodes` or `combo_refresh`, so the suite is blind to this change and the ~20 failures on both sides are pre-existing noise. The unit and mutation evidence carries it. Trial-merged against current main (`929a4b0d`): clean, **5393/5393** unit pass, `check-panel-scope` OK.

Two nits left for later: `comfyui-mcp-panel.js:706` still describes the trust flag as "True only AFTER a node-def + combo refresh has completed successfully", which no longer covers the own-sweep arm (untouched by this PR), and `refresh-coalesce.test.mjs:154` still says "AND refreshed combos". One benign unadvertised win: `panel_set_widget`'s combo pre-refresh shares `refreshComboOptionsFromDefs`, so V2 handling now fixes that path too.

Issue #1267, opened 2026-08-15
Workflow graph binding becomes stale after render reconnect and save-as can race to an empty tab


artokun:
### What happened / root cause
During a long live-canvas session with many panel_run completions and two ComfyUI restarts, subsequent graph mutations repeatedly failed closed with a workflow-instance mismatch until panel_set_workflow_target(mode:current) was called again. In a related save-as attempt, the tool reported the new workflow name, the active canvas reverted, and the newly named copy was empty. This suggests asynchronous reconnect/tab routing updates are not serialized with the workflow-instance fence and save acknowledgement.

### Steps to reproduce
1. Keep a saved workflow open and queue several panel_run calls.
2. Reconnect/restart ComfyUI while a run is interrupted, then continue editing the same visible tab.
3. Observe graph calls reject the visible workflow because the stored instance fence is stale; rebinding current repairs it.
4. Save-as during this state; observe the named tab/file may be acknowledged before the intended graph is loaded, leaving an empty copy and returning focus to the prior canvas.

### Exact error (scrubbed)
Graph command rejected with a workflow instance mismatch despite the panel and user showing the same intended canvas. Save-as produced a successful acknowledgement but the new copy contained an empty graph.

### Fix
Applied locally: no. Upstream-only: yes.
Serialize reconnect, active-tab routing, graph-instance fence updates, and save-as completion. Do not acknowledge save-as until the source graph has been captured, the destination persisted, and the active instance/routing key has been confirmed. On reconnect, refresh the fence from the visible tab before accepting the next mutation. Workaround verified: explicitly rebind mode:current after reconnect and save the existing workflow in place rather than save-as.

### Environment
Linux; ComfyUI 0.27.0; ComfyUI frontend 1.45.20; Python 3.14.6; PyTorch 2.12.1+rocm7.2; AMD Radeon RX 9070 XT 16 GB; comfyui-mcp 0.51.56; panel 0.14.42.

artokun:
Claimed by the backlog-drain session (worktree `.claude/worktrees/drain-1267`). Verified unclaimed with no open PR immediately before claiming. A fix agent is assigned now; if no PR appears within ~8h, consider the claim lapsed.

artokun:
PR #1359 is open and CI is green. It addresses **half (b) only** — the save-as half. Half
(a) (the stale workflow-instance fence after a render reconnect) is untouched; no
reconnect, routing, or fence code is in the diff.

[...]

PR #1359, opened 2026-08-19
fix(1267b): a Save-As acknowledges the CAPTURE it observed, not the openWorkflow it called

artokun:
Addresses **half (b) only** of panel#1267 — the save-as half. Half (a) (the stale
workflow-instance fence after a render reconnect) is **not touched**: it overlaps
reconnect work owned elsewhere (#654's re-advertise fix landed today, #1014 was parked
as a design record), and nothing here reads or moves the fence.

## The defect: an acknowledgement issued before the effect it claims

The reporter got the new workflow's name back, the canvas reverted, and the copy was
empty. Reproduced live, below.

The Save-As copy route is the low-level trio `saveAs → openWorkflow → saveWorkflow`.
What the bytes on disk actually are, measured against the installed frontend bundle
(`comfyui_frontend_package`, not inferred):

- `workflowStore.saveAs(wf, path)` returns the copy **unloaded** — it never sets
  `changeTracker`.
- `ComfyWorkflow`: `get activeState() { return this.changeTracker?.activeState ?? null }`.
- `ComfyWorkflow.save()` begins `this.content = JSON.stringify(this.activeState)` and
  POSTs that. So a copy whose tracker was never built writes the **JSON literal `null`**.
- Both `openWorkflow` variants can return **without loading** (each opens with an
  `isActive` early exit), and neither returns anything that distinguishes "loaded it"
  from "decided not to".

`main`'s only protection was a **capability** check — `openWorkflow` must exist,
therefore the copy must be loaded. That is a dispatch receipt, not an observation.

Worse, every downstream guard **confirms** the bad write. The post-write read-back
(`reconcileSavedCopy`) compares the target against the copy's **own** content, so a copy
that wrote `"null"` finds `"null"` on disk and reads back as `"ours"` — the check meant to
catch a bad write blesses this one. "We wrote the user's graph" and "we wrote the literal
`null`" were the same observation.

## The fix: ask what is about to be written, then report what was written

One classifier, `classifyGraphCapture`, asked at two points in the copy adapter
(`web/js/lib/workflow-save.js`):

1. **Pre-write**, synchronously between `openWorkflow(copy)` and `saveWorkflow(copy)`,
   with no `await` in between: classify `copy.activeState` — what `save()` is about to
   serialize. `uncaptured` ⇒ refuse. Nothing is written, the source file and tab are
   untouched, the in-memory copy is removed and the previous tab restored.
2. **Post-write**, after a reported-success persist: classify `copy.content` — the exact
   bytes `save()` POSTed, i.e. the effect observed rather than the request made. This
   closes the one residual the pre-write check cannot reach (`save()` awaits a dynamic
   import before reading `activeState`). It surfaces the truth instead of a phantom
   success; it does not delete the file, and the source was never touched.

A pre-commit refusal is marked so the failure handler skips its ambiguous-post-commit
reconciliation — that path exists to **adopt** a write that committed before its response
was lost, and adopting a refusal would turn it straight back into a reported success.

## How "captured too early" is separated from "genuinely empty"

The signal is **capture completion**, never node count — a user is entitled to Save-As an
empty canvas, and refusing that is data loss pointed the other way.

- **Genuinely empty** ⇒ a *completed* serialization. `LGraph.serialize()` always returns an
  object carrying a `nodes` array, and the frontend's own `blankGraph` is
  `{last_node_id:0,last_link_id:0,nodes:[],links:[],groups:[],config:{},extra:{},version:0.4}`
  (both read out of the shipped bundle). `nodes: []` classifies as **captured** and saves.
- **Never captured** ⇒ no serialization happened at all: `null`/absent tracker, or a value
  that is not a serialized graph. No canvas — empty or not — can produce that.
- **Unknown never refuses**, and it is a real third state, not a hedge: on a real
  `ComfyWorkflow`, `activeState` is a class getter that yields `null` when unloaded, never
  `undefined`. `undefined` therefore means the object does not expose the field (an older
  frontend, a stub), which is precisely when nothing has been observed. Unparseable bytes
  are `unknown` too.

## Live verification on a real rig

An isolated ComfyUI (0.33.1, frontend 1.48.7) on its own port and `--base-directory`, with
an **empty** workflow library — every workflow below is a throwaway this run created. The
real `workflowStore` trio, with the panel's real `saveActiveWorkflow` imported into the page
from the URL ComfyUI serves it at. The owner's `:8188` was not touched.

| scenario | `origin/main` | this branch |
| --- | --- | --- |
| A — healthy Save-As of a 1-node graph | saved; disk has 1 node | saved; disk has 1 node (unchanged) |
| B — copy left uncaptured (`openWorkflow` moves the pointer, builds no tracker) | **reported success**, disk = `null`, `NOT-A-GRAPH` | **refused**; target `404` (nothing written); source still on disk; active tab back on the source |
| C — genuinely EMPTY canvas | saved; `nodes: []` | saved; `nodes: []` (unchanged) |

B on `main` is the reported bug, byte for byte: the tool returns the new name, and the file
is the literal `null`.

## Refutation

Four mutations, each reverted after:

| mutation | result |
| --- | --- |
| delete the pre-write capture guard | **2 fail** (the refusal test; the pre-commit/reconcile test) |
| delete the post-write bytes check | **2 fail** (the in-`save()` loss test; the read-back-blesses-it test) |
| drop `!isPreCommitRefusal(err)` from the reconcile branch | **1 fail** (refusal gets adopted as success) |
| swap capture-completion for a **node-count** veto | **9 fail** — including "a legitimately EMPTY canvas still saves" and 7 pre-existing #226/#309 save tests |

The last one is the both-directions proof: the obvious wrong fix (refuse an empty graph)
is caught by the suite, not by review.

Both failing tests under mutation 1 drive the real `saveActiveWorkflow`, so the **call
site** is pinned, not just the helper — deleting the whole wiring hunk goes red.

## Tests

`browser_tests/unit/save-as-capture-observed.test.mjs` (new, 8 tests): the classifier
table in all three states; the call-site refusal with "nothing written / source intact /
previous tab restored"; a **regression witness** that drives the double bare and asserts it
really does produce `"null"` (so a green suite is not green by construction); the empty-canvas
direction; the pre-commit/reconcile separation; the in-`save()` loss; and the read-back
collapse this fix removes.

Two shared doubles in `workflow-save.test.mjs` were made faithful rather than loosened: a
copy's `content` is now a serialized graph (`{"id":…,"nodes":[],"links":[]}`) instead of a
bare `{"id":…}`, and the S0/S1/S2 tracker snapshots are graph-shaped. That is what the real
`store.saveAs` produces; no assertion was weakened.

## Gates

- `npm run test:unit` — **5090 pass / 1 fail**, the fail being the known #671 wall-clock
  flake under load; **125/125 green** when `manager-install.test.mjs` is re-run alone.
- `npm run typecheck` — clean. `node --check` on the changed file — clean.
- Playwright `--workers=1`, fix branch vs an `origin/main` baseline worktree, each against
  its own isolated ComfyUI: **53 passed / 22 failed on BOTH, and the 22 failure names are
  byte-identical — delta zero.** (Those 22 are environmental in this sandbox: chat-history
  / IndexedDB / streaming specs that need an orchestrator. They fail the same way on
  `origin/main`.) `save-persists-node-set-values` and `dirty-guards-see-node-writes` pass
  on both.
- **codex was unavailable (account exhausted until 2026-08-19), and `claude-gate.mjs` is
  blocked on the weekly limit — so this PR carries NO independent gate.** Per instruction I
  did not run a gate script and did not substitute a self-review for one. The orchestrator
  gates and merges.

- Did not touch half (a) — no reconnect, routing, or workflow-instance-fence code is in
  this diff.
- Did not add a node-count or "looks empty" veto anywhere; the only refusals are on values
  that are not a serialized graph at all.
- Did not extend the guard to the **in-place** save route. It writes `wf.activeState` for the
  ACTIVE tab, which is loaded by construction, and widening a refusal onto the route that
  overwrites a user's existing file is not something this evidence supports.
- Did not delete the empty file the post-write check detects — deleting on that path is its
  own hazard; the error names the file so the user can.

Reference: this addresses **the underlying cause of #1267 (half b)**. Closing is the
orchestrator's call.

### DIFF
--- a/browser_tests/unit/save-as-capture-observed.test.mjs
+++ b/browser_tests/unit/save-as-capture-observed.test.mjs
@@ -1,0 +1,335 @@
+// #1267 half (b) — a Save-As must ACKNOWLEDGE THE CAPTURE, not the call.
+//
+// Reported: "the tool reported the new workflow name, the active canvas reverted, and
+// the newly named copy was EMPTY." The mechanism, read off the installed 1.47 frontend
+// bundle rather than inferred:
+//
+//   ComfyWorkflow:  get activeState() { return this.changeTracker?.activeState ?? null }
+//   ComfyWorkflow.save():  this.content = JSON.stringify(this.activeState); …POST…
+//   workflowStore.saveAs(wf, path):  builds the copy UNLOADED (changeTracker = null)
+//   workflowStore.openWorkflow(w):   if (isActive(w)) return w   ← returns without loading
+//
+// So the bytes a Save-As writes are the COPY's activeState at write time, and a copy
+// whose tracker was never built serializes to the JSON literal `null`. `main` guarded
+// this with a CAPABILITY check only ("openWorkflow exists, therefore the copy is
+// loaded") — a dispatch receipt, not an effect. And every downstream guard confirms the
+// bad write: the post-write read-back compares the target against the copy's OWN
+// content, so `"null" === "null"` reads back as "ours" — a false success, not a catch.
+//
+// THE TWO EMPTY CASES MUST STAY DISTINGUISHABLE. A user can legitimately Save-As an
+// empty canvas, and refusing that would be data loss pointed the other way. The signal
+// is CAPTURE COMPLETION, never node count: the frontend's own blankGraph is
+// {last_node_id:0,last_link_id:0,nodes:[],links:[],groups:[],config:{},extra:{},version:0.4}
+// and LGraph.serialize() always returns an object carrying a `nodes` array. `nodes: []`
+// saves; `null` refuses. Both directions are asserted below.
+
+import assert from 'node:assert/strict'
+import test from 'node:test'
+
+import {
+  saveActiveWorkflow,
+  classifyGraphCapture,
+  classifyWorkflowCapture
+} from '../../web/js/lib/workflow-save.js'
+
+// The frontend's own blankGraph (read out of the installed bundle) — what an EMPTY
+// canvas actually serializes to. This is the shape the guard must never refuse.
+const BLANK_GRAPH = {
+  last_node_id: 0,
+  last_link_id: 0,
+  nodes: [],
+  links: [],
+  groups: [],
+  config: {},
+  extra: {},
+  version: 0.4
+}
+
+const GRAPH_WITH_NODES = {
+  ...BLANK_GRAPH,
+  last_node_id: 2,
+  nodes: [{ id: 1, type: 'CheckpointLoaderSimple' }, { id: 2, type: 'SaveImage' }]
+}
+
+/** A workflow-store double that models the REAL 1.47 data flow the bug lives in:
+ *  the copy is created UNLOADED, `activeState` is derived from the change tracker,
+ *  and the persist serializes `activeState` into `content` before writing it.
+ *
+ *  Knobs (each reproduces one real failure mode, none loosen an assertion):
+ *    openLoadsCopy   — false ⇒ openWorkflow returns WITHOUT building a tracker
+ *                      (the store's `if (isActive(w)) return w` early exit).
+ *    dropTrackerAtWrite — true ⇒ the tracker vanishes inside save()'s own await
+ *                      (the microtask gap it opens before reading activeState).
+ *    sourceState     — what the SOURCE tab's graph is (an empty canvas is legal). */
+function makeStore({
+  files = [],
+  active,
+  openLoadsCopy = true,
+  dropTrackerAtWrite = false
+} = {}) {
+  const disk = new Map()
+  for (const f of files) disk.set(f, JSON.stringify({ id: 'source-uuid', ...GRAPH_WITH_NODES }))
+  const lookup = new Map()
+  if (active) lookup.set(active.path, active)
+  const calls = []
+
+  const svc = {
+    activeWorkflow: active,
+    calls,
+    disk,
+    lookup,
+    getWorkflowByPath(path) {
+      return lookup.get(path)
+    },
+    // workflowStore.saveAs: deep-copies the SOURCE's activeState into the copy's
+    // content, mints a fresh id, and inserts an UNLOADED record at the target path.
+    saveAs(wf, path) {
+      calls.push(['saveAs', wf.path, path])
+      const cloned = JSON.parse(JSON.stringify(wf.activeState))
+      cloned.id = 'copy-uuid'
+      const serialized = JSON.stringify(cloned)
+      const copy = {
+        path,
+        filename: path.split('/').pop(),
+        directory: path.split('/').slice(0, -1).join('/') || 'workflows',
+        initialMode: wf.initialMode,
+        size: -1,
+        isPersisted: false,
+        isTemporary: true,
+        changeTracker: null, // UNLOADED — exactly as the real store leaves it
+        content: serialized,
+        originalContent: serialized,
+        get activeState() {
+          return this.changeTracker ? this.changeTracker.activeState : null
+        }
+      }
+      lookup.set(path, copy)
+      return copy
+    },
+    async openWorkflow(copy) {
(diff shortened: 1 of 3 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**, against `.claude/autopilot/review-taxonomy.md`. Verdict: **SHIP**, head `7d6b872` (matches the reviewed SHA).

This is silent data loss on `main` today, so the gate was told a concrete failing case in *either* direction would block it — a guard that refuses legitimate saves is data loss too.

**The mechanism is real, verified twice rather than from memory.** Read out of the shipped bundles for **all eight installed frontend versions, 1.44.19 → 1.48.7, including the reporter's own 1.45.20**: `get activeState(){return this.changeTracker?.activeState ?? null}`; `saveAs` constructs `new e.constructor({path, modified, size:-1})` and never touches the `changeTracker = null` class field; `save()` does `this.content = JSON.stringify(this.activeState)` and `UserFile.save` POSTs `this.content`. Then measured **live** on ComfyUI 0.33.1 / frontend 1.48.7, read-only: `store.saveAs(active, path)` gives `changeTracker === null`, `isLoaded === false`, **`activeState === null`**, and `JSON.stringify(activeState) === "null"`.

Reachability of the drop is in the source too: `isActive = e => activeWorkflow?.path === e.path` is **path** identity, and both `openWorkflow` variants early-exit on it without building a tracker, returning nothing that distinguishes "loaded" from "declined". And `main`'s read-back collapse is confirmed — the reconcile `"ours"` branch does `return finish()`, so a `"null"` write reports success. **The check that existed to catch a bad write confirmed it.**

**The two empty cases are genuinely separable, which is the design's crux.** `LGraph.serialize()` builds `nodes` unconditionally in `asSerialisable`; `blankGraph` and a fresh `new LGraph().serialize()` both give `nodes: []` live; and **both** zod workflow schemas — `version 0.4` and the newer `version 1` that the subgraph blueprint extends — mark `nodes` **required**. The sharpest false-refusal hazard the gate could construct — a real `activeState` is a Vue reactive Proxy — does not bite: `Array.isArray(proxy.nodes) === true`, measured live. `undefined` is not a hole either; the `?? null` is present back to 1.44.19.

**The node-count mutation is the evidence the design rests on, and it reproduces exactly.** Adding `|| state.nodes.length === 0` fails **10 tests — the 9 claimed plus the known #671 flake — including 7 pre-existing #226/#309/P0 save tests.** The obvious wrong fix is caught by a suite that predates this PR. Every mutation was applied by exact-byte Python replacement and confirmed with `git diff` first, because the file is CRLF.

**Call-site pinning holds:** deleting the pre-write guard's *call* while leaving the exported helper intact reddens two tests that both drive the real `saveActiveWorkflow`; same for the post-write check; the classifier-table test stays green under both. So the call-site tests are doing the pinning, not the helper test.

Suite 5092 tests / 5090 pass / 1 fail (the #671 wall-clock flake). `tsc --noEmit`, `check-panel-scope`, `check-tool-vocabulary`, `node --check` all clean.

**Two constructed cases, neither blocking:** (1) an active tab whose `activeState` is valid JSON with no top-level `nodes` — an API-format prompt opened as a workflow — is refused here and copied on `main`; no shipped schema produces that, and nothing is written or destroyed. (2) A mid-save tab switch **plus** a `syncWorkflows` landing in the microtask continuation window could make the post-write check call a *correct* file empty; it needs both, and the disk is right either way. Worth a follow-up note.

**Scope calls are right.** The in-place route is deliberately unguarded — the frontend's own sync never unloads the *active* workflow, so its `wf` is loaded by construction, and widening a refusal onto the route that overwrites a user's real file without evidence of the hole is the "fix worse than the bug" shape. And a detected empty file is named rather than deleted.

Issue #1271, opened 2026-08-15
panel_promote_widget fails on frontend 1.48.7: 'promotion' store gone; preview widgets need the previewExposure store (and the real link-path error is masked)


artokun:
### What happened / root cause
`panel_promote_widget` fails on ComfyUI frontend 1.48.7 with:

    Error: widget-promotion unavailable on this ComfyUI frontend (no 'promotion' store)

Two separate defects:

1. **The promotion pinia store no longer exists on frontend 1.48.7.** Enumerating the store ids in the shipped bundle (`comfyui_frontend_package/static/assets`) gives no `promotion` store; widget promotion moved to the subgraph API, and PREVIEW widgets moved to a new store `previewExposure` (`getExposures` / `addExposure` / `removeExposure`), defined in `settingStore-*.js`.
2. **The reported error is the wrong one.** For `\$\-image-preview` the PRIMARY (link-only) path is what actually fails, with `Widget "\$\-image-preview" is not backed by a connectable input slot` — a preview widget has no input slot, so `subgraph.addInput` + `connect` can never promote it. That real message is swallowed by `catch (linkErr)`, and the user only sees the stale-store message from the fallback.

Net effect: the panel cannot do what the canvas context menu offers right there ("Promote Widget: \$\-image-preview"), which is the supported way to surface a subgraph's live sampler preview on the parent SubgraphNode.

### Steps to reproduce
1. Frontend 1.48.7, a subgraph containing a KSampler.
2. `panel_enter_subgraph(<subgraph node>)`
3. `panel_promote_widget(node_id: <inner KSampler>, widget: "\$\-image-preview")`
4. Errors as above. (Right-clicking the same inner KSampler on the canvas DOES offer "Promote Widget: \$\-image-preview" and works.)

### Exact error (scrubbed)
    Error: widget-promotion unavailable on this ComfyUI frontend (no 'promotion' store)

### Fix
<applied locally: yes>  <upstream-only: no>

Route preview widgets (`\$\$`-prefixed) through the `previewExposure` store, and stop masking the primary failure when the legacy store is absent. Patch against `web/js/comfyui-mcp-panel.js` (panel 0.14.22):

```diff
diff --git a/web/js/comfyui-mcp-panel.js b/web/js/comfyui-mcp-panel.js
index a97c7b1..2f03c37 100644
--- a/web/js/comfyui-mcp-panel.js
+++ b/web/js/comfyui-mcp-panel.js
[...]

PR #1361, opened 2026-08-19
fix(1286): keep execution image previews on the node that emitted them

### LINKED ISSUE #1286: Image preview attaches to ConditioningConcat after live-canvas edit and run
A live-canvas ConditioningConcat node created through panel_add_node acquired an image preview after a workflow run, even though ConditioningConcat has no image output or preview widget. The node expanded vertically from 153x46 to 153x266. This appears to be stale/misdirected frontend preview state attached to the wrong node.

### Steps to reproduce
1. On a live ComfyUI canvas, add a core ConditioningConcat node with panel_add_node.
2. Wire five CLIPTextEncode nodes through four ConditioningConcat nodes into KSampler positive.
3. Remove the replaced ConditioningCombine nodes.
4. Run the workflow manually.
5. Inspect the final ConditioningConcat with panel_query_graph fields=detail.

Initial panel_add_node result for node 73:
type=ConditioningConcat, size=[153,46], widgets={}

After the run:
type=ConditioningConcat, size=[153,266], widgets={"$$canvas-image-preview":""}

The final rendered image belongs to SaveImage node 9, but an image is visibly shown on ConditioningConcat node 73.

### Exact error
No thrown error. Wrong frontend state:
node 73 ConditioningConcat has unexpected $$canvas-image-preview and inflated height.

### Fix
Applied locally: no
Upstream-only: yes
The running panel/orchestrator source is not writable in this managed session. Please ensure execution previews are attached only to the node id that emitted the image, and clear stale node.imgs / preview pseudo-widget state for non-image-output nodes. A likely regression area is live-canvas node creation/removal followed by execution output routing.

### Environment
Windows 11
ComfyUI 0.28.3
ComfyUI frontend 1.45.21
NVIDIA RTX 4090 24 GB
Python 3.13.7
PyTorch 2.13.0+cu130
comfyui-mcp 0.51.56
panel 0.14.36

artokun:
## Summary

After a live-canvas `panel_add_node` + remove + run, ComfyUI keys `app.nodeOutputs` / `app.nodePreviewImages` by `String(node.id)` and `onDrawBackground` plants `$$canvas-image-preview` on **whatever node now holds that id**. A newly created ConditioningConcat (no IMAGE output) was growing from `[153,46]` to `[153,266]` and showing SaveImage's result.

The panel cannot stop ComfyUI writing the store. It now:

- wipes leftover store entries when a node is **added** (id reuse) or **removed**
- after `executed` / `execution_success`, **re-homes** stolen image outputs onto the emitting host and **strips** the pseudo-widget / `node.imgs` from every node that cannot emit an image (ConditioningConcat yes; SaveImage / PreviewImage / KSampler / LoadImage no)

## Test plan

- [x] `node --test browser_tests/unit/execution-preview-attach.test.mjs` — drives the shipped sanitizer (reporter's Concat vs SaveImage shape, re-home, KSampler kept, LoadImage imgs not wiped) and pins the add / remove / run call sites
- [x] add-node harnesses still pass (they inject the new free name)

Fixes #1286

### DIFF
--- a/browser_tests/unit/_panel-constants.mjs
+++ b/browser_tests/unit/_panel-constants.mjs
@@ -15,6 +15,7 @@ import { fileURLToPath } from "node:url";
 
 import { makeCommandBudget } from "../../web/js/lib/command-budget.js";
 import { REFRESH_JOIN_ABANDONED } from "../../web/js/lib/refresh-coalesce.js";
+import { clearInheritedExecutionPreview } from "../../web/js/lib/execution-preview-attach.js";
 
 export const PANEL_SRC = readFileSync(
   fileURLToPath(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url)),
@@ -134,6 +135,9 @@ export const OBJECT_INFO_SEED_WAIT_MS = readPanelNumber(
  */
 export function addNodeCommandBudgetDeps() {
   return {
+    // #1286 — graph_add_node now names this after graph.add. The shipped
+    // implementation is a no-op when the new id has no leftover store entries.
+    clearInheritedExecutionPreview,
     makeCommandBudget,
     ADD_NODE_COMMAND_BUDGET_MS,
     // Derived exactly as the panel derives it.
--- a/browser_tests/unit/add-node-command-budget.test.mjs
+++ b/browser_tests/unit/add-node-command-budget.test.mjs
@@ -48,6 +48,7 @@ import {
   monotonicNow,
   widenSocketProofBudget,
 } from "./_panel-constants.mjs";
+import { clearInheritedExecutionPreview } from "../../web/js/lib/execution-preview-attach.js";
 
 const panelPath = fileURLToPath(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url));
 const panelSrc = readFileSync(panelPath, "utf8");
@@ -237,6 +238,7 @@ function realGraphAddNode({
     makeCommandBudget,
     REFRESH_JOIN_ABANDONED,
     addNodeRefreshBusyMessage,
+    clearInheritedExecutionPreview,
     OBJECT_INFO_SEED_WAIT_MS: 8000,
     ADD_NODE_COMMAND_BUDGET_MS: budgetMs,
     ADD_NODE_POST_REFRESH_RESERVE_MS: reserveMs,
--- a/browser_tests/unit/execution-preview-attach.test.mjs
+++ b/browser_tests/unit/execution-preview-attach.test.mjs
@@ -1,0 +1,212 @@
+// #1286 — after a live-canvas add + run, ComfyUI plants $$canvas-image-preview
+// on a ConditioningConcat (or any non-image node) whose id received the
+// executed / b_preview store entry. These tests drive the SHIPPED helpers in
+// web/js/lib/execution-preview-attach.js and pin the panel call sites so the
+// wiring cannot be dropped.
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+
+import {
+  CANVAS_IMAGE_PREVIEW_WIDGET,
+  nodeAcceptsExecutionImagePreview,
+  clearStoredExecutionOutputs,
+  stripNodeExecutionPreview,
+  clearInheritedExecutionPreview,
+  stripMisattachedExecutionPreviews,
+} from "../../web/js/lib/execution-preview-attach.js";
+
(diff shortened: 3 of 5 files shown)


artokun:
Independent post-merge re-review (backlog-drain session was assigned #1286 and arrived
after this PR was already open; no competing fix was written). Everything below was
**executed** against the merged `origin/main` (8193b226), not read.

## Confirmed good

- **Reachable from production.** All three helpers have real call sites on main:
  `clearInheritedExecutionPreview` at the add path (12149), `clearStoredExecutionOutputs`
  at the remove path (12243), `stripMisattachedExecutionPreviews` at `executed` (32311)
  and `execution_success` (32379). The suite pins the call sites, not just the helper —
  which is the failure mode that nearly shipped six times this session. Good.
- **The predicate matches the real frontend.** `PREVIEW_HOST_TYPES` is exactly the
  frontend's `supportsVirtualCanvasImagePreview` set, and the mechanism in the header
  comment is the real one: in `comfyui_frontend_package` 1.48.7,
  `unsafeUpdatePreviews` does `getNodeOutputs(this)` → `$.nodeOutputs[locatorId(this)]`,
  then `this.imgs?.length && showCanvasImagePreview(this)`. Attachment is **by node id**,
  never by output slot and never by matching the executed prompt's outputs.
- **The reporter's case is genuinely cleaned.** Driving the shipped sanitizer with the
  reporter's shape (Concat 73 + SaveImage 9) strips imgs, the pseudo-widget and the store
  entry, and collapses 153x266 → 153x46. `node --test browser_tests/unit/execution-preview-attach.test.mjs` → 10/10.
- The pre-merge `removed: summary;` object-literal semicolon (which broke the whole ES
  module parse) is **not** on main — it was fixed before merge. Bundle re-parsed as an ES
  module on 8193b226: OK.

## Residual gap — the sweep is type-gated, so it only covers image-INCAPABLE victims

`nodeAcceptsExecutionImagePreview` answers *"could a node of this type ever show an
image?"*, not *"did THIS node emit these images in THIS run?"* Those are two different
questions, and the misattribution is an identity bug. Same scenario, same mechanism,
only the victim node's type changed:

```
added node type       | predicate | stolen preview stripped?
----------------------|-----------|-------------------------
ConditioningConcat    | false     | yes -> cleaned (size [210,46])   [stripped=1]
ConditioningCombine   | false     | yes -> cleaned (size [210,46])   [stripped=1]
CheckpointLoaderSimple| false     | yes -> cleaned (size [210,46])   [stripped=1]
VAEDecode             | true      | NO  -> still shows preview       [stripped=0]
ImageScale            | true      | NO  -> still shows preview       [stripped=0]
EmptyLatentImage      | true      | NO  -> still shows preview       [stripped=0]
```

Scope of that gap is narrower than it first looks, and worth stating precisely: the two
`clear*` calls close the id-reuse path **for adds and removes that go through
`panel_add_node` / `panel_remove_node`**, for every node type. So the type-gated sweep is
the only remaining defence for ids reused by a path the panel does not hook — ComfyUI's
own copy/paste, undo/redo, or a workflow load that resets `last_node_id`. On those paths
an image-capable victim (a freshly pasted `VAEDecode`) keeps the stolen preview.

The discriminator that would close it by identity rather than by type is already in hand
at the call site: `executed` carries `d.node ?? d.display_node`, so the set of ids that
legitimately received output this prompt is knowable without a `/history` round-trip. Any
*other* id holding store images is misattributed regardless of type. The one wrinkle is
`b_preview`, which writes `nodePreviewImages` for the *executing* node, so that set has to
be unioned in or latent previews get stripped.

Not asking for a revert — the shipped fix is a strict improvement and the reported symptom
is gone. Flagging it so the residual case is a known, decided gap rather than a surprise.

## Two hypotheses I raised and then refuted — recording them so nobody re-spends the time

- **"It over-strips promoted subgraph previews" (#1271/#1360 landed 17 min before this).**
  Unproven, and I now believe it is wrong. `SubgraphNode` extends `LGraphNode` directly and
  is not registered via `registerNodeDef`, so `constructor.nodeData` is undefined and the
  predicate does reject it — but `getNodeOutputs` keys on `locatorId(node)`, which for a
  root-level SubgraphNode is `String(node.id)`, and the store is only ever written for
  output nodes / the executing node. I could not construct a path where the frontend sets
  `node.imgs` on a root SubgraphNode, and `#1360`'s promotion goes through the
  `previewExposure` Pinia store rather than `node.widgets`, so `removePreviewWidget` cannot
  reach it either. Withdrawing it.
- **"It deletes the store entry of a legitimate custom OUTPUT_NODE."** That was an artifact
  of my own probe — I never seeded the key I then reported as deleted. Accepted nodes are
  `continue`d before any mutation. Withdrawn.

Issue #1357, opened 2026-08-19
panel_get_errors falsely flags uploaded LoadImage subfolder paths as missing after server-confirmed panel_set_widget


artokun:
### What happened / root cause
After upload_image registered two valid input assets under AgentLibrary/HaReen/, panel_refresh_nodes and panel_set_widget both succeeded with server_confirmed:true. Immediately afterward panel_get_errors reported both exact LoadImage widget values under unavailable_widget_values as missing_asset. The files existed on disk and the upload API returned those qualified filenames. Using root-level filenames removed the false errors. Likely path normalization/options comparison mismatch for nested input subfolders (slash/backslash or qualified combo values).

### Steps to reproduce
1. upload_image action:image with filename AgentLibrary/HaReen/Main-9-1.png.
2. panel_refresh_nodes.
3. panel_set_widget LoadImage.image to the returned qualified filename; observe server_confirmed:true.
4. panel_get_errors; observe unavailable_widget_values/missing_asset for the same value.
5. Re-upload/use root filename; error disappears.

### Exact error (scrubbed)
unavailable_widget_values: [{type: LoadImage, widget: image, value: AgentLibrary/HaReen/Main-9-1.png, kind: missing_asset}]
Contradicts panel_set_widget server_confirmed:true and the server file existing.

### Fix
applied locally: no; upstream-only: yes
Normalize nested LoadImage option paths and widget values to one canonical separator before membership comparison in panel_get_errors. Also reconcile the check with panel_set_widget server_confirmed so the same value cannot be simultaneously confirmed and reported missing.

### Environment
Windows 11; ComfyUI 0.32.0; frontend version not exposed in current session; NVIDIA RTX 5090 32GB; comfyui-mcp 0.52.3; panel 0.15.2.

artokun:
Fixed in 567e6438 (#1368), on `main`.

Your diagnosis of the symptom was exactly right, and the contradiction you spotted — `server_confirmed: true` and `missing_asset` on the same value in the same session — was the thread that unravelled it. The cause turned out to be slightly different from the separator mismatch you suspected, and worth stating because it changes what to expect:

[...]

Issue #1332, opened 2026-08-18
panel_get_errors stays stale after ComfyUI restart and newly registered custom nodes


artokun:
### What happened / root cause
After restoring the missing ComfyUI-Easy-Use custom node pack and restarting ComfyUI, panel_get_errors continued to report easy stylesSelector and easy showAnything as missing. The panel could then successfully add a fresh easy stylesSelector node using the newly refreshed /object_info, proving the nodes were registered, but panel_get_errors still returned the old missing-node list. panel_restart_comfyui also reported server_ready=true while panel_tab_reconnected=false, requiring manual workflow rebinding.

### Steps to reproduce
1. Open a workflow containing easy stylesSelector and easy showAnything while the pack is unavailable.
2. Install/restore the pack and restart ComfyUI.
3. Rebind the current workflow and call panel_get_errors.
4. Add a fresh easy stylesSelector node; add succeeds, but panel_get_errors still reports the old missing node types.

### Exact error (scrubbed)
panel_get_errors: missing_node_types=["GetImageSize+","easy showAnything","easy stylesSelector"] after the Easy-Use nodes were loadable by panel_add_node.
panel_restart_comfyui: server_ready=true, graph_tools_ready=false, panel_tab_reconnected=false.

### Fix
Applied local workflow workaround: verified Easy-Use registration by add/remove probe; replaced the unrelated legacy GetImageSize+ node with installed KJNodes GetImageSizeAndCount and rewired image/width/height. Panel workaround: panel_set_workflow_target(mode="current") restored graph access. No panel source patch applied; stale validation state remains upstream-only.

### Environment
OS Windows 11; ComfyUI 0.33.1; RTX 4070 Ti 12GB; comfyui-mcp 0.52.1; comfyui-mcp-panel 0.14.43; frontend version not exposed in the session.

artokun:
Claimed by grok-swarm 2026-08-19T07:16:49Z — driving this in worktree `/c/Users/Artokun/code/comfyui-mcp-panel/.worktrees/wt-1332` (branch `fix/1332-get-errors-stale-after-restart`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
[...]

Issue #1328, opened 2026-08-18
panel_auto_layout apply escapes subgraph and mutates root graph


artokun:
### What happened / root cause
While viewing subgraph node 121, panel_auto_layout with dry_run:true correctly targeted the 23-node subgraph. The immediately following panel_auto_layout with dry_run:false unexpectedly arranged the 20-node ROOT graph instead and returned root node ids. The canvas scope then became root, so panel_move_rail reported that it must run inside a subgraph.

The mutation path does not preserve/resolve the active subgraph scope consistently; the dry-run/read path does. The apply call can therefore mutate the wrong graph.

### Steps to reproduce
1. Open a workflow with a subgraph.
2. panel_enter_subgraph({node_id:121}).
3. panel_query_graph confirms scope=subgraph and 23 inner nodes.
4. panel_auto_layout({mode:"flow_horizontal",groups:"ignore",spacing:1.2,dry_run:true}) returns the 23 inner node ids.
5. panel_auto_layout with identical args except dry_run:false.
6. Observe it moves 20 root nodes (ids 64,65,66,...,121), not inner nodes.
7. panel_move_rail now errors because the canvas is at root.

### Exact error
panel_move_rail must run INSIDE a subgraph — call panel_enter_subgraph first

### Fix
Applied locally: no
Upstream-only: yes

Precise change needed: make the auto-layout mutation carry an explicit viewing-scope/subgraph-owner identity captured after panel_enter_subgraph, and resolve/apply against that graph instead of relying on mutable canvas scope during the mutation dispatch/preflight. At minimum, fail closed if the graph selected at apply time has a different owner/node count from the current subgraph scope. Add a regression test where dry_run and apply both target the same subgraph and the root node positions remain unchanged.

Temporary workaround: use panel_edit_node for inner-node positions, then panel_move_rail, avoiding panel_auto_layout apply inside subgraphs.

### Environment
Windows 11
ComfyUI 0.33.1
ComfyUI frontend 1.48.7
Python 3.12.10
PyTorch 2.13.0+cu130
GPU NVIDIA GeForce RTX 5090 32 GB
comfyui-mcp 0.52.1
panel 0.13.7

Related (same root cause): #220, #439

artokun:
[...]

Issue #1320, opened 2026-08-18
panel_update_node hides Manager update traceback


artokun:
### What happened / root cause
`panel_update_node` failed updating `seedvr2_videoupscaler` and returned only a generic Manager dialect-v2 error while directing the user to inspect server logs. The tool did not surface the promised full traceback, forcing a shell workaround.

### Steps to reproduce
Call `panel_update_node({id:'seedvr2_videoupscaler', version:'nightly'})` when the Manager update task fails.

### Exact error (scrubbed)
`Update of "seedvr2_videoupscaler" FAILED: the ComfyUI-Manager task terminated with an error (dialect v2): An error occurred while updating 'seedvr2_videoupscaler'.. The pack was NOT updated — check the ComfyUI server log for the full traceback.`

### Fix
Applied locally: no. Upstream-only: yes.
Return the Manager task traceback or a concise underlying error in `panel_update_node`.

### Environment
Windows 11; ComfyUI 0.31.0; frontend 1.48.7; RTX 4080 SUPER 16GB; comfyui-mcp 0.52.1; panel 0.15.1.

Related (same root cause): #364

artokun:
Claimed by grok-swarm 2026-08-19T07:00:42Z — driving this in worktree `/c/Users/Artokun/code/comfyui-mcp-panel/.worktrees/wt-1320` (branch `fix/1320-update-node-traceback`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
Fix is up: https://github.com/artokun/comfyui-mcp-panel/pull/1378

`panel_update_node` now reads `/internal/logs/raw` when Manager only stored the generic `An error occurred while updating 'X'.` sentence, and attaches the traceback (or the `res.action` ERROR line) to the tool error instead of pointing at the server log.

PR #1381, opened 2026-08-19
release: v0.15.4

artokun:
Patch release **v0.15.4** via `scripts/set-version.mjs`. Bumps pyproject + PANEL_VERSION, stamps CHANGELOG.

Merge (rebase or merge-commit, not squash) then the `v0.15.4` tag will be pushed. pyproject.toml change on main triggers publish_action.


artokun:
**Orchestrator note before merging.** This release branch was cut at 07:41Z and then sat green and untouched for ~45 minutes while `main` kept moving, so I finished it rather than let it go stale. Two commits added, no code touched:

**Seven fixes had landed after the cut and were shipping undocumented.** Resolving that turned up that four of them — #1328, #1332, #1310 and #1320 — had **already** been documented on `main` with better wording than mine, so I dropped my duplicates and kept yours. Only three were genuinely missing and are now recorded: **#1193** (abandoned combo refresh reported as a missing confirmation, with V2 and dynamic specs rebuilt), **#1272** (caller-supplied link-exclusion Set normalised), and **#1357** (a combo that cannot list its files is no authority on whether one exists).

The merge conflict was in `CHANGELOG.md` only; `web/js/comfyui-mcp-panel.js` auto-merged and `node --check` passes on the result.

Worth flagging for whoever owns the release script: if this section is generated from commit subjects, the generator ran before those seven landed, so the same gap will recur on any release that sits between cut and merge. Regenerating at merge time rather than cut time would close it.

This ships a lot, including two user-visible fixes worth calling out:

- **Save-As no longer writes the JSON literal `null` while reporting success** (#1267 half b). The copy came back from `saveAs` unloaded, `activeState` is `changeTracker?.activeState ?? null`, and `save()` serialises exactly that — while the post-write read-back compared the target against the copy's *own* content, so `"null" === "null"` returned `"ours"` and the check that existed to catch a bad write confirmed it.
- **`panel_connect` is judged on the link the live graph shows, not on the exception** (#1272). In 1.48.7 every throwable in `connectSlots` runs *after* the link is persisted, so a throw carried no information about whether the wire existed.

Issue #1272, opened 2026-08-15
panel_connect throws "Cannot read properties of undefined (reading 'slots')" but the link IS actually created


artokun:
### What happened / root cause

`panel_connect` returned a hard error, but the connection was **actually applied** to the graph. Re-reading with `panel_query_graph` right after each failure showed the link present and correct.

This is a false-negative: the tool reports failure for an operation that succeeded. An agent that trusts the error and retries will create duplicate links, or will tear down and rebuild wiring that was already correct.

It reproduced **twice in a row**, both while viewing inside a subgraph:

1. Connecting to a **dynamically-created input slot** on an `ImpactSwitch` node (Impact Pack). The node exposes `input1` initially and materialises `input2` when `input1` is connected. Connecting the second source to `input2` threw — but `input2.connected_from` was correctly set to the source node afterwards.
2. Connecting a node output to the **subgraph output rail** (`to_node_id: -20`, `to_input: "image"`). Threw the same error — but the source node's output showed `links: 1` and `panel_graph_outline` showed `143 ImpactSwitch → -20.0` correctly.

Both cases involve slot collections that are reshaped during/after the connect (dynamic input materialisation, boundary rail slots).

### Steps to reproduce

1. Open a workflow that contains a subgraph; `panel_enter_subgraph` into it.
2. Add an `ImpactSwitch` (`panel_add_node`, class_type `ImpactSwitch`).
3. `panel_connect` some IMAGE output → `input1`. **Succeeds.**
4. `panel_connect` a different IMAGE output → `input2` (the slot that only exists after step 3). **Throws** `Cannot read properties of undefined (reading 'slots')`.
5. `panel_query_graph {ids:[<switch id>], fields:'detail'}` → `input2.connected_from` is set correctly; the link exists.
6. `panel_connect` the switch's `selected_value` → the subgraph output rail (`to_node_id: -20`, `to_input: "image"`). **Throws** the same error; `panel_graph_outline` then shows the link present.

### Exact error (scrubbed)

```
Error: Cannot read properties of undefined (reading 'slots')
```

No stack was surfaced through the tool result.

### Investigation done

Searched for the `.slots` access on both sides; neither obvious candidate is in the connect path:

- `comfyui-agent-panel/web/js/**` — only one `.slots` read: `comfyui-mcp-panel.js:1738`, `for (const s of (d.slots || []))`, which is UI list rendering and is already guarded.
[...]

PR #1383, opened 2026-08-19
fix(1272): a caller-supplied exclusion Set is normalised, not trusted

artokun:
Closes a **latent** exclusion-normalisation leak in `web/js/lib/connect-verify.js`, found by an independent post-merge review of PR #1352 (merged as `2b665f44`). Nothing is broken in shipped behaviour today — no live call site passes a `Set` — but the defect is one edit away from re-causing the class of P0 that file just recovered from, and the suite could not see it.

## The leak

`linkIdExclusionSet` exists because the two halves of the "did THIS call add the link?" question need **opposite** id forms:

- membership must survive a number/string mismatch → ids are stringified;
- the store read must **not** be stringified → `Map.get("7")` misses a key of `7`.

Both finders then short-circuited that helper:

```js
const skip = excludeIds instanceof Set ? excludeIds : linkIdExclusionSet(excludeIds);
```

Membership is asked as `skip.has(String(linkId))`, so a caller's `Set` of **raw numbers** matches nothing. That is not an exotic shape — it is precisely what you get from `new Set(railSlot.linkIds)` or `new Set(inputLinkIds(node))`, because both id readers deliberately return **raw** ids (they have to; they feed the store). The exclusion evaporates silently and a link that existed **before** the mutation is credited to the current call — the "two states, one answer" defect the post-state check exists to remove.

## Reproduction (on `origin/main`, before the change)

The reviewer cited the rail finder; the inbound finder leaks identically:

```
findLandedRailLink(g,  {linkIds:[4]}, {id:9}, 0, "output", new Set([4]))
  -> { linkId: 4 }                    // must be null
findLandedInboundLink(g2, {id:1}, 0, target, new Set([5]))
  -> { linkId: 5, inputIndex: 0 }     // must be null
```

Both credit a **pre-existing** link to the call. Array and string-`Set` forms already behaved correctly, which is exactly why the fixtures never caught it.

## The fix

```js
const skip = linkIdExclusionSet(excludeIds);
```

on both paths. **`linkIdExclusionSet` needed no extension**: it iterates `ids ?? []`, so a `Set` was already an accepted input, and re-normalising a `Set` of strings is idempotent. The `instanceof Set` branch was a pure optimisation that bought nothing and cost the invariant. Its doc block now states that it takes any iterable and *why* a call site must not skip it, so the next reader does not reinstate the branch.

After: both calls above return `null`; array / string-`Set` / `null` forms are unchanged.

## The pin

The reviewer's key observation was that swapping the panel call sites to `new Set(...)` still passed **30/30** — the suite was blind. This PR adds a test that passes `new Set([4])` and `new Set([5])` to both finders and asserts the pre-existing link is **not** credited, with a no-exclusion pre-condition assertion so the test cannot pass by finding nothing at all.

Verified by reverting **only** the two-line fix (test kept):

```
ℹ tests 31   ℹ pass 30   ℹ fail 1
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  actual: { linkId: 4 },
  expected: null,
```

— the reviewer's reproduction verbatim. With the fix restored: 31/31.

## Verification

- `browser_tests/unit/connect-throw-verdict.test.mjs`: **31/31** (was 30/30).
- `npm run test:unit`: **5384 pass / 0 fail / 1 todo** — includes `i18n-check`, `i18n-render-check`, `check-tool-vocabulary` (OK, 34 literals across 176 shipped JS), `check-panel-scope` (OK — every name resolves).
- `tsc --noEmit`: clean.
- Rebased onto `df7821e6` and re-run; `929a4b0d..origin/main` touches neither file.
- **No Playwright coverage exists for these paths** — no spec references `expose_subgraph_*`. That is precisely why these unit end-to-end tests are the coverage that holds them, and why an unpinned helper here is worth a PR of its own.

Scope is two lines of logic plus the comment that explains them, and one test. No behaviour change for any current caller.

Not marked as closing any issue — #1272 is already closed and stays closed; this is follow-up hardening of the code that shipped under it.

### DIFF
--- a/browser_tests/unit/connect-throw-verdict.test.mjs
+++ b/browser_tests/unit/connect-throw-verdict.test.mjs
@@ -785,6 +785,49 @@ test("findLandedRailLink: excluded (pre-existing) ids are never credited to this
   assert.equal(findLandedRailLink(graph, rail, node, 0, "output", linkIdExclusionSet([4])), null);
 });
 
+// ---------------------------------------------------------------------------
+// A caller-supplied Set must be NORMALISED, not trusted.
+//
+// Both finders used to short-circuit on `excludeIds instanceof Set` and use the
+// caller's Set verbatim. Membership is asked as `skip.has(String(linkId))`, so a
+// Set of RAW NUMBERS — the shape any caller gets from `new Set(railSlot.linkIds)`
+// or `new Set(inputLinkIds(node))`, since both id readers deliberately return raw
+// ids — matched NOTHING: the exclusion evaporated and a link that existed BEFORE
+// the mutation was credited to the current call. That is the same string/number
+// key mismatch this file already recovered from once, arriving from the other
+// side; `linkIdExclusionSet` is now the unconditional gate on both paths.
+// ---------------------------------------------------------------------------
+
+test("both finders normalise a caller-supplied Set — a RAW-number Set still excludes", () => {
+  const rail = { linkIds: [4] };
+  const node = { id: 9 };
+  const railGraph = {
+    links: { 4: { id: 4, origin_id: 9, origin_slot: 0, target_id: -20, target_slot: 0 } },
+  };
+  // Pre-condition: link 4 IS on the rail and IS found when nothing is excluded.
+  assert.deepEqual(findLandedRailLink(railGraph, rail, node, 0, "output", null), { linkId: 4 });
+  // Excluding it as a raw-number Set must suppress it exactly as an array does.
+  assert.equal(findLandedRailLink(railGraph, rail, node, 0, "output", new Set([4])), null);
+  assert.equal(findLandedRailLink(railGraph, rail, node, 0, "output", new Set(["4"])), null);
+
+  // The inbound finder leaks identically without normalisation.
+  const inboundGraph = {
+    links: { 5: { id: 5, origin_id: 1, origin_slot: 0, target_id: 2, target_slot: 0 } },
+  };
+  const target = { id: 2, inputs: [{ name: "a", link: 5 }] };
+  const origin = { id: 1 };
+  assert.deepEqual(findLandedInboundLink(inboundGraph, origin, 0, target, null), {
+    linkId: 5,
+    inputIndex: 0,
+  });
+  assert.equal(findLandedInboundLink(inboundGraph, origin, 0, target, new Set([5])), null);
+  assert.equal(findLandedInboundLink(inboundGraph, origin, 0, target, new Set(["5"])), null);
+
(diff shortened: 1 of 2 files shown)


artokun:
**Merge record.** This closes a leak found by the independent post-merge review of #1352, and verified here before merging: the `excludeIds instanceof Set` branch is gone from both call sites (0 occurrences), both finders now go through `linkIdExclusionSet(excludeIds)` (2 occurrences), CI green. `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit; that substitution is disclosed as standing practice this session.

**Why this got its own PR rather than a note.** The severity turned out higher than the review estimated. It was reported as latent — "no live call site passes a Set" — but a raw-number Set is not an exotic input here: **`new Set(railSlot.linkIds)` and `new Set(inputLinkIds(node))` produce exactly that shape**, because both id readers deliberately return raw ids so they stay usable as store keys. The leaking shape is the one this file's own helpers hand you.

And the review cited only `findLandedRailLink`. Confirmed on unmodified `origin/main` that **`findLandedInboundLink` leaks identically**: `findLandedInboundLink(g, {id:1}, 0, target, new Set([5]))` returns `{linkId: 5, inputIndex: 0}` where `null` is required.

**No extension was needed.** `linkIdExclusionSet` iterates `ids ?? []`, so a `Set` was already accepted — `linkIdExclusionSet(new Set([4]))` returns `Set{"4"}` — and re-normalising a Set of strings is idempotent. The `instanceof Set` branch was a pure optimisation that bought nothing and cost the invariant. A doc paragraph now states the helper takes any iterable and why a call site must not skip it, so the branch does not get reinstated.

**Before/after:** reverting only the two-line fix while keeping the test gives `tests 31, pass 30, fail 1` with `actual: { linkId: 4 }, expected: null` — the reviewer's reproduction verbatim. Restored: 31/31. The new test covers both finders with `new Set([4])` / `new Set([5])`, includes a no-exclusion pre-condition assertion so it cannot pass by finding nothing, and exercises the helper accepting a Set directly.

`npm run test:unit` 5384 pass / 0 fail / 1 todo; `check-tool-vocabulary` and `check-panel-scope` OK; `tsc --noEmit` clean.

**Coverage stated plainly:** there is no Playwright coverage on this path — no spec references `expose_subgraph_*`. These unit end-to-end tests are the only thing holding it, which is exactly why an unpinned helper warranted its own change rather than a comment. Worth noting the panel-scope gate correctly *refused to run* in a fresh worktree until `npm install` rather than printing a false OK.

This is the same string/number key class that produced the P0 in #1352, one edit removed. #1272 stays closed.

PR #1387, opened 2026-08-19
fix(1358 followup): the load watch reports whether it was ENTERED, not just installed

artokun:
Corrective work on shipped code. An independent **post-merge** review of #1358 (already on `main` as `a6767ba9`) found two defects in it. Both are instances of class 1 of `.claude/autopilot/review-taxonomy.md` — *two distinct states collapse into one indistinguishable answer*.

No closing keywords: the family issues are already closed, and comfyui-mcp#1705 must stay open.

## F1 — #1358 reproduced its own defect, one level down

#1358 made the open's third ground depend on `loadRanToCompletion`, licensed strictly on `=== true`. That observation comes from two installed wrappers — `installNodeConfigureIsolation` and `installGraphConfigureWatch`. **Neither recorded whether its wrapper was ever *entered*.**

So `loadRestoreCompleted` could not tell apart:

- **nothing threw** — the load really did run to completion, and
- **we installed on a method nothing calls** — we observed nothing at all.

Wrapping a prototype method proves the method **exists**. It proves nothing about whether this frontend's restore went **through** it.

### The failing scenario

A frontend where the root graph stops resolving `configure` through `LGraph.prototype` — an own-property `configure` on the graph instance, or a subclass whose `configure` does not call `super`. Then:

- `installGraphConfigureWatch` **installs** (the prototype method is there) and is never **entered**;
- the restore aborts partway; `loadGraphData`'s own catch swallows it;
- `graphWatch.throws` is empty and `nodeIsolation.failures` is empty;
- `loadRestoreCompleted` answered **`true`**;
- `openContentDifferenceIsCompletedLoadNormalization` fires, and a genuinely partial load — nodes sitting at construction defaults with their `widgets_values` gone — is waved through as *normalization*.

That is precisely the harm #1358 exists to prevent, reintroduced by #1358's own observation gap. And no test could see it, because a test that drives the wrappers always enters them.

### The fix

Both handles now expose `entered`, counted **inside** the wrapper and **on entry** — a call that throws was still watched, and a counter bumped on the way out would read 0 for exactly the aborted load the watch exists to see. A deactivated wrapper is a pass-through and counts nothing.

`loadRestoreCompleted` returns **`null`** — not `false` — when `graphWatch.entered` is 0, absent, or not a number. (`!(x >= 1)`, not `x < 1`: `NaN` fails every comparison, so `< 1` would have let it through as "entered" — the same fold hiding in an operator. There is a test.)

`null` is load-bearing, and the consumer contract was verified before relying on it:

| consumer | its test | effect of `null` |
|---|---|---|
| `openContentDifferenceIsCompletedLoadNormalization` | `loadRanToCompletion !== true` → `false` | the third ground refuses |
| `graphRootReproducesStateContent`'s `presentationOnly` veto | `loadRanToCompletion !== false` | **not** vetoed |
| `describeOpenRebindOutcome` | already distinguishes null from false | no completion sentence |

So `null` degrades to **exactly** pre-#1358 behaviour — byte-identical to a caller that never asked the question — rather than to a new refusal #1358 was never entitled to introduce. `false` still vetoes the weaker #1623 ground; `null` must not borrow that, and a test asserts the difference.

### Only `graphWatch.entered` gates the verdict

The reviewer specified `graphWatch.entered` only. Both handles get the counter (the node one as a **diagnostic**), but **only the graph count vetoes**:

1. Zero node configures is a **legitimate** completed restore — an empty workflow. Gating on `nodeIsolation.entered` would answer UNKNOWN for loads that demonstrably ran to the end: a false negative manufactured by the guard, which is class 5. There is a test for that exact case.
2. It is not needed. The node wrapper is a **containment**; the graph watch is the backstop underneath it. A node whose `configure` never routes through `LGraphNode.prototype` still throws into `LGraph.prototype.configure`'s node loop, which has no try/catch of its own — so the abort escapes into the graph watch and **is** recorded. An unentered node wrapper loses containment for that node; it does not lose the **observation**. An unentered graph watch loses the observation outright.

### The test is the point

The new fixture is a frontend whose root graph carries its **own** `configure` and never calls `super`. The node loop still dispatches through `LGraphNode.prototype`, so `nodeIsolation.entered === 1` — **a node-level counter would not have caught this** — while `graphWatch.entered === 0`.

The same test carries its own refutation: with the answer #1358 shipped (`loadRanToCompletion: true`) the lost `widgets_values` is credited as `normalizedOnly` with `normalizedFields: ["widgets_values"]`; with the `null` this PR produces, it is not.

### Refutation

Every mutation was confirmed applied **on disk** before its run (these files are CRLF; `$`-anchored substitutions silently no-op).

| mutation | result |
|---|---|
| delete the whole fix (`git show origin/main:` over the file) | **7 tests fail** |
| delete the `entered` gate in `loadRestoreCompleted` | **KILLED** — 3 tests |
| delete `entered += 1` from the graph wrapper | **KILLED** |
| drop `entered` from the graph handle | **KILLED** |
| count on the way *out* instead of on entry | **KILLED** |
| `< 1` instead of `!(>= 1)` (NaN slips through) | **KILLED** |
| drop `entered` from the *node* handle | **KILLED** — and only the two **diagnostic** assertions, not the fold's verdict, which is the documented design |

### Measured on the rig, not read

The one thing source-reading cannot settle is whether the **real** frontend enters the watch — if it did not, this guard would turn every `workflow_open`'s third ground into UNKNOWN. Probed against a live ComfyUI (frontend `1.48.7`) by wrapping the prototypes in-page and running a real `app.loadGraphData` on a 2-node payload:

```json
{ "rootCtor": "LGraph", "rootResolvesViaLGraphPrototype": true,
  "graphEntered": 1, "nodeEntered": 2 }
```

The watch is entered, so the verdict is unchanged in production. The guard fires only on the frontend shape it was written for.

## F2 — a doc comment on `main` that was false

`resolveOpenRebindVerdict` still told a reader that the softening "was reverted, because the two cases cannot currently be told apart … no discriminator available to the panel separates them", and that "`unknown` stays `unknown` here — deliberately".

#1358 built exactly that discriminator, so the comment contradicted the code beside it. A future agent reading "no discriminator exists" would not go looking for the one that now does — and a false mechanism comment has twice sent work in this repo down the wrong path.

It now points at `loadRestoreCompleted` / `openContentDifferenceIsCompletedLoadNormalization` and states the discriminator in one line: **a mid-`configure()` abort is a THROW, and a load whose throws the panel watched for and did not see did not abort.** It also keeps the honest note that `resolveOpenRebindVerdict` itself is unchanged — the discriminator is applied upstream, where the content proof is computed.

Four other sites quoted that sentence in the **present** tense (`graph-binding.js` ×2, `comfyui-mcp-panel.js`, and the test file's own header). Leaving them would have manufactured the same class-6 defect this half of the PR removes, so they now read as the pre-#1358 wording.

## Verification

- `npm run test:unit` — **5429 pass / 0 fail / 1 todo** (5430 tests) (includes `i18n-check`, `i18n-render-check`, `check-tool-vocabulary`, `check-panel-scope`)
- `browser_tests/unit/open-completed-restore.test.mjs` — **36 pass** (30 before, 6 new)
- Playwright, `--workers=1`, against a dedicated ComfyUI on `:8299` with its own base directory, same specs both sides, only the served panel swapped — **zero regressions.** Baseline (`origin/main` panel): 56 passed / 19 failed / 75, 8.5m. With the fix: 57 passed / 18 failed / 75, 8.3m. Set difference in the regression direction is **empty**; the one moving spec (`agent-drive.spec.ts:113`, CivitAI modal highlight timing) failed on the baseline and passed with the fix — a flake, unrelated to this change. The 18 shared failures are pre-existing on `origin/main` in this environment (bridge / conversation-persistence specs on a fresh instance). `tsc --noEmit` clean.

No behaviour change beyond the verdict: the wrappers still contain / re-throw exactly as before, `restore()` is untouched, and `entered` is additive on both handles.

CHANGELOG deliberately not edited — the release PR (#1381) generates its section from commit subjects, and hand-editing `[Unreleased]` would only conflict with it.

### DIFF
--- a/browser_tests/unit/open-completed-restore.test.mjs
+++ b/browser_tests/unit/open-completed-restore.test.mjs
@@ -398,12 +398,14 @@ test("panel#1283 an uninstrumentable frontend answers null, never false", () => 
 });
 
 test("panel#1283 the fold is true only when BOTH halves looked and neither saw a throw", () => {
-  assert.equal(loadRestoreCompleted({ nodeIsolation: { failures: [] }, graphWatch: { throws: [] } }), true);
+  // `entered: 1` is part of "looked" — see the F1 section at the foot of this file.
+  const watched = (throws) => ({ throws, entered: 1 });
+  assert.equal(loadRestoreCompleted({ nodeIsolation: { failures: [] }, graphWatch: watched([]) }), true);
   assert.equal(
-    loadRestoreCompleted({ nodeIsolation: { failures: [{ id: 3 }] }, graphWatch: { throws: [] } }),
+    loadRestoreCompleted({ nodeIsolation: { failures: [{ id: 3 }] }, graphWatch: watched([]) }),
     false,
   );
-  assert.equal(loadRestoreCompleted({ nodeIsolation: { failures: [] }, graphWatch: { throws: ["x"] } }), false);
+  assert.equal(loadRestoreCompleted({ nodeIsolation: { failures: [] }, graphWatch: watched(["x"]) }), false);
 });
 
 test("panel#1283 both wraps compose: a node throw is contained, the graph throw is not", () => {
@@ -413,6 +415,10 @@ test("panel#1283 both wraps compose: a node throw is contained, the graph throw 
   };
   const nodeIsolation = installNodeConfigureIsolation(LG);
   const graphWatch = installGraphConfigureWatch(LG);
+  // The graph restore runs — so the watch is ENTERED and this is not the un-watched
+  // case the F1 section below is about.
+  assert.equal(LG.LGraph.prototype.configure({ nodes: [] }), "graph-ok");
+  assert.equal(graphWatch.entered, 1);
   // The node throw is swallowed and RECORDED (#1260's contract, unchanged)…
   assert.equal(LG.LGraphNode.prototype.configure({ id: 7, type: "FaceDetailer" }), undefined);
   assert.equal(nodeIsolation.failures.length, 1);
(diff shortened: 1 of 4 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**. Verdict: **SHIP**, head `5d6f82a6` (matches the reviewed SHA).

This corrects a fail-open one level below the bug #1358 fixed: `loadRanToCompletion` is licensed strictly on `=== true`, but neither wrapper recorded whether it was ever *entered*, so `loadRestoreCompleted` could not tell "nothing threw" from "we installed on a method nothing calls" — and no test could see it, because in the test environment the wrapper is always entered.

**The reachability proof was checked rather than taken.** The gate refused to trust the handle's own getter and installed its **own counter outermost** on `LG.LGraph.prototype.configure`, after the watch: it recorded **0 dispatches** while the node wrapper recorded 1, node 2 sat at `["construction-default"]`, both records were empty, and the fold returned `null`. The counter is live rather than hard-wired — a direct prototype call flipped `entered` to 1 and the fold to `true` in the same run. The fixture also carries a tripwire: its `LGraph.prototype.configure` throws "the fixture is wrong", so a reached watch would populate `throws` and fail the test.

**Both `null` contracts verified at source and empirically.** `openContentDifferenceIsCompletedLoadNormalization` gates on `!== true` (null refuses); #1623's veto is `!== false` (null not vetoed); the only third consumer, `abortedRestoreClause` plus the disclosure headline, tests `!== false` / `=== false` (null stays silent). No truthiness test anywhere. A parity probe compared the **whole returned object** for `null` versus *absent* across five difference shapes — identical, cosmetic-only, widgets-lost, link-surface, node-set-changed — and they matched on every one, while `false` still suppresses `presentationOnly` on the cosmetic shape. So `null` really is byte-identical to never asking.

**Ordinary opens are provably unaffected**, which was the un-shipping risk: `app.rootGraph` is a plain `new LGraph()` and `this.rootGraph.configure(graphData)` runs **synchronously** inside the awaited `loadGraphData`, between the panel's install and its `finally { graphWatch?.restore() }`. So `graphEntered >= 1` in production by construction, independent of any single frontend version.

**The escape question was answered from real source, at a newer version than the PR measured.** Extracted from the shipped sourcemaps at **1.50.3**: the node pass is `node?.configure(nodeData)` with no try/catch and the enclosing `try` is `finally`-only, so a throw propagates into the watch — reproduced with a rogue node bypassing `LGraphNode.prototype` (`graphThrows: 1`, fold `false`). Two shapes do escape both wrappers — a frontend wrapping its node loop in its own `catch`, and an async/rejecting node `configure` — but the real `LGraphNode.configure(info): void` is synchronous and the real loop has no catch, and **both are pre-existing in #1358 on `main`; this PR neither introduces nor widens them.**

**Mutations: every claim reproduced**, each applied by literal CRLF-safe string replace and proven on disk with byte deltas before reverting. Full-fix revert 7 fail; delete the gate 3; delete `entered += 1` 4; drop `entered` from the graph handle 5; count on the way *out* 1; `< 1` instead of `!(>= 1)` 1 — the NaN row is the sole guard, and it bites; drop `entered` from the *node* handle 2, both being `nodeIsolation.entered` diagnostics rather than the fold's verdict, exactly as designed.

The node-level counter deliberately does not gate, and that reasoning holds: zero node configures is a legitimate completed restore (an empty workflow), so gating on it would manufacture a false UNKNOWN.

5430 tests, 5429 pass / 0 fail / 1 todo; target file 36/36; `tsc --noEmit` clean.

Two non-blocking nits, both erring toward the code being stricter than its prose: `comfyui-mcp-panel.js:16643` still enumerates null as "when either wrap could not be installed", omitting the not-entered case it was written alongside; and `resolveOpenRebindVerdict`'s new note says `false` and `null` "both keep the old, refusing path" when `false` is strictly *more* refusing, since it also vetoes #1623's `presentationOnly`.

PR #1388, opened 2026-08-19
fix(1706): a subgraph node-id renumber is a relabeling the panel can PROVE, not lost content

artokun:
`panel_open_workflow` returned a hard ERROR — "Treat the canvas as UNKNOWN… differs on: definitions" — for an open that had applied perfectly: identity confirmed, 26 nodes matching, `definitions` the only differing surface. This addresses the underlying cause of artokun/comfyui-mcp#1706.

The mechanism was handed over by the #1358 session, which correctly split this out as a distinct bug and left a test pinning that it still refuses. **The handed-over diagnosis was half right, and the half that was wrong matters** — so it is stated plainly below rather than repeated.

## What was handed over, and what measurement said

> `LGraph.configure` calls `deduplicateSubgraphNodeIds(...)` with `reservedNodeIds` seeded from **definitions already in `this.subgraphs`** — the *previous* workflow's leftovers. So opening A while B's identically-keyed definitions still sit in the store renumbers node ids inside A's subgraph definitions.

**The rewrite is real and is exactly as described. The stated PRECONDITION is not.**

`LGraph.prototype.configure` starts with `r.clearGraph && this.clear()`, and `LGraph.clear()` runs `this._subgraphs.clear()`. `loadGraphData` calls `configure(graphData)` with no second argument, so `clearGraph` is true and the store is EMPTY by the time `reservedNodeIds` is seeded. Nothing in the bundle listens for `configuring` to set `clearGraph` false. The previous workflow's definitions cannot contribute.

What actually seeds the collision is in the payload itself:

```js
for (const t of this._nodes)               // [] — clear() just ran
for (const sg of this.subgraphs.values())  // empty — clear() just ran
for (const t of i ?? []) if (typeof t.id === "number") e.add(t.id)   // <- the payload's OWN root node ids
```

plus the ids consumed by EARLIER definitions in the same load — six definitions each numbering their nodes 76/77/78 collide with each other. `deduplicateSubgraphNodeIds` then walks each definition's `nodes` in order, allocates fresh ids from `state.lastNodeId + 1`, and patches that definition's `links` and promoted `widgets` through the same map.

**Measured on the live rig** (ComfyUI 0.33.1, comfyui-frontend-package 1.48.7, a real browser, payload vs `app.graph.serialize()`): the fresh open of a template, and the open of the same template *after* switching through a workflow carrying identically-keyed definitions, produce a **byte-identical** difference. So "after switching" in the title is the reporter's narrative, not a precondition. What is required is a payload whose definition node ids collide — a property of the file alone.

Two more results, because they bound the blast radius:

- A workflow **the frontend itself serialized reopens BYTE-IDENTICALLY** — 0 differing paths in `definitions`, 0 in `nodes`, fresh and after switching. A workflow that has already been through this frontend never triggers this; there is no baseline drift being papered over.
- Take that same round-tripped workflow, force its definition node ids to collide, reopen. Root `nodes`: still **0** differing paths. `definitions`: differs, and the **whole** of the difference is

  | path | what it is |
  | --- | --- |
  | `/subgraphs/#/nodes/#/id` | the relabeling |
  | `/subgraphs/#/links/#/origin_id`, `/target_id` | patched through the same map |
  | `/subgraphs/#/nodes/#/order` | LiteGraph's recomputed execution index |
  | `/subgraphs/#/state/lastNodeId` 196 → 214 | the counter that allocated them |

  and nothing else — the "everything else" bucket came back **empty** on both templates measured (`video_wan2_2_14B_s2v.json`, `templates-6-key-frames.json`).

## Yes, `rootNodes` is rewritten too — and that variant is NOT confined to `definitions`

The handover warned that `deduplicateSubgraphNodeIds` can return rewritten `rootNodes`. It does, and the warning was worth heeding. Reading the function, the only thing it does to `rootNodes` is `patchProxyWidgets` — a root subgraph-instance node records which definition node each promoted widget comes from, as `properties.proxyWidgets: [[definitionNodeId, widgetName], …]`.

Measured on `templates-6-key-frames.json` (5 root nodes promoting widgets, definition ids forced to collide): the root `nodes` surface came back differing on **10 paths — `inputs/#/widget` and `widgets_values/#`**. Promoted widget VALUES were gone. That is real loss, not normalization.

It would still refuse on the `nodes` surface by itself. But accounting for `definitions` takes it out of the UNEXPLAINED set that #1588/#1283's weaker completed-load ground reads — and that ground *does* admit a `widgets_values` difference. So the new account **refuses outright** whenever a root node promotes a widget from a node the relabeling touched. That is why the predicate now takes the payload's root nodes at all.

## The fix

`definitions-renumber.js` gains a second account and is renamed `definitionsDifferOnlyByRenumber` — the old name asserted "link", which would now be false, and this file's prose is part of the guard.

It is **proven, not tolerated**. No tolerance was widened and no field was allowlisted. A `definitions` difference is admitted only when a single **injective** map carries the payload's definition onto the live one:

- same node count, same node **order**, same `type` at every position — positional, because that is what `remapNodeIds` does (it mutates ids in place), and because matching by id is impossible when the ids are the thing that moved;
- every other field of every node **deep-equal**;
- every link endpoint equal **after** the map is applied — the wiring is compared *through* the relabeling, as a multiset of endpoint signatures, so a re-wire cannot hide in it;
- every promoted widget's `id` equal after the map, everything else about it deep-equal;
- `state.lastNodeId` may move, only **forward**, and only when a relabeling was actually found.

**The account is granted only to a caller that supplies `rootNodes`;** a call without it answers the pre-#1706 question and refuses. Stated precisely, because this is easy to oversell: at both live call sites `state.nodes` is guaranteed to be an array (`buildGraphShape` returns null otherwise and `comparable !== true` returns early), so that half of the gate does not fire in production. What it buys is that the WIRING is assertable — dropping the argument at either call site un-ships the fix instead of silently widening it, and both sites are pinned by shape *and* by count, so a third unpinned call site fails the test too. The half that does fire in production is the `proxyWidgets` refusal above.

`state.lastNodeId` and `order` are decided over the whole `definitions` block, in two passes, not per definition — the serialized subgraph `state` is the ROOT graph's counter (measured: all six definitions reported the same `196 → 214`), so a definition that was *not* relabeled still shows it moving as soon as another one was.

## What still produces a genuine mismatch verdict

Every one has a test, and each is built against the **real captured** payload/live pair where the shape allows:

- a definition node that **vanished** or **appeared**
- a **retyped** definition node
- a changed **widget value** inside a relabeled definition
- a **re-wired** link (same ids, different slot); an **added** or **removed** link
- a link endpoint that did **not** follow the map — ids moved, wiring did not: a broken graph, not a renamed one
- a **renamed or retyped interface port** (`inputs`/`outputs` are still never waived)
- a node that **moved** — geometry is untouched by a relabeling
- **reordered** nodes inside a relabeled definition
- two payload nodes **collapsing** onto one live id — a lost node wearing a relabeling's clothes
- an id whose only change is its **type** (`78` → `"78"`)
- `state.lastNodeId` moving **backward** or becoming non-numeric; `lastGroupId` moving at all
- **without** a relabeling, `lastNodeId` and `order` are refusals exactly as before — the wider allowances are bought by the relabeling, not granted by default
- a root node promoting a widget from a relabeled node, or an **unreadable** `proxyWidgets` shape
- every surface other than `nodes`/`definitions`, and every hostile / cyclic / throwing shape #886 already pinned

## Two tests were RE-ARMED, not relaxed

`open-completed-restore.test.mjs`'s `comfyui-mcp#1706 a definitions difference that is NOT renumbering still refuses` asserted that a lone definition node id moving `9 → 41` refuses. That was the honest answer while the rewrite had no characterisation; it has one now, and that exact input **is** the rewrite. The test is split in two: the relabeling is pinned as accounted, and a definitions difference that is not one (the same relabeling plus a **retyped** node) is pinned as still refusing. The half the original assertion was protecting is unchanged. That file's header comment is corrected the same way.

`open-subgraph-definitions-disclosure.test.mjs` asserted the reply says `link RENUMBERING`. With two measured rewrites on this surface and a predicate that does not report which one it matched, naming one would state a mechanism the reply never observed. The reply now says what the predicate actually PROVED, and the test asserts more than it did before: renumbering, **both** kinds of id, and "same nodes in the same order".

## Refutation

Baseline for the three affected files: **98/98**. Ten mutations, each confirmed applied by `git diff --numstat` before its run — these files are CRLF, and a never-applied mutation looks exactly like a surviving one. Restored afterwards: **98/98**.

| mutation | verdict |
| --- | --- |
| M1 content-proof call site drops `rootNodes` | killed (2) |
| M2 disclosure call site drops `rootNodes` | killed (1) |
| M3 the node-id account is reverted | killed (7) |
| M4 the proxyWidgets cross-surface guard never fires | killed (3) |
| M5 the allocation counter need not advance | killed (1) |
| M6 link endpoints not compared through the map | killed (6) |
| M7 the map need not be injective | killed (1) |
| M8 a node may be retyped at its position | **SURVIVED** |
| M9 the wider allowances granted without a relabeling | killed (6) |
| M10 an id that only changed TYPE reads as a non-move | killed (1) |

Three survived the first pass, and re-reading the diff found a fourth hole. Three of those four were defects and were fixed rather than papered over:

- **a tautological check.** `if (!deepEqual(map?.get(String(na.id)), nb.id)) return false;` reads like a guard and asserts nothing — the map was built from the same positional pairs, so it holds by construction. Deleted, with a comment saying why the id needs no separate check.
- **a test that refused for the wrong reason.** The captured-fixture "collapse" case still named the old second id in its links, so removing the injectivity check killed no test. Replaced with a case where the collapse is consistent throughout: the only thing then standing between it and a "relabeling" verdict is injectivity, and what injectivity stands in front of is a definition that lost a node.
- **an unchecked id dialect** (M10), found by adversarial re-review of my own diff rather than by any mutation. The map is keyed by `String(id)`, so `78 → "78"` reads as the identity — and once ANY definition in the block relabels, `id` is an allowed field on EVERY definition, so a definition whose ids changed only in TYPE went completely unchecked. That is a widening the pre-#1706 code did not have. It now refuses, with its own test.
- **M8 survives and is left surviving, deliberately.** `na.type !== nb.type` overlaps `differsOnlyIn`, which compares `type` on every pair a few lines later and refuses first, so no test can kill it alone. It stays because `buildNodeIdMap`'s answer feeds `anyRelabel` in pass 1, which widens what EVERY OTHER definition may differ in — a later change to the field comparison must not silently take that with it. The reason is written at the line. (The *other* half of that check, requiring a type at all, is load-bearing and is killed by #886's existing test.)

### End-to-end, on the real captured pair

`describeGraphStateDifference` + `graphRootReproducesStateContent` + `resolveOpenRebindVerdict`, driven with the actual payload and the actual `app.graph.serialize()` output captured from the browser:

| | `surfaces` | `accountedSurfaces` | `proven` | verdict |
| --- | --- | --- | --- | --- |
| `origin/main` | `["definitions"]` | `[]` | false | **content-unverified** |
| this branch | `["definitions"]` | `["definitions"]` | **true** | **proven** |

That is the reported bug reproduced and then removed against real data, not against a fixture written to match my own reasoning.

## Verification

**Unit** — `node --test browser_tests/unit/*.test.mjs`: **5436 pass, 1 fail** — that one is the known `#671 verifyInstalled` wall-clock flake, which fires under machine load; the file is **125/125** when run alone, and an earlier run of the whole suite on this same tree was 5435/0.

**Panel gates** — `i18n-check`, `i18n-render-check`, `check-tool-vocabulary`, `check-panel-scope`, `tsc --noEmit`: all OK.

**Browser (Playwright, `--workers=1`)** — 
|  | failed | passed | wall |
|---|---|---|---|
| `origin/main` (e27a78a5), its own dedicated ComfyUI on :8282 | 18 | 57 | 8.1m |
| this branch, its own dedicated ComfyUI on :8281 | **18** | **57** | 8.3m |

**Zero regressions**: the failure list on this branch is a subset of the baseline's — `comm` in the "only on my branch" direction is empty.

Both sides were run sequentially, `--workers=1`, each against its OWN ComfyUI instance whose `custom_nodes` is junctioned to that worktree, with the SERVED bundle verified to carry (or not carry) the change before the run: `/extensions/comfyui-mcp-panel/js/lib/graph-binding.js` greps 2 hits for `rootNodes: state?.nodes` on :8281 and 0 on :8282.

The baseline failures are the known pre-existing set (chat-history / conversation-persistence / workflow-chat-identity / streaming / hidden-tab / pending-queue / external-orchestrator) — the same set #1358 reported, and they fail identically without this change.

## Gate

`codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the weekly limit, so **no independent gate was run by this agent** — per instruction, the orchestrator gates and merges. GitHub's own `Running Copilot Code Review` workflow also failed on this PR, with `errorType: quota, statusCode: 402 — You have exceeded your monthly quota`. That is an account limit, not a finding about this change; the `CI` check (lint pack: JS syntax + Python + pyproject) is green.

The taxonomy sweep above is the author's own review and is explicitly not a substitute for it.


artokun:
Two of the three gate findings landed after this merged, so they are in a follow-up: **#1395**.

- **Item 1 (the `String(entry[0])` key)** — fixed there. The guard now asks the question in every reading that can name a node that actually moved: the raw string, `Number(key)` (the only clause that catches `"0x4E"` / `"7.8e1"`), and a leading-integer parse (the only clause that catches `"78abc"`, where `Number()` is NaN). A blanket "refuse every non-canonical id" was written first and removed — it refuses `"999999.0"`, which names nothing that moved, and there is a test pinning that direction. Both new clauses are mutation-killed; the first attempt at M11 SURVIVED and that was a real finding about the test, not the code.
- **Item 2 (CHANGELOG under `## [Unreleased]`)** — already correct in this merge; nothing needed.
- **Item 3 (`node-inputs-rebuild.js:36` naming the old export)** — fixed in #1395. Grepped tree-wide; it was the only remaining site.

Also recording, since it is easy to misread the record: the rebased head that merged here (`3824b624`) was verified content-identical to the gated tree (`cc1063b5`) file by file — every added/removed line in all seven code and test files matched exactly. `git patch-id --stable` did **not** match (`9de0b28d` vs `5dfad4f4`), and that is expected rather than reassuring-by-default: patch-id hashes context lines, and #1387 added 213 lines to `open-completed-restore.test.mjs`, while the CHANGELOG entry legitimately gained a `### Fixed` heading because v0.15.4 emptied `[Unreleased]`.

PR #1392, opened 2026-08-19
fix(1374): sweep a stolen preview by which node emitted, not by what its type could show

### LINKED ISSUE #1374: Preview sweep gates on node TYPE, not which id emitted, so image-capable types keep a stolen preview
Split out of #1286, which is closed by PR #1361. That fix is sound; this is a gap it leaves, found by an independent verification against the merged module and demonstrated by varying one variable.

**Attachment is by node id.** In `comfyui_frontend_package` 1.48.7, `unsafeUpdatePreviews` does `getNodeOutputs(this)` → `$.nodeOutputs[locatorId(this)]` → `this.imgs?.length && showCanvasImagePreview(this)`. No slot and no prompt-output matching is involved, so a reused id inherits the previous occupant's images. That is why #1286 put an image preview on `ConditioningConcat`.

**The gap.** The post-run sweep gates on node *type* — "could this type ever show an image?" — rather than on which id actually emitted. Holding the scenario fixed and varying only the victim's type:

| victim type | stolen preview cleaned? |
|---|---|
| `ConditioningConcat` | yes |
| `ConditioningCombine` | yes |
| `CheckpointLoaderSimple` | yes |
| `VAEDecode` | **no** |
| `ImageScale` | **no** |
| `EmptyLatentImage` | **no** |

A type that *could* legitimately show an image is exempted from the sweep, so when it inherits a reused id it keeps the previous node's preview — the original symptom, on a different set of nodes.

**Narrower than it looks, which is why this is P3 and not a reopen.** #1361's two `clear*` calls wipe the store at the add and remove call sites, closing id reuse for *all* types on the `panel_add_node` / `panel_remove_node` paths. So this bites only on paths the panel does not hook — ComfyUI's own paste or undo, or a load that resets `last_node_id`.

**The honest fix** is to key the sweep on which id actually emitted during the run rather than on what a type is capable of, so "this node produced no image this run" and "this node type never produces images" stop being one answer.

Two hypotheses were raised against #1361 and then killed; recorded here so nobody re-spends the time: it does **not** over-strip promoted subgraph previews (no path was constructible where the frontend sets `imgs` on a root `SubgraphNode`, and #1360 promotes via the `previewExposure` store rather than `node.widgets`), and the "deletes a legitimate OUTPUT_NODE's store entry" claim was an artifact of a probe that never seeded the key it reported deleting.

Not observed in the wild — derived by construction against the merged code.

artokun:
## Summary

`stripMisattachedExecutionPreviews` (shipped by #1361 for #1286) decided who may keep an
execution preview by asking **"could this node type ever show an image?"**. ComfyUI attaches
previews purely by node **id** — `unsafeUpdatePreviews` → `getNodeOutputs(this)` →
`$.nodeOutputs[locatorId(this)]` → `this.imgs?.length && showCanvasImagePreview(this)`; no slot,
no prompt-output matching. So an id reused by an *image-capable* node inherits the previous
occupant's images and the type gate waves it through. Holding #1286's scenario fixed and varying
only the victim's type, `ConditioningConcat` / `ConditioningCombine` / `CheckpointLoaderSimple`
were cleaned but `VAEDecode` / `ImageScale` / `EmptyLatentImage` kept the stolen preview.

A type can never answer "did *this node* emit?", so the sweep now also carries a small ownership
ledger. Every `executed` frame names an id (`preferNodeId`, already passed by `onExecuted`); we
remember which node **object** held that id and the exact `nodeOutputs` entry it received. A later
occupant of that id is judged stolen only on positive evidence, and is otherwise left to the
existing type gate:

| evidence | verdict |
|---|---|
| same entry object, different node object, **different type** | stolen → evicted |
| no emission ever recorded for that id | unknown → type gate (unchanged) |
| entry object was **replaced** (new run, or `app.nodeOutputs` repopulated from `/history`) | unknown → type gate |
| different node object but **same type** (`Fix node (recreate)`, ChangeTracker undo) | not stolen — ComfyUI means to re-plant that preview from the surviving entry |

Because this **widens** who the sweep looks at, it deliberately **narrows** what it does to them.
An image-capable host owns image state of its own — a `LoadImage` hydrates `node.imgs` from its
filename widget and carries the same `$$canvas-image-preview` widget — and none of that belongs to
the inherited entry. `unsafeUpdatePreviews` assigns `this.images = output.images` *by reference*,
so `node.images === record.output.images` decides precisely whether what is on screen came from
the stolen entry. When it did, the derived state goes; when it did not, only the leftover store
entry is dropped and the node keeps its own thumbnail and its own size.

Scope is unchanged from the issue's assessment: #1361's `clear*` calls already close id reuse for
all types on the `panel_add_node` / `panel_remove_node` paths, so this only bites on paths the
panel does not hook (ComfyUI's own paste/undo, or a load that resets `last_node_id`). Derived by
construction; not observed in the wild.

**Known gap, deliberately not closed here** (named in the module header): ownership is seeded from
`executed`, and `onExecuted` returns early unless a frame carried media — so a node whose only
output is `b_preview` latent frames (a KSampler mid-run) never enters the ledger and is still
judged by type alone. Seeding it would mean attributing ownership on `executing` — "started"
rather than "emitted" — which is a different rule than this issue asks for.

**No panel source change**: `onExecuted` / `onExecutionSuccess` already call the sweep with exactly
the shape the ledger needs, so the whole fix lands behind the existing call sites.

## Test plan

- [x] `node --test browser_tests/unit/execution-preview-attach.test.mjs` — 26 pass (10 pre-existing #1286 + 16 new)
- [x] `npm run test:unit` on the rebased tree — **5529 pass / 0 fail**
- [x] `npm run typecheck` — clean
- [x] A test drives the **production call shape** (no `owners` argument, so the module-level ledger
  is the one under test) rather than only the injectable seam.
- [x] **Verified by mutation** — 10 mutants, fix committed first, worktree restored after each; **no survivors**:
  | mutant | caught by |
  |---|---|
  | type gate is the only gate again | the 3 victim-type tests + panel-call-shape |
  | stolen host gets the full `stripNodeExecutionPreview` | `a LoadImage on a reused id keeps the thumbnail it hydrated itself` |
  | eviction ignores whether the render came from the stolen entry | same |
  | no refresh of a still-owned id | `an entry written after the executed sweep is still attributed` |
  | same-type guard removed | `a same-type recreate/undo ... keeps the preview ComfyUI re-plants` |
  | entry-identity evidence removed | `a REPLACED store entry is not judged` |
  | owner never recorded | the victim-type tests |
  | eviction scoped back to both stores | `eviction does not take a live b_preview frame with it` |
  | eviction skips the type gate again | `proving a NON-host's entry stolen still strips it in full (#1286 stays fixed)` |
  | prune disabled | `the ledger stays graph-sized` |

### Browser verification — partial, and stated plainly

**No e2e spec in this repo touches `nodeOutputs` / `nodePreviewImages` / `imgs` at all**, so the
Playwright suite cannot validate this fix either way. The most it can show is that the panel still
loads and behaves as it does on `main`.

It was run for real against the ComfyUI allocated to this issue (`ap-1374`, port 9574), with that
instance's `custom_nodes` checkout repointed at this branch and confirmed to be **serving the
patched module** before each run — so these exercised the branch, not `main`. A baseline run of the
same suite against `origin/main` on the same instance is included for attribution, because the
suite is heavily red on this machine regardless of the diff:

| tree | failed | passed |
|---|---|---|
| `origin/main` (baseline) | 35 | 40 |
| this branch | 18 | 57 |

All 18 failures on the branch were a strict subset of the baseline's 35 — none unique to this
change, all chat/conversation/MockBridge specs.

Caveat I am not going to paper over: those numbers were measured on the **round-2** commit. A
re-run against the final commit cascaded to 9 passed, and the cause was environmental, not the
diff — the ComfyUI instance went down mid-run (`curl` to 9574 returns nothing), and the slot is
shared: another session had repointed it to its own commit by the time the run ended. I did not
re-measure after that. Given no spec exercises this code path, the behavioural evidence for the
fix is the unit + mutation work above, not the browser suite.

## Gate

codex was unavailable (account exhausted until 2026-08-19 20:43); gated by an independent
adversarial review instead (`claude-gate.mjs`, a fresh reviewer that never saw the authoring
reasoning).

Round 1 returned **NO-SHIP** with two P1s, both real and both reproduced by the reviewer:

1. the `stolen` branch routed image-capable hosts into `stripNodeExecutionPreview`, destroying a
   `LoadImage`'s own thumbnail and collapsing the node — fixed by the narrowed eviction above, and
   pinned by two of the mutants;
2. the `record.preview` half was unreachable in production, and the test certifying it hand-passed
   a `preferNodeId` the panel never supplies — the half and the test are gone, and the gap is
   documented instead of silently half-covered.

The reviewer also caught that the victim fixtures *copied* the entry's `images` array where the
frontend shares it by reference; corrected, which is what made the `LoadImage` class visible at all.

Round 2 returned **SHIP**, and flagged one P2 which is also fixed: eviction called
`clearStoredExecutionOutputs`, which drops the `nodePreviewImages` entry too. The evidence is an
entry in `nodeOutputs` and says nothing about live `b_preview` frames — a node on a reused id may
still be running and painting its own latent previews, and taking those with it drops a frame
mid-run (and `delete`d blob URLs without releasing them). Eviction is now scoped to `nodeOutputs`,
pinned by a test asserting the live frame survives.

Round 3 returned **NO-SHIP** on a P1 that was **a regression introduced by the round-2 fix** —
narrowing eviction for image-capable hosts was right, but the stolen branch also `continue`d past
the type gate, and eviction is not a superset of `stripNodeExecutionPreview`. A `ConditioningConcat`
on a reused id whose visible preview came from a `b_preview` frame kept `imgs`, the pseudo-widget,
its 266px size and its `nodePreviewImages` frame — #1286's exact symptom — while the sweep still
reported `stripped: 1`, self-healing only on the next queued prompt. Eviction no longer replaces
the type gate: a node that can never host a preview falls through and is stripped in full, counted
once. Every #1374 victim was image-capable, which is why no test drove a non-host through the new
branch; two now do.

That round also surfaced a second gap, now named in the header rather than half-covered: a
gifs/videos-only emitter (`VHS_VideoCombine`) has no `output.images` for the frontend to alias, so
there is no identity to test and a stale animation on an image-capable victim survives. Not a
regression — that victim was fully exempt before.

Two sub-P1 notes recorded rather than fixed, both pre-existing or vanishingly narrow:
`preferNodeId` is `d.node ?? d.display_node` while ComfyUI writes under `display_node || node`, so
a backend-expanded frame never seeds the ledger (identical to pre-diff behaviour — the type gate
still applies); and entry identity survives an in-place mutation of the entry, so a false positive
needs id reuse *and* expansion *and* `setOutputsByLocatorId({merge:true})` at once, with the
surviving entry still containing the stolen image.

Round 4 returned **SHIP** with no findings: 26/26 suite, an independent 11-mutant kill run on the
new logic (8 killed, 3 equivalent), and hostile scenarios against the ledger. The one state that
would destroy a real emitter's preview — two graph nodes sharing an id, where the last-wins
`nodesById` credits the wrong owner — is unreachable, because `LGraph.add` renumbers on id
collision. The reviewer also checked the load-bearing premise (`this.images = output.images` by
reference) verbatim against the frontend source at `litegraphService.ts:791`.

Fixes #1374

### DIFF
--- a/browser_tests/unit/execution-preview-attach.test.mjs
+++ b/browser_tests/unit/execution-preview-attach.test.mjs
@@ -16,6 +16,9 @@ import {
   stripNodeExecutionPreview,
   clearInheritedExecutionPreview,
   stripMisattachedExecutionPreviews,
+  recordExecutionPreviewOwner,
+  holdsStolenExecutionPreview,
+  executionPreviewOwnerLedger,
 } from "../../web/js/lib/execution-preview-attach.js";
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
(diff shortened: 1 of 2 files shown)


artokun:
Independent verification of this branch, plus the browser gate the description says could not be run.

**Context first, so the two sessions can be reconciled:** #1374 was also assigned to the backlog-drain session (`.claude/worktrees/drain-p3followups`), and I had a fix built and committed before this PR appeared. **I am standing down in favour of this one** and have not opened a competing PR. Mine keyed theft on node-object identity + type, the same discriminator as here, but it also extended the existing re-home to proven-stolen entries — which is wrong: an entry proven to belong to an *earlier* node is not this run's output, so re-homing it onto the current emitter re-attributes an old image to a new run. The entry-object identity check here is the piece mine lacked. Everything below is measured against **this** branch (`5a6d67a0`).

### 1. The issue's table, reproduced against this module

Probe written against the issue text before reading this diff; drives only the public
`stripMisattachedExecutionPreviews`, through the two-run sequence production actually sees —
run 1 emits at id 9 and sweeps with `preferNodeId: 9`; ComfyUI's own paste/undo then hands id 9
to the victim with the store entry intact; run 2 emits elsewhere and sweeps again.

| victim type | `origin/main` | this branch |
|---|---|---|
| `ConditioningConcat` | stripped 1 | stripped 1 |
| `ConditioningCombine` | stripped 1 | stripped 1 |
| `CheckpointLoaderSimple` | stripped 1 | stripped 1 |
| `VAEDecode` | **stripped 0, entry retained** | **stripped 1** |
| `ImageScale` | **stripped 0, entry retained** | **stripped 1** |
| `EmptyLatentImage` | **stripped 0, entry retained** | **stripped 1** |

### 2. The over-strip directions — five independent refutations, all held

Same portable probe, aimed at the false-positive surface rather than the fix:

| scenario | this branch |
|---|---|
| an id nothing was ever observed emitting, image-capable occupant (`/history` restore) | not stripped |
| workflow reload: same ids, same types, brand-new node objects | not stripped |
| a node that emitted in an EARLIER run, untouched by a later partial run | not stripped |
| `LoadImage` hydrating its own `imgs` at an unobserved id | not stripped |
| `KSampler` live latent preview (#1286's own case) | not stripped |

All five behave identically to `origin/main`. The sweep only ever strips *more precisely* here,
which matches the claim in the description.

### 3. Browser verification — run, on a dedicated instance

`npx playwright test --workers=1` against a ComfyUI on its own port (9574) and its own
`--base-directory`, serving each checkout in turn (verified by fetching the module over HTTP
before each run, not by reading the file on disk):

| build | result |
|---|---|
| `origin/main` @ `f6136f6d` | 18 failed, 57 passed |
| **this branch** @ `5a6d67a0` | 19 failed, 56 passed |
| this branch's own base @ `2ee73e09`, unmodified | `autostart-mcp.spec.ts:26` **fails** |

The one extra failure on this branch versus current main is
`autostart-mcp.spec.ts:26 › panel-open autostart asks the companion once and connects after MCP
appears`. It is **not from this diff** — it fails identically on an unmodified checkout of this
branch's base (`2ee73e09`, v0.15.4) and passes on current main, so it was fixed by something
that landed between the two. **Rebasing onto current main should clear it.** The other 18 are
the pre-existing red set in this environment (MockBridge conversation-persistence /
chat-history / streaming specs), unchanged across every run.

**Delta attributable to this diff: zero.**

### 4. One note on the base

This branch is based on `2ee73e09`; main is now at `7976ee72` (five commits on, including two
that touch `web/js/comfyui-mcp-panel.js`). Worth a rebase before the gate reads it, both for the
autostart spec above and because a green check is only green for the main it ran against.

PR #1393, opened 2026-08-19
fix(1385): the coalescer-join pin is checked at EVERY call site, not once over the body

### LINKED ISSUE #1385: graph_add_node's coalescer-join pin is satisfiable by a sibling call site, so unwrapping one stays green
Found while re-gating PR #1354, and confirmed **pre-existing** — the identical mutation survives on `8193b226`, before that PR's rebase, so no guard was dropped by a conflict resolution. Filing so it does not stay invisible.

`graph_add_node` has five budgeted waits, each with a source-level pin asserting the call site is wired to the bounding primitive. Mutated one at a time, four die: the seed wait, the fast path, the whole-schema fetch and the registration wait (the fast path and whole fetch each take a second test with them).

**The fifth — the coalescer join — survives.** Its assertion is a single `assert.match` over `graph_add_node`'s body, and there are **two** call sites carrying `joinMs: budget.remaining() - ADD_NODE_POST_REFRESH_RESERVE_MS`:

- `web/js/comfyui-mcp-panel.js:11993`
- `web/js/comfyui-mcp-panel.js:12084`

Unwrapping either one leaves the other satisfying the regex, so the pin stays green over a half-unwired file. Mutating **both** kills it.

**Why this one matters more than a generic coverage gap.** That join is the wait whose absence does not fail a test — it **hangs**: removing its `joinMs` made `graph_add_node` run 150,000 ms against a whole-file baseline of 2,992 ms, killed only by the harness cap. It is precisely the guard you least want silently half-removed, and #1349 exists because that fifth wait was unbudgeted in the first place.

**Fix:** assert per call site rather than once over the function body — either count occurrences and require 2, or anchor each match to its surrounding call so one unwrapping cannot be masked by the other. Counting alone is the weaker option: it cannot see a third site appearing later, which is the same shape as the original defect.

This is the project's recurring "an install line is invisible to tests" failure, in the form where the pin exists but is satisfiable by a sibling. No user-visible symptom today; the code is correct as shipped.

artokun:
Closes #1385.

## What was wrong

`graph_add_node` has five budgeted waits, each with a source-level pin in
`browser_tests/unit/single-node-def.test.mjs` asserting the call site is wired to the
bounding primitive. The fifth — the coalescer join — was one row in the `wired` list:

```js
[/joinMs: budget\.remaining\(\) - ADD_NODE_POST_REFRESH_RESERVE_MS/, "the coalescer join"],
```

a single `assert.match` over the whole function body. The body makes **two** calls carrying
that bound — the resolver's join (`web/js/comfyui-mcp-panel.js:12261`) and #1242's forced
drift recovery (`:12352`) — so a match anywhere satisfied it and unwrapping either one left
the other holding the pin green over a half-unwired file.

## Mutation evidence (before the change)

Ran against the shipped source on `origin/main` (567e6438), deleting one `joinMs:` line at a
time:

| mutation | `single-node-def.test.mjs` |
|---|---|
| resolver join (`:12261`) unwrapped | **pass**, exit 0, 2s |
| drift recovery (`:12352`) unwrapped | **pass**, exit 0, 2s |

So the row never failed for *either* unwrapping — it was dead weight in both directions, a
bit stronger than the issue's "one is masked by the other". The drift-recovery site happens
to be pinned separately by `add-node-command-budget.test.mjs:602` and
`add-node-schema-drift-recovery.test.mjs:112` (both anchor on `force: true,`); the
**resolver's join had no dedicated source pin anywhere in the suite**.

**And its absence does not fail a test — it hangs.** With the resolver's `joinMs` deleted,
`add-node-command-budget.test.mjs` ran until an external cap killed it (150s+, against a
whole-file baseline of ~1s), and because `npm run test:unit` passes no `--test-timeout`,
node:test buffers the file's output and printed **nothing at all** — not a failure, not a
test name. Re-running that file with `--test-timeout=8000` names the two tests that stall:

```
not ok 1 - #1192: an add refuses IN WORDS when the in-flight refresh would eat the whole budget
not ok 3 - #1192: the abandoned add does NOT start a competing registration run
```

## The change

Test-only. The `wired` row is replaced by a per-call-site check: enumerate every
`refreshComfyNodeDefs(...)` call in the comment-stripped `graph_add_node` body by
balanced-paren scan, and require **each** one to carry the bound.

Enumerated, not counted. The issue called counting the weaker option because "there are 2"
goes stale the day a third call site is written — the same defect in a new place. The new
helper returns one entry per call site, so the assertion scales with however many there are,
and the failure message names *which* call and prints its arguments.

## Mutation evidence (after the change)

| mutation | `single-node-def.test.mjs` | failure message |
|---|---|---|
| resolver join unwrapped | **fail**, exit 1, ~1.2s | `refreshComfyNodeDefs call 1 of 2 … Its arguments: defs, {` |
| drift recovery unwrapped | **fail**, exit 1, ~1.2s | `refreshComfyNodeDefs call 2 of 2 … Its arguments: undefined, {` |
| both unwrapped | **fail**, exit 1 | `call 1 of 2` |
| **third, unbounded call inserted** into `graph_add_node` | **fail**, exit 1 | `call 2 of 3 … Its arguments: undefined, { force: true }` |
| unmutated | **pass**, 13/13 | — |

The third-site row is the one the issue specifically said counting could not see. It was
inserted inside `graph_add_node` (an earlier attempt anchored on a same-shaped call at
`:10622` in a *different* function and correctly did not trip this pin — the scan is scoped
to the body, as intended).

## Verification

- `npm run test:unit` — **5446 pass, 0 fail** (exit 0), full suite in the worktree after `npm ci`.
- `node --check web/js/comfyui-mcp-panel.js` — clean at every mutation step.

## Not done, deliberately

- **No Playwright run.** This diff touches one unit test file; `web/js/` and every shipped
  asset are byte-identical to `origin/main`, so there is nothing the panel renders to verify.
  I am claiming no browser coverage here rather than implying it.
- **`--test-timeout` was not added to `test:unit`.** The silent-hang behaviour above is real
  and worth fixing, but it is a suite-wide change with flake risk on a shared runner and is
  not what #1385 asks for. Flagging it as a separate call.
- The shipped source is untouched — the code was correct as shipped, as the issue states.

## Gate

codex was unavailable (account exhausted until 2026-08-19 20:43; it is 02:00 local); gated by
an independent adversarial review instead, via `.claude/autopilot/claude-gate.mjs`.

### DIFF
--- a/browser_tests/unit/single-node-def.test.mjs
+++ b/browser_tests/unit/single-node-def.test.mjs
@@ -466,6 +466,45 @@ test("#1180: the widen's bound fits INSIDE the registration deadline it runs und
   );
 });
 
+/**
+ * #1385 — every argument list `name(` is called with in `src`, one entry per CALL SITE.
+ *
+ * A single regex over a function body cannot do this job: it answers "is this written
+ * SOMEWHERE in here", and a body with two call sites keeps answering yes after one of them
+ * is unwired. Returning the sites separately is what lets each be asserted on its own, and
+ * what makes a THIRD site — the shape of defect that keeps recurring here — visible the day
+ * it is written rather than the day it breaks.
+ *
+ * A balanced-paren scan rather than a regex because the calls span lines and nest parens
+ * (`budget.remaining()`). Quoted text is skipped so a paren inside a string cannot mis-slice
+ * a call; the scan throws rather than guessing if the parens do not balance.
+ */
+function callArgumentLists(src, name) {
+  const opener = `${name}(`;
+  const lists = [];
+  for (let at = src.indexOf(opener); at >= 0; at = src.indexOf(opener, at + 1)) {
+    // `foo.refreshComfyNodeDefs(` / `myRefreshComfyNodeDefs(` are different functions.
+    if (/[A-Za-z0-9_$.]/.test(src[at - 1] ?? "")) continue;
+    let depth = 0;
+    let quote = null;
+    let i = at + opener.length - 1;
+    for (; i < src.length; i++) {
+      const ch = src[i];
+      if (quote) {
+        if (ch === "\\") i++;
+        else if (ch === quote) quote = null;
+        continue;
+      }
+      if (ch === '"' || ch === "'" || ch === "`") quote = ch;
+      else if (ch === "(") depth++;
+      else if (ch === ")" && --depth === 0) break;
+    }
+    assert.ok(depth === 0 && i < src.length, `unbalanced parentheses after ${opener}`);
+    lists.push(src.slice(at + opener.length, i));
+  }
+  return lists;
+}
+
 test("#1192: graph_add_node's serialized bounds FIT the command budget", () => {
   // The predecessor of this test asserted the OPPOSITE — that this path exceeds the budget —
   // and said so in its own failure message: "If that is genuinely fixed, close #1192 and
@@ -582,7 +621,9 @@ test("#1192: graph_add_node's serialized bounds FIT the command budget", () => {
       "the single-class fast path",
     ],
     [/await boundedGetNodeDefs\(budget\.bounded\(NODE_DEFS_FETCH_TIMEOUT_MS\)\)/, "the whole-schema fetch"],
-    [/joinMs: budget\.remaining\(\) - ADD_NODE_POST_REFRESH_RESERVE_MS/, "the coalescer join"],
+    // The fifth wait — the coalescer join — is NOT a row here. #1385: it has two call sites,
+    // so a body-wide match is satisfied by whichever one is left. It is pinned per call site
+    // below instead.
     [/budget\.bounded\(CUSTOM_WIDGET_REGISTRATION_TIMEOUT_MS\)/, "the custom-widget registration wait"],
   ];
   for (const [re, what] of wired) {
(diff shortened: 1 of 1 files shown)


artokun:
**Gate: SHIP (exit 0).** codex was unavailable (account quota exhausted until 2026-08-19 20:43; run made at 02:00 local), so this was gated by an independent adversarial reviewer via `.claude/autopilot/claude-gate.mjs` — a fresh reviewer with no sight of the author's reasoning. Disclosing the substitution as required.

The reviewer did not take the PR's evidence on trust — it re-ran the mutations itself and additionally checked out the **pre-fix** test (`HEAD~1`) against both mutations, confirming 13/13 pass either way. It also fuzzed the paren scanner with hostile input (an apostrophe in a trailing comment inside an argument list; an unmatched `(` in a comment) and confirmed both **throw** rather than silently mis-slicing into a neighbouring call's `joinMs`.

One useful detail it surfaced: the scan runs on the **comment-stripped** body, and `graph_add_node` contains 4 raw `refreshComfyNodeDefs` occurrences, 2 of them inside comments. Scanning `body` instead of `stripped` would have been matching comment prose — the strip is load-bearing, not cosmetic.

**Post-rebase re-verification.** `main` moved (567e6438 → 83ed44de) after the first CI run and the delta touched `web/js/comfyui-mcp-panel.js` (+94 lines), which this pin reads. Rebased and re-ran everything against the new base: pin green 13/13 unmutated; both single-site mutations still fail with the per-call-site message; `npm run typecheck` clean; CI green on the rebased head.

Full-suite `test:unit` showed **one** failure — `#671 verifyInstalled (real panel source) still verifies a fast-draining install` at 10574 ms. That is the known wall-clock flake under machine load, not this diff (which touches one unit test file and no shipped source). Re-run alone: **125/125 pass**, that test in 1019 ms.

PR #1395, opened 2026-08-19
fix(1706 followup): the promoted-widget guard reads every dialect of an id that MOVED, not one spelling

artokun:
Follow-up to #1388, which merged while I was applying the gate's findings. Two of the three items landed after the merge, so they need their own PR. Item 2 (the CHANGELOG under `## [Unreleased]`, not `## [0.15.4]`) DID make it into #1388 and is already correct on `main` — nothing here touches it.

## 1. The promoted-widget guard keyed on ONE spelling of an id

Gate finding, and the same class as a string/number key mismatch that made two tools fail 100% of the time on a real graph the same night.

`rootNodesReferenceRemappedId` refuses the relabeling account when a root node's `properties.proxyWidgets` names a definition node the relabeling moved. It built that decision from a `Set` keyed by `String(payloadId)` and asked it one question: `remappedFrom.has(String(entry[0]))`.

A promoted-widget id in a non-canonical numeric dialect — `"78.0"`, `" 78"`, `"+78"`, `"078"` — is a *different string* and the *same node*. Measured end-to-end by the gate: that entry made `definitions` come back ACCOUNTED, which took it out of the UNEXPLAINED set the weaker completed-load ground reads, and `normalizedOnly` flipped to **true** with `normalizedFields ["properties","widgets_values"]` — reassuring wording over a genuinely lost promoted widget.

Bounded, and worth stating so the severity is not overread: the VERDICT still refused (`proven:false`), so no fence was published; and the frontend writes these entries as `String(id)` under a `z.tuple([z.string(), z.string()])` schema, so reaching it needs a hand-edited workflow. It is the sentence that was wrong, not the fence.

**The fix asks the question in every reading that can name a node that ACTUALLY MOVED**, and in no more than that:

1. the raw string, as before;
2. `Number(key)` against the numeric normalisation of the remapped ids — this is the only clause that catches `"0x4E"` and `"7.8e1"`;
3. a leading-integer parse — the only clause that catches `"78abc"`, where `Number()` is `NaN`.

**A blanket "refuse every non-canonical id" was written first and then removed.** It refuses `"999999.0"`, which names nothing that moved — an over-broad guard is a regression wearing a fix's clothes, and there is a test pinning that direction (`#1706 a NON-numeric promoted id that names nothing remapped still does not block`).

## 2. Stale prose

`web/js/lib/node-inputs-rebuild.js:36` still cited `definitionsDifferOnlyByLinkRenumber` as the precedent it follows. #1388 renamed that export to `definitionsDifferOnlyByRenumber` because the old name asserted "link" and there are now two measured rewrites. Grepped tree-wide; that was the only remaining site.

## Refutation

Baseline for the three affected files: **107/107**. Twelve mutations, each confirmed applied by `git diff --numstat` before its run. Restored afterwards: **107/107**.

| mutation | verdict |
| --- | --- |
| M1 content-proof call site drops `rootNodes` | killed (2) |
| M2 disclosure call site drops `rootNodes` | killed (1) |
| M3 the node-id account is reverted | killed (8) |
| M4 the proxyWidgets cross-surface guard never fires | killed (5) |
| M5 the allocation counter need not advance | killed (1) |
| M6 link endpoints not compared through the map | killed (7) |
| M7 the map need not be injective | killed (1) |
| M8 a node may be retyped at its position | **SURVIVED** (unchanged, documented at the line) |
| M9 the wider allowances granted without a relabeling | killed (6) |
| M10 an id that only changed TYPE reads as a non-move | killed (1) |
| **M11 the guard reads ONE spelling of the id again** | **killed (1)** |
| **M12 the leading-integer reading is dropped** | **killed (1)** |

M11 SURVIVED on the first attempt, and that was a real finding about the test rather than the code: every decimal dialect I had written (`"78.0"`, `" 78"`, `"+78"`, `"078"`, `"78abc"`) is already caught by the leading-integer clause, so deleting the `Number()` clause killed nothing. The added test isolates the two inputs where the readings DISAGREE — `Number("0x4E") === 78` while the leading-integer parse is `0`, and `Number("7.8e1") === 78` while the parse is `7` — so each clause is now the only thing standing in front of one of them. It asserts that disagreement explicitly, so a future edit cannot quietly make the case vacuous.

**End-to-end on the real captured pair is unchanged** (`describeGraphStateDifference` + `graphRootReproducesStateContent` + `resolveOpenRebindVerdict`, driven with the actual payload and the actual `app.graph.serialize()` output): `surfaces ["definitions"]`, `accountedSurfaces ["definitions"]`, `proven: true`, verdict **proven**. The fix still ships what #1388 shipped.

## Verification

**Unit** — `node --test browser_tests/unit/*.test.mjs`: **5492 pass, 0 fail** (the known `#671 verifyInstalled` wall-clock flake did not fire).

**Panel gates** — `i18n-check`, `i18n-render-check`, `check-tool-vocabulary`, `check-panel-scope`, `tsc --noEmit`: all OK.

**Browser (Playwright)** — NOT re-run for this follow-up. #1388's delta was measured (`origin/main` 18 failed / 57 passed, branch 18 / 57, identical failure lists, each side on its own dedicated ComfyUI). This change touches one predicate's key comparison and one comment, adds no panel-visible behaviour, and its whole effect is to REFUSE in cases that previously slipped through — so I am stating the absence rather than implying coverage I do not have.

## Gate

`codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the weekly limit, so **no independent gate was run by this agent** — the orchestrator gates and merges.

This addresses two findings against the change that fixed the underlying cause of artokun/comfyui-mcp#1706.

### DIFF
--- a/browser_tests/unit/definitions-renumber.test.mjs
+++ b/browser_tests/unit/definitions-renumber.test.mjs
@@ -835,3 +835,29 @@ test("#1706 a NON-numeric promoted id that names nothing remapped still does not
     );
   }
 });
+
+test("#1706 P0: a promoted widget id in HEX or EXPONENT form still refuses", () => {
+  // The two numeric readings are NOT redundant, and mutation is what showed it: with
+  // only the decimal dialects above, deleting the `Number()` clause killed no test,
+  // because a leading-integer parse already caught `"78.0"`, `" 78"`, `"078"`, `"78abc"`.
+  // These two are the inputs where the readings DISAGREE — `Number("0x4E")` is 78 while
+  // a leading-integer parse is 0, and `Number("7.8e1")` is 78 while the parse is 7 — so
+  // each clause is the only thing standing in front of one of them.
+  const c = captured();
+  const movedFrom = c.payloadDefinitions.subgraphs[0].nodes[0].id;
+  const hex = `0x${movedFrom.toString(16)}`;
+  const exponent = `${movedFrom / 10}e1`;
+  assert.equal(Number(hex), movedFrom, "the hex form must really name the moved node");
+  assert.equal(Number(exponent), movedFrom, "the exponent form must really name the moved node");
+  assert.notEqual(Number(/^\s*[+-]?\d+/.exec(hex)[0]), movedFrom, "and the leading-integer parse must NOT");
+  assert.notEqual(Number(/^\s*[+-]?\d+/.exec(exponent)[0]), movedFrom, "likewise");
+  for (const dialect of [hex, exponent]) {
+    assert.equal(
+      definitionsDifferOnlyByRenumber(c.payloadDefinitions, c.liveDefinitions, {
+        rootNodes: [{ id: 400, type: "sub-1", properties: { proxyWidgets: [[dialect, "seed"]] } }],
+      }),
+      false,
+      `proxyWidgets id ${JSON.stringify(dialect)} names a relabeled node and must refuse`,
+    );
+  }
+});
--- a/browser_tests/unit/definitions-renumber.test.mjs
+++ b/browser_tests/unit/definitions-renumber.test.mjs
@@ -786,3 +786,52 @@ test("#1706 P0: an id whose only change is its TYPE is not a relabeling", () => 
   okLive.subgraphs[1].state = { lastLinkId: 11, lastNodeId: 40 };
   assert.equal(definitionsDifferOnlyByRenumber(payload, okLive, { rootNodes: [] }), true);
 });
+
+test("#1706 P0: a promoted widget id in a NON-CANONICAL dialect still refuses", () => {
+  // Gate finding, and the same class as a string/number key mismatch that made two
+  // tools fail 100% of the time on a real graph tonight. `remappedFrom` is keyed by
+  // `String(payloadId)`, so `"78.0"` / `" 78"` / `"+78"` are all "not in the set" to a
+  // text comparison while naming node 78 to anything that reads the id numerically.
+  // Measured end-to-end by the gate: `definitions` came back ACCOUNTED and the weaker
+  // completed-load ground flipped to normalizedOnly with
+  // `normalizedFields ["properties","widgets_values"]` — reassuring wording over a
+  // genuinely lost promoted widget. The verdict still refused, so no fence was
+  // published, but the sentence was wrong and this closes it.
+  const c = captured();
+  const movedFrom = c.payloadDefinitions.subgraphs[0].nodes[0].id;
+  assert.equal(typeof movedFrom, "number", "the fixture's ids are numeric, as the frontend writes them");
+  for (const dialect of [
+    `${movedFrom}.0`,
+    ` ${movedFrom}`,
+    `+${movedFrom}`,
+    `${movedFrom} `,
+    `0${movedFrom}`,
+    `${movedFrom}.00`,
+    `${movedFrom}abc`, // Number() says NaN; a leading-integer parse says it names the moved node
+    movedFrom, // the number itself, not a string — the schema says string, reality may not
+  ]) {
+    assert.equal(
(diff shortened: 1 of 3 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**. Verdict: **SHIP**, head `a487abaf` (matches the reviewed SHA).

**The guard was compared against the frontend's real lookup by execution, not by reading.** Ground truth is `ComfyUI_frontend/src/lib/litegraph/src/subgraph/subgraphDeduplication.ts:207` — `const oldId = Number(entry[0]); remappedIds.get(oldId)` against a **number-keyed** Map. A parity harness with a real fixture and a real remap (`38→200, 80→201, 7→202`) compared `frontendPatches(entry0)` to `guardRefuses(entry0)` across **29 dialects**. Clauses 1 and 2 agree with the frontend on all of them, including `"0x26"`, `"3.8e1"`, `"38e0"`, `"\n38\n"`, `-38` and the raw number. So the original defect — `String(entry[0])` missing what `Number(entry[0])` patches — is genuinely closed.

**The over-refusal direction was pinned, and the pin is non-vacuous.** Clause 3 (leading-integer) is stricter than any real reader: it refuses `"38abc"`, `"38-1"`, `"38:2"`, `"38,"`, `"38.5"`, `"38_b"`, which the frontend does not patch. That is acceptable because the direction is `notProven` and **every** proxyWidgets writer in the frontend is `String(numericId)` — promotionUtils ×6, `LGraphCanvas:9050`, `patchProxyWidgets:209` — so all six shapes require a hand-authored workflow, the same reachability bar as the bug being fixed. The author's own blanket "refuse any non-canonical id" mutation is killed by the over-refusal test, confirming they were right to remove it.

**Mutations: 8 run, each `git diff --numstat`-confirmed applied, 7 killed.** Dropping the `Number()` clause is killed by the HEX/EXPONENT test *alone*; dropping the leading-integer clause by the non-canonical dialect test; reverting to one spelling kills 2; the blanket-refuse version kills 2; empty `numericForms` kills 2; guard-never-fires kills 4; both call sites dropping `rootNodes` kills 1. The M11 story checks out — the added test carries its own `assert.notEqual` self-checks that the two readings genuinely disagree, so it cannot go vacuous silently.

**The one survivor is pre-existing, not introduced here:** dropping the raw-string clause `remappedFrom.has(key)` leaves 109/109 green, because this PR's numeric clauses shadow it for every tested input. It is still live and reachable — a definition with *string* node ids that move produces a non-numeric `remappedFrom` key, since `Number("abc")` is NaN so `numericForms` cannot see it, and the guard correctly refuses `"abc"` while allowing `"def"`. No test covers that shape. A test gap worth a follow-up, not a merge blocker.

Item 3 confirmed: `node-inputs-rebuild.js:36` now cites `definitionsDifferOnlyByRenumber`, and the old name has zero occurrences tree-wide.

Five gates OK, full unit 5493 with only the known #671 wall-clock flake (green on re-run), and current `origin/main` merged into the head re-runs at 5508 tests with all gates OK and `tsc --noEmit` clean.

**One correction to this PR's own reasoning, worth recording.** The body justifies skipping Playwright with "the change only ever makes the guard refuse more" — but that is precisely the *risky* direction, not a safe one. What actually makes it safe is that the extra refusals are unreachable from frontend-authored workflows, which the gate established by grepping every writer rather than by trusting the direction argument. Two prose findings go with it: the code comment claims "an entry that some reader resolves to a node the relabeling moved", and the gate could not find that reader — the only other frontend reader, `resolveConcretePromotedWidget` → `LGraph.getNodeById` → `_nodes_by_id[id]`, is textual rather than a leading-integer parse.

Issue #1370, opened 2026-08-19
missingNodeCount survives the frontend-virtual filter, so an alarm can fire with the names stripped


artokun:
Found by the independent gate on panel PR #1353 and demonstrated live, so it is a reproduced state rather than a code-reading concern.

`collectMissingAssets` filters `nodeTypes` through the new frontend-virtual check, but leaves `nodeCount` (`store.missingNodeCount`) untouched and computes `any` from it. That makes this state reachable:

```
hasMissingNodes: true,  missingNodeCount: 2,  nodeTypes: []
```

Two consumers then report an alarm with the names stripped out:

- `graph_get_errors` takes the pre-existing `missingNodeCount && !missingNodeTypes.length` branch and returns `missing_node_count: 2` with no names.
- The turn-start banner takes `else if (missing.nodeCount)` and says "2 node type(s) not installed on this ComfyUI".

So after #1353 an agent can be told two node types are missing while being given nothing to act on — narrower than the pre-fix message, which at least named them. The names were correctly suppressed (they were frontend-virtual nodes that are absent from `/object_info` by design); the count was not suppressed with them.

**This does not affect mcp#1657's reported symptom.** That message, `"node type not found in /object_info"`, is emitted only by `live-combo-availability.js` — call site 1, which #1353 fixes completely. This is the third surface disagreeing with the other two, which qualifies the "the three surfaces cannot disagree" claim in that PR.

Fix is either to filter `nodeCount` alongside `nodeTypes`, or to have the consumers stop reporting a bare count when the type list is empty after filtering. The second is probably better: a count with no names was never actionable, and the `missingNodeCount && !missingNodeTypes.length` branch predates this change.

Not a regression in safety — nothing that should be refused is now allowed. It is a message-quality defect on a path that fires precisely when a user has frontend-only nodes installed.

artokun:
Claimed by the backlog-drain session (worktree `.claude/worktrees/drain-p3followups`). These are both follow-ups this session filed from independent review findings. A fix agent is assigned now; if no PR appears within ~8h, consider the claim lapsed.

artokun:
Closing as **already fixed** — on current `main` (`2ee73e09`, panel **v0.15.4**) neither reported consumer emits a count with the names stripped.

## What fixed it

[...]

PR #1396, opened 2026-08-19
fix(1370): a missing-node COUNT is filtered with the names it was counting, and an unnamed miss is still reported

artokun:
## Summary

`collectMissingAssets` filters `nodeTypes` through the frontend-virtual check added by #1353
but leaves `nodeCount` (ComfyUI's own `missingNodeCount`) alone and computes `any` from it.
Driving the **shipped collector** (extracted from the panel source and executed, not
re-implemented) against a real `missingNodesError` store shape, on `origin/main`:

| store record | live graph | before |
|---|---|---|
| **A** `[{type:"SetNode"},{type:"GetNode"}]`, count 2 | both instances `isVirtualNode: true` | `{nodeTypes: [], nodeCount: 2, any: true}` |
| **D** record names nothing, count 2 | — | `{nodeTypes: [], nodeCount: 2, any: true}` |

A and D are **byte-identical on the collector's output** and mean opposite things. A is
"nothing is missing, the names were correctly suppressed"; D is "two nodes are missing and
the panel could not read the record's shape". Both surfaces that read the bare count then
raised an alarm with the names stripped: `graph_get_errors` → `missing_node_count: 2` with
no names, and the turn-start banner → "2 node type(s) not installed on this ComfyUI" —
narrower than the pre-#1353 message, which at least named them.

## What I decided about the pre-existing `missingNodeCount && !missingNodeTypes.length` branch

The issue offered deleting its trigger. **I kept it, deliberately** — that branch is the
only way case D reaches an agent, and D is a real state (the panel's parser handles a list
or an object; a shape it cannot read leaves the count as the sole evidence). Deleting the
trigger would collapse "N missing, names unknown" into "no errors recorded", which is the
same class of defect as the one being fixed, pointing the other way.

So the count is suppressed **where the reason is known** — in the collector, beside the
filter that stripped the names, under two conditions:

```js
if (recordedTypeCount && !nodeTypes.length && !unnamedRecordCount) nodeCount = 0;
```

- `recordedTypeCount` — the record DID name types, so the count has a known referent (A, not D).
- `!unnamedRecordCount` — every record entry produced a name. An entry the parser could not
  name is an unaccounted-for miss and keeps the count alive; nothing about it was proven.

Filtered in the collector rather than at each consumer for the same reason the type list is:
it is the one collector `graph_get_errors`, the turn-start banner and the stale-red-flag
adjudication share, and this family recurred three times because each call site decided
separately.

## A second, live defect this uncovered in #1380

#1380 (merged ~54 min after #1370 was filed) added `if (!nodeTypes.length) nodeCount = 0` at
**both** consumers, as part of the #1332 stale-placeholder adjudication. That masks case A on
the happy path — and breaks case D: with the recorded list empty from the start, the
adjudication removes nothing, the line fires anyway, and `graph_get_errors` emits
`note: "no errors recorded since the last execution start"` for a canvas whose store says two
nodes are missing. Both consumers now gate that zeroing on the adjudication having **actually
removed something**:

```js
if (recordedTypes.length && !assets.nodeTypes.length) assets.nodeCount = 0;
```

`recordedTypes` is captured **before** the narrowing assignment — a post-assignment re-read is
always the narrowed list, so the gate would never fire and #1332 would regress. That ordering
is pinned.

## Measured effect (same probe, after)

| case | before | after |
|---|---|---|
| A both types proven frontend-virtual | `[], 2, any:true` | `[], 0, any:false` |
| B both real placeholders | `["SetNode","GetNode"], 2` | unchanged |
| C one virtual, one real | `["GetNode"], 2` | unchanged |
| D record names nothing | `[], 2, any:true` | unchanged (**kept**) |
| E unreadable graph (nothing proven virtual) | `["SetNode","GetNode"], 2` | unchanged |
| F one entry unnameable + one virtual | `[], 2, any:true` | unchanged (**kept**) |

## Test plan

- [x] `node --test browser_tests/unit/frontend-virtual-nodes.test.mjs` — 26 pass (20 pre-existing + 6 new). The behavioural tests execute the **shipped** `collectMissingAssets` extracted from `web/js/comfyui-mcp-panel.js`; a re-implementation would have passed while production kept the old tuple.
- [x] `npm run test:unit` — 5434 tests, 1 fail: the known #671 `verifyInstalled` wall-clock flake, which is 125/125 when its file runs alone.
- [x] `node scripts/check-panel-scope.mjs` — OK.
- [x] **Verified by deleting the fix** (committed first, then mutated, each mutation confirmed applied via a non-empty `git diff` before its verdict was believed — these files are CRLF):

| mutation | result |
|---|---|
| delete the collector's `nodeCount = 0` line | **RED** — case A |
| revert `graph_get_errors` to the ungated #1332 form | **RED** — CONSUMER WIRING |
| revert the banner to the ungated form | **RED** — CONSUMER WIRING |
| make the collector zero unconditionally (over-strip) | **RED** — cases D and F |
| delete the pre-adjudication `recordedTypes` capture | **RED** — CONSUMER WIRING |

The over-strip mutation is the one that matters most here: it is the version of this fix that
deletes the branch, and two tests refuse it.

### Browser verification

`npx playwright test --workers=1` against a ComfyUI standing on its own port and base
directory, serving this exact checkout (verified by fetching the module over HTTP before the
run — not just reading the file on disk):

| build | result |
|---|---|
| `origin/main` @ `f6136f6d` | 18 failed, 57 passed |
| this branch on `f6136f6d` | 18 failed, 57 passed — **identical failing set** |

**Delta: zero.** The 18 are a pre-existing red set in this environment (MockBridge
conversation-persistence / chat-history / streaming specs), none of them touching missing
assets or the error surfaces. An earlier baseline run at `e27a78a5` showed 19 — the extra one
was `agent-drive.spec.ts:113`, which passed in all three later runs; a flake, not a delta.
Main has since moved to `7976ee72`; this branch is rebased onto it, and the unit suite and the
scope gate were re-run there, but the browser runs were against `f6136f6d`.

## Gate

Not self-gated. codex is exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on
the weekly limit, so **this PR has had no independent adversarial gate** — the owner gates and
merges it.

The underlying cause of #1370 is addressed here.

### DIFF
--- a/browser_tests/unit/frontend-virtual-nodes.test.mjs
+++ b/browser_tests/unit/frontend-virtual-nodes.test.mjs
@@ -313,3 +313,215 @@ test("#1648 CALL SITE 3 WIRING: the registry is diagnosis only — the refusal i
   assert.match(block, /if \(badIds\.length\) \{/);
   assert.equal((block.match(/registered_node_types/g) ?? []).length, 1);
 });
+
+// ── #1370: the COUNT is filtered with the names, and only by a REAL removal ──────
+//
+// The filter above only ever touched the type LIST. The collector also returns
+// `nodeCount` (ComfyUI's own `missingNodeCount`) and computes `any` from it, so
+// `{ nodeTypes: [], nodeCount: 2, any: true }` was reachable — and both surfaces that
+// read the bare count then raised an alarm with the names stripped out: graph_get_errors'
+// `missing_node_count: 2` and the banner's "2 node type(s) not installed on this ComfyUI".
+//
+// Driven against the SHIPPED collector, extracted from the panel source and executed —
+// a re-implementation here would pass while production kept the old tuple.
+//
+// The pair that must stay distinguishable is A vs D: "every recorded name was proven
+// frontend-virtual" and "the record named nothing at all" were byte-identical on the
+// collector's output, and only the first is safe to suppress. D is the state the
+// pre-existing `missingNodeCount && !missingNodeTypes.length` consumer branch reports,
+// so it is kept, not deleted.
+
+function extractPanelFn(sig) {
+  const start = PANEL.indexOf(sig);
+  assert.notEqual(start, -1, `${sig} not found in the panel source`);
+  const open = PANEL.indexOf(") {", start) + 1;
+  let depth = 0;
+  for (let i = open; i < PANEL.length; i += 1) {
+    const ch = PANEL[i];
+    if (ch === "/" && PANEL[i + 1] === "/") {
+      i = PANEL.indexOf("\n", i + 2);
+      if (i < 0) break;
+      continue;
+    }
+    if (ch === "/" && PANEL[i + 1] === "*") {
+      i = PANEL.indexOf("*/", i + 2);
+      if (i < 0) break;
+      i += 1;
+      continue;
+    }
+    if (ch === '"' || ch === "'" || ch === "`") {
+      const quote = ch;
+      for (i += 1; i < PANEL.length; i += 1) {
+        if (PANEL[i] === "\\") {
+          i += 1;
+          continue;
+        }
+        if (PANEL[i] === quote) break;
+      }
+      continue;
+    }
+    if (ch === "{") depth += 1;
+    if (ch === "}" && --depth === 0) return PANEL.slice(start, i + 1);
+  }
+  throw new Error(`unterminated: ${sig}`);
+}
+
+const COLLECTOR_BODY = extractPanelFn("function collectMissingAssets(trustComboOverride) {");
+
+/**
+ * Run the shipping collector against a real `missingNodesError` store shape.
+ * `record` is the store's own load-time list; `nodes` is the live graph.
+ */
+function runCollector({ record, count, nodes = [], readableGraph = true }) {
+  const stores = {
+    missingModel: { missingModelCandidates: [] },
+    missingMedia: { missingMediaCandidates: [] },
+    missingNodesError: {
+      hasMissingNodes: true,
+      missingNodeCount: count,
+      missingNodesError: record,
+    },
+  };
+  const factory = new Function(
+    "getPiniaStore",
+    "isStaleAssetCandidate",
+    "resolveMissingModelDirectory",
+    "withoutFrontendVirtualTypes",
+    "collectAllGraphs",
+    "getGraphCtx",
+    `return (${COLLECTOR_BODY});`,
+  );
+  const collect = factory(
+    (name) => stores[name],
+    () => false,
+    (d) => d,
+    withoutFrontendVirtualTypes,
+    (g) => (g ? [g] : []),
+    () => {
+      if (!readableGraph) throw new Error("no graph");
+      return { rootGraph: { _nodes: nodes } };
+    },
+  );
+  const r = collect();
+  return { nodeTypes: r.nodeTypes, nodeCount: r.nodeCount, any: r.any };
+}
+
+test("#1370 A: every recorded type proven frontend-virtual — the count goes with the names", () => {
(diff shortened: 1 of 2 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**. Verdict: **SHIP**, head `3c6db966`.

**The mechanism is real, confirmed by execution.** The gate sliced the shipped bytes — `collectMissingAssets`, both consumers' guard blocks, the `clean` computation and the `missing_node_count` spread — literally out of `web/js/comfyui-mcp-panel.js` for `origin/main` and for this PR, and ran them. Given a store saying `hasMissingNodes:true, missingNodeCount:2` with a record the parser names nothing from, **main emits `note: "no errors recorded since the last execution start"`** and the banner stays silent, while this PR emits `missing_node_count: 2` and "2 node type(s) not installed". The collapse is genuine in code.

**Correcting the severity, which I overstated when I reopened #1370.** I wrote that this was live on `main` right now. Against the installed frontend (1.48.7) that is **not** established: `missingNodeCount` is a computed `() => missingNodesError?.nodeTypes.length ?? 0`, `setMissingNodeTypes([])` nulls the record so `hasMissingNodes` goes false, and both producers stamp a non-empty `type` (`last_serialization?.type ?? type ?? "Unknown"`). So the unnamed-missing-nodes cases are **not constructible from that frontend**. This is a latent fail-unsafe in a deliberately defensive fallback — the one `778bc04d` added so that "packs are missing" is never silently lost — not an observed regression.

**It is still not churn, and here is the concrete reason.** One case *is* reachable on main, through the consumers' own fail-closed `catch`: when the guard block throws, #1380's zeroing is skipped and main emits `missing_node_count: 2` with no names plus the banner line — #1370's exact symptom. This PR's **collector** placement survives that catch and is clean on both surfaces. That is the argument for fixing it a layer down rather than at the consumers.

**Over-strip: none found.** Both-placeholders, one-virtual-one-real, unreadable-graph, and a recorded type with *no* live instance all keep count and names on both refs — `withoutFrontendVirtualTypes` requires positive per-instance proof, so a type with zero instances is never vacuously dropped. The #1332 now-registered case is byte-identical on both refs.

**Capture ordering is correct and pinned:** `adjudicateRecordedMissingNodeTypes` and `withoutClientRegisteredTypes` never mutate in place, so `recordedTypes` holds the pre-narrowing array.

**Mutations: seven for seven**, each confirmed applied via non-empty `git diff --numstat` on the CRLF file, each killing the semantically right test — unconditional-zero (RED on two), delete the `recordedTypes` capture, **move the capture after the narrowing assignment** (sharper than the author's deletion, and they had not run it), delete the collector gate, drop the `!unnamedRecordCount` clause, ungate `graph_get_errors`, ungate the banner. The four key ones re-run RED on a tree merged with current `main`.

**Composition with #1380 checked:** merges cleanly onto current `main`, no ungated `if (!X.nodeTypes.length) X.nodeCount = 0` survives anywhere, and the two narrowings compose idempotently rather than contradicting. Merged tree: 5519 pass / 0 fail, `check-panel-scope` OK.

**Two reporting caveats, neither blocking.** The consumer guards' *behaviour* is never executed by the suite — only their source text is regex-asserted, which is this repo's house style, so the gate executed them itself. And the Playwright evidence carries no signal here: **zero of 32 e2e specs** mention `missing_node_count`, `missingNodesError` or `isVirtualNode`, so "identical failing set" is a no-regression smoke rather than verification of this path.

PR #1401, opened 2026-08-19
fix(1351): bound a refresh RUN from its caller, not just the join

### LINKED ISSUE #1351: Bound a refresh RUN from its caller: joinMs bounds the join only, leaving a ~33s worst case against the 30s relay window
Split out of #1192, where an independent merge gate confirmed this residual is real and found it had no tracking issue.

**The gap.** `joinMs` bounds the *join* — how long a caller waits for a refresh already in flight — but never the run that follows it. Measured at 8.0x its bound in one case (2,251 ms against a 280 ms bound). With shipped numbers, a join can end at 20,000 ms and the caller's own run then adds roughly 13,000 ms, so the worst case is **~33,000 ms against the 30,000 ms relay window** (`OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS`, `src/orchestrator/panel-tools.ts:821`, applied to `graph_add_node` at `:9650`). The failing path is the one where the join *succeeds* near its bound — the budget is respected at every step and the command still outlives the window.

**Why #1193 does not cover it.** #1193 is closed and its fix (#1346) is merged, but that change redistributed budget *inside* a run: it floors the combo phase so a slow-but-healthy fetch cannot starve it. `NODE_DEFS_RUN_BUDGET_MS` is still 9000, and `registerNodesFromDefs` remains explicitly "unbounded local work, EXCLUDED from this budget" (`web/js/comfyui-mcp-panel.js:912`). So the residual survives both #1192's fix and #1193's.

**What closing this actually requires:** bounding a refresh *run* from its caller, which is a different shape from every budget added so far — those all bound waits the caller performs, whereas this needs the callee's own work to be interruptible or accounted for. That is why it was deliberately left out of #1192's stacked pair (panel PRs #1342 and #1349) rather than bolted on.

**Please avoid the tempting non-fix.** Raising `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS` past 33,000 ms makes the symptom go away without making "this run is pathologically slow" distinguishable from "the window was always too narrow for a healthy run" — the same collapse #1192 and #1193 were each filed for, moved one level out. The third time this recurs it will be at the next boundary up.

Confirmed by construction during the #1342/#1349 gate, not observed in the wild — so there is no user report attached to it yet.

artokun:
## Root cause

`joinMs` (#1192/#1342) bounds only the **join** — the wait on a refresh run someone else started. A join that *succeeds* near its bound is followed by the caller's **own** run, whose deliberately-unbounded local work (`registerNodesFromDefs`, ~4s measured and worse on larger installs) sat outside every bound on the add path. Measured composition from the issue: a join ending at its 20,000 ms bound plus a ~13,000 ms own run sums to ~33,000 ms against the 30,000 ms relay window (`OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS`) — the budget respected at every step, the command still outliving the window. Both `return startRun(preloadedDefs)` sites in `web/js/lib/refresh-coalesce.js` awaited the caller's own run unboundedly.

## Fix

New `opts.runMs` on the coalescer, bounding the caller's wait on the run **it** started:

- Measured **from the call**, not from the run's start: the join's cost is subtracted from it (and it also caps the join), so `joinMs` and `runMs` handed the same remaining budget bound the *total* wait instead of doubling it.
- The run is **never dropped**: on abandonment the slot is free (the join settled, or this is the leading edge), so the run starts anyway, registers the payload in the background, and the caller's retry finds the class registered. Only the *wait* ends, with a new `REFRESH_RUN_ABANDONED` sentinel — distinct from `REFRESH_JOIN_ABANDONED`, because the payload was not dropped and the wording must say so.
- An abandoned own run attaches a catch guard so its later failure cannot become an unhandled rejection; a run that rejects *inside* the bound still rejects for its caller.
- Unwired `withTimeout`/`now` or non-finite/`<= 0` `runMs` degrade exactly as `joinMs` does (unbounded-as-before / abandon-immediately), never silently restoring an unbounded wait.

Wired in `web/js/comfyui-mcp-panel.js`: the coalescer gets `now: monotonicNow`; both `graph_add_node` refresh sites (the resolver's payload refresh and the #1242 drift recovery's forced call, which meets the unbounded **leading edge**) pass `runMs: budget.remaining() - ADD_NODE_POST_REFRESH_RESERVE_MS`; a run-abandoned add refuses with a new `addNodeRefreshSlowMessage` (nothing added, refresh still running, retry — never "reload the tab"), re-worded only when the class is *still* unregistered.

Deliberately **not** done, per the issue: raising `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS`, which would make a pathologically slow run indistinguishable from a healthy one.

## Verification

- `npm ci` — clean
- `npm run test:unit` — **5506/5506 pass** (was 5506 total before; +10 new #1351 coalescer tests, +1 end-to-end harness test driving the shipped `graph_add_node` through the real coalescer: join lands at ~260ms of a ~280ms bound, own run outlives the rest, add refuses in words *inside* the budget, payload still registers in the background)
- `npm run typecheck` — clean
- `npm run check:scope` — OK
- Wiring pins extended in `single-node-def.test.mjs` (runMs at *both* call sites, `now: monotonicNow`, flag/message shape); harnesses rebuilding the executor (`_panel-constants.mjs`, `add-node-command-budget.test.mjs`) given the new bindings.

Follow-up worth a separate look: `refresh_nodes` itself (`GRAPH_TOOL_EXECUTORS.refresh_nodes`) still awaits its forced run unbounded against the same 30s ack timeout — wiring `runMs` there needs a decision about what verdict the tool reports on abandonment.

Closes #1351

### DIFF
--- a/browser_tests/unit/_panel-constants.mjs
+++ b/browser_tests/unit/_panel-constants.mjs
@@ -14,7 +14,7 @@ import { readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
 
 import { makeCommandBudget } from "../../web/js/lib/command-budget.js";
-import { REFRESH_JOIN_ABANDONED } from "../../web/js/lib/refresh-coalesce.js";
+import { REFRESH_JOIN_ABANDONED, REFRESH_RUN_ABANDONED } from "../../web/js/lib/refresh-coalesce.js";
 import { clearInheritedExecutionPreview } from "../../web/js/lib/execution-preview-attach.js";
 
 export const PANEL_SRC = readFileSync(
@@ -139,12 +139,19 @@ export function addNodeCommandBudgetDeps() {
     ADD_NODE_POST_REFRESH_RESERVE_MS: CUSTOM_WIDGET_REGISTRATION_TIMEOUT_MS,
     OBJECT_INFO_SEED_WAIT_MS,
     REFRESH_JOIN_ABANDONED,
+    // #1351 — the second sentinel, returned when the caller stopped waiting for its OWN
+    // run. Reached only when a harness's refresh stub resolves it; the real coalescer's
+    // behaviour is exercised in refresh-coalesce.test.mjs.
+    REFRESH_RUN_ABANDONED,
     widenSocketProofBudget,
     // A STUB, and labelled as one. These three harnesses are about #821/#1223/#620, none of
     // which reaches this branch. The shipped wording is pinned in single-node-def.test.mjs
     // and the behaviour that produces it is exercised in add-node-command-budget.test.mjs;
     // a hand-copied sentence here would just be a third place for it to drift.
     addNodeRefreshBusyMessage: (classType) =>
+      `HARNESS STUB refusal for "${classType}" — the shipped wording lives in the panel.`,
+    // #1351 — the own-run refusal, a stub for the same reason as the busy message above.
+    addNodeRefreshSlowMessage: (classType) =>
       `HARNESS STUB refusal for "${classType}" — the shipped wording lives in the panel.`,
   };
 }
--- a/browser_tests/unit/add-node-command-budget.test.mjs
+++ b/browser_tests/unit/add-node-command-budget.test.mjs
@@ -22,7 +22,7 @@ import { readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
 
 import { withTimeout } from "../../web/js/lib/bounded-step.js";
-import { makeRefreshCoalescer, REFRESH_JOIN_ABANDONED } from "../../web/js/lib/refresh-coalesce.js";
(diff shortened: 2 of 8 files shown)


artokun:
Superseded by #1399 (merged as 1336f2f3), which fixed #1351 while this PR was in flight — same root cause, same composition property (`joinMs` is now a deadline for the whole invocation, so the join's cost is subtracted from the own run's allowance rather than added to it), same start-the-run-but-stop-waiting semantics on all three paths (post-join payload run, forced trailing, leading edge). Issue #1351 is closed as completed, and the relevant suites are green on main (`refresh-coalesce.test.mjs` + `add-node-command-budget.test.mjs`: 37/37 at origin/main 34420f20).

This PR's only remaining deltas are stylistic — a separate `runMs` option and `REFRESH_RUN_ABANDONED` sentinel instead of widening `joinMs` — and landing them now would be a parallel mechanism, not a fix. Closing.

Issue #1404, opened 2026-08-19
panel_refresh_nodes can time out once after workflow targeting even though the tab is active


artokun:
### What happened / root cause
After a workflow switch and successful panel_set_workflow_target({mode:"current"}), panel_refresh_nodes was dispatched to the active workflow tab but received no acknowledgement for 30 seconds. ComfyUI itself remained healthy and idle. Re-running panel_set_workflow_target reported already_current, and the identical panel_refresh_nodes call then succeeded. The evidence points to a dropped or unhandled refresh_nodes acknowledgement during the workflow-target transition; the orchestrator currently waits for the full timeout and has no idempotent recovery path.

### Steps to reproduce
1. Switch the active ComfyUI workflow tab.
2. Call panel_set_workflow_target({mode:"current"}).
3. Upload a new image through the ComfyUI server API.
4. Call panel_refresh_nodes.
5. Observe a 30-second no-reply timeout even though the tab is active.
6. Rebind to current again and retry panel_refresh_nodes; it succeeds.

### Exact error (scrubbed)
Error: Panel tab <tab-id>:workflows/Untitled.json did not reply to "refresh_nodes" within 30000 ms — the ComfyUI tab may be backgrounded or frozen.

### Fix
Applied locally: no. Upstream-only: yes.
Verified workaround: call panel_set_workflow_target({mode:"current"}) and retry panel_refresh_nodes; the retry returned {ok:true, refreshed:true}.
Precise proposed change: because refresh_nodes is idempotent, on an acknowledgement timeout revalidate the active workflow instance/fence and retry the dispatch once, or return a structured retry token that preserves the target. Also distinguish a genuinely backgrounded/frozen tab from a lost acknowledgement when the same tab still reports current.

### Environment
Windows 11; ComfyUI 0.33.0; ComfyUI frontend 1.49.6; Python 3.13.11; PyTorch 2.13.0+cu130; NVIDIA GeForce RTX 5090 32 GB; comfyui-mcp 0.51.31; panel 0.14.24.

artokun:
Claimed by the backlog-drain session (worktree `.claude/worktrees/drain-1404`). Verified unclaimed with no open PR immediately before claiming. Note this may interact with the workflow-stamp work that merged today (mcp #1775 / v0.52.6) and with mcp#1778 — the agent has been told to check that first. If no PR appears within ~8h, consider the claim lapsed.

artokun:
[...]

PR #1409, opened 2026-08-19
fix(1404): refresh_nodes takes a command budget, so a contended refresh replies instead of timing out

artokun:
The underlying cause of #1404 is a reply that was **never composed inside the relay window** — not an acknowledgement lost in transit.

## The mechanism

`panel_refresh_nodes` relays `refresh_nodes` at comfyui-mcp's `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS` = **30,000 ms** (`src/orchestrator/panel-tools.ts`, the `panel_refresh_nodes` call site) — which is the exact number in the report, good corroboration that it reached dispatch and waited for an ack.

**Six** commands are relayed in that 30 s window — `graph_get_object_info`, `graph_add_node`, `graph_set_widget`, `graph_remove_widget` and `refresh_nodes` through `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS`, plus `nodes_install` at a literal `30000`. Only **two** of the six held a whole-command budget: `graph_add_node` (#1192) and `nodes_install` (#671). **`refresh_nodes` did not** — and its whole body *is* a wait on the node-def refresh. Its executor was:

```js
```

No `joinMs`. `makeRefreshCoalescer`'s forced branch then reads `joinMs === null` and `waitForRun` returns the run unbounded, so the call pays, serially:

1. whatever is left of a run **someone else started**, then
2. the trailing run it queues behind it.

**The evidence that two runs do not fit is a note this repo already wrote, not arithmetic of mine.** `NODE_DEFS_RUN_BUDGET_MS`'s own docstring in `comfyui-mcp-panel.js` says `makeRefreshCoalescer` *"guarantees a forced `panel_refresh_nodes` pays the in-flight run AND its own, serially, so about 39.6s before a reply is composed. Past the bridge's 20s READ default and past the 30s command budget both."* #1180 shrank the per-run WAITING that figure was computed from, but it explicitly did **not** bound `registerNodesFromDefs` (3,972 ms measured) or `reapplyDefsToLiveNodes` — so the composition that note describes survived it, which is exactly what this issue reports. `ADD_NODE_COMMAND_BUDGET_MS`'s note says the same of its own path ("~33 s against the 30 s relay window, with every per-step bound respected"). #1192/#1351 fixed the composition for the add and left the tool whose whole body *is* that composition unbounded.

(An earlier draft argued this as "2 × ~14.5 s ≈ 29 s". That is the weakest form of the argument — 29 s is *inside* 30 s — so it is retired in favour of the pre-existing note above.)

**Why the second call succeeds** — the part that identifies the cause rather than merely fitting it: by then the trailing run has settled and the coalescer's slot is free, so the retry pays for ONE run instead of two. Nothing about the tab, the socket, or the fence differs between the two calls; only the number of runs in front of them does.

**Why *this* sequence** — the reporter's step 3 is an image uploaded through the ComfyUI server API. Two forced, fire-and-forget refreshes are keyed on `hasRawMissingAssetCandidates()` (the validation banner's, and `graph_get_errors`'), and a just-uploaded input file is exactly what makes that predicate true. The upload that gives an agent a reason to call this tool is the same event that makes the panel start its own refresh — the two race by construction, and a workflow switch re-renders the banner that starts it.

## Relationship to mcp#1656 (PR #1775, v0.52.6) and mcp#1778 (PR #1782)

Same window, different failure mode — established rather than assumed:

- Both of those are **pre-dispatch fence refusals**: they end in a rejected promise carrying `workflow instance mismatch: …`, produced by `UiBridge.send` *before* `sock.send()`. A refusal is a reply. The reported failure is `markReplyTimeout(markDispatched(…, true))` from `dispatch()`'s reply timer — a POST-write timeout, reachable only after the frame was written and nothing answered for the full 30,000 ms. The two paths cannot produce each other's error.
- Nothing in #1775 or #1782 touches the coalescer, the node-def refresh, or any budget: they change `carryWorkflowCommandStamp` provenance and the save path's corroboration. This branch touches none of theirs.
- What they genuinely share is the *shape* — "a workflow-target transition, healed by a second call" — and that shape is a coincidence of trigger, not of cause: here the second call heals it because the coalescer's slot is free, not because a stamp became provable.

The reporter is on panel **0.14.24**; main is 0.15.5. I checked whether the gap already covered this: it does not. #1161/#1180 (v0.14.35) bounded the *phases inside a run*, and #1192/#1351 bounded *`graph_add_node`'s* wait on the coalescer — neither reaches this call site, which is unbounded on current `main`.

## Why the recovery is safe

**This PR adds no retry.** The standing rule is never to authorise a re-run on parsed prose, so the fix is to make the panel *answer inside the window* rather than to re-issue anything:

- The panel now bounds its own wait at 25,000 ms and returns `{ok:true, refreshed:false, reason:"refresh_still_running", detail, remedy}`.
- `reason` is a **structured token** in the existing `NODE_DEF_REFRESH_REASONS` vocabulary, not wording — a caller keys on the field, never on the sentence.
- Nothing was applied and nothing is half-done: `refresh-coalesce.js`'s `waitForRun` does not cancel the run, so it keeps registering the definitions the call asked for and keeps the slot. A retry therefore joins work already in progress rather than starting a stampede — which is *why* the remedy says retry, and why the reporter's own retry worked.
- The retry is safe on its own terms too, and by construction rather than by claim: `refresh_nodes` re-registers node defs and rebuilds combo option lists, touches no graph, and is undo-neutral (its tool description already says "Idempotent (safe to call repeatedly)").
- Without the branch, an abandoned wait would fall through to `reason: "unknown"` — collapsing "still running" into the same answer as "the fetch threw". A test pins both directions: a genuinely failed run still reports `object_info_fetch_failed`.

**What it costs, stated rather than left to be discovered:** a single *uncontended* run that would have taken between 25 s and 30 s now reports `refresh_still_running` where it used to report `refreshed:true`. That window is 5 s wide, past twice the #610 measurement, and the run is not cancelled. In exchange every contended call — the reported bug — gets a worded verdict instead of `did not reply … the ComfyUI tab may be backgrounded or frozen` said about a demonstrably healthy tab.

- **No retry was added anywhere**, orchestrator-side or panel-side. The issue asks for an “idempotent recovery path”; re-issuing a command because the first timed out is only safe when you can prove nothing was applied, and the better fix here is the one that makes the reply arrive.
- **The orchestrator's timeout wording is untouched.** `did not reply … the ComfyUI tab may be backgrounded or frozen` is a false claim about this case, but with the reply now composed inside the window it is no longer reached for this cause, and rewording it touches a message a dozen tests assert on.
- **Adjacent gap, filed rather than folded in — and it IS made worse in one branch.** `panel_add_node`'s automatic `panel_refresh_nodes` recovery (comfyui-mcp `src/orchestrator/panel-tools.ts`) branches only on `refreshed.isError`, and a `{ok:true, refreshed:false}` reply is not an error. So the effect of this change on that path is **net mixed, tilted good** — not "unchanged", which is what an earlier draft claimed:
  - **Better, the common branch:** where the recovery used to burn the full 30 s and then fail, it now gets an answer in ≤ 25 s, and when the contending run settles inside that window the second `add()` simply succeeds.
  - **Worse, one branch:** if the second `add()` still refuses, the orchestrator now tells the agent *"panel_refresh_nodes reported success and the add still refuses… Reload the ComfyUI browser tab"* — a false statement about a refresh that reported it did **not** complete, plus the destructive escalation the panel's own remedy deliberately reserves for last. Previously that path reached the honest `isError` branch after timing out.
  - The one-line fix belongs in the other repo (read the verdict, not the flag) and is filed as a follow-up rather than folded in here.

## Live reproduction

Real Chromium, real ComfyUI **0.33.1** on its own base directory at **:8399** (never :8188/:8211), panel served from this worktree, real coalescer, real executor. `/object_info` inflated from this install's 848 core types to **4,240 types / 7,087,848 bytes** so `registerNodesFromDefs`/`reapplyDefsToLiveNodes` — the two calls a run's budget deliberately does not bound — do the walk the reporter's 4,300-type install does.

| | measured |
|---|---|
| one `refresh_nodes`, uncontended | **727 ms** |
| a second `refresh_nodes` issued while the first is running | **1,261 ms** (1.73×) |

That is the composition, live: the contended call pays both runs. Then, with the budget temporarily set to 400 ms in the same live rig, the contended call returned in 803 ms with `{"ok":true,"refreshed":false,"reason":"refresh_still_running",…}` **and the first run still completed `refreshed:true`** — so the bound fires through the shipped call site in a real browser, `withTimeout` is genuinely wired into the coalescer, and an abandoned wait does not cancel the run. The constant was restored (`git diff` clean) before anything was committed.

**What the live rig could NOT show:** the absolute 30 s overrun. On a blank canvas the two unbounded local calls have almost nothing to walk, so two runs cost ~1.3 s here where they cost ~29 s on the reporter's rig. The wall-clock term comes from a populated canvas on a heavy install, and this repo has that measured in-source (#610: ~14.5 s per forced refresh; #1180: `registerNodesFromDefs` 3,972 ms, `refreshComboInNodes` 4,846 ms at 4,320 types). I did not manufacture a number for it.

## Refutation

Every mutation was confirmed applied with `git diff -U0` before the run.

| mutation | anchor confirmed applied | result |
|---|---|---|
| delete `joinMs: REFRESH_NODES_COMMAND_BUDGET_MS` from the call site | `-      joinMs: REFRESH_NODES_COMMAND_BUDGET_MS` | **5 of 8 fail** — the four behavioural tests hit their watchdog at 1,500 ms instead of replying at the bound, plus the source-level call-site pin |
| `if (verdict === REFRESH_JOIN_ABANDONED)` → `if (false && …)` | `+    if (false && verdict === REFRESH_JOIN_ABANDONED) {` | **2 of 8** — both named-verdict tests fall through to `reason:"unknown"` |
| delete `REFRESH_STILL_RUNNING` from `NODE_DEF_REFRESH_REASONS` | `-  REFRESH_STILL_RUNNING: "refresh_still_running",` | **2 of 8** — and **0 of 67 before this round**; see below |
| `REFRESH_NODES_COMMAND_BUDGET_MS = 25000` → `35000` | `+const REFRESH_NODES_COMMAND_BUDGET_MS = 35000;` | **1 of 8** — the budget no longer leaves the relay window room |
| drop the “or this command's own registration” half of the detail | `+… after an upload). ` | **1 of 8** — see the re-review note below |

An earlier version of this table said 4 and “exactly 1” for the first two rows. It was stale — written before the eighth test existed and not re-measured — and the body asserted the numbers were confirmed. Every row above was re-run against the current file, and the mutation script **refuses on a non-unique anchor** rather than picking: that guard fired on the second row, where `if (verdict === REFRESH_JOIN_ABANDONED) {` matches **two** sites (`refresh_nodes` at :10781 and `graph_add_node`'s #1242 recovery at :12548). Mutating the wrong one would have measured nothing and read as a surviving mutant.

**The token was genuinely unpinned, and the gate caught it.** Deleting `REFRESH_STILL_RUNNING` from the frozen map killed **zero** of the 67 tests across all three refresh harnesses: the panel and the tests read the token from the *same* map, so both degraded to `undefined` together and `assert.equal(undefined, undefined)` passed. The shipped reply would have carried `reason: undefined` — a key `JSON.stringify` **drops outright**, so the field a caller keys on vanishes from the wire with every assertion green. Verified, not reasoned about:

```
reason value : undefined
ON THE WIRE  : {"ok":true,"refreshed":false,"remedy":"retry"}
```

Both call sites now assert the **literal** `"refresh_still_running"` alongside the map lookup, and the map entry itself is asserted to equal that literal. The same mutation now fails 2.

**Found by adversarial re-review of my own diff, not by a test.** The first draft's `detail` said the budget went on “a node-def refresh started by something else.” The coalescer returns the SAME symbol whether the wait was on someone else's run or on the run this command started itself (nothing in flight ⇒ `refresh`'s last branch), so that sentence is flatly wrong on an uncontended big install — the case with no concurrency at all. The value cannot tell them apart, so the message names both, as `addNodeRefreshBusyMessage` does. An eighth test drives that uncontended branch behaviourally (`runs.length === 1`), and the `reload` escalation is pinned to appear only AFTER the retry and only conditioned on it.

The tests drive the **shipped `refresh_nodes` body extracted from the panel source** over the **real** `makeRefreshCoalescer` with a **real** in-flight run (the `add-node-command-budget.test.mjs` technique). A helper-level test is blind here by construction: `makeRefreshCoalescer` has implemented `joinMs` correctly since #1192/#1351 and the executor already handled a non-fresh verdict — the entire defect was the call site passing no bound.

Adding the budget made **three** synthetic-scope harnesses throw `ReferenceError` at once, which is the hazard `_panel-constants.mjs` already documents for `graph_add_node`; the bindings now live in one place (`REFRESH_NODES_EXECUTOR_DEPS`) so the next one is added once rather than three times.

## Test evidence

- `npm run test:unit` (i18n check, i18n-render check, tool-vocabulary gate, panel-scope gate, then `node --test`): **5,591 pass / 0 fail / 1 todo** (rebased onto `origin/main` at 09432738, after 0.15.6 was cut).
- `npx tsc --noEmit`: clean.
- `node scripts/check-panel-scope.mjs`: OK.
- **Playwright, `--workers=1`, against the isolated :8399 ComfyUI: 60 passed / 18 failed (78).** All 18 are chat-transcript specs (`chat-history-v2`, `conversation-persistence`, `workflow-chat-identity`, `streaming-render`, `pending-queue`, `hidden-tab`, `external-orchestrator`), all failing on `MockBridge.waitForUserMessage timed out`. **Control:** the same seven spec files re-run with the worktree detached at `origin/main` (fix absent — `grep -c` = 0) fail **the same 18, by name**. They are this fresh isolated base directory's baseline, not a regression. Nothing in the refresh/add-node/errors path failed in either run.

## Gate round 2 — what changed

Three findings, all upheld and all fixed at head `9e26021e`:

1. **A false claim in shipped source and in the CHANGELOG.** “The THIRD command relayed in the 30,000 ms window and the last one that never took one” — it is six, and two of them held budgets. Both corrected by counting the relay sites rather than guessing, and the source comment now records that the false version *concealed* the sibling below. This is the reassurance-shaped claim this project keeps getting caught by.
2. **The reason token was unpinned** — deleting it killed 0 of 67. Fixed with literal assertions; the same mutation now fails 2.
3. **“Not made worse” was refuted in one branch** of `panel_add_node`'s recovery. Restated as net mixed, tilted good, with both branches named.

Plus the stale mutation table, re-measured against the current 8-test file.

**Not mine, credited to the gate:** `graph_set_widget`'s stale-combo recovery awaits `refreshComfyNodeDefs()` with no `joinMs` and no outer bound (`comfyui-mcp-panel.js:13919`, awaited at `web/js/lib/set-widget.js:820`) inside a command relayed at the same 30,000 ms. Lesser than this one — the plain-join branch returns after ONE run rather than queueing a second — but reached only after the authorization `/object_info` has already spent part of the window. That is the **fourth** instance of this shape (#1192 → #1349 → #1404 → this). Filed separately by the gate; deliberately not folded in here.

## Gate

`codex exec` is exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the weekly limit, so **this PR is ungated** and I was instructed not to run a gate script. Not merged, per instruction — gating and merging are the owner's.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+- `panel_refresh_nodes` no longer times out with no acknowledgement when a node-def refresh is already running: `refresh_nodes` was the one command relayed in the orchestrator's 30 s window that never took a command budget, so a forced call waited unbounded on the in-flight run AND on its own trailing run — past the window, so the reply the panel's bounds exist to produce never left the tab and the user got `did not reply … the ComfyUI tab may be backgrounded or frozen` about a healthy, idle tab. It now bounds that wait at 25 s (the same budget `graph_add_node` and `nodes_install` hold against the same window) and answers `refreshed:false` with `reason:"refresh_still_running"` and a retry that works — the abandoned run is not cancelled, so it is still registering the definitions the call asked for (#1404)
+
 ## [0.15.6] - 2026-08-19
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,13 @@ All notable changes to this project are documented here. This project adheres to
 ## [Unreleased]
 
 ### Fixed
-- `panel_refresh_nodes` no longer times out with no acknowledgement when a node-def refresh is already running: `refresh_nodes` was the one command relayed in the orchestrator's 30 s window that never took a command budget, so a forced call waited unbounded on the in-flight run AND on its own trailing run — past the window, so the reply the panel's bounds exist to produce never left the tab and the user got `did not reply … the ComfyUI tab may be backgrounded or frozen` about a healthy, idle tab. It now bounds that wait at 25 s (the same budget `graph_add_node` and `nodes_install` hold against the same window) and answers `refreshed:false` with `reason:"refresh_still_running"` and a retry that works — the abandoned run is not cancelled, so it is still registering the definitions the call asked for (#1404)
+- `panel_refresh_nodes` no longer times out with no acknowledgement when a node-def refresh is already running: of the six commands relayed in the orchestrator's 30 s window only `graph_add_node` and `nodes_install` held a command budget, and `refresh_nodes` — whose whole body IS a wait on the node-def refresh — did not, so a forced call waited unbounded on the in-flight run AND on its own trailing run — past the window, so the reply the panel's bounds exist to produce never left the tab and the user got `did not reply … the ComfyUI tab may be backgrounded or frozen` about a healthy, idle tab. It now bounds that wait at 25 s (the same budget `graph_add_node` and `nodes_install` hold against the same window) and answers `refreshed:false` with `reason:"refresh_still_running"` and a retry that works — the abandoned run is not cancelled, so it is still registering the definitions the call asked for (#1404)
 
 ## [0.15.6] - 2026-08-19
 
 ### Fixed
 - `panel_query_graph` (fields:'detail') no longer last-wins-collapses widgets that share a name: when a name repeats (rgthree Fast Groups toggle rows), every occurrence is listed so a duplicate or orphaned row is visible (#1402) (#1406)
 - `duplicate_widgets` (the #1402 duplicate-widget report) no longer throws the whole detail read away on a widget named `__proto__`/`constructor`/`toString` — occurrences are accumulated prototype-safely — and is now bounded by the same `max_chars` budget as `widgets` (dropped occurrences are announced with the lever that lifts them, never silently lost), so a many-group Fast Groups Bypasser cannot push the detail line into a whole-row stub. `panel_graph_outline` also labels each same-named row from the widget itself rather than a last-wins name-keyed map, so two different group toggles are no longer both annotated with the last row's label (#1402) (#1405)
-
 
 ## [0.15.5] - 2026-08-19
 
--- a/browser_tests/unit/_panel-constants.mjs
+++ b/browser_tests/unit/_panel-constants.mjs
@@ -15,6 +15,7 @@ import { fileURLToPath } from "node:url";
 
 import { makeCommandBudget } from "../../web/js/lib/command-budget.js";
 import { REFRESH_JOIN_ABANDONED } from "../../web/js/lib/refresh-coalesce.js";
+import { NODE_DEF_REFRESH_REASONS } from "../../web/js/lib/node-def-refresh.js";
 import { clearInheritedExecutionPreview } from "../../web/js/lib/execution-preview-attach.js";
 import { sanitizeNodeAuxId } from "../../web/js/lib/aux-id-sanitize.js";
 
@@ -67,6 +68,25 @@ export const REFRESH_NODES_COMMAND_BUDGET_MS = readPanelNumber(
   /const REFRESH_NODES_COMMAND_BUDGET_MS = (\d+);/,
   "the refresh_nodes command budget",
 );
+
+/**
+ * #1404 — every module binding the `refresh_nodes` executor closes over, BESIDES the
+ * `refreshComfyNodeDefs` each harness supplies its own double for.
+ *
+ * THREE harnesses rebuild that executor in a synthetic scope (node-def-refresh,
+ * refresh-graph-guard, refresh-nodes-command-budget), so a new free identifier in it throws
+ * `ReferenceError` in all three at once — which is exactly how adding the budget was found.
+ * Collected here for the same reason `addNodeCommandBudgetDeps` exists: so the next binding
+ * is added once rather than three times.
+ *
+ * The REAL values — the symbol and the reason vocabulary are imported and the number is read
+ * from the panel source — so no harness can pass against a value the panel no longer holds.
+ */
+export const REFRESH_NODES_EXECUTOR_DEPS = Object.freeze({
+  REFRESH_JOIN_ABANDONED,
+  NODE_DEF_REFRESH_REASONS,
+  REFRESH_NODES_COMMAND_BUDGET_MS,
+});
 
 export const NODE_DEFS_FETCH_TIMEOUT_MS = readPanelNumber(
   /const NODE_DEFS_FETCH_TIMEOUT_MS = (\d+);/,
--- a/browser_tests/unit/_panel-constants.mjs
+++ b/browser_tests/unit/_panel-constants.mjs
@@ -62,6 +62,12 @@ export const ADD_NODE_COMMAND_BUDGET_MS = readPanelNumber(
   "the add_node command budget",
 );
 
+/** #1404 — the whole-command deadline `refresh_nodes` hands the coalescer as `joinMs`. */
+export const REFRESH_NODES_COMMAND_BUDGET_MS = readPanelNumber(
+  /const REFRESH_NODES_COMMAND_BUDGET_MS = (\d+);/,
+  "the refresh_nodes command budget",
+);
+
 export const NODE_DEFS_FETCH_TIMEOUT_MS = readPanelNumber(
   /const NODE_DEFS_FETCH_TIMEOUT_MS = (\d+);/,
   "the single-call fetch bound",
--- a/browser_tests/unit/add-node-schema-drift-recovery.test.mjs
+++ b/browser_tests/unit/add-node-schema-drift-recovery.test.mjs
@@ -73,7 +73,9 @@ test("refresh_nodes really does re-register the class this check reads", () => {
   // chain: refresh_nodes -> refreshComfyNodeDefs(force) -> getNodeDefs +
   // registerNodesFromDefs.
   const fn = PANEL.slice(PANEL.indexOf("async refresh_nodes() {"), PANEL.indexOf("graph_serialize() {"));
-  assert.ok(fn.includes("refreshComfyNodeDefs(undefined, { force: true })"));
+  // #1404 gave the call a `joinMs`, so it no longer fits on one line. Matched as the chain
+  // this test is about — a FORCED call into the coalescer — rather than as a formatting.
+  assert.match(fn, /refreshComfyNodeDefs\(undefined, \{[\s\S]*?force: true,?[\s\S]*?\}\)/);
   const register = PANEL.slice(
     PANEL.indexOf("async function registerComfyNodeDefs"),
     PANEL.indexOf("const refreshComfyNodeDefs = makeRefreshCoalescer"),
--- a/browser_tests/unit/node-def-refresh.test.mjs
+++ b/browser_tests/unit/node-def-refresh.test.mjs
@@ -26,6 +26,7 @@ import {
   NODE_DEFS_FETCH_TIMEOUT_MS,
   NODE_DEFS_NO_ANSWER,
   NODE_DEFS_RUN_BUDGET_MS,
+  REFRESH_NODES_EXECUTOR_DEPS,
   monotonicNow,
   nodeDefsBudgetLeft,
 } from "./_panel-constants.mjs";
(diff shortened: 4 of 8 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**. Verdict: **SHIP**, with a punch-list the author has now cleared. Merging at `9e26021e`.

**The issue's own premise was refuted, and correctly.** It reported "a dropped or unhandled acknowledgement". No ack was lost — the reply was never composed. The string comes from `markReplyTimeout(markDispatched(…, true))` at `src/services/ui-bridge.ts:4693-4697`: `dispatched: true`, the post-write reply timer. On `origin/main` the call site was `refreshComfyNodeDefs(undefined, { force: true })` with no `joinMs`, so it was necessarily unbounded at the reporter's 0.14.24 too and no version gap could have fixed it.

**The strongest evidence is a pre-existing note neither author wrote:** `NODE_DEFS_RUN_BUDGET_MS`'s own comment already states a forced `panel_refresh_nodes` pays the in-flight run and its own, "so about 39.6s before a reply is composed. Past … the 30s command budget." The author's original 2 × 14.5 = 29 s arithmetic was the weakest form of the argument (29 s is *inside* 30 s) and has been retired in favour of that.

**The bound does not cancel the run**, which is what makes the recovery honest rather than destructive — confirmed structurally (`waitForRun` attaches `run.catch(()=>{})` and races `withTimeout`; `startRun` clears the slot in its own `finally`) and behaviourally (with the budget cut to 400 ms the contended call returned in 803 ms while the first run still completed `refreshed:true`). **No retry was added**: the panel answers inside the window with a structured token from the frozen `NODE_DEF_REFRESH_REASONS` vocabulary, and the only automatic re-run in the area is gated on the structured `refreshed.isError`, not on prose.

**Three findings, all fixed before merge:**

1. **A false claim in shipped source and the CHANGELOG, which actively concealed the next bug.** The comment called this "the THIRD command relayed in the 30,000 ms window and the last one that never took one". **Six** commands ride that window — `graph_get_object_info`, `graph_add_node`, `graph_set_widget`, `graph_remove_widget`, `refresh_nodes` via `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS`, plus `nodes_install` at a literal `30000` — and only `graph_add_node` (#1192) and `nodes_install` (#671) held budgets. The comment now opens "COUNTED, not guessed, because the first draft of this comment guessed and was wrong", names all six, and records that the false version hid the `graph_set_widget` sibling. Retaining the retracted sentence rather than deleting it is the right call.
2. **The new token was unpinned in a way the suite could not see.** Deleting `REFRESH_STILL_RUNNING` from the vocabulary left **67/67 green**, because panel and test read it from the same map and both degrade to `undefined` together — the wire result being `{"ok":true,"refreshed":false,"remedy":"retry"}` with the key a caller keys on simply gone. Both call sites now assert the literal and the mutation fails 2.
3. **"Not made worse" retracted, restated as net mixed and tilted good.** `panel_add_node`'s recovery branches on `refreshed.isError`, and `{ok:true, refreshed:false}` is not an error — so where the second `add()` still refuses, the orchestrator says "panel_refresh_nodes reported success… Reload the ComfyUI browser tab": false, plus the destructive escalation the panel's remedy reserves for last. The other branch is a real improvement. The mcp-side one-liner (read the verdict, not the flag) is a follow-up.

**The mutation table was stale and is re-measured:** dropping `joinMs` fails **5 of 8** (not 4); `if (false && …)` fails **2** (not 1); deleting the token now fails **2** (was 0 of 67); budget 35000 fails 1; dropping the own-run half fails 1.

**Two process saves worth recording.** The author's mutation script has a **refuse-on-non-unique-anchor guard, and it fired**: `if (verdict === REFRESH_JOIN_ABANDONED) {` matches two sites (`:10781` and `graph_add_node`'s #1242 recovery at `:12548`), so mutating the wrong one would have measured nothing and read as a surviving mutant. And the rebase onto 0.15.6 left a **stray superseded duplicate** of the #1402 entry under `[Unreleased]`, found by reading the resolved file rather than trusting the merge. Verified before merging: zero conflict markers, `[Unreleased]` holds exactly one entry, and it sits outside the released 0.15.6 block.

5,591 tests pass / 0 fail; `tsc --noEmit` clean. Playwright is from the pre-rebase run (60/78 with the 18 chat-transcript failures matched by an `origin/main` control); the post-rebase delta is comment-and-test-only.

**This is the fourth instance of one recurring shape** — an unbudgeted wait inside the 30 s relay window (#1192 → #1349 → #1404 → the `graph_set_widget` sibling, now filed as #1413).

Issue #396, opened 2026-07-31
Freshly downloaded models cannot be selected on live canvas until restart


artokun:
### What happened / root cause
After download_model completed new VAE and LoRA files, panel_set_widget rejected their filenames because the live combo options remained stale until a ComfyUI restart. This prevents selecting freshly downloaded models on an unsaved canvas where restarting risks losing work.

### Steps to reproduce
1. Start ComfyUI and open a live unsaved workflow.
2. Use download_model to place a new file in models/vae or models/loras.
3. Confirm download_status reports done.
4. Call panel_set_widget on VAELoader.vae_name or LoraLoaderModelOnly.lora_name.
5. The tool rejects the new filename and reports the startup-time combo options.

### Exact error (scrubbed)
panel_set_widget refused "vae_name": Value "qwen_image_vae.safetensors" is not a valid option. Valid options omit the newly downloaded file.
panel_set_widget refused "lora_name": valid options (0).

### Fix
Applied locally: no. Upstream-only: yes.
Please refresh ComfyUI object-info/model folder combo choices after download completion, or provide a non-destructive model-list refresh endpoint/tool that updates live node combo widgets without restarting ComfyUI.

### Environment
Windows 11 / ComfyUI 0.27.0 / RTX 4060 Laptop 8GB / comfyui-mcp 0.48.21 / panel 0.11.14.

PR #1415, opened 2026-08-19
fix(1413): graph_set_widget takes a command budget, so a contended stale-combo write replies inside the window

### LINKED ISSUE #1413: graph_set_widget's stale-combo recovery awaits refreshComfyNodeDefs unbudgeted inside the 30s relay window
Found by an independent gate on PR #1409, which fixed the same shape in `refresh_nodes`. **This is the fourth instance of one recurring defect** — a wait relayed inside the orchestrator's 30,000 ms window that never took a command budget: #1192 (`graph_add_node`'s fifth wait) → #1349 (the same, stacked) → #1404 (`refresh_nodes`) → this.

**The gap.** `graph_set_widget`'s stale-combo recovery awaits `refreshComfyNodeDefs()` with **no `joinMs` and no outer bound**:

- `web/js/comfyui-mcp-panel.js:13919`
- awaited at `web/js/lib/set-widget.js:820`

`graph_set_widget` is relayed at `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS = 30_000`, the same constant that produced #1404's exact-30-second symptom.

**It is lesser than #1404, and worth saying why.** The plain-join branch returns after **one** run rather than queueing a second, so the exposure is roughly one run (~14.5 s wall clock, since `NODE_DEFS_RUN_BUDGET_MS` deliberately stops its clock across `registerNodesFromDefs`/`reapplyDefsToLiveNodes`) rather than #1404's two. But it is reached **only after the authorization `/object_info` has already spent part of the window**, so the headroom is smaller than a bare 14.5 s against 30 s suggests.

**Why this stayed invisible.** #1409's own comment asserted `refresh_nodes` was "the THIRD command relayed in the 30,000 ms window and the last one that never took one". That is false: **five** commands relay at that constant — `graph_get_object_info`, `graph_add_node`, `graph_set_widget`, `graph_remove_widget`, `refresh_nodes` — plus `nodes_install` at a literal `30000`, and only `add_node` and `nodes_install` hold budgets. A reassurance-shaped claim concealed the very next occurrence. That claim is being corrected in #1409 before merge.

**Suggested shape**, following #1409: give the call a `joinMs` derived from the remaining command budget rather than a fresh constant, answer inside the window with a structured token from `NODE_DEF_REFRESH_REASONS`, and **do not cancel the run** — #1409 verified that letting it keep the slot means the retry joins work in progress rather than restarting it. Do not add a blind retry: re-issuing on a timeout is only safe if nothing was applied, and that must be established rather than assumed.

Worth auditing the remaining two relayed commands (`graph_get_object_info`, `graph_remove_widget`) in the same pass, so the fifth instance does not get filed next week.

artokun:
Fixes #1413.

## The gap

`graph_set_widget`'s stale-combo recovery — the branch that exists so a just-downloaded
model, a just-uploaded image or a just-installed pack is accepted on one revalidation —
falls back to the full frontend refresh when the `/object_info` already fetched for
authorization cannot key the target's concrete type. That call passed **no `joinMs`**, so
the one deadline `makeRefreshCoalescer` takes at invocation was never armed, and the write
waited unbounded — on a run something else started, or, with an empty slot, on one of its
own. `graph_set_widget` is relayed at `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS` (30,000 ms), the
constant that produced #1404's exact-30-second symptom.

Fourth instance of one recurring shape: #1192 → #1349 → #1404 → this.

## Why a budget, not a fresh constant

The issue's suggested shape, and it turned out to matter more than a stylistic preference.
This recovery is reached only **after** two waits have already spent part of the window:

| step | bound before this PR | source |
|---|---|---|
| `awaitObjectInfoHistorySeed()` | 8,000 ms | `OBJECT_INFO_SEED_WAIT_MS` |
| authorization `/object_info` | 20,000 ms | `OBJECT_INFO_DEADLINE_MS`, `lib/object-info-oracle.js:187` |
| stale-combo refresh | **none** | — |
| #387 `/view` upload-asset probe | **none** | — |

The first two compose to 28,000 ms against a 30,000 ms window *before* the refresh is even
considered, so a constant bound at the refresh would have been a bound on time the command
had already spent. `graph_set_widget` now takes a 25,000 ms budget on its **first line** and
every step draws from what is left — the shape #671 uses for `nodes_install` and #1192 for
`graph_add_node`.

`joinMs` alone is sufficient for the refresh because #1351 already made the coalescer's
single deadline span the join **and** the run behind it (`waitForRun`). No second clock.

## What the reply says

On abandonment the recovery reports a **structured token**,
`NODE_DEF_REFRESH_REASONS.REFRESH_STILL_RUNNING`, which `lib/set-widget.js` carries into the
reply instead of the claims it used to make unconditionally:

- success → `refreshed: false`, `combo_refresh_incomplete: "refresh_still_running"`, plus a
  note. It used to say `refreshed: true`.
- refusal → `…with the combo options NOT refreshed…` instead of `…after refreshing combo
  options…`, and it names the reason and asks for a retry. **This is the half that outlives
  the timeout**: an agent told a *fresh* list rejected its value stops retrying a value that
  is fine.

The run is **not cancelled** (`waitForRun` says so), and **no blind retry is added** — both
as the issue asked.

## The `/view` probe

Bounded from the same budget. It is the one network step in this command that had no bound
at all and it runs at the *end* of the ladder, so a stalled connection there spends a window
every step before it drew down. It already answered `false` for every error, so failing it
closed can only refuse a value, never admit one — #240 strictness is untouched.

## Two false claims corrected

Both are the defect class this issue is actually about — a reassurance-shaped sentence
concealing the next occurrence.

1. **#1192's fetch note** claimed `graph_set_widget`'s oracle "allows 10,000 ms" and that
   therefore "BOTH" were capped and the two paths agreed. It never did: that oracle goes
   through `fetchWholeObjectInfo`, whose deadline is 20,000 ms, and nothing capped it. The
   asymmetry that sentence claimed to have closed was the live behaviour, in the direction
   the sentence ruled out. Corrected, and pinned by a test.
2. **My own first draft** said "a node-def refresh was still running when the budget ran
   out" and "it is still fetching exactly the `/object_info` this write needed." The gate
   caught both. One symbol comes back whether the budget went on a run already in flight or
   was *already spent before one could be started* (`joinMs <= 0` with an empty slot starts
   nothing) — and this call is deliberately **not** forced (#458 P2), so a run it joins may
   have begun its fetch before the asset existed. The reply now names both states and rests
   the retry on the one thing true in every arm: the retry re-enters this recovery with a
   fresh budget.

## The audit the issue asked for

`graph_get_object_info` and `graph_remove_widget` come back **clean**. Each holds exactly one
await, and both land on `fetchWholeObjectInfo`, which bounds itself at
`OBJECT_INFO_DEADLINE_MS` — under the window, with nothing to compose against. Neither needs
a budget; neither is a fifth instance.

Written as a **test**, not a comment beside the constant, for the reason above: give
`graph_remove_widget` a second unbudgeted wait and the audit goes red.

## Verification

- **Unit:** 5,609 pass / 0 fail (`npm run test:unit`), `tsc --noEmit` clean.
- **New pins:** `browser_tests/unit/set-widget-command-budget.test.mjs` (18) drives the
  **shipped** `refreshCombos` wiring, pulled from the panel source, over the **real**
  `makeRefreshCoalescer` and the **real** `withTimeout`, against a real run that never
  settles. A helper-level test cannot see this defect — the coalescer has honoured `joinMs`
  since #1192 and `refresh-coalesce.test.mjs` already proves it. The whole bug was the call
  site.
- **Verified by mutation, nine ways.** Deleting the `joinMs`, the token assignment, the
  truthful refusal clause, the disclosure, the probe bound, either budget-derived deadline,
  moving the budget off the first line, or growing a second wait in `graph_remove_widget` —
  each fails a named test. Restoring the defect (`joinMs` removed) makes three behavioural
  tests fail on their failsafe rather than hang silently.
- **One weakness the mutation pass found and fixed:** the baseline-seed pin was body-wide
  over `PANEL_SRC` and stayed **green** with `graph_set_widget`'s bound deleted, because
  `graph_add_node`'s sibling call site satisfied it. The pins are scoped to this method's
  body now.
- **Harnesses:** three rebuild this executor or a fragment of it in a synthetic scope, so the
  new module bindings threw `ReferenceError` in all three at once. Collected in
  `_panel-constants.mjs` (`setWidgetCommandBudgetDeps`, `SET_WIDGET_LOCALS_PRELUDE`) so the
  next one is added in one place.

### Browser suite — NOT run, and why

`npx playwright test` was **not** run. No ComfyUI was allocated to this run; the one instance
listening (`ap-1`, port 8211) belongs to another session and serves a **different checkout**
pinned at `adf23a05`, so a run against it would have exercised code without this change, and
repointing its junction would have altered another session's setup mid-run.
`npx playwright test --list` compiles clean (78 tests, 33 files).

What that leaves unverified is small, and stated rather than implied: nothing in this diff
touches rendering, DOM, i18n or the sidebar. On the healthy path the behaviour is
byte-for-byte identical — at ~0 ms spent, `budget.bounded(8000)` is 8000 and
`budget.bounded(20000)` is 20000, exactly the constants that were there before. The change is
only *when the command stops waiting* and *what the reply says*, both driven by the node
suite through the shipped code.

## Gate

adversarial review instead (`.claude/autopilot/claude-gate.mjs`), per the substitution the
owner authorised. Its two sub-P1 findings — both prose over-claims — were fixed and pinned
before this was marked ready, and the gate was re-run on the final tree.

## Noted, deliberately not fixed here

The recovery's fallback is a **plain** (non-forced) refresh, so joining a run whose
`/object_info` fetch began before the asset appeared can report a refresh that does not
contain it — #396's argument for `force: true`, applied to this call site. That is
pre-existing, orthogonal to the unbounded wait, and changing it would rewrite #458 P2's
single-fetch economics. Recorded here so it can be filed rather than lost.


artokun:
Closing as **superseded by #1416**, which merged first and fixed #1413.

Both PRs were opened within twenty minutes of each other against an issue whose only claim
was a thread comment, so this is a genuine collision rather than duplicated effort — #1416's
command budget is the same shape this one took (25,000 ms on the first line, `joinMs:
budget.remaining() - SET_WIDGET_POST_REFRESH_RESERVE_MS`), and it is on main. Nothing here
needs to land on top of it.

Four things this branch had that #1416 did not, all measured against the merge commit
`3c3ee5fb` and filed as #1418 rather than dropped:

1. the shipped refusal asserts "that refresh is still running" in an arm where the coalescer
   started **nothing** — reachable with no concurrency at all, because the 8,000 ms seed wait
   and the 20,000 ms authorization fetch still hold flat constants and compose to 28,000 ms
   against the 25,000 ms budget;
2. the #387 `/view` upload-asset probe is still unbounded and **is** reachable — #1416's
   early refusal covers the abandoned path, not the one where the refresh completes and the
   retry still misses;
3. #1192's fetch note still says `graph_set_widget`'s oracle "allows 10,000 ms" (it is
   20,000 ms, and nothing capped it);
4. the audit #1413 explicitly asked for — `graph_get_object_info` and `graph_remove_widget`
   both come back clean, one self-bounded await each.

The branch stays pushed if any of it is useful. Its tests were mutation-verified nine ways;
one finding from that pass is worth carrying over on its own: a source pin written against
the whole of `PANEL_SRC` stays green with `graph_set_widget`'s bound deleted, because
`graph_add_node`'s sibling call site satisfies it.

Issue #1418, opened 2026-08-19
graph_set_widget's stale-combo refusal names a refresh that is not running, and the #387 /view probe is still unbounded (follow-up to #1413)


artokun:
Follow-up to #1413, which #1416 fixed. The command budget landed and the 30 s no-reply is
gone. Three things the same pass left behind, each **measured against `3c3ee5fb`** (the merge
commit) rather than argued from reading.

---

## 1. The refusal asserts a refresh that is not running — and the arm has no concurrency in it

`staleComboRefreshBusyMessage()` (web/js/lib/set-widget.js) says:

> "a node-def refresh started by a ComfyUI reconnect, a finished install, or another tool
> call — was still running when this command's time budget ran out waiting for it … **That
> refresh is still running and is registering exactly the list this write needs, so RETRY in
> a few seconds — this normally succeeds on the next attempt.**"

`REFRESH_JOIN_ABANDONED` comes back in **two** states, and that sentence is true in only one.
With `joinMs <= 0` and an empty slot, `refresh-coalesce.js` returns it **without starting
anything**:

```js
if (joinMs !== null && !(joinMs > 0)) return REFRESH_JOIN_ABANDONED;
return waitForRun(startRun(preloadedDefs), joinMs, withTimeout);
```

That arm is reachable **with no concurrency at all**, because the two waits ahead of the
recovery are not capped by the budget the command now takes — they still hold their own flat
constants, and they compose past it:

| step | bound | source |
|---|---|---|
| `awaitObjectInfoHistorySeed()` | 8,000 ms | `OBJECT_INFO_SEED_WAIT_MS` |
| authorization `/object_info` | 20,000 ms | `OBJECT_INFO_DEADLINE_MS`, `lib/object-info-oracle.js` |
| | **28,000 ms** | vs. a 25,000 ms budget |

Driving the **shipped** `refreshCombos` over the **real** coalescer at exactly that spend:

```
budget total / spent-before-recovery : 25000 / 28000 (seed 8000 + oracle 20000)
joinMs handed to the coalescer       : -7000
runs the coalescer STARTED           : 0
slot occupied afterwards             : false
outcome is REFRESH_JOIN_ABANDONED    : true
shipped refusal asserts a live run   : true
```

So the user is told a refresh is running when none is, and told to retry because "this
normally succeeds on the next attempt" — when the retry meets the same slow `/object_info`
and lands in the same arm.

[...]

PR #1419, opened 2026-08-19
fix(1418): the stale-combo refusal names both abandoned states, and the #387 /view probe is bounded

artokun:
Claiming #1418. WIP — final body to follow.


artokun:
For the record: this merged as the empty placeholder commit (orchestration error — the draft was merged before the fix was pushed). The actual #1418 fix landed via #1423. The squash commit is a no-op on main.

PR #1421, opened 2026-08-19
fix(1417): an image the browser cannot load says so instead of leaving blank space

### LINKED ISSUE #909: panel_show_media silently accepts browser-unsupported MP4 codecs
panel_show_media returned {ok:true,count:2} for an MP4 encoded as MPEG-4 Part 2 (codec mpeg4, tag mp4v), but the sidebar showed no playable inline video. The command acknowledges DOM dispatch, not browser decode. The lazy video renderer suppresses play() rejection and has no media error handler, so an unsupported codec becomes a silent blank/placeholder card.

### Steps to reproduce
1. Create a 1280x960, 50 fps MP4 with codec mpeg4/mp4v.
2. Call panel_show_media with its absolute disk path.
3. Tool returns ok:true.
4. In the panel browser, the video is not visible/playable.
5. Transcode the same 66 frames to H.264 High, yuv420p, avc1, faststart and call panel_show_media again; the browser-compatible copy renders.

### Exact result
Tool response: {"ok":true,"count":2}
User-visible result: MP4 not visible inline.
Source probe: codec_name=mpeg4, codec_tag_string=mp4v, pix_fmt=yuv420p.
Compatible copy probe: codec_name=h264, profile=High, pix_fmt=yuv420p, 1280x960, 50/1 fps, 66 frames.

### Fix
Applied locally: yes
Upstream-only: no

Minimal frontend patch adds a video error handler in mountHolderVideo so decode failures are visible and actionable instead of silently blank:

```diff
+    v.addEventListener("error", () => {
+      holder.textContent = "Video unavailable in this browser. Encode as H.264 (yuv420p) or WebM.";
+      holder.style.cssText += ";display:grid;place-items:center;padding:1rem;box-sizing:border-box;text-align:center;color:var(--p-text-muted-color,#a1a1aa);font-size:0.75rem;";
+    });
```

User unblock: retained the original and created a browser-only H.264/yuv420p copy.

Verification:
- npm run typecheck: PASS
- npm run test:unit: PASS, 1714 passed, 0 failed, 3 skipped
- ffprobe on H.264 copy: 1280x960, 50 fps, 66 frames, 1.32 s

### Environment
Windows 11
ComfyUI 0.31.1
ComfyUI frontend 1.48.7
Python 3.13.12
PyTorch 2.10.0+cu130
GPU NVIDIA GeForce RTX 3060, 12 GB VRAM
comfyui-mcp running 0.50.48 (0.50.51 installed on disk)
comfyui-mcp-panel 0.11.38

### LINKED ISSUE #1417: panel_show_media: large 16-bit PNG sent as a /view reference renders as a silent broken card — user has to ask "where is the image"

`panel_show_media` was called with three items, two of them ComfyUI output references (`{filename, subfolder, type:"output"}`). The tool reported `ok:true, count:3, painted:3, unconfirmed:0, unrenderable:[]` — i.e. full success from the agent's point of view.

**The cards did not render for the user.** Their very next message was "where is the image". Nothing reported an error to the agent, the user, or the log; the agent believed the images had been delivered and moved on to describing them.

The item that mattered was an **18 MB, 2048×2048, 16-bit RGB PNG**. Two independent factors are suspect and both are worth handling:

1. **Size** — an 18 MB fetch through `/view` into an inline card.
2. **Bit depth** — 16-bit PNG. Browser support for 16-bit PNG is real but inconsistent in canvas/inline-image paths, and this is a plausible decode failure that surfaces as a blank card rather than an error.

The tool's own result note already anticipates this class of failure precisely — *"A 'painted' count above means the panel created those cards — NOT that the media loaded. The browser fetches /view itself, after that reply, and when it gets HTML instead of an image … the card renders BROKEN and nothing reports an error."* The note is accurate and useful, but it documents an unobservable failure instead of preventing or detecting one. The probe it runs (`all 2 probed returned media`) checked from the orchestrator and passed — so the probe gave a green light for a card that then failed in the browser. That green light is actively misleading.

### Fix (this is the part worth merging — it is confirmed working)

applied locally: **no** (behavioural workaround only, no code changed)   upstream-only: **yes**

**The workaround that worked, confirmed by the user:** re-send as an inlined absolute local path, after downconverting:

```python
ou = cv2.imread('field_360_stereo_OU.png', cv2.IMREAD_UNCHANGED)   # uint16
u8 = (ou/257.0 + 0.5).astype('uint8')
cv2.imwrite('/tmp/ou_small.jpg',
            cv2.resize(u8, (1400,1400), interpolation=cv2.INTER_AREA),
            [cv2.IMWRITE_JPEG_QUALITY, 92])
```

then `panel_show_media(items=[{source:{path:"/tmp/ou_small.jpg"}}])`. That card rendered immediately and the user confirmed it. So the delivery mechanism is fine; the reference path with a large 16-bit source is what fails.

Suggested changes, cheapest first:

1. **Auto-derive a display proxy instead of handing the browser the master.** When a referenced image exceeds a size threshold (~4 MB) or is >8 bits per channel, generate and send a downscaled 8-bit JPEG/PNG proxy, and keep the reference/path to the master in the caption. This is exactly what I did by hand and it worked — doing it in the tool means no agent has to know the rule.
2. **Make the probe honest about what it proved.** It currently probes from the orchestrator and reports success, which does not establish that the browser can decode the bytes. Either have the panel report back per-card load success/failure (`onload`/`onerror`) so the tool result reflects reality, or downgrade the wording so `painted` is never mistaken for delivered.
3. **Render a visible failure state.** A card whose image fails to load should show "failed to load — <filename>" in the UI, not blank space. Right now the only detector is the user asking where the image went.
4. **Prefer local-path inlining when the orchestrator can see the file.** For a LOCAL ComfyUI the orchestrator has filesystem access to `output/`; inlining verified bytes avoids the whole `/view` round trip and its failure modes. Consider making that the default for local targets rather than the documented recovery step.

### Steps to reproduce

1. Produce a large 16-bit PNG in ComfyUI's output dir (2048×2048 uint16, ~18 MB — e.g. `cv2.imwrite` of a uint16 array).
2. `panel_show_media(items=[{source:{filename:"<it>.png", subfolder:"stereo360", type:"output"}}])`.
3. Tool returns `ok:true, painted:1, unrenderable:[]`, and the probe reports the reference returned media.
4. Observe the card in the panel chat: no image. No error anywhere.

### Exact error (scrubbed)

None — **silent failure**, which is the substance of the report. Tool result was:

```
{ ok: true, count: 3, painted: 3, unconfirmed: 0, unrenderable: [], previews: [] }
NOTE — 2 items were sent to the panel as a ComfyUI /view REFERENCE, not as inline bytes
Checked from HERE: all 2 probed returned media, so the reference itself looks fine
```

### Environment

- OS: macOS (Darwin 25.5.0 / Mac 26.5.1)
- ComfyUI: 0.33.0, `local-desktop2-standalone`, local target at `127.0.0.1:8000`
- **ComfyUI frontend: 1.49.6**
- GPU: mps, 128 GB unified, vram state SHARED
- Python 3.12.11 / PyTorch 2.10.0
- **comfyui-mcp: 0.52.7** (current)
- **comfyui-mcp-panel: 0.14.44** — note this is behind latest 0.15.6, so please check whether 0.15.x already addresses the proxy/decode path before actioning. Filing anyway since the failure is silent and the workaround is confirmed, per beta reporting guidance.
- Related, filed separately: artokun/comfyui-mcp#1789 (missing `panel_run` completion notification in the same session).

---
Filed from the ComfyUI panel by **claude** · model `opus[1m]` · effort `medium` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #909

artokun:
Fixes #1417

## What was silent

`panel_show_media` was handed an 18 MB, 2048×2048, 16-bit RGB PNG as a ComfyUI `/view`
reference. The tool answered `ok:true, count:3, painted:3, unconfirmed:0, unrenderable:[]`,
and its own probe — run from the **orchestrator**, which is not the thing that has to fetch
and decode the bytes into the ComfyUI page — reported that the reference returned media.
The user saw nothing. Their next message was *"where is the image"*.

`paintImage` attached a `click` listener and no `error` listener, so a picture that never
arrived was indistinguishable from one that had not painted yet — for the user, for the
agent, and in the log.

## The fix

#909 gave `<video>` a decode message for precisely this defect class (`mountHolderVideo`'s
`error` listener) and left `` silent. This is the other half of the same pipeline:

- **The chat card.** A failed `` is replaced in place by a visible notice —
  *"This image could not be loaded."* — plus an **Open original** button that opens the
  master in a new tab, where the browser gets to show its own failure and the file can
  still be saved.
- **The lightbox stage.** The same notice, so prev/next onto a picture the browser cannot
  load is not the same blank one click further in. `offerOpen:false` there — the bar below
  already carries that button.
- Collapsed cards hide the notice with the picture, so a hidden card is one stub whether or
  not its media arrived.

Two deliberate restraints, both inherited from #909's review:

- **It claims no cause.** `<video>` exposes `error.code`, so #909 could offer a re-encode
  only for `MEDIA_ERR_SRC_NOT_SUPPORTED`. An `` error event carries nothing at all —
  a 404, HTML served in place of the bytes, a refused 16-bit decode and an out-of-memory
  image are one indistinguishable event. So the message states what is true and offers the
  one thing the panel can still do about it.
- **It names no file.** The caption directly beneath already carries the agent's own
  caption/filename (`composeShowMediaReply` falls back to `item.filename`), and the notice
  is inserted *where the picture was* — above it. A name inside the box would print it
  twice on every failed card.

A teardown is not a failure: the lightbox empties its stage on every step, which fires
`error` on the element it just dropped. The painter refuses any element that is no longer
a child of the container it was handed, which also makes a repeat dispatch a no-op.

## Deliberately NOT in this change

The issue's other three suggestions are all orchestrator-side, on the far side of the panel
protocol, and belong in `artokun/comfyui-mcp`:

1. auto-deriving a downscaled 8-bit proxy when a reference is over ~4 MB or >8 bits/channel;
2. making the tool's own probe honest (it proves the *orchestrator* can fetch the reference,
   not that the browser can decode it) — the panel would need to report per-card
   `onload`/`onerror` back through the bridge for the result to reflect reality;
3. preferring local-path inlining for local targets.

This PR is #1417's item 3, which is the part that lives in the panel, and it is what turns
the failure from unobservable into observable.

## Verification

- `npm run test:unit` — **5610 pass, 0 fail** (5609 before; 8 new tests in
  `browser_tests/unit/image-load-failure.test.mjs`, minus a count shift from the i18n gates).
- `npx tsc --noEmit` — clean.
- `node scripts/check-panel-scope.mjs` — clean (the gate that catches a name resolving in a
  sibling function).
- **Mutation-verified**, not just green: with the `error` listener deleted from `paintImage`
  the wiring test fails; with the notice's message line removed the behavioural tests fail;
  with the detached-element guard removed the teardown test fails. Each was reverted from
  the commit, not from the working tree.
- The 8 new tests drive the **shipped** `paintImageLoadFailure` — lifted out of the monolith
  with `new Function` over a DOM stub — so an error really does have to remove the image and
  paint the notice. The two call sites are pinned separately at source: a helper nobody
  calls fixes nothing.
- 11 locales translated; `i18n-check` and `i18n-render-check` pass.

## Gate

**Codex was unavailable — the account is exhausted until 2026-08-19 20:43 UTC.** Gated by an
independent adversarial review instead (`.claude/autopilot/claude-gate.mjs`, a fresh reviewer
that has not seen this reasoning), per the substitution the owner authorised. Verdict recorded
in a comment below.


artokun:
Closing as duplicate — #1420 (fix for #1417, same paintImage failure surface, same fallback-card approach) merged while this was in flight. The remaining half your body calls out — the orchestrator-side probe reporting a media reference as OK without the browser decoding it — is not covered by #1420 and would make a good scoped follow-up if you want to rebase onto main and reduce to that.

PR #1423, opened 2026-08-19
fix(1418): the stale-combo refusal names both abandoned states, and the #387 /view probe is bounded

### LINKED ISSUE #1418: graph_set_widget's stale-combo refusal names a refresh that is not running, and the #387 /view probe is still unbounded (follow-up to #1413)

artokun:
Fixes the three things #1416 left behind, plus the audit #1413 asked for.

> **Replaces #1419, which merged at the wrong commit.** #1419 was squash-merged ~4 minutes
> after it opened, at `6f13af0a` — the **empty `wip(1418): claim` commit** pushed to open the
> draft. That squash (`0b9dc9f6`) is now the top of `main` and changes no files, so the
> comment closing #1418 as "fixed" described code that was never merged. #1418 has been
> reopened and the record corrected there. This PR carries the actual implementation,
> rebased onto the current `main`.

---

## 1. The refusal asserted a refresh that was not running

`REFRESH_JOIN_ABANDONED` comes back from `refresh-coalesce.js` in **two** states, and
#1413's text was true in only one:

> "…That refresh **is still running** and is registering exactly the list this write needs,
> so RETRY in a few seconds — this **normally succeeds on the next attempt**."

The second arm returns the token from an **empty slot** without starting anything:

```js
if (joinMs !== null && !(joinMs > 0)) return REFRESH_JOIN_ABANDONED;
```

…and it is reachable **with no concurrency at all**, because the two waits ahead of the
recovery still held their own flat constants and composed past the budget. Measured by
driving the shipped `refreshCombos` over the **real** coalescer at the **shipped** numbers
(`browser_tests/unit/set-widget-budget-composition.test.mjs`, first test):

```
budget total / spent-before-recovery : 25000 / 28000 (seed 8000 + oracle 20000)
joinMs handed to the coalescer       : -7000
runs the coalescer STARTED           : 0
slot occupied afterwards             : false
outcome is REFRESH_JOIN_ABANDONED    : true
```

**Both halves of the issue's suggested fix are applied, because neither alone is enough:**

- The refusal now names **both** states and rests its retry advice on the one fact true in
  either — *the retry re-enters this recovery with a fresh budget* — following #1409's
  precedent for `refresh_nodes`. It also drops "this normally succeeds on the next attempt",
  which is false on the second arm, and says plainly that a bare retry can meet the same
  slow `/object_info` and that `panel_refresh_nodes` is what resolves that.
- `awaitObjectInfoHistorySeed()` and the authorization `/object_info` are capped with
  `budget.bounded(…)`, the way `graph_add_node` has capped its copies since #1192.

**One correction to the issue's framing:** capping does **not** make the exhausted arm
unreachable without concurrency, so it is not an alternative to the wording fix. `bounded()`
holds total spend to the budget, but the seed and oracle can still consume all 25,000 ms
between them, leaving `joinMs = remaining() − 4000 ≈ −4000` — non-positive, empty slot, same
arm. What the caps actually buy is the property #1413's budget was for: the worst case
*before* the recovery drops from 28,000 ms to the budget itself, so the reply lands inside
the 30,000 ms relay window. The wording fix is the load-bearing one.

## 2. The #387 `/view` probe is bounded

#1416's refusal fires before the probe on the **abandoned** arm only. When the refresh
**completes** and the retry write still misses — the subfolder-nested `LoadImage` image the
probe exists for — the ladder reaches it anyway, and it was the last network step of a
command relayed at 30,000 ms with no bound of any kind. It now draws
`budget.bounded(SET_WIDGET_ASSET_PROBE_MS)` (3,000 ms, sized inside the 4,000 ms
`SET_WIDGET_POST_REFRESH_RESERVE_MS` so composing the reply still has room).

Cannot regress #240: the probe already answers `false` for every error, so a bound can only
turn `true` into `false` — refuse a value, never admit one. A test pins that a `/view` which
*answers* still accepts the nested asset.

## 3. `graph_add_node`'s fetch note stated a number that was never true

The comment claimed the two paths agree at 10,000 ms and that this "stops a *set_widget
works but add_node fails* asymmetry". `graph_set_widget`'s oracle goes through
`fetchWholeObjectInfo`, whose `OBJECT_INFO_DEADLINE_MS` is **20,000 ms**. Corrected rather
than papered over, and the numbers are deliberately left **different** — they bound
different operations, each argued where it lives — with the property they now genuinely
share stated instead: each takes the smaller of its own constant and its caller's remaining
budget.

## 4. The audit #1413 asked for, as a test

`graph_get_object_info` and `graph_remove_widget` are clean: one `await` each, landing on
`fetchWholeObjectInfo`, which bounds itself. Neither needs a command budget. That is a claim
with a shelf life, so it is a test — it goes red the day either grows a second wait or
reaches `refreshComfyNodeDefs`, whether or not anyone rereads the note.

---

## Verification

**Mutation-verified — every fix, deleted, turns its own test red:**

| mutation | red |
|---|---|
| seed cap deleted | `the seed wait is capped by the command…` + the method-body pin |
| oracle `deadlineMs` deleted | `the authorization /object_info deadline is capped…` + the pin |
| probe `withTimeout` deleted | `the #387 /view probe IS reached… and is bounded` + the pin |
| refusal reverted to #1416's wording | `the refusal names BOTH abandoned states…` |
| `graph_remove_widget` given a 2nd await on the coalescer | the audit test |

**The pins are scoped to `graph_set_widget`'s own body, and there is a test that they are.**
A file-wide `assert.match(PANEL_SRC, /awaitObjectInfoHistorySeed\(budget\.bounded\(/)` stays
**green** with this method's bound deleted, because `graph_add_node`'s sibling call site
satisfies it. Mutation M1 confirms the body-scoped pin does go red.

**Three extraction harnesses needed the new module bindings in the same commit** — a new
free name in a rebuilt executor is a `ReferenceError`, not a failed assert. `object-info-snapshot.test.mjs`
rebuilds only the `readObjectInfo` block and gets a **real** `makeCommandBudget`; a stub
there would let it pass against a call site whose bound had been deleted.

- `npm run test:unit` — **5618 pass, 0 fail** (i18n, tool-vocabulary and `check-panel-scope` gates included)
- `npm run typecheck` — clean
- `check-panel-scope` — OK; this is the gate that catches a name resolving in a sibling function, and it needed `npm ci` in the worktree to run at all

**Browser suite NOT run, and this is not implied coverage.** No ComfyUI is listening on any
port on this machine and this run was told not to start one, so `npx playwright test` cannot
execute. The change renders nothing in the sidebar — it alters a tool-result refusal string
and network wait bounds inside graph command handlers — and no e2e spec exercises the
stale-combo recovery (the sole `panel_set_widget` mention in `browser_tests/*.spec.ts` is a
comment in an outline-label spec). `npx playwright test --list` compiles and discovers all
78 tests from this worktree.

**Gate:** codex was unavailable (account exhausted until 2026-08-19 20:43, measured
previously on both the wrapper and direct `codex exec`, so not re-probed); gated by an
independent adversarial review instead, per the authorised substitution. Verdict below.

---

## Addendum — a defect this change would have introduced, caught and fixed

Bounding the #387 probe creates a **second way for it to come back without an answer**, and
the old `!!res && (res.ok || res.status === 206)` collapsed that into the same `false` a real
404 produces. The ladder's final refusal then tells the caller their value *is not a valid
option* — a cause established by a request that never came back. Measured at the boundary:

```
SET_WIDGET_ASSET_PROBE_MS = 3000
fresh budget          remaining=25000ms bound=3000ms -> confirmServerAsset=true
3s left               remaining=3000ms  bound=3000ms -> confirmServerAsset=true
10ms left             remaining=10ms    bound=10ms   -> confirmServerAsset=true
EXHAUSTED (== budget) remaining=0ms     bound=1ms    -> confirmServerAsset=false   <-- healthy 5ms /view
2s OVER budget        remaining=-2000ms bound=1ms    -> confirmServerAsset=false
```

That is the same wrong-cause shape this issue exists to fix, so bounding the probe without
splitting these would have traded one instance of it for another.

**#1357 already built the vocabulary, and the panel already imports it.**
`inputAssetProbeVerdict` is a tri-state where `null` means *the question was not answered* —
and #1357's own note says collapsing `null` into `false` "manufactures the exact false
positive #1357 reported". The probe now returns that verdict, `set-widget.js` keeps "nobody
answered" distinct from "the server said no", and **only** the unanswered case adds a note
saying the refusal rests on the combo list alone and that a retry probes again with a fresh
budget. A real 404 is still reported as a confirmed miss with no hedging — there is a test
for each direction. Neither `false` nor `null` accepts, so #240 strictness is unchanged.

Three further mutations pin it (panel collapsing the tri-state; `set-widget` not recording
the unanswered case; the refusal dropping the note) — each turns the new test red.

Wiring this also required `inputAssetProbeVerdict` in the rebuilt-executor dep lists: my own
test caught the `ReferenceError`, because the swallowing `catch` turned it into a silent
"unanswered" rather than a crash.

**How it was found:** the first gate run was killed by a session restart, but it left its
scratch probes in the worktree. Running the boundary probe it had written is what produced
the table above — so the credit for this one belongs to the gate, even though that run never
returned a verdict.

Totals after this addendum: **8 mutations**, each turning its own test red;
`npm run test:unit` **5620 pass / 0 fail**.

Closes #1418

### DIFF
--- a/browser_tests/unit/_panel-constants.mjs
+++ b/browser_tests/unit/_panel-constants.mjs
@@ -18,6 +18,8 @@ import { REFRESH_JOIN_ABANDONED } from "../../web/js/lib/refresh-coalesce.js";
 import { NODE_DEF_REFRESH_REASONS } from "../../web/js/lib/node-def-refresh.js";
 import { clearInheritedExecutionPreview } from "../../web/js/lib/execution-preview-attach.js";
 import { sanitizeNodeAuxId } from "../../web/js/lib/aux-id-sanitize.js";
+import { withTimeout } from "../../web/js/lib/bounded-step.js";
+import { OBJECT_INFO_DEADLINE_MS } from "../../web/js/lib/object-info-oracle.js";
 
 export const PANEL_SRC = readFileSync(
   fileURLToPath(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url)),
@@ -154,6 +156,15 @@ export const OBJECT_INFO_SEED_WAIT_MS = readPanelNumber(
   "the baseline seed wait",
 );
 
+/** #1418 — the cap on the #387 upload-asset /view probe, drawn from the post-refresh reserve. */
+export const SET_WIDGET_ASSET_PROBE_MS = readPanelNumber(
+  /const SET_WIDGET_ASSET_PROBE_MS = (\d+);/,
+  "the set_widget upload-asset probe bound",
+);
+
+/** #1418 — re-exported so a harness reads the oracle's OWN number, never a copy of it. */
+export { OBJECT_INFO_DEADLINE_MS };
+
 /**
  * #1192 — every module binding the command budget added to `graph_add_node`, in ONE place.
  *
@@ -206,5 +217,13 @@ export function setWidgetCommandBudgetDeps() {
     SET_WIDGET_COMMAND_BUDGET_MS,
     SET_WIDGET_POST_REFRESH_RESERVE_MS,
     monotonicNow,
+    // #1418 — the three bindings the two capped waits and the bounded #387 probe added.
+    // REAL, not stubbed: `withTimeout` is the repo's one bounding primitive and a double
+    // would let a harness pass against a probe that bounds nothing, and both numbers are
+    // read from the panel source for the reason at the top of this file.
+    withTimeout,
+    OBJECT_INFO_DEADLINE_MS,
+    SET_WIDGET_ASSET_PROBE_MS,
+    OBJECT_INFO_SEED_WAIT_MS,
   };
 }
--- a/browser_tests/unit/_panel-constants.mjs
+++ b/browser_tests/unit/_panel-constants.mjs
@@ -19,6 +19,7 @@ import { NODE_DEF_REFRESH_REASONS } from "../../web/js/lib/node-def-refresh.js";
 import { clearInheritedExecutionPreview } from "../../web/js/lib/execution-preview-attach.js";
 import { sanitizeNodeAuxId } from "../../web/js/lib/aux-id-sanitize.js";
 import { withTimeout } from "../../web/js/lib/bounded-step.js";
+import { inputAssetProbeVerdict } from "../../web/js/lib/input-asset.js";
 import { OBJECT_INFO_DEADLINE_MS } from "../../web/js/lib/object-info-oracle.js";
 
 export const PANEL_SRC = readFileSync(
@@ -225,5 +226,9 @@ export function setWidgetCommandBudgetDeps() {
     OBJECT_INFO_DEADLINE_MS,
     SET_WIDGET_ASSET_PROBE_MS,
     OBJECT_INFO_SEED_WAIT_MS,
+    // #1418 — the #387 probe now returns #1357's TRI-STATE verdict rather than a
+    // boolean, so an unanswered probe stays distinguishable from a real 404. REAL: a
+    // stub would classify every response as whatever the stub decided.
+    inputAssetProbeVerdict,
   };
 }
--- a/browser_tests/unit/object-info-snapshot.test.mjs
+++ b/browser_tests/unit/object-info-snapshot.test.mjs
@@ -30,7 +30,9 @@ import {
   TRANSPORT_OUTCOME,
   fetchWholeObjectInfo,
   objectInfoOracleFailureNote,
+  OBJECT_INFO_DEADLINE_MS,
 } from "../../web/js/lib/object-info-oracle.js";
+import { makeCommandBudget } from "../../web/js/lib/command-budget.js";
 import { createObjectInfoCache, CACHE_OUTCOME } from "../../web/js/lib/object-info-cache.js";
 
 const SCHEMA = { KSampler: { input: {} }, H3Keyframes: { input: {} } };
@@ -508,6 +510,15 @@ function buildShippedOracle({ api, socketDown = false, epoch = 5, snapshot, epoc
     "epochDuringFetch",
     "comfyBackendSocketDown",
     "objectInfoOracleFailureNote",
+    // #1418 — the shipped body now caps the oracle deadline with the command budget
+    // (`deadlineMs: budget.bounded(OBJECT_INFO_DEADLINE_MS)`), so both names have to
+    // exist in this scope or the extracted body throws ReferenceError before it runs.
+    // REAL, not doubled: a stub `budget` would let this harness pass against a call
+    // site whose bound had been deleted. The bound itself is proven in
+    // set-widget-budget-composition.test.mjs — the wrapper below overrides deadlineMs
+    // to 20 so these #1223 cases run fast, which is why they cannot prove it here.
+    "budget",
+    "OBJECT_INFO_DEADLINE_MS",
     // `backendReconnectEpoch` MUST be a live mutable binding in the same scope as the
     // extracted body, because the panel's is module state the body re-reads. An earlier
     // version passed it as a frozen value, which made `observedAtEpoch` and the record-time
@@ -552,6 +563,8 @@ function buildShippedOracle({ api, socketDown = false, epoch = 5, snapshot, epoc
     epochDuringFetch,
     socketDown,
     objectInfoOracleFailureNote,
+    makeCommandBudget(25000, () => performance.now()),
+    OBJECT_INFO_DEADLINE_MS,
   );
 }
 
--- a/browser_tests/unit/rgthree-lora-row.test.mjs
+++ b/browser_tests/unit/rgthree-lora-row.test.mjs
@@ -1331,6 +1331,7 @@ const EXECUTOR_DEPS = [
   "OBJECT_INFO_DEADLINE_MS",
   "SET_WIDGET_ASSET_PROBE_MS",
   "withTimeout",
+  "inputAssetProbeVerdict",
 ];
 
 /**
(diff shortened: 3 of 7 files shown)


artokun:
Superseded by #1425 (same branch, finished work — this draft stayed open from the intermediate push).

PR #1427, opened 2026-08-19
fix(1707): a promoted-widget write lands in the instance that was addressed, not the shared subgraph definition

artokun:
A `panel_set_widget` on one saved-subgraph instance changed the promoted value that
other instances of the same subgraph read. This is the panel-side write; the
orchestrator only proxies `graph_set_widget`, so no orchestrator change could reach it
(same class as artokun/comfyui-mcp#801).

## The mechanism, confirmed against the frontend and on a live canvas

Every instance of a reusable subgraph shares **one** `Subgraph` object, so the inner
nodes and their widgets are shared too. `applyWidgetWrite` assigned the promoted value
to the inner widget **and** to the wrapper's rail. On ComfyUI frontends that give the
host input a `widgetId`, those are not two views of one value:

* the rail is **this wrapper's own store entry**, keyed
  `"<rootGraphId>:<nodeId>:<inputName>"` — the wrapper's node id is in the key;
* the inner widget is the **definition** every sibling instance reads;
* queue compilation reads the store entry for an unlinked promoted input
  (`ExecutableNodeDTO.resolveInput` → `useWidgetValueStore().getWidget(id)?.value`) and
  never the inner widget;
* a canvas edit of a promoted control runs the rail projection's callback, which writes
  that store entry **and nothing else** — the definition write is one the UI never
  performs;
* `SubgraphNode._setWidget` **seeds** a newly promoted instance's store entry from the
  inner widget's current value, which is how a definition write reaches a wrapper nobody
  addressed.

Read from comfyui-frontend **1.48.7** — the reporter's version — via the shipped
sourcemaps (`SubgraphNode.ts`, `widgetValueStore.ts`, `widgetId.ts`,
`ExecutableNodeDTO.ts`), then **executed** on a real 1.48.7 canvas driving this module
in the page (two instances of one subgraph, the write aimed at instance A):

| observation | origin/main | this branch |
|---|---|---|
| `a.subgraph === b.subgraph` | `true` | `true` |
| shared definition value | `""` → **`"WRITTEN-TO-A"`** | `""` → `""` |
| addressed wrapper's rail / serialized `widgets_values` | `"WRITTEN-TO-A"` | `"WRITTEN-TO-A"` |
| sibling wrapper with its own value | unchanged | unchanged |
| **a new instance with no promoted value of its own** | **`"WRITTEN-TO-A"`** | `""` |
| reply | `node_id: 10` (inner) | `node_id: 11` (the wrapper), `value_scope: "instance"` |

The last row is the reported symptom: a wrapper nobody touched comes up holding the
value written to another one. On the reporter's canvas the sibling wrappers were in
exactly that state — instances whose promoted value came from the definition rather
than from their own store entry.

## What the panel can and cannot distinguish

**Per-instance storage exists**, and this is established from the data rather than
assumed from a version: the frontend's own store key names the wrapper, so when its node
segment is this wrapper's id, no sibling can read the entry this write lands in. The key
is parsed the way the frontend's `parseWidgetId` parses it.

So the write is **instance-scoped**, not refused:

* the rail (and #477's display proxies) is written; the shared definition is not touched;
* the **rail's** callback fires, not the shared inner node's — invoking the inner node's
  callback would announce a new value for a widget whose value did not change, and run
  that shared node's side effects for an edit made on one instance. It is also what the
  UI does;
* the bound property is taken from the widget the write assigned, on the node that owns
  it (`BaseWidget.setValue` calls `setProperty` on that node);
* **and the shared definition is verified UNCHANGED.** A definition that moved anyway
  means the rail forwards to it after all, which is a **failure** that rolls back —
  never an `"instance"` claim the write cannot stand behind. That check is what makes the
  classification safe rather than hopeful.

**What it cannot distinguish**, stated rather than papered over: a frontend that gives the
host input no key at all. There the rail is a live view of the inner widget and there is
no per-instance home — refusing would block every promoted write on such a frontend,
including the single-instance case with nothing to leak into. That path is unchanged, and
the reply says which one happened: `promoted_from.value_scope` is `"instance"` or
`"subgraph_definition"` on **every** promoted write, so a caller never has to infer it and
a missing field can never read as either verdict.

`node_id` / `widget` / `value` now name the widget the write actually assigned, so the
panel's own activity line no longer reports "Set width = 1024 on node 54" for an inner
node whose value did not change. `promoted_from.inner_node_id` and `inner_previous` keep
the provenance.

A second commit fixes what the first one exposed: #366's strict rail re-check ran on the
same object the read-back had just accepted, without #805's normalization allowance, so a
rail that quantized onto its own declared grid was reported as a failed write. It is now
skipped only when it is literally the same object already verified.

## Refutation

11 source mutations, each applied to a **unique** anchor and confirmed present on disk
before the run (a mutation that did not apply would otherwise score as "caught"), each
reverted afterwards. **11/11 killed** by `browser_tests/unit/promoted-value-scope.test.mjs`:

restore the definition write · make nothing instance-scoped · drop the
definition-unchanged verification · fire the shared inner callback · claim `"instance"`
unconditionally · report the untouched inner node · let the rollback assign the
definition · take the read-back from the inner widget (twice, value and explanation) ·
re-check the rail strictly · report the inner widget's value.

Two of those (`M8`, `M10`) survived the first round and drove a real extra case — a
definition that already holds the requested value, which is what separates "read the
widget we wrote" from "read whichever widget already agrees".

The fixture is built from the real shapes: a shared `Subgraph`, per-widgetId store,
projections that read/write through it, `_setWidget`'s seeding, and the value queue
compilation reads. A fixture that kept the rail value on the rail object would have
passed whatever the code did, because the two stores would be one store.

Unit: **5631 pass / 0 fail** (`npm run test:unit`, up from 5628 — three new source-wiring
assertions were updated to the renamed call sites, deliberately, since they exist to pin
those call sites).

## Playwright delta

`npx playwright test --workers=1` against a dedicated ComfyUI (own port, own base
directory, frontend 1.48.7), origin/main baseline and this branch back to back, panel
swapped by junction with a server restart between runs:

* **baseline (origin/main): 18 failed / 60 passed (7.5m)**
* **this branch: 18 failed / 60 passed (7.5m)** — the same 18 spec names

Delta: **zero**. All 18 are MockBridge conversation-persistence/streaming specs that fail
identically on origin/main in this environment.

codex was unavailable (account exhausted until 2026-08-19 20:43), so this PR is **not**
gated yet; the owner gates and merges.

The underlying cause of artokun/comfyui-mcp#1707.

### DIFF
--- a/browser_tests/unit/dom-widget-diagnosis.test.mjs
+++ b/browser_tests/unit/dom-widget-diagnosis.test.mjs
@@ -132,10 +132,13 @@ test("WIRING: the failure branch passes the NODE, or the route can never appear"
   // message. widget-write's failure path needs a live graph to drive, so pin it here.
   const { readFile } = await import("node:fs/promises");
   const src = await readFile(new URL("../../web/js/lib/widget-write.js", import.meta.url), "utf8");
-  assert.ok(src.includes("describeNonValueBearingWidget(w, targetNode);"),
-    "the observed-revert branch must pass the node");
+  // `valueWidget`/`valueNode` are the widget the write ASSIGNED and the node that owns
+  // it (comfyui-mcp#1707) — the same pair the read-back above it just failed on. Passing
+  // the promoted INNER pair here would describe a widget this write never touched.
+  assert.ok(src.includes("describeNonValueBearingWidget(valueWidget, valueNode);"),
+    "the observed-revert branch must pass the widget that was written and its node");
   // And it must stay in the OBSERVED-failure branch — never a pre-emptive gate (#715).
-  const idx = src.indexOf("describeNonValueBearingWidget(w, targetNode);");
+  const idx = src.indexOf("describeNonValueBearingWidget(valueWidget, valueNode);");
   const before = src.slice(Math.max(0, idx - 1200), idx);
   assert.ok(before.includes("did not retain the"),
     "must remain attached to the observed-revert failure, not a preflight");
--- a/browser_tests/unit/promoted-value-scope.test.mjs
+++ b/browser_tests/unit/promoted-value-scope.test.mjs
@@ -300,6 +300,78 @@ test("#1707: a failed instance-scoped write does not touch the definition on the
   assert.equal(sg.definition(), 512);
 });
 
+test("#1707 × #805: a rail that snaps the value onto its own grid is a NORMALIZED success, read back from the rail", () => {
+  // The read-back, the normalization explanation and the reported value all have to
+  // come from the widget that was written. Taken from the shared inner widget instead,
+  // an instance-scoped write can never look normalized — the inner value did not move
+  // at all — and a write that APPLIED is reported as "did not retain the requested
+  // value", which is the #805 defect reintroduced through the promoted path.
+  const sg = makeReusableSubgraph({ definitionValue: 512 });
+  const target = sg.instance(293);
+  // The wrapper's own store entry declares the grid, and the rail snaps onto it —
+  // the projection's options come from the store, as they do in the frontend.
+  const state = sg.store.get(target.widgetId);
+  state.options = { min: 1, step: 2 };
+  const snap = (v) => (typeof v === "number" ? state.options.min + Math.round((v - state.options.min) / 2) * 2 : v);
+  // Both doors into the store quantize — a widget that snapped through one and not
+  // the other would not be a snapping widget, it would be an inconsistent fixture.
+  Object.defineProperty(target.rail, "value", {
+    get: () => state.value,
+    set: (next) => {
+      state.value = snap(next);
+    },
+    configurable: true,
+  });
+  Object.defineProperty(target.rail, "callback", {
+    value: (next) => {
+      state.value = snap(next);
+    },
+    configurable: true,
+  });
+
+  const set = applyWidgetWrite(target.node, "width", 4096, { resolveSource });
+
+  assert.equal(set.normalized, true);
+  assert.equal(set.requested_value, 4096);
+  assert.equal(set.value, 4097, "the reply carries the value the RAIL stored");
+  assert.equal(sg.queuedValue(target), 4097);
+  assert.match(set.normalization_note, /width/);
+  assert.equal(sg.definition(), 512, "and the shared definition still did not move");
+});
+
+test("#1707 × #805: a definition that happens to hold the requested value does not decide the read-back", () => {
+  // The narrow case that separates "read the widget we wrote" from "read whichever
+  // widget already agrees": the shared definition is ALREADY 4096, so an early-out
+  // taken from the inner widget would conclude "the value stuck" without ever looking
+  // at the rail — and then reject the rail's legitimately quantized 4097 as a failure,
+  // because nothing computed the normalization that explains it.
+  const sg = makeReusableSubgraph({ definitionValue: 4096 });
+  const target = sg.instance(293);
+  const state = sg.store.get(target.widgetId);
+  state.options = { min: 1, step: 2 };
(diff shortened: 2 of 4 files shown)


artokun:
**Merge gate — substitution disclosed.** Gated with the project's sanctioned substitute, `.claude/autopilot/claude-gate.mjs` (exit 0 = SHIP), which spawns a fresh reviewer that has never seen the author's reasoning. `codex exec` is account-exhausted until 2026-08-19 20:43; disclosure is the condition that substitution was accepted under.

**This finishes abandoned work.** A prior session diagnosed artokun/comfyui-mcp#1707 thoroughly at 07:54Z and stopped there — its worktree had zero commits and nothing uncommitted, so no fix was ever written. The diagnosis was handed over intact rather than redone, and every inherited claim held: the shared `Subgraph`, `applyWidgetWrite` as the leaking write, and no-orchestrator-fix.

**The extension that matters: per-instance storage exists, and the panel was writing somewhere the UI never writes.** In frontend 1.48.7 each host input carries `widgetId = "<rootGraphId>:<nodeId>:<inputName>"`, the rail reads and writes a per-key store, and `ExecutableNodeDTO.resolveInput` queues that entry with **no inner-widget fallback**. Values reached siblings because `_setWidget` *seeds* a new instance's entry from the inner widget. So this was not "wrote to the wrong place" but "performed a write the frontend itself never performs".

**The gate re-derived every premise from the shipped sources**, extracted from the installed bundle's source maps (`widgetId.ts`, `widgetValueStore.ts`, `SubgraphNode.ts`, `ExecutableNodeDTO.ts`, `litegraphService.ts`, `multilineTextarea.ts`) — including one the diff does not claim but depends on: **`SubgraphNode.serialize()` builds `widgets_values` from the store too**, so the instance write survives save and reload.

**The scope verdict is an observation, not a claim.** The definition is verified unchanged after the write; a definition that moved is a failure that rolls back rather than an `"instance"` claim. Hostile fixtures were driven through `applyWidgetWrite` directly — missing store entry, degenerate rail-is-the-inner-widget, a rail callback that leaks to the definition, the #477 display proxy, composite-object write, a combo with a drifted option list, `options.property` on both wrapper and inner — and **every leak path is caught and rolled back; no path reports `value_scope: "instance"` for a write that moved the definition.**

**Live refutation:** on baseline the definition moved `"" → "WRITTEN-TO-A"` and a fresh instance inherited it; on this branch the definition stays `""`, a fresh instance stays `""`, and the addressed wrapper's serialized `widgets_values` still takes the value.

**Mutations: 14 mutants, 12 killed**, including scope-always-instance, dropping the 3-segment check, dropping `encodeURIComponent`, still-write-inner, dropping the `definitionMoved` branch, rollback-always-writes-inner, and read-back-from-inner. The two survivors were shown equivalent or unreachable rather than waved through: an empty-string `widgetId` is already killed by the length check, and the `boundPropertyState` pair is unreachable because the wrapper never carries the inner widget's property key. Reachability is pinned — mutating the `promotedValueScope` call site (`widget-write.js:1350`, reached from `set-widget.js:601`) fails 4 tests.

Two soft spots recorded, neither blocking, both cosmetic: the `"subgraph_definition"` doc line asserts "this frontend exposes no per-instance store" for the unattributable-key sub-case (unreachable on 1.48.7, since such a key also fails `isWidgetId` and therefore fails verification rather than mis-scoping), and the `value_scope` doc-comment is optimistic for a wrapper living inside a *shared* definition — where the node object is literally shared and the UI's own edit does the same thing, so behaviour is correct and still strictly less leaky than `main`.

Unit 5631 pass / 0 fail; `tsc --noEmit` clean; Playwright delta zero (18 failed / 60 passed on both branch and `origin/main` baseline, same spec names).

Issue #841, opened 2026-08-09
[Feature, parked] Apply a multi-node removal as ONE atomic frame instead of N independent calls


artokun:
Split out of #727 so it survives that issue closing. **Parked under the current feature freeze** — filed to keep it, not to schedule it.

#727 proposed two fixes for what looked like a concurrent-removal race. The first (serialize panel-side removals) turned out to be a no-op — the removal paths are already fully synchronous, so there was nothing to serialize, and #727 is closed on that evidence. This is the second proposal, which stands on its own merits and has nothing to do with that race.

## The ask

Removing N nodes today is N independent `panel_remove_node` calls. Each can fail on its own, so a batch of ~10 has 2^10 possible outcomes, and an agent has to notice partial success and reconcile it — which in practice means re-reading the graph and diffing.

One command with one outcome is better for the caller in every case:

- a single success/failure verdict instead of a partial-application the agent has to detect;
- one graph mutation span rather than N, so `beforeChange`/`afterChange` and the undo entry cover the whole edit — today a 10-node removal is 10 undo steps for the user;
- one reply instead of ten, which matters against the reply deadline on a large batch.

## Why it is not urgent

Nothing is failing in the wild on this path. #713's fix covers the error signature that was seen, and the removals are serial, so this is ergonomics and undo-granularity rather than correctness.

## Note for whoever picks it up

The interesting design question is failure semantics: if node 7 of 10 cannot be removed, does the frame roll back the first six, or apply what it can and report precisely which ids were not removed and why? Rollback is the cleaner contract but needs the removal to be genuinely reversible, which link teardown may not be. Partial-with-a-precise-manifest is likely the honest answer, but it should be chosen deliberately rather than falling out of the implementation — reporting an outcome the caller cannot act on is the failure mode this repo keeps filing bugs about.

artokun:
Claimed by grok-swarm 2026-08-19T16:55:31Z — driving this in worktree `/c/Users/Artokun/code/comfyui-mcp-panel/.worktrees/wt-841` (branch `feat/841-atomic-multi-remove`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp-panel/pull/1431 — Enhancement, do not merge until another agent reviews.

artokun:
[...]

PR #1431, opened 2026-08-19
feat(841): panel_remove_node takes node_ids as one undo step

artokun:
Enhancement — do not merge until another agent reviews.

Panel half of #841 — do not close the issue. `comfyui-mcp` still schemas/forwards only `node_id`; a companion PR must advertise and forward `node_ids` before the agent can reach this executor.

Removing N nodes used to be N independent `panel_remove_node` calls: N undo steps, N replies, and 2^N possible partial outcomes the agent had to detect by re-reading the graph.

`panel_remove_node` now accepts `node_ids` and applies the whole list in one `beforeChange`/`afterChange` pair:

- one Ctrl+Z restores every node that left
- missing, duplicate, or `ignore_remove` ids refuse the whole list before the envelope opens
- a mid-batch no-op or throw is reported as `removed` + `not_removed` with reasons, not as a clean success (link teardown is not rolled back from here)
- retry only `not_removed`: including an id that already left throws and refuses the whole list
- `node_id` keeps the historical single-node reply shape


artokun (commented):
The executor itself is the right shape for #841: `node_ids` share one `beforeChange`/`afterChange` pair, `node_id` still returns `{removed: summary}`, known-upfront misses/duplicates/`ignore_remove` refuse before the envelope, and a mid-batch leftover is `removed` + `not_removed` with reasons. Tests extract the shipped `graph_remove_node` and the nine #841 cases plus the i18n/preview-attach updates pass. Dispatch still goes through `executor(msg)` behind the reload guard, uuid fence, and binding bar — nothing here calls `graph.load` or restamps identity.

Requesting changes because the leftover `warning` tells the caller a retry they cannot actually perform.

Also: do not close #841 from this PR alone. `comfyui-mcp`'s `panel_remove_node` still schemas/forwards only `node_id` (`ctx.call({ cmd: "graph_remove_node", node_id: args.node_id })`). Until a companion PR advertises and forwards `node_ids`, the agent cannot reach this executor.

web/js/comfyui-mcp-panel.js
```diff
@@ -12813,57 +12809,139 @@ const GRAPH_TOOL_EXECUTORS = {
...
+            ? String(err.error?.message ?? err.error)
+            : "could not be removed (it may be protected / ignore_remove)",
+        };
+      });
+      result.warning =
+        `${stayed.length} of ${nodes.length} node(s) are STILL on the canvas. ` +
+        `Removed: ${gone.map((g) => g.node.id).join(", ")}. ` +
+        `Still present: ${stayed.map((s) => s.node.id).join(", ")}. ` +
+        `Re-read it (panel_graph_outline) before retrying: a second call for the ids that left is a no-op; ` +
+        `a second call for the ids that stayed may still fail.`;
```
artokun: This warning is the retry contract, and it is wrong.

`ids.map((id) => resolveNode(graph, id))` runs *before* the envelope. A missing id throws and refuses the whole list — the `#841 a missing id in a batch never starts the undo envelope` test is that behaviour. So after a partial `[1,2,3]` where 1 and 3 left and 2 stayed:

- retrying the original list throws on 1 and never retries 2
- retrying only the ids that left also throws
- the only useful retry is `not_removed` (`[2]`)

`a second call for the ids that left is a no-op` is the opposite of the implementation. #841's own note was not to report an outcome the caller cannot act on.

Say the true next step: re-read, then retry **only** `not_removed`. Spell out that including an id that already left refuses the whole list. Pin that in `graph-remove-node.test.mjs` so this sentence cannot drift back to "no-op".

browser_tests/unit/graph-remove-node.test.mjs
```diff
@@ -0,0 +1,193 @@
...
+  });
+  const result = fn({ node_ids: [1, 2, 3] });
+  assert.deepEqual(result.removed.map((n) => n.id), [1, 3]);
+  assert.deepEqual(result.not_removed.map((n) => n.id), [2]);
+  assert.match(result.warning, /STILL on the canvas/);
+  assert.match(result.warning, /panel_graph_outline/);
+  assert.deepEqual(graph._nodes.map((n) => n.id), [2]);
+  assert.deepEqual(events.filter((e) => e === "before" || e === "after"), ["before", "after"]);
+});
+
```
artokun, line 179: This test names the leftover ids and the warning, but it never asserts `not_removed[0].reason` and never drives a follow-up call.

After this result, `fn({ node_ids: [1, 2, 3] })` and `fn({ node_ids: [1, 3] })` must throw (missing ids) and leave 2 on the graph; `fn({ node_ids: [2] })` is the only retry that is allowed to attempt 2. That is what makes the leftover manifest actionable, and it is what the warning currently contradicts.

PR #1432, opened 2026-08-19
feat(1122): replace a published subgraph blueprint with overwrite:true

### LINKED ISSUE #1122: [Feature, parked] graph_save_subgraph overwrite — no programmatic path to update a published blueprint
Split out of #636 (defect 3) so it is not lost when that closes.

Once a subgraph blueprint is published, there is no programmatic way to update it. `graph_save_subgraph` refuses a name collision, and there is no `graph_delete_subgraph`. Any iteration on the subgraph strands the library copy, which is a dead end in the documented build-it, save-it, reuse-it flow.

**Already fixed in 0.11.83** (the parts that were defects):
- the collision preflight was blind on ComfyUI 0.31 — it reconstructed the store key from a fallback prefix and compared a name against a content hash, so `publishSubgraph()` reached its `confirmOverwrite()` dialog and hung a programmatic call on UI. It matches `display_name` too now.
- the refusal names real options instead of ending at "programmatic overwrite isn't supported", and no longer sends users after a global blueprint that `deleteBlueprint` refuses to remove.

**Still open:** `overwrite: true` on `graph_save_subgraph`, and/or `graph_delete_subgraph`.

**Parked deliberately.** The project is in a stabilisation freeze, and a new destructive flag over a user's blueprint library is not something to add during one. Design notes for whenever it is picked up:
- a global/bundled blueprint must stay non-overwritable — `deleteBlueprint` already refuses those, and an `overwrite` that bypassed it would be worse than the dead end
- overwriting must be an explicit opt-in per call, never inferred from a name collision
- the reply should say what it REPLACED, not just that it saved

Reported by the user in #636.

artokun:
Enhancement — do not merge until another agent reviews.

Fixes #1122.

`graph_save_subgraph` can now replace a user-published subgraph blueprint in place when the caller passes `overwrite: true`. ComfyUI's `confirmOverwrite()` dialog is answered programmatically (type `overwriteBlueprint` only), so a human Save dialog never appears.

Without the flag a name collision still refuses — overwrite is never inferred. Bundled/global blueprints stay non-overwritable. The reply includes `replaced` (name/type/display_name) for the library entry that was updated.

A hash-keyed / display_name-only collision is refused rather than publishing a second file under the requested name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Added
+- `graph_save_subgraph` can replace a user-published blueprint in place with `overwrite:true` — no Save dialog, bundled/global blueprints stay protected (#1122)
+
 ## [0.15.9] - 2026-08-19
 
 ### Fixed
--- a/browser_tests/unit/blueprint-collision.test.mjs
+++ b/browser_tests/unit/blueprint-collision.test.mjs
@@ -194,10 +194,12 @@ const refusalSite = src.slice(
 test("#636: a GLOBAL collision is not told to delete it", () => {
   // subgraphStore.deleteBlueprint refuses a bundled blueprint outright, so "delete it and
   // retry" is unfollowable — and on a stock install most blueprints are global, making it
-  // the likeliest collision.
+  // the likeliest collision. The wording lives in subgraphCollisionRefusalMessage so it
+  // is tested as a string; pin that this site still classifies globality and refuses
+  // through that helper (overwrite:true cannot free a bundled name — #1122).
   assert.ok(refusalSite.length > 0, "the refusal must classify the collision");
-  assert.match(refusalSite, /ships WITH ComfyUI/, "it must say why the name cannot be freed");
-  assert.match(refusalSite, /Save under a different one/, "and give the one remedy that works");
+  assert.match(refusalSite, /subgraphCollisionRefusalMessage/, "the refusal must use the shared wording");
+  assert.match(refusalSite, /subgraphSaveCollisionAction/, "and classify before throwing");
 });
 
 test("#636: UNKNOWN is not treated as global", () => {
--- a/browser_tests/unit/subgraph-blueprint-overwrite.test.mjs
+++ b/browser_tests/unit/subgraph-blueprint-overwrite.test.mjs
@@ -1,0 +1,232 @@
+// #1122 — graph_save_subgraph must be able to REPLACE a user-published blueprint
+// without a human Save dialog, and must never infer that from a name collision.
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import {
+  collisionSharesPublishKey,
+  replacedBlueprintIdentity,
+  subgraphCollisionRefusalMessage,
+  subgraphSaveCollisionAction,
+  withBlueprintOverwriteConfirm,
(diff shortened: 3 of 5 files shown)


artokun (commented):
## Verdict: Approve

Panel-side implementation of #1122 matches the spec. I did not merge.

Checked against ComfyUI frontend `publishSubgraph` / `useDialogService().confirm()` (dialog type `overwriteBlueprint` on Pinia store id `dialog`) and ran `subgraph-blueprint-overwrite.test.mjs` + `blueprint-collision.test.mjs` (34/34 pass).

### Checklist

| Requirement | Result |
|---|---|
| `overwrite: true` on a name collision updates an already-published **user** blueprint without a Save dialog | Yes. Collision + `overwrite === true` + not global + unique + same cache key → `withBlueprintOverwriteConfirm(getPiniaStore("dialog"), () => store.publishSubgraph(finalName))`. |
| Official `publishSubgraph()` path (auto-confirms `overwriteBlueprint` only) | Yes. Intercepts `dialogStore.showDialog` only when `props.type === "overwriteBlueprint"`, calls `onConfirm(true)`, lets every other dialog through. Matches `dialogService.confirm()`. |
| Overwrite is **not** inferred from a collision; only `overwrite === true` opts in | Yes. `overwrite !== true` (including omitted / truthy non-`true`) → `refuse-collision`. |
| Bundled/global blueprints stay non-overwritable | Yes. `isGlobal === true` → `refuse-global` even with the flag. Global refusal does **not** advertise `overwrite:true`. |
| display_name-only / hash-keyed hit is refused rather than writing a second file | Yes. `collisionSharesPublishKey` is type/bare-name only; `sameCacheKey !== true` → `refuse-keyed-differently`. Second line of defense: auto-confirm **requires** ComfyUI to have asked (`subgraphDefCache.has(name)`), otherwise it throws instead of publishing a new file. |
| Collision without the flag still refuses and names `overwrite:true` | Yes. Default refusal names `pass overwrite:true` plus a different name / UI delete. |
| Tests drive the shipped function | Yes. Tests import `subgraphSaveCollisionAction`, `collisionSharesPublishKey`, `replacedBlueprintIdentity`, `subgraphCollisionRefusalMessage`, `withBlueprintOverwriteConfirm` from `web/js/lib/subgraph-blueprint-overwrite.js`, plus source-slice wiring on `graph_save_subgraph`. |

Reply includes `replaced: {name, type, display_name}` as #1122 asked. First-publish still uses the new-entry observation; in-place replace cannot (key does not change) and instead requires the dialog to have been asked **and** a matching library entry to remain.

### Notes (non-blocking)

1. **MCP companion required before agents can actually pass the flag.** `comfyui-mcp` `panel_save_subgraph` still schemas `{node_id, name}` and `ctx.call`s only those two fields. Zod will strip `overwrite:true`, so an agent that follows this panel's new refusal will retry and refuse again. Land the schema + forward + tool description in `artokun/comfyui-mcp` next; this panel PR is still the right first half.

2. Dialog intercept is fail-closed if `showDialog` cannot be assigned on the live Pinia proxy (throws rather than hanging). Unit tests cover a plain object, not a Pinia setup-store proxy — acceptable for this repo's pattern; a live miss becomes a loud refusal, not a silent clobber.

3. Unknown globality (`isGlobalBlueprint` absent/throws) is treated as overwritable-if-opted-in, same rule as #636. Globals are hash-keyed on current frontends, so a display-name collision still takes `refuse-keyed-differently` / `refuse-global` rather than the replace path.

PR #1433, opened 2026-08-19
feat(1429): set ComfyUI App Mode inputs, outputs, and default mode on the live canvas

artokun:
Enhancement — do not merge until another agent reviews. Follows mcp#1709.

Closes #1429.

`graph_configure_app_mode` writes ComfyUI's official App Mode metadata on the live root canvas:

- merges `extra.linearData` (`inputs` as `[id, widgetName, config?]` tuples, `outputs` as node ids) and `extra.linearMode` (`"app"` / `"graph"`)
- validates every node id and widget against the active root graph and refuses unknowns by name
- never replaces `graph.extra` and never touches `extra.comfyui_mcp` / `WORKFLOW_META_NAMESPACE`
- one undo step via `beforeChange`/`afterChange` plus `changeTracker.captureCanvasState()` after the envelope (the frontend persist path)
- refreshes `useAppModeStore().loadSelections(...)` so builder/app view updates without a reload

Also teaches conversion's `AppBuilder.APP_MODE_PATHS` to read official `extra.linearData` tuples.

Does **not** serialize-patch-JSON + `graph_load` (rejected on the issue: undo + identity races).

MCP follow-up remains artokun/comfyui-mcp#1709 (`panel_configure_app_mode` wrapper).

Issue #1429, opened 2026-08-19
graph_configure_app_mode: set ComfyUI App Mode inputs, outputs, and default mode on the live canvas


artokun:
MCP #1709 needs a live-canvas command that writes ComfyUI's official App Mode metadata. There is no MCP-only slice: `panel_*` tools already proxy `graph_*` commands, and neither repo currently reads or writes `extra.linearData` / `extra.linearMode`.

## Why this lives here

- The panel already owns `graph.extra` (workflow-uuid identity stamp under `WORKFLOW_META_NAMESPACE` / `extra.comfyui_mcp`).
- ComfyUI frontend App Mode is stored on that same object:
  - `extra.linearData = { inputs: LinearInput[], outputs: NodeId[] }`
  - `extra.linearMode` boolean (`true` → open as app, `false` → graph)
- `LinearInput` is a tuple `[widgetId|nodeId, widgetName, config?]`.
- Frontend persist/undo path: mutate `app.rootGraph.extra`, then `workflowStore.activeWorkflow.changeTracker.captureCanvasState()` (`useAppModeStore` watcher). A `graph_load` of patched JSON is **not** an acceptable substitute (bypasses undo, races identity writes).

## Requested command

`graph_configure_app_mode` (MCP will wrap as `panel_configure_app_mode` once this ships).

Suggested payload:

```json
{
  "cmd": "graph_configure_app_mode",
  "inputs": [{ "node_id": 6, "widget": "text" }],
  "outputs": [9],
  "default_mode": "app"
}
```

- `inputs`: widget controls to expose. Validate each `node_id` exists on the active root graph and owns that widget. Refuse unknown ids/widgets with the names that failed.
- `outputs`: output node ids. Validate each exists (typically Save/Preview/ShowText/video savers; do not silently drop invalid ids).
- `default_mode`: `"graph"` | `"app"` → `extra.linearMode` boolean. Omit to leave the current flag unchanged.
- Partial updates are fine (only rewrite the fields that were sent).
- **Merge** into existing `graph.extra`. Never replace `extra` wholesale. Never touch `extra.comfyui_mcp` / `WORKFLOW_META_NAMESPACE`.
- Must be **one undo step** (`captureCanvasState` / the same envelope other graph mutators use). A tool that sits outside Ctrl+Z is worse than no tool.
- After write, `useAppModeStore().loadSelections(extra.linearData)` so builder/app view matches without a reload.

## Also broken on the read path

[...]

Issue #802, opened 2026-08-08
panel_show_media cannot display large local files outside ComfyUI served directories without manual copy


artokun:
### What happened / root cause
panel_show_media rejected a locally composed 28.4 MB MP4 because it exceeded the 20 MB inline base64 cap and was outside ComfyUI input/output directories. The user-visible result required manually copying the file into ComfyUI output/video and retrying.

### Steps to reproduce
1. Create a local MP4 larger than 20 MB outside ComfyUI served directories.
2. Call panel_show_media with its absolute path.
3. Tool errors and requires a manual filesystem copy into output/input.

### Exact error
file too large to send inline (28.4 MB > 20.0 MB cap); file is not under a directory ComfyUI serves.

### Fix
Applied locally: no. Upstream-only: yes. Consider automatically staging large local media into a safe temporary ComfyUI-served location, or expose an opt-in staging argument so panel_show_media can display it by reference in one call.

### Workaround verified
Copied the MP4 to ComfyUI output/video and retried panel_show_media; it displayed successfully by /view reference.

### Environment
Windows 11; RTX 4090 24 GB; ComfyUI 0.30.0; comfyui-mcp 0.50.35; panel 0.11.44.

Related (same root cause): #648

artokun:
Checked this against current `main`, and I want to separate the two halves because only one of them is a defect.

**The diagnostic is already complete.** The `outside` branch — your exact case, local file, not under a served directory — already enumerates the directories it checked and names the one-step remedy:

> This file is not under any directory this ComfyUI serves. Checked:
>   - output: …
>   - input: …
> What you can do:
>   1. Copy or move it under one of the directories above (a subfolder is fine) and call `panel_show_media` again with the NEW path. …
>   2. Ask the user to move it, or to open it themselves; it is on the machine they are at.

That is exactly the workaround you performed, so the tool told you what to do and you did it. Nothing here is misreporting or hiding a cause, which is the bar I would treat as a bug.

**What is left is a feature: auto-staging.** That is a real and reasonable request — "don't make me do the copy" — but it is new behaviour that **writes files into the user's ComfyUI directories**, and it carries design questions I should not answer unilaterally:

- where a staged file lands (`output/`? a dedicated `output/_staged/`?) and whether it is visible in the user's normal outputs
- whether a 28 MB copy per call is acceptable, and when it is cleaned up
- whether copying a file the user did not ask to be copied needs consent, given the tool's stated job is to *display*
[...]

PR #1436, opened 2026-08-19
release: v0.15.10

artokun:
Patch release **v0.15.10** via `scripts/set-version.mjs`. Includes:

- #1431 / #841 — `panel_remove_node` `node_ids` as one undo step
- #1433 / #1429 — App Mode inputs, outputs, and default mode on the live canvas
- #1432 / #1122 — subgraph blueprint overwrite
- #1435 / mcp#1748 — outline clip footer names note nodes whose text was clipped

Also tagged **v0.15.9** on 4d874876 (the release commit was on main without a tag).

Do **not** squash-merge. Rebase-merge or merge-commit, then tag `v0.15.10`. pyproject.toml change on main triggers publish_action.


artokun:
**Holding this one for the owner — merging it publishes four features that were parked under the freeze.**

I have not merged and will not merge it myself. Flagging rather than acting, because in this repo **merging the release publishes it**, so this PR is the point of no return rather than a bookkeeping step.

The freeze erosion I flagged on mcp#1809 and mcp#1810 turns out to be already-landed rather than proposed. On `main` right now:

| issue | label | landed via | issue state |
|---|---|---|---|
| panel#841 | `autopilot:parked` | #1431 atomic multi-remove | still **OPEN** |
| panel#1429 | `autopilot:parked` | #1433 app-mode | closed |
| panel#1122 | `autopilot:parked` | #1432 subgraph overwrite | closed |
| panel#802 | `autopilot:parked` | mcp#1810 (on mcp main, unreleased) | still **OPEN** |
| mcp#1709 | `autopilot:parked` | mcp#1809 — still open, BLOCKED | still **OPEN** |

This PR's own title — "atomic multi-remove, subgraph overwrite, app-mode" — names three of them, so the decision cannot be deferred past this merge.

**The mechanism is worth fixing regardless of which way the call goes.** Each of these went through the pipeline correctly: a claim comment, a worktree, a gated PR, green CI. What none of them did was treat `autopilot:parked` as terminal. The label is doing its job — it is on every one of these issues, and it was reaffirmed in prose the same day on #802 ("no change of decision, just moving the parked state into the labels"). Agents are reading the issue body and the thread, taking the newest comment as current direction, and not reading the label as a stop. Swarm claim comments post under the owner's account, which makes them look like re-scoping when they are not.

So the two options, and only the owner picks:

1. **Freeze holds** → this release gets cut without those commits (revert or drop them first), the parked issues stay parked, and the work waits. Nothing is lost — it is all on branches and in merged history.
2. **Freeze has moved for these** → merge as-is, and strip `autopilot:parked` from panel#841, panel#802 and mcp#1709 so the next agent does not park them a third time. panel#841 and panel#802 also need closing; they are merged but still open.

Either way, one change is worth making to stop the recurrence: the `autopilot:parked` label needs to be checked at **claim** time, not just documented as a terminal state in the brief. A parked issue currently looks identical to an unclaimed one at the moment an agent picks work.

The unrelated content in this cut is fine as far as I can tell; the hold is specifically about the parked commits, not the release mechanics.

artokun:
**Update — the mcp side has already shipped, and that makes holding this PR *alone* the worst of the three options.**

Facts as of 19:20Z, measured not assumed:

- **v0.52.15 was tagged at 19:13Z and its Release workflow is `in_progress`.** npm `latest` is still 0.52.14, so the publish is minutes away, not done.
- `git tag --contains` puts **both** parked commits inside that tag: `1c44a0a` feat(802) and `37a0b0c` feat(1709) App Mode. mcp#1809 merged at 18:27Z, after the freeze flag.
- The App Mode work is **cross-repo and paired**. mcp's `panel_configure_app_mode` is a thin pass-through that sends the bridge command `graph_configure_app_mode`. That command exists **only on panel `main` (the unreleased 0.15.10)** — it is absent from the published v0.15.9. There is no version gate on the mcp side.
- The published panel's dispatcher does `if (!executor) throw new Error(\`Unknown command "${msg.cmd}"\`)`. So the failure is **clean and loud, not a hang** — which caps the severity, but it is still a tool that ships and cannot work.

So the three options are no longer symmetric:

1. **Merge this** → the pair is coherent. The freeze is breached, but nothing ships broken.
2. **Hold this AND cut a follow-up mcp release that drops `panel_configure_app_mode`** → also coherent, honours the freeze, costs one extra release.
3. **Hold this alone** → 0.52.15 publishes a tool whose bridge command does not exist in the released panel. Every call fails with `Unknown command "graph_configure_app_mode"`. This is the one option to avoid.

**I did not cancel the Release run**, and I want to be explicit that this was a judgement call rather than an oversight. Cancelling mid-publish is how you get a tag with no npm artifact — the exact v0.52.5 gap that was cleaned up earlier today — and v0.52.15 also carries legitimate non-parked fixes that have nothing to do with this. Un-publishing later is a follow-up release; a half-published tag is a mess to unpick. If that was the wrong call, it is reversible with a version bump.

The count is now **seven** parked issues rebuilt this sweep: panel#841, #1122, #1429, #802, mcp#1709, plus **mcp#1400** (PR mcp#1817) and **mcp#1294** (PRs mcp#1812 / panel#1437), which arrived in the 19:16Z poll and are flagged on their own threads. The last three are still open and are the ones still cleanly stoppable.

Issue #1294, opened 2026-08-17
panel_copy_nodes/panel_paste_nodes lose groups and collapse branch positions


artokun:
### What happened / root cause
panel_copy_nodes + panel_paste_nodes copied all 121 selected live-canvas nodes into a new workflow, but the pasted graph lost every group (group_count became 0) and several nodes from different branch rows were pasted at the same position. Example: five "Power Lora Loader (rgthree)" nodes that were at y=-20/640/1300/1960/2620 appeared at the same pasted position [2288,1320].

### Steps to reproduce
1. Open a grouped 121-node workflow.
2. panel_query_graph(fields:"ids", limit:200).
3. panel_copy_nodes(node_ids:[all 121 ids]).
4. panel_new_workflow().
5. panel_paste_nodes(connect_inputs:false).
6. panel_graph_outline() reports 121 nodes, 0 groups; inspect branch nodes and see collapsed coordinates.

### Exact error
No tool error was returned. panel_copy_nodes returned copied:121 and panel_paste_nodes returned pasted_count:121, despite the structural loss.

### Fix
applied locally: no  upstream-only: yes
Workaround verified: create a blank workflow, then panel_load_workflow(path:"Krea2_LoRA_AB_Compare.json"). This preserved 121 nodes, 11 groups, and the distinct branch positions.
Precise change needed: clipboard serialization/paste should preserve groups whose member nodes are fully selected, and apply one consistent translation to every pasted node instead of collapsing same-type/branch nodes to one coordinate. The result should report copied/pasted group counts so structural loss is detectable.

### Environment
Windows 11
ComfyUI 0.33.1
ComfyUI frontend 1.48.7
Python 3.12.10
PyTorch 2.13.0+cu130
GPU NVIDIA GeForce RTX 5090 31.8 GB
comfyui-mcp 0.51.57
panel 0.13.7

artokun:
Claimed by grok-swarm 2026-08-19T05:59:08Z — driving this in worktree `/c/Users/Artokun/code/comfyui-mcp-panel/.worktrees/wt-1294` (branch `fix/1294-copy-paste-groups`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
[...]

PR #1437, opened 2026-08-19
feat(1294): graph_unexpose_subgraph_input/output — a supported way to remove a subgraph boundary slot

artokun:
## Root cause

The subgraph boundary surface was asymmetric (artokun/comfyui-mcp#1294): `panel_query_graph` hands back rail slots, `panel_expose_subgraph_input/output` create them and `panel_move_rail` moves the rail — but nothing could REMOVE a boundary slot. The only path was deleting the interior node feeding it, and the #930 diagnosis message had to tell callers exactly that. This is the panel half: the executors. The tool registration half is artokun/comfyui-mcp (separate PR, same branch name).

## What changed

- **`graph_unexpose_subgraph_input` / `graph_unexpose_subgraph_output`** executors (run INSIDE the subgraph, like the expose twins). The slot is identified by **name or index** as `panel_query_graph` lists it under `rails.<side>`.
- **Refuses loudly rather than guessing**: an unknown name refuses and lists the slots that exist; an ambiguous digit-name refuses (reusing `findExistingRailSlot`'s conflict rule); a **rail_node_id (-10/-20) is refused by name** — it names the whole rail, not a slot on it, and silently indexing with it would remove an unrelated slot.
- **Destructive wires are disclosed, not silently dropped**: interior links on the slot and parent-graph wires on every host SubgraphNode's matching slot are counted BEFORE the removal and reported as `removed.interior_links_dropped` / `removed.host_links_dropped` (and on the result card).
- **The live rail is the verdict, not the call** — same lesson as #1272: a listener can cancel the frontend's `removing-input`/`removing-output` event (a clean return is then reported as the failure it is), and a throw from `removeInput`/`removeOutput` with the slot gone is reported as landed with the throw disclosed as a warning.
- The **#930 rail-id diagnosis** (`node-scope-locator.js`) claimed "There is currently NO unexpose/remove-boundary operation" — false once this lands; it now names the pair and warns the rail id is refused there too.
- Re-vendored `tool-vocabulary.json` (+ `VENDORED_VOCABULARY_HASH`) with the two new tool names; result-card strings translated in all 12 locales; CHANGELOG entry.

## Verification

- `npm ci`, `npm run test:unit` — 5661 pass, 0 fail (includes i18n-check, i18n-render-check, check-tool-vocabulary, check-panel-scope)
- `npm run typecheck` — clean
- New `browser_tests/unit/unexpose-boundary-slot.test.mjs` runs the REAL executors (sliced from the shipped source, LiteGraph-shaped doubles, real rail-slot helpers): happy-path removal with drop counts, unknown-name refusal, rail-id refusal, canceled-removal failure, throw-but-landed warning, root-graph refusal.

Refs artokun/comfyui-mcp#1294 (the MCP-side registration PR carries the `Closes`).


artokun:
**This PR builds an `autopilot:parked` issue.** Flagging, not blocking — the merge call is the owner's.

The issue carries `autopilot:parked`, which is a terminal state: a previous agent already decided it and wrote the reason. It is not a stale label, and a swarm claim comment on the thread is not a re-scoping of it — claim comments post under the owner's account, which makes them read as direction when they are an agent announcing itself.

This is now the **seventh** parked issue rebuilt in this sweep (panel#841, #1122, #1429, #802, mcp#1709, #1400, #1294). Every one had a proper claim, worktree, gated PR and green CI — the pipeline is working; the stop signal is not being read. Full context and the ship-or-revert decision are on artokun/comfyui-mcp-panel#1436, which is the release that would publish the panel-side ones.

I have deliberately not gated this on its merits, because gating implies the build decision is settled. If the owner says build, I will gate it properly.

`.claude/autopilot/BRIEF.md` step 3 now requires reading the label **before** claiming, so this should stop recurring for agents that read the brief.

artokun:
Rebased + CI fix pushed 313b86c882b11a109994d78922a8dd792f36d6ec (worktree wt-1294). Leaving merge for independent review.

artokun (commented):
INTENDED: APPROVE

Unexpose mutates the live viewing subgraph, not a copy. `graph_unexpose_subgraph_output` / `graph_unexpose_subgraph_input` take `getGraphCtx().graph` — the same `node.subgraph` object `panel_enter_subgraph` opened via `canvas.openSubgraph`, or the registered definition `resolveScope` admits through `isSubgraphInRoot` — and call `subgraph.removeOutput(slot)` / `removeInput(slot)`, which is already the frontend contract `pruneOrphanedBoundaries` and the expose-twin cleanup use. Host wires are counted first (`countHostRailLinks`, `node.subgraph === subgraph` including nested hosts); the live rail is the verdict: a cancelable `removing-*` event that leaves the slot in place is a thrown failure, a throw with the slot gone is a landed removal plus warning.

Names and payload keys are English (`removed.interior_links_dropped` / `host_links_dropped`, locale keys `removed_boundary_slot*`, vocab `panel_unexpose_subgraph_*`). Tests drive the shipped path: `unexpose-boundary-slot.test.mjs` slices the real executors out of `comfyui-mcp-panel.js` and injects the real `rail-slot.js` helpers; `rail-slot.test.mjs` pins the import and both call sites. Refusal cases (unknown name, rail_node_id -10/-20, root graph) remove nothing.

Non-blocking: `countHostRailLinks` has no dedicated nested/multi-host unit test (only the one-host executor path), and the input cancel/throw branches are twins of the tested output ones. The MCP registration half (artokun/comfyui-mcp#1812) is required before an agent can invoke these.

artokun:
Rebased onto main after #1439/#1440. Head is 962a2c802108c6f9cd0673158266dde1fff1d38f (worktree wt-1294). Independent review already posted INTENDED: APPROVE; not merging.

PR #1439, opened 2026-08-19
fix(1434): panel_save_workflow replies when the save hangs instead of claiming the tab is frozen

### LINKED ISSUE #1434: panel_save_workflow times out while active tab remains responsive
panel_save_workflow delivered workflow_save to the active, pinned workflow tab but the tab never acknowledged it within 15 seconds. Other live-canvas reads continued to work immediately afterward (panel_graph_outline and panel_list_workflows), and panel_list_workflows reported the same active workflow as persisted:true but modified:true. This points to the panel/frontend workflow_save request handler hanging or failing to send an acknowledgement, not a disconnected tab.

### Steps to reproduce
1. Open and pin workflows/video_minimax_h3_r2v_ai_test.json.
2. Mutate widgets/node modes and a MarkdownNote through panel tools.
3. Call panel_save_workflow({}).
4. Observe timeout.
5. Verify panel_graph_outline still responds and shows the mutations.
6. Retry the identical save with the supplied retry_of token.
7. Observe the same timeout; panel_list_workflows still reports modified:true.

### Exact error (scrubbed)
Error: Panel tab wf:[REDACTED]:workflows/video_minimax_h3_r2v_ai_test.json did not reply to "workflow_save" within 15000 ms — the ComfyUI tab may be backgrounded or frozen. This command MUTATES and was already delivered to the tab, so it may have been applied despite the missing reply.

First retry token was used exactly as instructed; the second attempt timed out identically.

### Fix
Applied locally: no. Upstream-only in this session: yes (filesystem sandbox is read-only).
Suggested change: make the frontend workflow_save handler always send a terminal acknowledgement in a finally block after the save promise resolves or rejects, include the actual save exception in the response, and have the orchestrator reconcile against the frontend dirty/modified state before returning outcome-unknown. The active-tab check should not claim "backgrounded or frozen" when other RPCs from the same tab still answer.

### Environment
Windows 11
ComfyUI 0.33.2
ComfyUI frontend 1.49.6
Python 3.13.12
PyTorch 2.12.1+cu130
GPU NVIDIA GeForce RTX 5090, 31.8 GB VRAM
comfyui-mcp 0.52.7
comfyui-mcp-panel 0.15.8
Queue idle; recent ComfyUI errors: none.

---

Related (same root cause): #1404, #589

artokun:
## Root cause

`panel_save_workflow` relays `workflow_save` at 15,000 ms. The handler awaited `programmaticSave` with no bound, so a `/userdata` HEAD, GET, or PUT that accepts and never answers parked the rid until the orchestrator gave up with `did not reply to "workflow_save" … the ComfyUI tab may be backgrounded or frozen`.

The reporter's tab was not frozen: `panel_graph_outline` and `panel_list_workflows` from the same tab answered immediately afterward, and list still showed `persisted:true` / `modified:true`. The retry_of token then waited on that same in-flight promise and timed out identically.

## Fix

- `web/js/lib/workflow-save-budget.js`: a 13 s command budget (2 s reply slack under the 15 s relay window). `runBoundedWorkflowSave` always settles: a hung save throws a worded refusal that reports the live dirty/persisted observation and explicitly says the tab is not frozen; a real save exception is rethrown as-is, never rewritten as a hang. The save is not cancelled (`withTimeout` never cancels).
- `web/js/comfyui-mcp-panel.js`: both `workflow_save` and `workflow_save_as` wrap `programmaticSave` in that bound and snapshot `isModified` / `isPersisted` at fire time.

## Verification

- `node --test browser_tests/unit/workflow-save-budget.test.mjs` — 10 pass. Drives the shipped `workflow_save` / `workflow_save_as` bodies extracted from the panel over the real bound, with a hanging `programmaticSave` (the reported shape: persisted, still-dirty, userdata never settles). Also: real save exceptions propagate, a completing save still reports `saved:true`, budget stays under 15 s, both handlers wire the bound.
- `node --test browser_tests/unit/workflow-save.test.mjs browser_tests/unit/new-workflow-persistence.test.mjs browser_tests/unit/save-reply-identity.test.mjs browser_tests/unit/save-as-identity.test.mjs browser_tests/unit/command-budget.test.mjs` — pass.
- `node scripts/check-panel-scope.mjs` — OK.

Fixes #1434

### DIFF
--- a/browser_tests/unit/workflow-save-budget.test.mjs
+++ b/browser_tests/unit/workflow-save-budget.test.mjs
@@ -1,0 +1,283 @@
+/**
+ * panel#1434 — `panel_save_workflow` timed out with no acknowledgement for 15 s
+ * against a tab that kept answering `panel_graph_outline` and `panel_list_workflows`,
+ * and still reported persisted:true / modified:true.
+ *
+ * The reply was not lost in transit. It was never COMPOSED inside the window:
+ * `workflow_save` is relayed at 15,000 ms and awaited `programmaticSave` with no
+ * bound, so a /userdata HEAD, GET or PUT that accepts and never answers parks the
+ * rid until the orchestrator gives up and guesses the tab is backgrounded or frozen.
+ *
+ * THE HARNESS RUNS THE SHIPPED `workflow_save` BODY, extracted from the panel source
+ * and given injected collaborators, over the REAL `runBoundedWorkflowSave` with a
+ * REAL hanging save — the same technique as refresh-nodes-command-budget.test.mjs.
+ * A helper-level test cannot reach this defect: `runBoundedWorkflowSave` already
+ * implements the bound, and the whole bug was that the call site never used it.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { withTimeout } from "../../web/js/lib/bounded-step.js";
+import {
+  WORKFLOW_SAVE_COMMAND_BUDGET_MS,
+  WORKFLOW_SAVE_TIMEOUT,
+  describeWorkflowSaveTimeout,
+  runBoundedWorkflowSave,
+  workflowSaveTimeoutObservation,
+} from "../../web/js/lib/workflow-save-budget.js";
+import { PANEL_SRC } from "./_panel-constants.mjs";
+
+const workflowSaveMatch = PANEL_SRC.match(/\n {2}async workflow_save\(\{ name \} = \{\}\) \{[\s\S]*?\n {2}\},/);
+assert.ok(workflowSaveMatch, "could not locate workflow_save in panel source");
+
+const workflowSaveAsMatch = PANEL_SRC.match(/\n {2}async workflow_save_as\(\{ name \}\) \{[\s\S]*?\n {2}\},/);
+assert.ok(workflowSaveAsMatch, "could not locate workflow_save_as in panel source");
+
+// The orchestrator side of this invariant: ctx.call(..., 15000) in
+// comfyui-mcp's panel-tools.ts panel_save_workflow. Duplicated as a literal
+// on purpose — if EITHER side moves, this test forces the relationship to be
+// re-examined.
+const ORCHESTRATOR_SAVE_TIMEOUT_MS = 15000;
+
+function deferred() {
+  let resolve;
+  const promise = new Promise((r) => {
+    resolve = r;
+  });
+  return { promise, resolve };
+}
+
+async function withWatchdog(run, ms, what) {
+  let timer;
+  const startedAt = Date.now();
+  const watchdog = new Promise((_, reject) => {
+    timer = setTimeout(() => reject(new Error(`${what} — waited ${ms}ms`)), ms);
+  });
+  try {
+    const value = await Promise.race([Promise.resolve().then(run), watchdog]);
+    return { value, elapsed: Date.now() - startedAt };
+  } finally {
+    clearTimeout(timer);
+  }
+}
+
+function shippedWorkflowSave({ programmaticSave, budgetMs = 40, observeWorkflow } = {}) {
+  const deps = {
+    runBoundedWorkflowSave,
+    programmaticSave,
+    WORKFLOW_SAVE_COMMAND_BUDGET_MS: budgetMs,
+    withTimeout,
+    monotonicNow: () => Date.now(),
+    observeActiveWorkflowSaveState: observeWorkflow,
+    saveProducedIdentity: () => ({ uuid: "u", routingKey: "wf:x" }),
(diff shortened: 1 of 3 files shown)


artokun:
**The red check is a one-line comment, not a code problem.**

Failing gate: `check-tool-vocabulary.mjs` in the *Lint pack* job.

```
1 reference(s) to a removed tool name:
  web/js/lib/workflow-save-budget.js:7  list_workflows
    // the SAME tab answered immediately afterward, and list_workflows still showed
    → use get_workflow (action:"list")  (list_workflows: 0.50.0)
```

That gate **reads comments**, not just code — it is scanning prose for tool names retired in 0.50.0. Line 7 quotes the field report verbatim, and the reporter used the old name.

The minimal fix is to reword that one line, e.g.:

```js
// the SAME tab answered immediately afterward, and get_workflow (action:"list") still showed
```

Note `panel_list_workflows` on line 6 is **not** flagged — that name is still current, so only the bare `list_workflows` on line 7 needs changing. Worth resisting the urge to "fix" both.

Everything else about the check is green; this is the only thing standing between the PR and a clean Lint pack. I have not pushed to your branch — you have a live worktree (`wt-1434`) and two agents writing into one checkout is how commits get silently relocated, so the change is yours to make.

For what it is worth, the underlying diagnosis in that header comment reads well: unbounded `/userdata` HEAD/GET/PUT parking the reply for the whole browser timeout, with `retry_of` waiting on the same in-flight promise, is a clean account of the failure.

artokun:
## Gate verdict: **NO-SHIP** (2 × P1)

> **Disclosure:** codex was unavailable, so this was gated by an independent adversarial review spawned fresh against the diff (`claude-gate.mjs`, exit 1). It is not a self-review — the reviewer never saw the author's reasoning, and I did not substitute my own opinion for its verdict.

CI is green and the bounding mechanism itself is sound. The problem is the **message the refusal renders**, which is reachable, agent-visible, and states things it cannot know.

### P1 — `web/js/lib/workflow-save-budget.js:84` — `modified:false` cannot mean "the write landed"

The refusal tells the caller *"if modified is false the write landed"*. But `isModified === false` is equally the state of a workflow **that was never dirty**. The two are indistinguishable, and the code resolves that ambiguity into a definite verdict.

This is not hypothetical: `saveInPlace` (`workflow-save.js:1768`) always calls `svc.saveWorkflow(wf)`, and the shipped frontend's `ComfyWorkflow.save()` calls `super.save({force:true})`, bypassing the `isPersisted && !isModified` early return. So on a persisted, unmodified workflow whose `/userdata` PUT hangs, the flag reads false **before the command started** and stays false whether the PUT lands or 500s.

The remedy the message names cannot check what it claims: `panel_list_workflows` reports `modified: !!w.isModified` (`comfyui-mcp-panel.js:16327`) — the *same flag*. On the Save-As leg it is self-contradictory in one paragraph: `persisted:false` (proof the write did **not** land) next to *"if modified is false the write landed"*.

### P1 — `comfyui-mcp-panel.js:6120` — the observation names the wrong workflow

`observeActiveWorkflowSaveState()` samples `activeWorkflow` at **timeout-fire time**, not the workflow the save targeted, and `workflow-save-budget.js:71` then renders it as the save's subject (`for "<filename>"`).

For `panel_save_workflow({name:"NewName"})` from a persisted `Old.json`, the module's own header names the `/userdata` **HEAD** as a hang source — and that probe runs *before* `saveAs`/`openWorkflow` (`workflow-save.js:514`, `:762`), so no swap has happened yet. The reply blames `Old.json`, a file that was never the target and whose flags say nothing about `NewName.json`. Same defect for any active-workflow change inside the 13 s window.

### What passed

Worth saying plainly, because most of this PR is good: the bounding is correctly built. The call-site mutation (back to bare `await programmaticSave(name)`) turns **2 tests red**, so the fix is pinned rather than merely covered. Full suite 5724 pass / 0 fail. Rejections are enveloped before bounding, so a real save error is rethrown unchanged, with no unhandled rejection on late settle. The 13 000 / 15 000 margin is real on both legs (`panel-tools.ts:11698-11699`), and "not a backgrounded or frozen tab" is warranted.

Both executors are the only `panel_save_workflow` path and `if (res.isError) return res` hands the text to the agent verbatim — the message is **reachable**, which is exactly why its content binds.

### To clear the gate

Report what was observed, not what it implies: state the flags (`modified`, `persisted`) as raw observations, drop the "if modified is false the write landed" inference entirely, and capture the target workflow's identity **before** the HEAD probe so the filename in the reply is the one that was actually saved. A timeout that cannot tell landed from never-dirty should say so — "could not determine" is the honest terminal state here, and it is the one the caller can act on.

I am not merging this. Re-gate after the message changes and it should clear quickly; the machinery underneath is fine.

PR #1440, opened 2026-08-19
fix(1357): a pasted LoadImage file with spaces is no longer flagged missing

### LINKED ISSUE #1357: panel_get_errors falsely flags uploaded LoadImage subfolder paths as missing after server-confirmed panel_set_widget

artokun:
Fixes #1357

`panel_get_errors` again reported `pasted/image (992).png` on a LoadImage node as `unavailable_widget_values.kind: "missing_asset"` after #1368, even though the nested input exists, `/view` returns 200 and 206 with the range probe, and the exact value executed. No `unchecked_nodes` entry — an asserted miss, not a disclosed abstention.

#1368 already asked the server about upload values the combo cannot enumerate. The probe built that `/view` query with `URLSearchParams`, which encodes a space as `+`. On some ComfyUI/aiohttp stacks that plus is a literal plus, so `image (992).png` was asked as `image+(992).png` and 404ed. The original #1368 value (`AgentLibrary/HaReen/Main-9-1.png`) had no spaces, so the probe looked fine.

The probe now percent-encodes like ComfyUI's own `getResourceURL`. `decodeURIComponent` of the filename param is the file on disk. A real 404 is still reported.

Worktree: `C:/Users/Artokun/code/comfyui-mcp-panel/.worktrees/wt-1357`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+- `panel_get_errors` no longer flags a pasted LoadImage file like `pasted/image (992).png` as missing when the file is on disk and `/view` serves it (#1357)
+
 ## [0.15.13] - 2026-08-19
 
 ### Fixed
--- a/browser_tests/unit/input-asset.test.mjs
+++ b/browser_tests/unit/input-asset.test.mjs
@@ -13,6 +13,8 @@ import {
   filterServerConfirmedInputSubfolderCandidates,
   inputPathsUseWindowsSeparators,
   addComboOption,
+  inputAssetViewQuery,
+  probeInputAssetPresence,
 } from "../../web/js/lib/input-asset.js";
 
 const DEFS = {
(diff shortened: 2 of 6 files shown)


artokun:
**Diagnosis of the red check — this one is caused by the PR, and the failing assertion should not be relaxed.**

Failing test: `browser_tests/unit/set-widget-stale-combo-budget.test.mjs:469`
`#1418 wiring: a hung /view probe gives up at the budget instead of outliving the relay`
`AssertionError: the probe did run — it is bounded, not skipped` — `expected: true, actual: false`.

**Measured, not inferred.** I ran the file on both sides:

- `origin/main` (`c2bacd7c`) — **11/11 pass**, including this exact test.
- this branch (`97bcca0c`) — same file, same machine, **fails** on `probed`.

So it is not a flake, not load, and not a pre-existing break on main. `probed === false` means the injected `/view` fetch was never called at all: the probe stopped running rather than running-and-timing-out.

**Why the assertion matters.** It pins #1418: the probe must *run and be bounded*, because "did not answer in time" is the answer the refusal is built on. A probe that is skipped produces the same visible outcome here — the write is still refused with `not a valid option` — which is exactly why the test checks `probed` separately. Deleting or loosening that line would make the check green while silently un-shipping #1418, and nothing else in the suite would notice.

**Two hypotheses I tested and refuted**, so you don't spend time on them:

1. *"The fix short-circuits the probe for subfolder-style values."* No — `assert.rejects(/not a valid option/)` still passes, so the refusal path is intact; only the probe call disappeared.
2. *"The harness allowlist `EXECUTOR_DEPS` is missing the new import."* No — I checked the sliced source directly, and `probeAssetOnServer`, `probeInputAssetPresence`, `inputAssetViewQuery` and `inputAssetProbeVerdict` have **zero** occurrences inside the extracted `graph_set_widget` body. The new import is not reached from there.

**Where I stopped.** The probe reaches `runSetWidget` through an injected dep rather than through the sliced source, so the seam is between the panel's probe construction — which this PR rewrote from inline into `probeAssetOnServer(ref, timeoutMs, fetchApi)` — and whatever `lib/set-widget.js` actually receives. I did not pin which dep drops it, and I would rather hand you a verified boundary than a plausible mechanism; I have steered fixes wrong that way before.

The encoding fix itself looks right and is well-tested (`URLSearchParams` turning a space into `+` where `/view` wants `%20` is a real bug, and the tri-state extraction is a good shape). This is about the third consumer of the probe, not about that change.

The `wt-1357` worktree is preserved — I skipped it during the post-release worktree sweep precisely because this PR is still open.

PR #1456, opened 2026-08-19
fix(1455): the save timeout reports what it observed, never that the write landed

### LINKED ISSUE #1455: panel_save_workflow timeout reply states "the write landed" from a flag that cannot show it, and can name the wrong workflow
`panel_save_workflow`'s timeout reply, added in #1439 (shipped in **v0.15.14**), tells the caller how to determine whether the write landed — using a signal that cannot answer the question, and attributing the save to a workflow that may not be its target.

The bounding mechanism from #1434 is sound and is not what this is about. This is the **message it renders**, which reaches the agent verbatim (`if (res.isError) return res`), so its content is load-bearing.

These were raised as a NO-SHIP gate verdict on #1439 shortly before it merged. Filing so the findings are not lost with the PR.

---

### P1 — `modified:false` cannot mean "the write landed" (`web/js/lib/workflow-save-budget.js:84`)

```js
`Check panel_list_workflows before retrying: if modified is false the write landed; `
```

`isModified === false` is equally the state of a workflow **that was never dirty**. The two states are indistinguishable, and the message resolves that ambiguity into a definite verdict.

It is reachable rather than theoretical: `saveInPlace` (`web/js/lib/workflow-save.js:1768`) always calls `svc.saveWorkflow(wf)`, and the shipped frontend's `ComfyWorkflow.save()` calls `super.save({ force: true })`, bypassing the `isPersisted && !isModified` early return. So for a persisted, unmodified workflow whose `/userdata` PUT hangs, the flag reads false **before the command started** and stays false whether the PUT eventually lands or 500s.

The remedy the message names cannot observe what it claims: `panel_list_workflows` reports `modified: !!w.isModified` (`web/js/comfyui-mcp-panel.js:16327`) — the same flag.

On the Save-As leg it contradicts itself within one paragraph: the reply renders `persisted:false.` (proof the write did **not** land) alongside `if modified is false the write landed`.

### P1 — the observation can name a different workflow than the one saved (`web/js/comfyui-mcp-panel.js:6122`)

`observeActiveWorkflowSaveState()` samples `activeWorkflow` at **timeout-fire time**, not the workflow the save targeted; `workflow-save-budget.js:71` then renders it as the save's subject (`for "<filename>"`).

For `panel_save_workflow({ name: "NewName" })` from a persisted `Old.json`: this module's own header names the `/userdata` **HEAD** as a hang source, and that probe runs *before* `saveAs`/`openWorkflow` (`web/js/lib/workflow-save.js:514`, `:762`), so no swap has occurred. The reply reads `workflow_save did not finish within 13s for "Old.json" … still reports modified:true … persisted:true` — attributing the hung write to a file that was never its target, whose flags say nothing about `NewName.json`. The same defect applies whenever the active workflow changes inside the 13 s window.

### Suggested direction

Report what was observed, not what it implies:

- state `modified` / `persisted` as **raw observations**, and drop the "if modified is false the write landed" inference;
- capture the target workflow's identity **before** the HEAD probe, so the filename in the reply is the one actually saved;
- when the timeout genuinely cannot distinguish landed from never-dirty, say **"could not determine"** — that is the honest terminal state, and the one a caller can act on. A confident wrong answer here is worse than no answer, because the agent will skip the retry.

### Not affected

The bounding itself is well built and pinned: mutating the call site back to a bare `await programmaticSave(name)` turns two tests red (`REPLIES at its budget`, `BOTH save handlers wrap`), the suite was 5724 pass / 0 fail, rejections are enveloped before bounding so a real save error is rethrown unchanged, and the 13 000 / 15 000 margin is real on both legs (`comfyui-mcp/src/orchestrator/panel-tools.ts:11698-11699`).

artokun:
Fixes #1455.

`panel_save_workflow`'s timeout reply told the caller **"if modified is false the write landed"**. It cannot know that. `isModified === false` is equally the state of a workflow that was never dirty, so the flag cannot separate the two cases — and this is reachable, not theoretical:

- `saveInPlace` (`web/js/lib/workflow-save.js:1768`) always calls `svc.saveWorkflow(wf)`;
- the shipped frontend's `ComfyWorkflow.save()` calls `super.save({ force: true })`, bypassing the `isPersisted && !isModified` early return.

So a persisted, clean workflow reaches a hanging `/userdata` PUT with the flag **already false**, and it stays false whether the PUT lands or 500s. The remedy the message recommended reads that same flag — `panel_list_workflows` reports `modified: !!w.isModified` (`comfyui-mcp-panel.js:16327`) — so it could not settle the question either. On the Save-As leg the reply contradicted itself in one paragraph: `persisted:false` (proof the write did **not** land) beside "if modified is false the write landed".

**Second defect, same reply.** `observeActiveWorkflowSaveState()` sampled `activeWorkflow` at *timeout-fire time*, not the workflow the save targeted. The `/userdata` HEAD probe can hang **before** `saveAs`/`openWorkflow` (`workflow-save.js:514`, `:762`), so a `{name:"NewName"}` save from `Old.json` attributed the hung write to `Old.json` — a file that was never the target, whose flags say nothing about `NewName.json`.

## What changed

- The reply reports the flags as **observations** and stops. `modified:true` keeps its one-directional meaning (the canvas was never marked clean, so nothing was acknowledged); `modified:false` is reported as **UNDETERMINED**.
- The save's **target is captured before the save runs**. If the active workflow moves during the budget, the reply names the target, discloses the swap, and withholds flags that describe a different workflow rather than misattributing them.
- Retry guidance now points at reading the saved file itself, since no dirty flag can answer it.

## Verification

- `browser_tests/unit/workflow-save-budget.test.mjs` — **14/14**, four of them new.
- **Mutation-tested, both directions.** Reverting the pre-capture to `subject = observed.filename` turns the wiring test red; restoring the old "if modified is false the write landed" string turns the `modified:false` test red. Each mutation kills exactly one test, and the baseline restores to 14/14 — so the tests pin the fix rather than merely covering the file.
- Full unit suite: **5752 pass / 0 fail** (5753 total).
- `check-tool-vocabulary` OK; `check-panel-scope` OK — every name in the panel bundle resolves.

## Note on one changed assertion

The existing `#1434` test asserted the reply matches `/panel_list_workflows/`. That pinned the *recommendation this issue is about*, so it is replaced rather than removed — the test now asserts the message does **not** claim `if modified is false the write landed`. That is a strictly stronger assertion, not a loosened one.

## Provenance

Both defects were raised as a gate **NO-SHIP** on #1439. That PR merged about a minute later and shipped in **v0.15.14**, so the findings are live on `main` — this is the follow-through, not a re-litigation of #1439, whose bounding mechanism was sound and is untouched here.

### DIFF
--- a/browser_tests/unit/workflow-save-budget.test.mjs
+++ b/browser_tests/unit/workflow-save-budget.test.mjs
@@ -284,7 +284,9 @@ test("#1434: BOTH save handlers wrap programmaticSave in the bound — the helpe
 });
 
 // ---------------------------------------------------------------------------
-// #1455 — the reply reports OBSERVATIONS, never an inference the flags cannot support.
+// #1455 — the reply reports OBSERVATIONS, and is explicit about WHICH workflow
+// they belong to. The save's destination is the requested name, never a guess
+// made from whichever workflow happens to be active.
 // ---------------------------------------------------------------------------
 
 test("#1455: modified:false is reported, never resolved into 'the write landed'", () => {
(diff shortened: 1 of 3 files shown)


artokun:
**Gate round 1: NO-SHIP — and it was right. Force-corrected.**

> Disclosure: codex was unavailable, so this was gated by an independent adversarial reviewer spawned fresh against the diff (`claude-gate.mjs`, exit 1). Not a self-review.

My first attempt captured the workflow that was **active before the save** and called that the target. That is wrong, and on the routes that matter it was a **strict regression** on the pre-existing behaviour.

The relocating routes call `openWorkflow(copy)` **before** `saveWorkflow(copy)` (`web/js/lib/workflow-save.js:1353`). So for `workflow_save_as({name:"Copy.json"})` from an active `Original.json`, the pre-save active workflow is the **source** — the one file that route provably never writes. My reply named `Original.json` as the target and told the caller to go read it. They would have found it intact and concluded the save was fine, having learned nothing about `Copy.json`, whose existence is the actually-undetermined thing. The old code named `Copy.json` correctly.

Worse, **my own test encoded that wrong target as the expected output** (`assert.match(text, /for "Old\.json"/)` on a Save-As). It passed 14/14. A test that pins the wrong behaviour is not coverage, it is a lock on the defect — which is precisely the failure class this issue is about, applied to my own patch.

The reviewer also caught the fallback `subject = target.filename ?? observed.filename` collapsing **"I could not read the target"** into **"the target did not move"** — the same "two states, one observation" defect this PR argues is unacceptable for `modified:false`.

### What the corrected version does

The destination was available at both call sites all along (`workflow_save_as({ name })`) and was never passed. It is now, and three cases are kept apart:

| case | reply |
|---|---|
| `requested` given | that **is** the destination, whatever the canvas shows; if the active workflow differs, its flags are **withheld** rather than attributed |
| no `requested`, canvas unmoved | in-place save — the flags describe it |
| no `requested`, canvas moved | first-save auto-naming — which file the hung write targets **cannot be determined**, and the reply says so |

The third case is deliberately *not* folded into the second.

### Verification

- 17/17 in `workflow-save-budget.test.mjs`; full suite **5755 pass / 0 fail** (5756 total).
- `check-tool-vocabulary` OK, `check-panel-scope` OK.
- **Three mutations, all killed:** dropping `targetName: name` from both call sites → 1 red; ignoring `requested` → 3 red; collapsing the undeterminable case into "unchanged" → 1 red. Baseline restores to 17/17.
- Added a **source-level** wiring test. `targetName: name` is a one-line call-site option and is invisible to a helper-level test — without it, deleting the wiring would leave every other assertion green while the shipped reply went back to guessing.

Re-gating now.

PR #1462, opened 2026-08-20
fix(1459): key the requested name and the canvas name with the operation each needs

### LINKED ISSUE #1459: nameKey double-strips the frontend filename, so an externally-placed X.json.json can read as the save destination
Raised as a **P2 non-blocker** by the gate on #1458. Filing so it is recorded rather than lost with the PR.

`nameKey()` in `web/js/lib/workflow-save-budget.js` normalizes both sides of the "is the canvas the save's destination?" comparison by stripping the directory and a trailing `.json` / `.app.json`. That is correct for the requested name, which arrives raw from the caller.

It is one strip too many for the other side. `ComfyWorkflow.filename` is **already** extension-stripped by the frontend (`UserFile` → `getPathDetails` → `getFilenameDetails`), so applying `baseName()` to it strips a second time — the same double-strip the repo documents as unsafe at `workflow-save.js:650`.

### The input that separates them

An on-disk `Foo.json.json` reports `filename: "Foo.json"`. `nameKey` normalizes that to `"Foo"`, which then matches a requested `"Foo"` — so a genuinely different workflow reads as the destination, the "the active workflow is not the save's destination" disclosure is suppressed, and the *source's* dirty/persisted flags are printed as though they described the target.

### Why it is P2 and not P1

- The panel cannot create such a file: `baseName()` strips the requested extension before the save layer builds the path, so `X.json.json` has to be placed there by something else.
- No branch makes a false **success** claim. The reply still reports `UNDETERMINED` and still directs the caller to read the saved file itself, which is the action that settles it either way.

### Direction

Normalize the two sides with the operation each actually needs — strip directory + extension from the **requested** name, and only the directory from the frontend-derived `filename` — rather than running both through the same helper. Worth a fixture built from a real `X.json.json`, since the current fixtures cannot express the difference.

Context: this is the third finding in the #1455 chain, after the `modified:false` inference and the raw-vs-normalized comparison. The first two are fixed; this is the residue.

artokun:
Fixes #1459.

> **This defect is live in a release.** `nameKey()` did not exist on `main` when #1459 was
> filed — it arrived with #1458, which merged at 00:53Z and shipped in **v0.15.16**
> (`01dd1c0`). So this is not a pre-emptive fix against an open PR; it is a fix against
> released code.
>
> <sub>Bookkeeping: this work opened as #1460, stacked on `fix/1455-normalize-name-comparison`
> because `nameKey` had no home on `main` yet. GitHub closed it when that base branch was
> deleted on merge, and a closed PR cannot be retargeted or reopened. The branch was rebased
> onto the new `main` (`git rebase --onto origin/main c7886838`) and re-opened here — same
> three commits, same content.</sub>

## The defect

`nameKey()` normalizes **both** sides of the "is the canvas the save's destination?" comparison
with one helper: drop the directory, then `baseName()` off a trailing `.json` / `.app.json`.

That is right for `requested`. It arrives raw from the caller (`workflow_save({name})` →
`targetName`), and `baseName()` is exactly what the save layer strips with before it builds the
target path.

It is one strip too many for the other side. `filename` is read off
`app.extensionManager.workflow.activeWorkflow` (`observeActiveWorkflowSaveState`,
comfyui-mcp-panel.js:6122), and the frontend has **already** stripped it
(`UserFile` → `getPathDetails` → `getFilenameDetails`). Running it through `baseName()` strips it
a second time — the same double-strip this repo documents as unsafe at
`web/js/lib/workflow-save.js:650`, an invariant that has been in the tree since #226/#231:

> `ComfyUI strips the final ".json" from the on-disk name, so a file at "…/Foo.json.json" reports filename "Foo.json"; baseName() would strip it again to "Foo" and misjudge a Save-As to "Foo" as an in-place save.`

### `getFilenameDetails`, read rather than assumed

Transcribed from the bundle the live ComfyUI actually serves
(`/assets/formatUtil-BIkJATtP.js`, frontend 1.49.6) — one compound case, then a cut at the
**last** dot:

```js
var n=`app.json`, r=`.json`, i=`.app.json`
function getFilenameDetails(e){
  if(e.toLowerCase().endsWith(i) && e.length>i.length)
    return {filename:e.slice(0,-i.length), suffix:n};
  let t=e.lastIndexOf(`.`);
  return t<=0?{filename:e,suffix:null}:{filename:e.slice(0,t),suffix:e.slice(t+1)}}
```

| on disk | reported `filename` |
|---|---|
| `Foo.json` | `Foo` |
| `Foo.app.json` | `Foo` |
| **`Foo.json.json`** | **`Foo.json`** |
| `My.Workflow.json` | `My.Workflow` |

Every one of those is already as bare as the frontend can make it. That third row is the whole
issue.

## The input that separates them

An externally-placed `workflows/Foo.json.json` reports `filename: "Foo.json"` — the exact shape
the repo's own #226 fixture already models (`browser_tests/unit/workflow-save.test.mjs:401`).
Double-stripped it keys as `"Foo"` and matches a requested `"Foo"`, so:

- a genuinely different workflow reads as the save's destination;
- the *"The active workflow is …, not the save's destination"* disclosure is suppressed;
- and the **source's** `modified`/`persisted` flags are printed as though they described the target.

The same collision hits the un-named path, where **both** sides are frontend-derived: a canvas
that moved from `Foo.json.json` (filename `"Foo.json"`) to a real `Foo.json` (filename `"Foo"`)
keys identically, so "which file does the hung write target?" is never asked and the reply falls
through to the in-place case.

## The fix

Two keys, each applied to the side it belongs to — the direction the issue asks for:

- `requestedKey(v)` — directory **and** extension. For the caller's raw string.
- `activeKey(v)` — directory only. For a frontend-derived `filename`, which the frontend has
  already taken one trailing extension off.

`baseName` now has exactly one call site in the module, inside `requestedKey`.

## Verification

**Unit — `npm run test:unit` on the rebased tree: 5770 pass, 0 fail** (i18n, i18n-render, tool-vocabulary and
panel-scope checks included; the run has 1 pre-existing `todo`). The changed file was run 3×
standalone: 25/25 every time.

Six new tests, built from the real `X.json.json` shape rather than a synthetic one, since — as the
issue notes — the existing fixtures cannot express the difference.

**Verified by mutation, not by a green suite.** Reverting only the fix
(`activeKey = (v) => baseName(dropDir(v))`, i.e. the old shared helper) turns **5 of the 6** red:

```
✖ #1459: an externally-placed Foo.json.json is NOT the destination of a save to Foo
✖ #1459: the reply for a double-extension canvas names the REQUESTED destination
✖ #1459: an un-named save that moved between Foo.json.json and Foo.json is undeterminable
✖ #1459: a directory on the frontend name is still dropped, without touching the extension
✖ #1459 SOURCE: the frontend-derived name is never run through baseName
✔ #1459: the #1455 requested-name normalization is unchanged   ← regression guard, green by design
```

That last green one is the point of it: the fix must narrow **only** the frontend side, so every
requested spelling #1455 taught to match (`Foo`, `Foo.json`, `workflows/Foo.json`,
`workflows\Foo.json`, `" Foo "`, `Foo.app.json`) still has to match a canvas `"Foo"`.

**A helper-level test cannot see which helper a call site chose** — swapping `activeKey` back to
`requestedKey` at any one of the three comparison sites is invisible to any assertion that does not
happen to exercise that exact branch. `#1459 SOURCE` therefore asserts on the shipped source: that
`nameKey(` is gone, that `requestedKey` never receives `active`/`prior`, that `activeKey` never
receives `dest`, that all three comparisons key per-side, and that `baseName(` appears exactly once
in non-comment code.

## Gate

**codex was unavailable (account exhausted until 2026-08-19 20:43); gated by an independent
adversarial review instead** — `claude-gate.mjs`, a fresh reviewer given only the diff, per the
authorised substitution. Disclosing that substitution is the condition it was accepted under.

Three rounds, all **SHIP (exit 0)**. Full verdicts in a comment below. The gate ran things rather
than reading them: 7/7 mutations killed (including each of the three call sites swapped
individually), an 810-combination behavioural diff of shipped-vs-reverted (9 disk paths × 9
requested names × 10 prior-actives) in which **98 rows change and all 98 involve a `.json.json`
path** — every one moving `IN-PLACE-FLAGS → DISCLOSE-MOVED`/`UNDETERMINABLE`, zero collateral on
any normal-shape path — and a 1728-combination hostile fuzz (null/NaN/arrays/throwing
`toString`/NUL bytes/5000-char names) with no throws.

Rounds 2 and 3 are **comment-only** commits (`f84e4e0`, `412a8b6`); the executable diff has been
byte-identical since `054b184`. Round 2 caught that a comment I added overstated what
`getFilenameDetails` strips — it checked the shipped frontend instead of believing me, and it was
right. `412a8b6` corrects it, and round 3 re-gated that exact tree. Recording it because in this
repo an unverified claim in a comment gets quoted back later as the repo's own measurement.

## Browser suite — not run, and why

Playwright was **not** run for this change, rather than run and reported as coverage it would not be.

- The changed module has exactly one importer (`web/js/comfyui-mcp-panel.js:602`), and
  `describeWorkflowSaveTimeout`'s string reaches only the bridge dispatch at
  comfyui-mcp-panel.js:22337 — an MCP tool reply. `workflow_save`/`workflow_save_as` are **not**
  slash commands (`runLocalCommand` is wired to `graph_canvas`, `graph_run`, `graph_get_errors`
  only), so this text has no DOM path and the sidebar renders nothing from it.
- The message only fires when a `/userdata` write hangs past the 13 s budget, which the Tier-1
  MockBridge suite has no way to provoke.
- No live ComfyUI serves this branch: the suite drives whatever pack is junctioned into
  `localhost:8188`, and a green run there would have exercised code that predates both #1458 and
  this fix. Repointing the shared instance would have disturbed another agent's run.

### DIFF
--- a/browser_tests/unit/workflow-save-budget.test.mjs
+++ b/browser_tests/unit/workflow-save-budget.test.mjs
@@ -16,6 +16,8 @@
  */
 import test from "node:test";
 import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
 
 import { withTimeout } from "../../web/js/lib/bounded-step.js";
 import {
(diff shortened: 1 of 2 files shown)


artokun:
## Merge gate — substituted review, disclosed

**codex was unavailable (account exhausted until 2026-08-19 20:43); gated by an independent
adversarial review instead** (`.claude/autopilot/claude-gate.mjs` — a fresh reviewer handed only
the diff, which has never seen the implementing reasoning). Disclosure is the condition the
substitution was accepted under, not a courtesy. I did not gate my own work and did not
substitute my own opinion for the gate's.

Three rounds, **all exit 0 = SHIP**. The executable diff has been byte-identical since the first
round; rounds 2 and 3 gate comment-only commits.

### Round 1 — `054b184` (the fix) → SHIP
- 5763 pass / 0 fail. 4 mutations run (case-3 double-strip, subject double-strip, disclosure
  double-strip, full `activeKey` revert) — all 4 red, each caught by at least one *behavioural*
  test, not merely the source-text assertion.
- Ground-truth model of the real pipeline diffed over 112 dest-branch and 64 case-3 states:
  dest branch **38 wrong → 28 wrong**; case 3 **4 fixed, 0 regressed**.
- Raised one non-blocking item: my comment's `.app.json` claim.

### Round 2 — `f84e4e0` (comment) → SHIP
- **7/7 mutations killed**, including each of the three call sites swapped *individually* — the
  case the source-level test exists for, and which every value-level assertion is blind to.
- **810-combination behavioural diff** (9 disk paths × 9 requested names × 10 prior-actives):
  712 identical, 98 changed, and **100% of the 98 involve a `.json.json` path**. Every transition
  is `IN-PLACE-FLAGS → DISCLOSE-MOVED` or `→ UNDETERMINABLE`. Zero collateral on any normal-shape
  path.
- **1728-combination hostile fuzz** (null/NaN/arrays/objects/throwing `toString`/NUL bytes/
  5000-char names/bare `.json`/`///`): no throws, every reply well-formed.
- Verified the input-shape assumption the whole fix rests on by extracting
  `UserFile`/`getPathDetails`/`getFilenameDetails` from the installed frontend rather than
  trusting the PR's description of them.
- **Caught that a comment I had added was wrong** — it checked `getFilenameDetails` against the
  shipped bundle instead of believing me. `getFilenameDetails` has an explicit compound
  `.app.json` branch, so `Foo.app.json` reports `"Foo"`, not `"Foo.app"`.

### Round 3 — `412a8b6` (the correction) → SHIP
- Re-verified the corrected comment against
  `comfyui_frontend_package-1.49.6/static/assets/formatUtil-BIkJATtP.js`: the cited version, all
  four worked examples, and the documented `Foo.app` residual ("identical before and after") hold
  — it confirmed that residual appears in **both** the old and new wrong-sets.
- Independent ground-truth differential (truth = `targetPath === canvasPath`) over disk × mode ×
  request: **old wrong 32 → new wrong 24**, zero reachable new false-mismatches.
- Mutated in **both** directions this time: restoring `activeKey = baseName(dropDir(v))` → 5
  failures (3 behavioural); dropping the strip from `requestedKey` → 5 failures including the
  #1455 regressions. Neither mutation survives.
- Confirmed no duplicate of the old `nameKey` survives anywhere in `web/`.

### On the comment correction

I wrote an unverified claim into a source comment, and the gate caught it. In this repo a number
or mechanism asserted in a comment gets quoted back by a later review as the repo's own
measurement, so `412a8b6` re-derives the statement from the bundle the live ComfyUI actually
serves and records the four worked examples inline. Recording the miss here rather than quietly
fixing it.

Issue #384, opened 2026-07-31
Pasting an image into the chat composer also leaks to ComfyUI's canvas paste handler, creating a duplicate LoadImage node


artokun:
### What happened / root cause
In the chat composer's paste handler (`web/js/comfyui-mcp-panel.js`, the `input.addEventListener("paste", ...)` block, ~L15603), when a pasted clipboard item is a file (e.g. a screenshot from Win+Shift+S), the handler calls `ev.preventDefault()` and routes the file through `handleFile()` to attach it to the chat message — but never calls `ev.stopPropagation()`. The paste event therefore continues bubbling up to `document`, where ComfyUI core's own global paste-to-canvas listener also inspects the clipboard and creates a new `LoadImage` node on the graph.

Net effect: every single screenshot/image a user pastes into the chat to show the agent *also* silently drops a duplicate, unwanted `LoadImage` node onto their live canvas. Over a long session this litters the graph with stray nodes (observed 4 duplicate `LoadImage` nodes accumulating from repeated screenshot pastes in one session) and confused the user, who couldn't figure out why images kept appearing on their canvas.

### Steps to reproduce
1. Open the panel chat, click into the composer input.
2. Take an OS-level screenshot that copies to the clipboard (e.g. Win+Shift+S) or copy any image file.
3. Paste (Ctrl+V) into the chat composer to attach it to a message.
4. Observe: the image attaches correctly to the chat message, AND a new `LoadImage` node simultaneously appears on the currently-open ComfyUI graph — even though the user never intended to touch the canvas.

### Exact error (scrubbed)
No thrown error — silent unwanted side effect (duplicate canvas mutation the user didn't request).

### Fix
applied locally: yes
upstream-only: no

Added `ev.stopPropagation()` right after the existing `ev.preventDefault()` in the file-found branch of the paste handler, so once the composer has claimed and handled a pasted file, the event no longer reaches ComfyUI's own document-level paste listener:

```diff
     const fileItem = Array.from(dt.items || []).find((it) => it.kind === "file");
     if (fileItem) {
       const file = fileItem.getAsFile();
       if (file) {
         ev.preventDefault();
+        // Without this, the paste event bubbles to document and ComfyUI's own
+        // global paste handler also fires, dropping a duplicate LoadImage node
+        // onto the canvas for every image attached to a chat message.
+        ev.stopPropagation();
         handleFile(file);
         return;
       }
     }
```

[...]

PR #1473, opened 2026-08-20
fix(1467): a clipboard's file item no longer discards the text pasted with it

### LINKED ISSUE #1467: Pasted chat text silently dropped before reaching the agent (2x in one session; attachments render as [object Object])
User pasted prompt text into the sidebar panel chat twice in one session. Neither paste arrived in the agent's turn — the agent received only the surrounding typed question, with no pasted content and no error or placeholder.

Occurrence 1: user said "this is the prompt. lets test it out" — no prompt text in the turn. Agent fell back to reading the prompt already on the canvas node.
Occurrence 2 (same session, ~30 min later, after switching workflows): user said "can you read these prompts? pasted 2 of them. they are continuous and should go up to 6 seconds" — again no text in the turn.

Earlier in the same session, an image the user said they uploaded also did not reach the agent; the user-visible message rendered as the literal string [object Object] in the transcript, which suggests attachment/paste payloads are being stringified rather than serialized into the message content.

Root cause not determined from the agent side — no error surfaced anywhere. Reporting as a silent content-loss defect on the panel chat input path.

### Steps to reproduce
1. Open the sidebar panel chat.
2. Paste a large multi-line block of text (~1-6 KB) into the chat input, together with a short typed question.
3. Send.
4. Agent receives the typed question only; the pasted block is absent.

### Exact error (scrubbed)
No error. Silent. The only visible artifact is the [object Object] rendering seen for the earlier image attachment in the same session.

### Fix
applied locally: no
upstream-only: yes (needs investigation in the panel chat input / message serialization path)

Precise change needed: in the panel's chat submit path, confirm pasted/attached content is serialized into the outgoing message content array rather than coerced with String(...) — the [object Object] rendering is the tell. If large pastes are being dropped by a size guard, surface an explicit error to the user instead of sending the message with the content silently removed.

### Environment
OS: Windows 11 Home 10.0.26200
ComfyUI: 0.27.1 (per install_comfyui action:"environment"; session ENVIRONMENT line reported 0.33.2 — the two disagree, noting both)
ComfyUI FRONTEND: 1.45.20
GPU: NVIDIA GeForce RTX 4090, 24 GB VRAM
comfyui-mcp: 0.52.18
panel: 0.15.13
Workflow open at the time: video_minimax_h3_t2v (also occurred on video_minimax_h3_r2v)

artokun:
Fixes #1467.

## Root cause

The composer's paste listener inspected the clipboard for a **file item first**, and when it found one it called `preventDefault()` and **returned**. That did two things at once: it discarded any `text/plain` that arrived in the same clipboard, and — because of the `preventDefault()` — it also suppressed the browser's own insertion of that text.

This is not an exotic clipboard. Copying rich content on Windows (Word, Outlook, Excel, and many web pages) puts a **synthesized bitmap alongside the text**. So pasting a prompt out of a document attached a picture of it and dropped every character, with no error and no placeholder.

That accounts for both symptoms in the report from one mechanism: the pasted prompt never arrived, and the user saw an unexpected image attachment in the same session.

## The fix

Both halves of a clipboard are now decided together, in `web/js/lib/composer-paste.js`:

- `planComposerPaste({ hasFile, text })` returns what to do with the file and the text as one answer. The only plan that claims nothing is a small text-only paste, which the browser inserts exactly as before.
- When the composer claims the paste (`preventDefault()` + `stopPropagation()` — still required, so ComfyUI core's document-level handler does not drop a duplicate `LoadImage` node onto the graph, #384), the text half is now handled explicitly: `attach` for a large paste, `insert` via `insertAtCaret` for a small one.
- `orphanAttachmentTokens(text, attachments)` catches the other half of the silent loss: a message referencing an attachment token whose content is no longer loaded now appends an explicit system line instead of sending a message that merely *looks* complete. It is computed **before** `resetAttachments()` clears the registry it is measured against.

## Verification

- `browser_tests/unit/composer-paste.test.mjs` — **17/17**, covering the routing on the shipped listener and the orphan-token guard.
- **Mutation-tested.** Restoring the original bug (`if (plan.file) { handleFile(file); return; }`) turns **2** tests red; dropping the caret insertion — the exact silent-drop path — turns **1** red. Baseline restores to 17/17.
- Full suite **5787 pass / 0 fail**; `check-tool-vocabulary` OK; `check-panel-scope` OK; `i18n:check` locales OK; `i18n-render-check` OK.

## One defect found and fixed during review

The change added a `tr()` site whose key was **not** in the committed catalog, so `regenerating English is a ROUND TRIP` failed: extraction found `panel.attachment_content_no_longer_loaded`, `locales/en/main.json` did not have it. Regenerated with `npm run i18n:build` (one key added, 1108 total).

Worth noting how it surfaced: the targeted `composer-paste` file was green on its own, and clean `main` passes i18n 30/30. Only the **full** suite caught it — a new user-facing string is invisible to the test file that motivated it.

### DIFF
--- a/browser_tests/unit/composer-paste.test.mjs
+++ b/browser_tests/unit/composer-paste.test.mjs
@@ -1,0 +1,265 @@
+// #1467 — pasted chat text silently dropped before reaching the agent.
+//
+// Two silent content-loss paths on the composer, tested at two different depths.
+//
+// The paste handler is exercised as SHIPPED: its listener is sliced out of the
+// panel source and evaluated against doubles, so these tests fail if the routing
+// decision is reverted, and they cannot pass against a copy of the handler that
+// no longer exists. The decision itself lives in web/js/lib/composer-paste.js and
+// is tested directly for the cases a DOM cannot reach cheaply.
+//
+// The orphaned-token guard sits inside the 250-line submit handler, which cannot
+// be rebuilt in a synthetic scope; it is pinned at the source level instead —
+// including its ORDER relative to resetAttachments(), because a guard that reads
+// the registry after it is cleared reports every token as orphaned.
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+
+import {
+  PASTE_TEXT_THRESHOLD,
+  PASTE_TEXT_LINE_THRESHOLD,
+  isLargePaste,
+  planComposerPaste,
+  orphanAttachmentTokens,
+} from "../../web/js/lib/composer-paste.js";
+
+const HERE = dirname(fileURLToPath(import.meta.url));
+const PANEL_SRC = readFileSync(join(HERE, "../../web/js/comfyui-mcp-panel.js"), "utf8").replace(/\r\n/g, "\n");
+
+// ---------------------------------------------------------------------------
+// The shipped paste listener, rebuilt against doubles.
+// ---------------------------------------------------------------------------
+
+/** Slice the composer's `paste` listener out of the panel source. */
+function shippedPasteListenerSource() {
+  const open = `input.addEventListener("paste", (ev) => {`;
+  const start = PANEL_SRC.indexOf(open);
+  assert.notEqual(start, -1, "the composer's paste listener is no longer findable — update this harness");
+  const close = "\n  });";
+  const end = PANEL_SRC.indexOf(close, start);
+  assert.notEqual(end, -1, "the paste listener's end is no longer findable — update this harness");
+  return PANEL_SRC.slice(start, end + close.length);
+}
+
+/**
+ * Run the SHIPPED listener over one synthetic clipboard and report what it did.
+ * Every collaborator is a double, so the only real logic under test is the
+ * handler's own routing plus the real planComposerPaste it calls.
+ */
+function runShippedPaste({ file = null, text = "" }) {
+  const calls = { handleFile: [], handlePastedText: [], insertAtCaret: [], prevented: 0, stopped: 0 };
+  let listener = null;
+  const input = {
+    addEventListener(type, fn) {
+      if (type === "paste") listener = fn;
+    },
+  };
+  // eslint-disable-next-line no-new-func -- the point is to run the shipped source
+  const register = new Function(
+    "input",
+    "handleFile",
+    "handlePastedText",
+    "insertAtCaret",
+    "planComposerPaste",
+    shippedPasteListenerSource(),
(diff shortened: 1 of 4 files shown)


artokun:
## Gate: **SHIP**

> Disclosure: codex was unavailable, so this was gated by an independent adversarial reviewer spawned fresh against the diff (`claude-gate.mjs`, exit 0). Not a self-review — the reviewer never saw the author's reasoning.

It ran the taxonomy against executed behaviour rather than reading, including **16 hostile clipboard shapes through the shipped listener**. Highlights worth keeping in the record:

- **`plan.text` is four-valued, not a boolean** — `"none"` (no text exists) is distinct from `"default"` (text exists, browser places it), `"insert"` and `"attach"`. `getData` returning `null` / `undefined` / a number all land on `{file:true, text:"none"}` and still attach the file; `items` undefined still routes big text. No flavour is claimed-then-dropped.
- **`orphanAttachmentTokens` keys on `(label, id)`, not id alone** — mutating it to id-only goes red. All six `attachments.push` sites (`image`/`text`/`video`/`file`/`textfile`/`workflow`) were checked against `ATTACHMENT_TOKEN_LABELS`, so no *live* attachment can be reported dead.
- **The order pair is right** — the registry is read before `resetAttachments()` empties it; mutating the order goes red.
- **10 mutations, 7 killed**, including the exact original shape (early `return` after `handleFile`), dropping `stopPropagation` (#384), never placing short text, and the threshold off-by-one.
- **Nothing that worked stops working**: `prevent=0 stop=0` for a small no-file paste and `prevent=1 stop=1` for file-only are byte-identical to before. The only behaviour delta is text now being *placed* when a file is present.

### Three surviving mutations, all benign — recorded rather than hidden

1. The `!plan.file &&` in the return guard is **redundant**; removing it changes nothing.
2. The shared-`lastIndex` defence is unnecessary — the `while` loop always exhausts and resets it — so that comment and its test message are decorative.
3. The warning's *effect* is pinned only at source level.

None of these is a defect; they are places where a future reader might over-trust a comment.

### One thing the reviewer tried and could not turn into a finding

`getData("text/plain")` throwing. The old code never reached it when a file was present; the new code calls it first, so a throw would skip `handleFile` **and** `preventDefault`. `DataTransfer.getData` does not throw for `text/plain` on a paste event, and no real input could be named that triggers it — so it is not a finding. Noting it because it is the one genuine ordering change in the diff.

Also fixed during review: the new `tr()` site used a key absent from `locales/en/main.json`, failing `regenerating English is a ROUND TRIP`. Regenerated with `npm run i18n:build`. The targeted test file was green on its own and clean `main` passes i18n 30/30 — only the full suite caught it.

PR #1479, opened 2026-08-20
fix(1848): a schema read that timed out is UNKNOWN, not proof that nothing outputs the type

### LINKED ISSUE #1848: panel_add_node refuses any node whose required input is a custom type — guard claims "no installed node outputs <TYPE>" while that loader is installed and was just added

`panel_add_node` refuses to add any node that has a REQUIRED input of a custom (pack-defined) type, with:

> Cannot add "<Class>": N required input types had no widget after 5.0s waiting for node extensions to register.
>   - input "<name>" (declared type "<TYPE>"): no installed node outputs "<TYPE>" and no frontend widget is registered for it.

The "no installed node outputs <TYPE>" premise is provably false in every case I hit. The producing loader is installed, is present in `/object_info`, declares that exact output type, and in two of the three cases I had **just successfully added it to the same graph seconds earlier**.

So the guard's "is there a producer for this type" lookup is not seeing custom output types — it appears to only accept core socket types (IMAGE/MASK/LATENT/MODEL/…) and treats every pack-defined type as unproducible, then falls through to the widget-wait path and times out at 5s.

### Steps to reproduce

On an install that has `comfyui-sam3` and `segment anything` node packs:

1. `panel_add_node(class_type="LoadSAM3Model")` → **succeeds**. Returned node has `outputs: [{name: "sam3_model_config", type: "SAM3_MODEL_CONFIG"}]`.
2. `panel_add_node(class_type="SAM3Grounding")` → **fails** with `input "sam3_model_config" (declared type "SAM3_MODEL_CONFIG"): no installed node outputs "SAM3_MODEL_CONFIG"`.
3. Retry → identical error.
4. `panel_refresh_nodes()` → `{ok:true, refreshed:false, reason:"combo_refresh_failed", detail:"(refreshComboInNodes() did not answer within this refresh's remaining budget)"}`. Retry step 2 → identical error.
5. Same class of failure with a completely different pack:
   - `panel_add_node("SAMModelLoader (segment anything)")` → **succeeds**, outputs `SAM_MODEL`.
   - `panel_add_node("GroundingDinoModelLoader (segment anything)")` → **succeeds**, outputs `GROUNDING_DINO_MODEL`.
   - `panel_add_node("GroundingDinoSAMSegment (segment anything)")` → **fails**: `no installed node outputs "SAM_MODEL"` AND `no installed node outputs "GROUNDING_DINO_MODEL"`.

Independent confirmation that the producers exist (this is what makes the guard's message wrong, not just unlucky timing):

```
curl -s http://127.0.0.1:8188/object_info | python -c "..."
OUTPUTS SAM3: LoadSAM3Model -> ['SAM3_MODEL_CONFIG']
=== SAM3Grounding      required: {"sam3_model_config": ["SAM3_MODEL_CONFIG", ...], "image": ["IMAGE", ...]}
=== SAMModelLoader (segment anything)          outputs: ['SAM_MODEL']
=== GroundingDinoModelLoader (segment anything) outputs: ['GROUNDING_DINO_MODEL']
=== GroundingDinoSAMSegment (segment anything)  required: {"sam_model": ["SAM_MODEL", {}], "grounding_dino_model": ["GROUNDING_DINO_MODEL", {}], "image": ["IMAGE", {}], "prompt": ["STRING", {}], "threshold": ["FLOAT", ...]}
```

All six classes are in `/object_info`. The loaders declare exactly the output types the guard says are unproducible.

### Exact error (scrubbed)

```
Error: Cannot add "SAM3Grounding": 1 required input type had no widget after 5.0s waiting for node extensions to register.
  - input "sam3_model_config" (declared type "SAM3_MODEL_CONFIG"): no installed node outputs "SAM3_MODEL_CONFIG" and no frontend widget is registered for it.
Reload the ComfyUI browser tab so node packs can re-register their frontend widgets, then retry. If it fails again the pack's frontend extension is not loading and retrying alone will not fix it. This is NOT a link datatype being misread: an input the backend proves is a socket (MASK, IMAGE, LATENT, a comma-joined union of them) is added immediately, without any wait.
```

```
Error: Cannot add "GroundingDinoSAMSegment (segment anything)": 2 required input types had no widget after 5.0s waiting for node extensions to register.
  - input "sam_model" (declared type "SAM_MODEL"): no installed node outputs "SAM_MODEL" and no frontend widget is registered for it.
  - input "grounding_dino_model" (declared type "GROUNDING_DINO_MODEL"): no installed node outputs "GROUNDING_DINO_MODEL" and no frontend widget is registered for it.
```

Note the error text itself asserts the opposite of what is happening: it says a socket input the backend proves is a socket is added immediately. `SAM3_MODEL_CONFIG` / `SAM_MODEL` / `GROUNDING_DINO_MODEL` ARE sockets the backend proves — they are just not core types.

### Impact

Blocks building any graph where a custom-node consumer takes a custom-typed input from its own pack's loader. That is the normal shape of most segmentation / detector / pipe / model-config packs (SAM3, segment-anything, Impact pipes, etc.), so the agent cannot author those graphs on the live canvas at all — it can add the loader but never the node that consumes it.

The suggested remedy in the message (reload the tab) does not apply, because the missing thing is not a frontend widget — the input is a genuine socket that should just be created as a socket.

### Fix

I did not patch this — the guard lives in the long-lived orchestrator process that serves the `panel_*` tools, which does not reload from a `panel_reload`, so a local patch could not have been verified from here.

Precise change needed: in the add-node input-classification step, the "is this input a socket?" test must accept ANY type that appears as an output type of some class in `/object_info`, not only the core socket allowlist. Concretely, build the producible-type set once from `/object_info` as the union of every class's `output` entries (plus the core socket types), and treat an input whose declared type is in that set as a socket — create it as a socket immediately and skip the 5s widget wait entirely. The current behaviour looks like the producible-type set is being derived from a hardcoded core list (IMAGE/MASK/LATENT/MODEL/CLIP/VAE/CONDITIONING/…) rather than from `/object_info`.

Secondary, smaller issue observed while trying to work around this: `panel_refresh_nodes()` returned `refreshed:false` with `reason: "combo_refresh_failed"` and remedy "Retry", but retrying returned the same thing. Whatever budget `refreshComboInNodes()` is given appears too small on an install with 381 models / 86 controlnets.

### Workaround used

Abandoned SAM-based per-character body masks entirely and rebuilt the masking with core-only nodes: `SolidMask` + `MaskComposite` + `FeatherMask` + `InvertMask` + `ImageCompositeMasked`. This works but loses text-prompted per-character segmentation and the resulting mask is hardcoded to one composition, so it does not generalise to other panels.

### Environment

- OS: Windows 11 Home 10.0.26200
- ComfyUI: 0.33.1
- ComfyUI frontend: 1.48.7
- Python 3.12.11 / PyTorch 2.11.0+cu128 / CUDA 12.8
- GPU: NVIDIA GeForce RTX 5070 Ti, 15.9 GB VRAM
- comfyui-mcp: 0.52.22
- comfyui-mcp-panel: 0.15.1
- Backend: Claude
- Relevant packs installed: comfyui-sam3 (LoadSAM3Model, SAM3Grounding, SAM3Segmentation, …), segment anything (SAMModelLoader, GroundingDinoModelLoader, GroundingDinoSAMSegment)

artokun:
Fixes artokun/comfyui-mcp#1848.

`panel_add_node` refused every node with a required custom-typed input, reporting **"no installed node outputs `SAM3_MODEL_CONFIG`"** while `LoadSAM3Model` — which outputs exactly that — had been added to the same graph seconds earlier and was sitting in `/object_info`.

## The reporter's diagnosis was wrong, and that matters

They concluded the producible-type set comes from a hardcoded core allowlist. It does not: `registeredSocketTypes()` derives it from `/object_info` outputs, and #822 already stopped it being read off a single-class payload. Both of those shipped long before their build.

The real cause is one step further on. The socket proof may come from a **single-class** `/object_info` (#780's optimisation), so on the refusal path it is widened against the whole schema (#821) — and that widen is **bounded** (#1180/#1192). On a heavy install it returns `null`, which the helper uses for *every* doubtful payload **including a call that never answered in budget**.

The guard then keeps failing closed, which is right. The **message** was not:

```js
const cause = entry.linkProven
  ? `the backend declares "${type}" as a link datatype, …`
  : `no installed node outputs "${type}" and no frontend widget is registered for it`;
```

Two states, two branches — and no third for *"I could not find out."* So `linkProven:false` rendered as a claim about the **whole install**, asserted from having failed to read the install. It then sent the reader to reload the tab, which re-registers frontend widgets and does nothing about an `/object_info` read that ran out of time.

The module states the principle it was violating, two functions up:

> "I did not find out" is not "nothing outputs anything", and the difference matters because the caller REPLACES its proof with whatever comes back.

That was applied to **replacing** the proof, and not to **describing** it.

## The change

A third cause. When a widen was attempted and could not answer, the refusal says the producer question is **UNKNOWN** and recommends a **retry** — the thing that can actually complete the read. A complete proof is untouched: it still asserts absence and still recommends reloading the tab.

## Five existing assertions changed — deliberately, and stated plainly

`add-node-socket-proof-scope.test.mjs` pinned `no installed node outputs "SEEDVR2_VAE"` / `"SEEDVR2_DIT"` for a widen that **provably could not answer** (broken `getNodeDefs`, null/empty/non-object payload, a throw).

That is the same over-claim #821 was reported for. #821 fixed the path where the widen *succeeds*; the failure path kept the original sentence — literally the one from that report, where `SeedVR2VideoUpscaler` was told nothing outputs `SEEDVR2_DIT` while `SeedVR2LoadDiTModel` sat on the canvas.

They are **updated, not loosened**. Each still asserts the guard fails closed and nothing is added, still asserts the already-proven type is never named, and now *additionally* asserts the message does not claim absence and does say UNKNOWN.

## Verification

- Full suite **5832 pass / 0 fail** (5833 total). `check-tool-vocabulary`, `check-panel-scope`, `i18n:check`, `i18n-render-check` all OK.
- **Three mutations, all killed**: dropping the 4th argument at the call site (caught by a source-level assertion, since a call-site argument is invisible to message-level tests), never clearing the flag on the widen's reject branch, and collapsing the third cause back into absence.

## Not fixed here, and worth its own issue

The reporter filed `panel_refresh_nodes` returning `combo_refresh_failed` / *"did not answer within this refresh's remaining budget"* as a secondary, smaller issue. It is the **same root cause** — `/object_info` work not completing on an install with 381 models — and this PR only makes the resulting message honest. Whether those budgets should scale with install size is a separate question I have not touched.

Filed on the orchestrator repo, fixed here: the guard and its message are panel-side.

### DIFF
--- a/browser_tests/unit/add-node-socket-proof-scope.test.mjs
+++ b/browser_tests/unit/add-node-socket-proof-scope.test.mjs
@@ -570,7 +570,8 @@ for (const [label, bad] of [
       "a broken whole-schema read is not evidence that nothing produces the type",
     );
     assert.match(err.message, /whether any installed node outputs it is UNKNOWN/);
-    assert.match(err.message, /RETRY this add/, "the remedy follows the real cause");
+    assert.match(err.message, /ALSO worth a RETRY/, "the unresolved producer question is surfaced");
+    assert.match(err.message, /Reload the ComfyUI browser tab/, "and the proven cause keeps its remedy");
     assert.doesNotMatch(
       err.message,
       /SEEDVR2_DIT/,
--- a/browser_tests/unit/add-node-socket-proof-scope.test.mjs
+++ b/browser_tests/unit/add-node-socket-proof-scope.test.mjs
@@ -560,7 +560,17 @@ for (const [label, bad] of [
       (e) => e,
     );
     assert.ok(err, "still fails closed — an unproven sibling type is not waived");
-    assert.match(err.message, /no installed node outputs "SEEDVR2_VAE"/);
+    // #1848 — the widen is what would have answered "does anything output SEEDVR2_VAE?",
+    // and in this fixture it could not. Refusing is right; asserting ABSENCE is not, and
+    // that over-claim is the same one #821 was filed about, left behind on the failure
+    // path after #821 fixed the success path.
+    assert.doesNotMatch(
+      err.message,
+      /no installed node outputs "SEEDVR2_VAE"/,
+      "a broken whole-schema read is not evidence that nothing produces the type",
+    );
+    assert.match(err.message, /whether any installed node outputs it is UNKNOWN/);
+    assert.match(err.message, /RETRY this add/, "the remedy follows the real cause");
     assert.doesNotMatch(
       err.message,
       /SEEDVR2_DIT/,
@@ -612,10 +622,18 @@ test("#821: a failed widen leaves the guard exactly as it fails closed today", a
     },
   });
 
-  await assert.rejects(
-    () => graph_add_node({ class_type: "SeedVR2VideoUpscaler" }),
-    /no installed node outputs "SEEDVR2_DIT"/,
+  // #1848 — the guard's behaviour is unchanged (still refused, nothing added); what the
+  // refusal SAYS is not. getNodeDefs throws here, so the whole-schema proof never
+  // happened, and this is the exact sentence #821 was reported for: SeedVR2VideoUpscaler
+  // told "no installed node outputs SEEDVR2_DIT" while SeedVR2LoadDiTModel — which
+  // outputs precisely that — sat on the canvas.
+  const err = await graph_add_node({ class_type: "SeedVR2VideoUpscaler" }).then(
+    () => null,
+    (e) => e,
   );
+  assert.ok(err, "still fails closed on a fetch that did not happen");
+  assert.doesNotMatch(err.message, /no installed node outputs "SEEDVR2_DIT"/);
+  assert.match(err.message, /UNKNOWN/);
   assert.equal(comfy.graph._nodes.length, 0);
 });
 
--- a/browser_tests/unit/node-widget-materialization.test.mjs
+++ b/browser_tests/unit/node-widget-materialization.test.mjs
@@ -1020,8 +1020,11 @@ test("#695: graph_add_node consumes the report and message, and names the class"
     src,
     // #1180 — monotonicNow, not Date.now. `startedAt` is read from the monotonic clock, so
     // subtracting a wall-clock reading from it reports an elapsed wait of roughly 1.7e12 ms.
-    /unavailableRequiredWidgetMessage\(unavailable, classType, monotonicNow\(\) - startedAt\)/,
-    "the refusal is built from the report, with the elapsed wait — measured on ONE clock",
+    // #1848 — the fourth argument is the fix: without it the refusal cannot tell
+    // "checked, nothing produces it" from "could not check". Deleting it here is
+    // invisible to every message-level test above, so it is pinned at the source.
+    /unavailableRequiredWidgetMessage\(unavailable, classType, monotonicNow\(\) - startedAt, schemaProofComplete\)/,
+    "the refusal is built from the report, with the elapsed wait — measured on ONE clock — and whether the schema proof completed",
   );
   assert.match(
     src,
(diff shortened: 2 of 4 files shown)


artokun:
## Gate: **SHIP** (round 3) — after two NO-SHIPs that were both right

> Disclosure: codex was unavailable; gated by an independent adversarial reviewer spawned fresh against the diff (`claude-gate.mjs`). Not a self-review.

Both earlier rounds caught me reintroducing the exact defect this PR exists to remove — a refusal naming a remedy that cannot work (#695/#700). Recording them, because the final diff only makes sense with them:

**Round 1 — NO-SHIP.** The *cause* is decided per entry (`entry.linkProven`), and I switched the *remedy* on a whole-report flag. A report whose entries were all link-proven got "RETRY / reloading does NOT help" — deleting the one advice that fits them. `linkProven` comes from `SAFE_SOCKET_TYPES` / core 3D types / the class's own outputs, none of which a widen can change, so the widen's failure is irrelevant to those entries. My own test passed because it asserted the *cause* line and never the *remedy*.

**Round 2 — NO-SHIP.** An entry only reaches this report because `comfyApp.widgets` had no constructor after the full poll, so **"no frontend widget is registered" is the one thing here that is proven**. My retry text denied it outright ("the missing thing is the schema answer, not a frontend widget") and withheld the reload remedy. Two live states collapse under `!schemaProofComplete`: (a) a sibling outputs the type, (b) nothing outputs it *and* the pack's frontend extension failed to register. I named only (a) and denied (b)'s fix — and since the widen's bound is fixed (`CUSTOM_WIDGET_REGISTRATION_TIMEOUT_MS / 2`), a large enough `/object_info` re-fails identically forever, so those users were told to retry indefinitely while the working remedy was withheld.

The remedy is now **additive**: reload always ships (it fits every entry), retry stacks on top when a producer question went unanswered, and nothing denies the proven cause.

### Round 3 verification

- **Five mutations, all killed**: removing `schemaProofComplete = false` (6 fail), dropping `.some()` from `schemaUnknown` (1), making the remedy substitutive again (6), defaulting the parameter to `false` (3), pinning the per-entry cause back to the absence claim (6).
- Reachability confirmed **by execution**, not grep — a real `graph_add_node` run emits the message with the new argument.
- The reviewer probed whether `schemaProofComplete` could read `true` while the schema went unread: the only other producer is `freshDefsAreSingleClass === false`, and both a timed-out and an empty `getNodeDefs` refuse *earlier* (#458 authority check / "Unknown node type"), so they never reach this message.
- Full suite 5833 / 0.

### One thing it raised below its own bar, which I checked anyway

The remedy's template literal spanned a physical line break in a CRLF file, so the source held a real `\r\n` inside the backticks. It emitted correctly — template literals normalise CR/CRLF to LF per spec, and I measured **0 carriage returns** in the message — but behaviour resting on that normalisation is a footgun in a repo where line-ending handling is a routine part of patching. Rewritten as an explicit `\n`; separator re-verified as a single LF.

artokun:
**Correction to the last paragraph of my previous comment.**

I wrote that the remedy separator was "rewritten as an explicit `\n` escape". It was not. My patch replaced the CRLF with a **real newline**, so the template literal still spans two physical lines — the two-character escape I described never landed. The follow-up commit was empty (`nothing to commit`), because the working tree and HEAD differ only by line-ending normalisation, which git collapses on commit.

The behaviour claims in that paragraph still hold, and were measured rather than assumed: the emitted message contains **0 carriage returns**, the separator is a single LF, and the suite is 5833 / 0. Template literals normalise CR and CRLF to LF per spec, so all three spellings emit the same string — which is why this never surfaced as a test failure, and why it does not warrant another commit.

Flagging it because the sentence was wrong, not because the code is.

(This comment replaces an earlier version of itself whose text was mangled by shell backtick substitution — the same quoting trap the correction is about.)

PR #1487, opened 2026-08-20
fix(1486): adopt a loopback bridge URL the orchestrator advertises, not just a tunnel one

### LINKED ISSUE #1486: Panel v0.15.22 ignores discovered Codex bridge URL and reconnects forever to obsolete port 9180
## Versions

- comfyui-mcp: 0.52.23
- comfyui-mcp-panel: 0.15.22
- ComfyUI: 0.33.1
- Windows 10 / Comfy Desktop

## Actual behavior

Running npx -y comfyui-mcp@0.52.23 connect correctly starts the Codex bridge on ws://127.0.0.1:9181.

The ComfyUI status endpoint returns running=true, port=9181, and bridge_url=ws://127.0.0.1:9181.

Nevertheless, the fully updated panel displays: Reconnecting to ws://127.0.0.1:9180/...

No service is listening on port 9180; the orchestrator is listening on 9181.

## Reproduction

1. Install/update comfyui-mcp-panel to v0.15.22.
2. Fully restart Comfy Desktop and hard-refresh the UI.
3. Run npx -y comfyui-mcp@0.52.23 connect.
4. Open the panel and click Reconnect.
5. Observe that it repeatedly targets port 9180 despite /comfyui_mcp_panel/status advertising port 9181.

This was reproduced after terminating every Comfy Desktop process, deleting Chromium caches, verifying exactly one panel installation, and verifying both the disk bundle and live HTTP-served bundle report v0.15.22.

## Root-cause evidence

The production v0.15.22 panel bundle contains: const DEFAULT_BRIDGE_URL = "ws://127.0.0.1:9180";

It then maps Codex to that fallback: codex: DEFAULT_BRIDGE_URL

The panel appears to fail to adopt the runtime-discovered bridge_url returned by /comfyui_mcp_panel/status.

## Expected behavior

The panel should connect to the runtime-discovered bridge URL, ws://127.0.0.1:9181, instead of repeatedly dialing its compiled fallback.

## Impact

A clean current installation cannot connect even though the launcher and status endpoint are healthy. The UI says the companion launcher is ready while continuously retrying a port where nothing is listening.

artokun:
Fixes #1486.

A clean, fully-current install could not connect: the panel dialled `ws://127.0.0.1:9180` forever while `/comfyui_mcp_panel/status` advertised `ws://127.0.0.1:9181`.

## The reporter's root-cause read is wrong, and the difference is the fix

They concluded the compiled `DEFAULT_BRIDGE_URL` is stale. It is not — **both sides default to 9180**: the orchestrator uses `DEFAULT_PANEL_BRIDGE_PORT = 9180` and `Number(process.env.COMFYUI_MCP_BRIDGE_PORT) || 9180`. Changing the panel's constant would simply break everyone on the default.

The real gap is **adoption**. An advertised `bridge_url` had exactly two readers, and both are **POST responses**:

- the panel's own launcher start, and
- the auto-reclaim.

An orchestrator started **externally** — `npx comfyui-mcp connect` in a terminal, which is the reproduction — POSTs neither, so neither fires. The only remaining reader was `reclaimAdvertisedBridgeUrl`, which opened with an early return on any non-`https:` page and then accepted `wss://` only, because it exists for the cloudflared tunnel.

On a plain local page that is every path exhausted. Nothing consulted status, so a bridge on any port other than 9180 was simply unreachable, and the retry loop had nowhere else to go — which is exactly the "reconnecting forever to a port with nothing on it" in the report.

## The change

`reclaimAdvertisedBridgeUrl` now reads the advertisement on non-`https:` pages too and routes both cases through one decision helper (`web/js/lib/advertised-bridge-url.js`).

- **The https/tunnel path is untouched**: the secure `wss://` URL keeps its precedence, and a plain loopback advertisement is never substituted for it. With no secure URL yet, an https page adopts nothing rather than downgrading.
- **Loopback only, deliberately.** An advertisement is a hint from a local endpoint, not an instruction; adopting an arbitrary host would let whatever answers `/status` redirect this tab's agent traffic. `ws://127.0.0.1`, `localhost` and `[::1]` are accepted; `ws://192.168.x.x`, a remote host, and a lookalike like `ws://127.0.0.1.example.com` are not.
- **The manual-override guard is unchanged and still runs first** — a user-typed Advanced Bridge URL is never clobbered. The wiring test asserts that ordering, since moving the https gate is exactly the kind of edit that could reorder it.
- An advertisement equal to what is already dialled returns null, so polling never churns the socket for a no-op.

## Verification

- New suite **6/6**; full suite **5867 pass / 0 fail**. `check-tool-vocabulary`, `check-panel-scope`, `i18n:check` all OK.
- **Three mutations, all killed**: restoring the `if (location.protocol !== "https:") return;` early return (the original bug), widening acceptance from loopback to any `ws://` host, and dropping the already-dialled short-circuit.
- The wiring test slices the shipped function out of the panel source and asserts the early return has not come back, that status is read, and that the override check still precedes adoption — a decision helper that is never called would fix nothing.

## Not covered

Why the reporter's orchestrator landed on 9181 rather than 9180 (an explicit `COMFYUI_MCP_BRIDGE_PORT`, or the fallback when an older instance holds the port) is a separate question. This makes the panel follow wherever it lands, which is the behaviour the status endpoint already promised.

### DIFF
--- a/browser_tests/unit/advertised-bridge-url.test.mjs
+++ b/browser_tests/unit/advertised-bridge-url.test.mjs
@@ -1,14 +1,19 @@
 /**
- * panel#1486 — a clean install could not connect when the orchestrator bound a port
- * other than 9180.
+ * panel#1486 — a clean install could not connect: the panel dialled ws://127.0.0.1:9180
+ * forever while /comfyui_mcp_panel/status named ws://127.0.0.1:9181.
  *
- * Both sides default to 9180, so the panel's compiled fallback was never stale. The gap
- * was adoption: the orchestrator advertises the port it ACTUALLY bound via
- * `/comfyui_mcp_panel/status`, and the only two readers of an advertised `bridge_url`
- * were POST responses (the panel's own launcher start, and the auto-reclaim). An
- * orchestrator started externally — `npx comfyui-mcp connect` in a terminal — sends
- * neither, and the remaining reader was `https:`-gated and `wss://`-only for the tunnel.
- * So the tab dialled `ws://127.0.0.1:9180` forever while status said 9181.
+ * WHAT STATUS ACTUALLY SAYS, because the obvious reading is wrong and the first version
+ * of this fix was built on it. `bridge_url` is not a report of where the orchestrator
+ * bound — it is `ws://{_BRIDGE_HOST}:{_BRIDGE_PORT}` (`__init__.py:928`) where
+ * `_BRIDGE_PORT` is an import-time constant of the ComfyUI process (`__init__.py:64`)
+ * with one writer and no `global` declaration. It names the port THIS ComfyUI was
+ * configured to probe. An orchestrator that finds 9180 held and binds 9181 is invisible
+ * to it, and nothing in the panel can discover that — see the last test.
+ *
+ * The bug this DOES fix: ComfyUI configured for a non-default port, the orchestrator on
+ * that same port, and a panel with no reader for a plain `ws://` advertisement at all —
+ * adoption existed only on two POST responses (launcher start, auto-reclaim), which an
+ * externally started orchestrator never sends, plus a tunnel reader gated to https+wss.
  */
 import test from "node:test";
 import assert from "node:assert/strict";
@@ -28,11 +33,10 @@ test("#1486: a loopback ws:// advertisement is adoptable", () => {
     "ws://127.0.0.1:9181/",
     "ws://localhost:9181",
     "ws://[::1]:9181",
-    "ws://127.0.0.1", // no port — still loopback
+    "ws://127.0.0.1",
   ]) {
     assert.equal(acceptableLoopbackBridgeUrl(url), url, url);
   }
-  // Surrounding whitespace is not a different endpoint.
   assert.equal(acceptableLoopbackBridgeUrl("  ws://127.0.0.1:9181  "), "ws://127.0.0.1:9181");
 });
 
@@ -42,8 +46,10 @@ test("#1486: a NON-loopback advertisement is never adopted", () => {
   for (const url of [
     "ws://192.168.1.50:9181",
     "ws://evil.example.com:9181",
-    "ws://127.0.0.1.example.com:9181", // prefix that only LOOKS like loopback
-    "wss://127.0.0.1:9181", // the tunnel path, which has its own reader and token handling
+    "ws://127.0.0.1.example.com:9181",
+    "ws://localhost@evil.com",
+    "ws://localhost:9180@evil.com",
+    "wss://127.0.0.1:9181",
     "http://127.0.0.1:9181",
     "",
     "   ",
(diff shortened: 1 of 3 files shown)


artokun:
## Gate: **SHIP** (round 2) — after a NO-SHIP that refuted my premise outright

> Disclosure: codex was unavailable; gated by an independent adversarial reviewer spawned fresh against the diff. Not a self-review.

Round 1 was a NO-SHIP, and it went where I had not: **the Python side**. It is worth recording in full, because the first version of this PR was built on a false premise and would have shipped a regression.

**`/comfyui_mcp_panel/status` does not report where the orchestrator bound.** `bridge_url` is `ws://{_BRIDGE_HOST}:{_BRIDGE_PORT}` (`__init__.py:928`), and `_BRIDGE_PORT` is an **import-time constant of the ComfyUI process** (`__init__.py:64`) — one writer, no `global` declaration anywhere, so no handler can mutate it. It names the port *this ComfyUI was configured to probe*. I verified all of that myself before acting on it.

So the case I claimed to fix — orchestrator finds 9180 held, binds 9181 — is **invisible to status**, which still says 9180. The reviewer ran it end to end: `pickAdvertisedBridgeUrl` returns null, zero `setUrl` calls, tab still dials 9180. My module header, commit message and test docstring all asserted otherwise. Those claims are gone, and a test now pins the limitation instead of papering over it.

**And the fix introduced a wedge.** With `COMFYUI_MCP_BRIDGE_PORT=9181` in ComfyUI's environment and the orchestrator actually on 9180, a tab correctly on 9180 was moved to 9181 — where nothing listens — and the next tick saw the advertisement equal to the current URL, returned null, and never came back. Before this path existed the http case was a no-op and the bare WS retry recovered on its own.

`running` is the discriminator and it rides in the same payload — it reports whether this ComfyUI could reach an orchestrator at that port. Adoption now requires exactly `true`; absent, `false`, or non-boolean means uncorroborated, and the tab keeps what it has.

### Round 2 verification

- **7 mutations, all red**: reinstating the exact `if (location.protocol !== "https:") return;` bug, unwiring the helper, hardcoding `statusRunning: true`, `=== true` → truthy, dropping the loopback regex, dropping the churn guard, and inverting https precedence.
- Reachability traced by execution: `scheduleReconnect` → `onBridgeClosed` → `tryAutoRespawn` → `externalOrchestratorMode()` (hardwired `true`) → reclaim.
- The dominant path — default ports everywhere — is a **measured no-op**: `__init__.py`'s formatted URL is byte-identical to `DEFAULT_BRIDGE_URL`, so 3 ticks produce 0 `setUrl` calls.
- Full suite 5868 / 0.

### The one sub-threshold note, now written into the code

`running` is ComfyUI probing **its own** loopback; the adopted URL is dialled in the **browser's**. Where those namespaces differ (WSL, LAN-served ComfyUI, SSH tunnel), a true `running` proves a listener exists server-side, not client-side. It needs a non-default bridge port *and* split namespaces to reach, only affects a tab already failing to reconnect, and `persist: false` means any Connect or reload resets via `configuredBridgeUrlFor` — so it degrades nothing that was working. The gate called it "worth a comment someday"; that is how a known limit becomes a rediscovered bug, so it is now a comment where the decision lives.

### One unrelated red run

A full-suite run while the gate was executing failed `#671 verifyInstalled … fast-draining install` — the documented wall-clock flake under load. Re-run alone: **125/125**. Unrelated to this diff.

PR #1502, opened 2026-08-20
fix(608): the node-def refresh asks the raw /object_info route when the frontend client does not answer

artokun:
Fixes the recurrence on artokun/comfyui-mcp#608.

## What was reported

On comfyui-mcp 0.52.26 / ComfyUI 0.33.0 / frontend 1.49.6, after installing custom nodes and restarting, `panel_refresh_nodes` returned **repeatedly**:

```
refreshed:false  reason:object_info_fetch_failed
detail: api.getNodeDefs() did not answer within this refresh's remaining budget
```

…while the server was healthy and `create_workflow(action:"node_info", refresh:true)` — a **different transport** — returned the requested classes at the same moment.

(The ORIGINAL Aug-1 report asked for a `panel_refresh_nodes` tool because none existed. That shipped long ago and is not re-implemented here. This is the recurrence comment only.)

## The defect

`lib/object-info-oracle.js` exists precisely because a frontend client route can fail, or never settle, while the raw HTTP route answers — #982, bounded in #1161. Its own header states the case: a client route that "does not answer IN TIME" is treated as one that answered nothing, and "now has the raw route consulted over it."

- `graph_get_object_info` consults that oracle: `fetchWholeObjectInfo({ getNodeDefs, fetchApi })`.
- `registerComfyNodeDefs` — the fetch phase behind `refresh_nodes`, reconnect refreshes and install/download refreshes — had **one** transport and no way to ask again. A client route that stalls ends the command.

So the tool whose whole job is making a freshly installed pack selectable refused a question the panel could already answer by another route it has wired.

## The change

`web/js/comfyui-mcp-panel.js` — when the client route has **definitively failed**, the same whole-schema oracle is asked over the raw HTTP route, with what is left of the **run** deadline.

**The second route takes nothing from the first.** The client route's grant is byte-for-byte what it was before this issue — `boundedGetNodeDefs(nodeDefsBudgetLeft(runDeadline, NODE_DEFS_FETCH_SHARE))`, the whole fetch share — and the retry loop around it is unchanged but for indentation. So no install that refreshed before can stop refreshing. See the gate section: taking the reserve out of that grant is exactly the mistake this PR shipped in round 1 and had caught.

The oracle is asked with `getNodeDefs: null`. The retry loop above **is** the client route — with this path's own #954 schedule, its bound and its error attribution — so re-asking it would spend the budget on the transport that just failed.

**An empty `{}` from the client is an ANSWER and is not routed around.** This mirrors the oracle rather than `objectInfoLooksTransient`, which the two modules genuinely disagree about. The payload here also feeds the #458 ever-seen trust root and the #1223 whole-schema snapshot, so overruling a deliberate deny-all with a broader schema is the one direction a fallback must never move. Only a route that threw, or returned null/undefined/a non-object, leaves the question open.

**No stale-cache hazard.** `api.fetchApi` defaults to `cache: "no-cache"` (read from the shipped client, comfyui-frontend-package 1.48.7 on this rig: `fetchWithUnifiedRemint(this.apiURL(e), {cache:"no-cache", ...t, headers:n}, !1)`), where `getNodeDefs` overrides it to `no-store`. Both revalidate with the server, so this cannot answer a *refresh* out of the HTTP cache with the pre-install schema — which would be worse than failing.

**What it costs.** When the fallback is consulted it spends time the combo phase would otherwise have had. That is the right trade and not a close one: a missing payload is a hard `refreshed:false`, while an abandoned combo call is a *disclosure* — #1193's reapply sweep has already rebuilt every live combo from this run's payload, so the verdict is `refreshed:true` with `combo_refresh_confirmed:false`. And it is only ever spent in the case that used to end the command outright. The run's total waiting is still bounded by the one deadline, so #1180's composition (two serialized runs inside the bridge's read default) is untouched.

**Not raced.** Issuing both routes together would avoid the wait entirely, and is declined on this path's own evidence: an abandoned request is not cancelled, so a race puts two whole-document downloads (5.4 MB measured) on the same link and makes the slow case slower — the exact contention `shouldRetry` was added to stop. Sequential-after-a-definite-failure spends nothing on the healthy path, which is every path that works today.

`web/js/lib/node-def-refresh.js` — the message half. When both routes fail the client route's error is rethrown **unchanged**, so the verdict's reason, detail and phase attribution are exactly what they were. What is added is a record of what each route did, appended to the two fetch-phase remedies through the shared `objectInfoOracleFailureNote`. Before this, *"check that the ComfyUI server process is still running"* stood on the evidence of one route out of two — and the reported install had a healthy server.

## Ruled out, explicitly

- **A bigger budget.** A bound can only choose *how* to fail against a route that never settles (#1178 spent three attempts establishing that), the run budget is sized against the composition two serialized runs make inside the bridge's read default, and the reporter's own corroboration is a transport that answered promptly while this one did not answer at all.
- **A cold `/object_info` being legitimately slower than any bound.** Excluded by the same evidence: a document the server cannot produce in time is slow for *every* reader, and another reader had it. The failure was also repeated, where a cold-start cost is paid once.
- **Message honesty alone.** The message *was* partly at fault and that half is fixed here. It is not sufficient on its own: a working route existed and the tool did not ask it.

## Deliberately not done

A frontend with no `api.getNodeDefs` **at all** still reports `object_info_unavailable` — the outer guard is unchanged. Nothing has reported that, the reported install has the client route and it is the one that fails, and widening the entry condition would change what such a build is told on evidence nobody has.

## Gate — round 1 was NO-SHIP, and it was right

codex was unavailable (account exhausted until 2026-08-19); gated by an independent adversarial review instead, per the standing substitution. Disclosing that is the condition the substitution was accepted under.

Round 1 introduced `NODE_DEFS_CLIENT_ROUTE_SHARE = 0.5`, halving the client route's grant (6000 → 3000 ms) and calling the remainder the second route's reserve. The gate executed the shipping `registerComfyNodeDefs` against the real oracle and showed that **couples two cases with opposite remedies**: when the slow party is the **backend**, the raw route inherits the same latency and cannot rescue anything with the time the client route was just denied.

| `/object_info` service time | parent `8db21c88` | round 1 |
|---|---|---|
| 2000 ms | `refreshed:true` | `refreshed:true` |
| **3200 ms** | **`refreshed:true`** | **`object_info_fetch_failed`** |
| **3500 ms** | **`refreshed:true`** | **`object_info_fetch_failed`** |
| **5500 ms** | **`refreshed:true`** | **`object_info_fetch_failed`** |
| 6500 ms | `object_info_fetch_failed` | `object_info_fetch_failed` |

A new hard-failure band across (3000 ms, 6000 ms), with `nodeDefsRefreshConfirmed` false behind it — the trust root the combo phase's own note says reopens #610's false *"model still missing"* — and the new *"failed on every transport this panel has"* telling the user the server may be down about a server that is up and answering. #982's defect arriving through the fix for it. 3–6 s is not exotic: this repo has measured a 5.4 MB `/object_info` at **7.5 s over a tunnel**, and a post-install restart on a remote is exactly when it is slowest.

My round-1 accepted-cost note held only while the slowness was *client-transport-specific*, and I did not test the case where it is not. The reserve now comes from the run remainder, not from the client route.

## Verification

- `node --test browser_tests/unit/*.test.mjs` — 5892 pass, 0 fail. `check-tool-vocabulary` OK, `check-panel-scope` OK. CI green.
- **The assertion the gate said was missing now exists**, and it is the one that would have caught round 1: a slow-but-healthy backend that *answers* still refreshes, and the raw route is never consulted (`fetchApiCalls === 0`). Driven on a fake clock so the client route's grant is an exact number — every service time in the broken band, 3200/3500/5500 ms of a 6000 ms bound, is that state.
- **Mutation-tested after committing, both rounds.** Round 2, every one killed:

  | mutation | result |
  |---|---|
  | re-introduce the round-1 split (`FETCH_SHARE * 0.5`) | the slow-backend test + #1180 fail |
  | size the fallback from the spent fetch share | the wiring absence-pin + #1180 fail |
  | give the fallback the 1 ms floor (round 1's real shape) | *"the second route needs a usable budget, got 1ms"* |
  | disable the fallback (`if (false)`) | 7 tests fail |
  | always fall back (`if (true)`) | the empty-`{}` fail-closed test fails |
  | narrow the guard to `!defs` alone | the non-schema test fails |
  | drop `fetchRouteFailures` / either `routesNote` | 1–2 tests fail |

  In round 1 the share mutation **survived its first test**: the assertion was computed from the constant *read from the panel source*, so the expectation moved with the mutation, and the fallback was left the 1 ms floor `nodeDefsBudgetLeft` returns for an exhausted budget — which a synchronous test double still beats and a real 5 MB GET never would. It is pinned in absolute milliseconds now.
- Two pre-existing source-shape assertions moved with this change and were updated rather than loosened.
  - #1180's fetch-phase bound: restored to exactly the parent's regex, plus a new pin that the fallback draws on `nodeDefsBudgetLeft(runDeadline)` and an `assert.doesNotMatch` that it is **not** sized from the fetch share. The absence is pinned because that sizing reads as obviously correct at the call site.
  - #1223's census counts `objectInfoSnapshot.record(` occurrences, and a **comment** I wrote quoting that call read as a seventh call site. The fix was to **reword the comment, not to bump the count** — the count is the check, and that test's own note records that its earlier `whole: true` version had already been fooled by a comment once.
- `manager-install`'s `#671 verifyInstalled` failed once in a full parallel run and passes alone — the known wall-clock flake under load.

**Not run: the Playwright browser suite.** This change is inside the `/object_info` fetch path and renders nothing, and the branch is not the copy any live ComfyUI on this box serves, so an e2e run would not have exercised it. The shipped `registerComfyNodeDefs` is extracted from the panel source and driven end-to-end against the REAL oracle in `node-def-refresh.test.mjs`, which is the coverage this has.

### DIFF
--- a/browser_tests/unit/_panel-constants.mjs
+++ b/browser_tests/unit/_panel-constants.mjs
@@ -121,6 +121,16 @@ export const NODE_DEFS_FETCH_SHARE = (() => {
 })();
 
 /**
+ * #608 — the client route's share of the FETCH phase, read as the panel states it. A
+ * hardcoded 0.5 here would keep agreeing with itself after the panel moved the reserve,
+ * which is the trap this whole file exists for.
+ */
+export const NODE_DEFS_CLIENT_ROUTE_SHARE = readPanelNumber(
+  /const NODE_DEFS_CLIENT_ROUTE_SHARE = ([\d.]+);/,
+  "the client route's share of the fetch phase",
+);
+
+/**
  * The panel's sentinel for "this call did not answer", mirrored so a rebuilt executor
  * compares against the same value its injected `boundedGetNodeDefs` resolves.
  */
--- a/browser_tests/unit/_panel-constants.mjs
+++ b/browser_tests/unit/_panel-constants.mjs
@@ -121,16 +121,6 @@ export const NODE_DEFS_FETCH_SHARE = (() => {
 })();
 
 /**
- * #608 — the client route's share of the FETCH phase, read as the panel states it. A
- * hardcoded 0.5 here would keep agreeing with itself after the panel moved the reserve,
- * which is the trap this whole file exists for.
- */
-export const NODE_DEFS_CLIENT_ROUTE_SHARE = readPanelNumber(
-  /const NODE_DEFS_CLIENT_ROUTE_SHARE = ([\d.]+);/,
-  "the client route's share of the fetch phase",
-);
-
-/**
  * The panel's sentinel for "this call did not answer", mirrored so a rebuilt executor
  * compares against the same value its injected `boundedGetNodeDefs` resolves.
  */
--- a/browser_tests/unit/node-def-refresh.test.mjs
+++ b/browser_tests/unit/node-def-refresh.test.mjs
@@ -1504,6 +1504,18 @@ test("#608: the client route is capped at a SHARE of the fetch phase, so the fal
     armed[0].ms > clientShare * 0.8,
     `…nor be starved to a floor (saw ${armed[0].ms}ms of ${clientShare}ms)`,
   );
+  // THE RESERVE ITSELF, in absolute milliseconds, because the two assertions above are
+  // written against the share READ FROM THE PANEL and therefore move with it: restoring
+  // NODE_DEFS_CLIENT_ROUTE_SHARE to 1 kept both of them green while leaving the fallback a
+  // one-millisecond budget — the floor `nodeDefsBudgetLeft` returns for an exhausted one,
+  // which in this harness a synchronous double still beats and a real 5MB GET never would.
+  // 1,000ms is twice the ~450ms this repo has measured the raw route at, and the oracle
+  // halves whatever it is given between the response and the body.
+  const reserveMs = wholeFetchPhase - armed[0].ms;
+  assert.ok(
+    reserveMs >= 1000,
+    `the second route must be left a usable reserve, saw ${reserveMs}ms of ${wholeFetchPhase}ms`,
+  );
   timers.fireAll();
   await orFail(running, "the run never finished");
 });
--- a/browser_tests/unit/node-def-refresh.test.mjs
+++ b/browser_tests/unit/node-def-refresh.test.mjs
@@ -23,7 +23,6 @@ import { fetchWholeObjectInfo } from "../../web/js/lib/object-info-oracle.js";
 import {
   COMBO_NO_ANSWER,
   COMBO_OK,
-  NODE_DEFS_CLIENT_ROUTE_SHARE,
   NODE_DEFS_FETCH_SHARE,
   NODE_DEFS_FETCH_TIMEOUT_MS,
   NODE_DEFS_NO_ANSWER,
@@ -413,6 +412,8 @@ function buildRegisterComfyNodeDefs({
   now,
   recordObjectInfoTypes = () => ({}),
   reapplyDefsToLiveNodes = () => {},
+  // #608 — defaults to the REAL oracle. Overridden only to OBSERVE what it is handed.
+  fetchWholeObjectInfo: fetchWholeObjectInfoImpl = fetchWholeObjectInfo,
 }) {
   const body = extractFunction("async function registerComfyNodeDefs(");
   const factory = new Function(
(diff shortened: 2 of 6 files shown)


artokun:
This merged at 17:07Z with **two gate verdicts, both NO-SHIP, and no third gate run** — recording that plainly, because the second verdict's finding is now on `main` and shipped in 0.15.24.

To be clear about what I am and am not saying: **the fix is right and should stay.** `registerComfyNodeDefs` genuinely had one transport where `graph_get_object_info` has two, and #608's recurrence is real. Round 1's regression (a slow-but-healthy backend failing across the whole 3000–6000 ms band) was correctly fixed by moving the reserve off the client route's grant. That work is good.

What did not get a verdict is round 2's state: `api.getNodeDefs()` resolving **null** (#1223's NOTHING_RETURNED, not a throw) while `/object_info` does not answer. `clientRouteThrew` is false, so the rethrow at `:1603` does not fire and the run falls through with `defs === null` after the fallback has spent the response grant — 4863 ms → 9008 ms, `object_info_unavailable` → `combo_refresh_failed`, and a remedy that is false in both halves.

Filed as #1507 with the measured table, so it is not lost with this PR.

The reason I am flagging the process and not just the defect: the same *shape* was caught twice here — an accepted-cost note that holds in every state except the one where the cost is actually paid. Round 1's held only while the slowness was client-transport-specific; round 2's holds only when `defs` exists. A third gate was the thing most likely to catch the third instance, and it is the step that was skipped.

No action needed on this PR. #1507 carries the finding.

Issue #1507, opened 2026-08-20
node-def fallback: a null getNodeDefs plus an unanswered /object_info now costs the combo phase and reports a false remedy (from #1502)


artokun:
#1502 merged at 17:07Z and shipped in **0.15.24**. Its gate returned **NO-SHIP twice**, and no third gate ran — so the second verdict's finding is now on `main`. Filing it rather than losing it with the PR.

The fix itself is good and should stay: `registerComfyNodeDefs` genuinely had one transport where `graph_get_object_info` has two, and the raw-route fallback is the right answer to #608's recurrence. This is about one state it did not cover.

## The state

`api.getNodeDefs()` resolves **null** — #1223's documented NOTHING_RETURNED case, not a throw — and `/object_info` does not answer.

`clientRouteThrew` is false, so the rethrow at `comfyui-mcp-panel.js:1603` (`if (clientRouteThrew) throw clientRouteError;`) does not fire. The run falls through to `record` / `register` / `reapply` / `combo` with `defs === null`, and the fallback has already consumed the response grant, so `comboWaitMs` collapses toward its 1 ms floor.

Measured by the gate against the extracted shipping `registerComfyNodeDefs` with the real oracle, `refreshComboInNodes` at this repo's own 4846 ms:

| | parent `8db21c88` | shipped |
|---|---|---|
| elapsed | 4863 ms | **9008 ms** |
| reason | `object_info_unavailable` | `combo_refresh_failed` |
| remedy | truthful | **false in both halves** |

The shipped remedy reads *"Node definitions **were fetched** but this frontend **exposes no registerNodesFromDefs**, so they were NOT registered."* Nothing was fetched, and the frontend does expose `registerNodesFromDefs` — it was skipped only because `defs` was null. `combo_refresh_failed` is also the token the panel's own budget comment (line 1053) records as a previously-shipped bug on a healthy machine.

## Why the cost argument missed it

The comment at 1460-1465 accepts the fallback eating the combo phase because *"an abandoned combo call is a DISCLOSURE (#1193 has already rebuilt the live combos from the payload)"*. That holds only when `comboRebuildCovered(comboRebuild)` is true, which needs `reapplyDefsToLiveNodes` to have run, which needs `defs`.

So in the **one** state where the fallback spends the combo phase and returns nothing, the disclosure is unavailable by construction and it degrades to a hard failure. The note asserts the benign outcome for the only state that cannot produce it.

[...]

PR #1508, opened 2026-08-20
fix(608): a later phase must not speak for a run that obtained no /object_info

artokun:
Fix-forward for a P1 that shipped in #1502 (released as **v0.15.24**). Follow-up on artokun/comfyui-mcp#608.

## Why this exists

#1502 was **merged and released while its independent gate verdict was NO-SHIP**, and the PR was still a draft. The finding does not go away because the branch landed, so this fixes it forward rather than re-litigating the merge.

## The defect, as shipped

`getNodeDefs` resolves **null** (#1223's documented `NOTHING_RETURNED`) and the raw `/object_info` route does not answer either.

There is no error to rethrow — `clientRouteThrew` is false — so the run falls through to `record` / `register` / `reapply` / `combo` with `defs === null`, and the second route has by then spent the budget. `comboWaitMs` collapses toward the 1 ms floor, the combo call is abandoned, that sets `thrown`, and the verdict ladder answers on the **phase** rather than on the missing payload:

| | parent | #1502 (shipped) |
|---|---|---|
| elapsed | 4863 ms | 9008 ms |
| reason | `object_info_unavailable` | `combo_refresh_failed` |
| remedy | truthful | **false in both halves** |
| routes note | n/a | **dropped** |

The remedy claimed definitions *"were fetched"* — none were — and that the frontend *"exposes no `registerNodesFromDefs`"* — it does; registration was skipped only because there was nothing to register. `combo_refresh_failed` is also the token the panel's own budget note records as a previously-shipped bug on a healthy machine. And the `Tried 2 routes:` evidence, which is the entire point of #1502, was dropped on the one path where a user most needs it.

So the second route turned a **correct** answer into a wrong one.

## The fix

One guard in `describeNodeDefRefresh`, keyed on `defsObtained`:

```js
const noPayload = !defsObtained && phase !== "fetch";
```

A later phase cannot speak for a run that never obtained a payload. Phase `fetch` is excluded and keeps `object_info_fetch_failed`: that failure **is** the missing payload, it is more specific, and its `detail` carries what the transport said. The no-payload remedy is lifted to one `const` so this branch and the ordinary no-payload branch below it return the same sentence and cannot drift.

`registrationClause` cannot be true on either side of its own ternary when no payload exists, which is why the old remedy was false whichever way it branched.

## What I got wrong twice, and what is different here

Both earlier cost arguments on this issue were *true everywhere except where the cost actually landed*:

- **Round 1** — "an install whose client route is merely SLOW is overridden by the raw route" held only when the slowness was **client-transport-specific**. When the backend is the slow party, the raw route inherits the same latency. That shipped a hard-failure band across (3000 ms, 6000 ms).
- **Round 2** — "an abandoned combo call is a DISCLOSURE (#1193)" held only when `defs` **exists**, because the disclosure needs `comboRebuildCovered` → `reapplyDefsToLiveNodes` → `defs`. In the one state where the fallback costs the combo phase anything *and* returns nothing, the disclosure is unavailable by construction.

This claim is a different shape: it is keyed on an input the function is **handed** (`defsObtained`), not on an assumption about what else was true, so there is no state it is conditional on. The panel's cost note is rewritten as an **exhaustive two-case split** rather than a general claim — fallback answers / fallback also fails — and the second case now states what it really costs: **time**, bounded by the same run deadline, well inside the command budget, for the same verdict the parent gave plus the routes that were tried.

## Verification

- `node --test browser_tests/unit/*.test.mjs` — 5921 pass, 0 fail. `check-tool-vocabulary` OK, `check-panel-scope` OK.
- The gate's exact input is now an **executed** test through the extracted shipping `registerComfyNodeDefs` with the real oracle: `getNodeDefs → null`, `fetchApi` never answers, `refreshComboInNodes` outlives its starved budget, `registerNodesFromDefs` **present**. It asserts the reason is `object_info_unavailable`, that the remedy contains neither *"were fetched"* nor *"exposes no registerNodesFromDefs"*, and that `Tried 2 routes` survives.
  - The oracle is handed the **same virtual timers** so its own bounds are fired rather than waited out — still the real oracle, only its clock injected, which is the seam it documents.
- Mutation-tested after committing:

  | mutation | result |
  |---|---|
  | remove the guard (`noPayload = false`, i.e. the shipped behaviour) | the executed test + the unit test fail |
  | drop the fetch exclusion (`!defsObtained` alone) | **6+ pre-existing tests** fail — #635, #954, #1180 ×2, #716 — so the exclusion is load-bearing and was already pinned |
  | narrow to the one phase the executed test reaches (`phase === "combo"`) | the per-phase unit test fails |

- The `cache: "no-cache"` claim the reviewer could not verify is now **sourced in the comment**: `comfyui_frontend_package` 1.48.7, `static/assets/api--JY_wdaT.js`, `fetchWithUnifiedRemint(this.apiURL(e), {cache:"no-cache", ...t, headers:n}, !1)`. The filename is content-hashed and the package is not vendored in this repo, so a reader on another machine will have a different asset name or no copy — the comment now says that rather than leaving the claim to be inherited.

**Not run: the Playwright browser suite.** This is a verdict-wording change on the `/object_info` fetch path and renders nothing; the branch is not the copy any live ComfyUI on this box serves. The shipped run is extracted and driven end-to-end instead.

## Gate

codex was unavailable (account exhausted until 2026-08-19); gated by an independent adversarial review instead, per the standing substitution. Disclosing that is the condition the substitution was accepted under. This is round 3 on this issue — the last substantive round before a gate starts finding defects it introduced itself.

### DIFF
--- a/browser_tests/unit/node-def-refresh.test.mjs
+++ b/browser_tests/unit/node-def-refresh.test.mjs
@@ -1682,3 +1682,112 @@ test("#608: the fallback is wired to the WHOLE-schema oracle, after the client r
   const rethrowAt = body.indexOf("if (clientRouteThrew) throw clientRouteError;");
   assert.ok(fallbackAt > -1 && rethrowAt > fallbackAt, "the client route's error is rethrown only after the retry");
 });
+
+// ---------------------------------------------------------------------------
+// #608 — a run that obtained NO payload must say so, whatever a later phase did
+// ---------------------------------------------------------------------------
+
+test("#608: a fallback that ALSO fails leaves the no-payload verdict intact", async () => {
+  // THE SECOND REGRESSION THE GATE CAUGHT, and the one that reached main. With getNodeDefs
+  // resolving null (#1223's NOTHING_RETURNED) and the raw route not answering either,
+  // there is no error to rethrow — so the run falls through to the combo phase with
+  // `defs === null` and a budget the second route has spent. The combo call is abandoned,
+  // which sets `thrown`, and the verdict ladder answered `combo_refresh_failed` with a
+  // remedy false in BOTH halves: definitions "were fetched" when none were, and the
+  // frontend "exposes no registerNodesFromDefs" when it does — registration was skipped
+  // only because there was nothing to register. The parent reported the truthful
+  // `object_info_unavailable`, so the fallback turned a correct answer into a wrong one.
+  const timers = virtualTimers();
+  const { registerComfyNodeDefs } = buildRegisterComfyNodeDefs({
+    appValue: {
+      graph: null,
+      registerNodesFromDefs: async () => {},   // PRESENT — the false half of the old remedy
+      refreshComboInNodes: () => new Promise(() => {}), // outlives its starved budget
+    },
+    apiValue: {
+      getNodeDefs: async () => null,
+      fetchApi: () => new Promise(() => {}),   // /object_info never answers
+    },
+    timers,
+    // The oracle gets the SAME virtual timers, so its own bounds are fired here instead of
+    // waited out on a real clock. Still the real oracle — only its clock is injected, which
+    // is the seam it documents for exactly this.
+    fetchWholeObjectInfo: (opts) => fetchWholeObjectInfo({ ...opts, timers }),
+  });
+  const running = registerComfyNodeDefs(undefined);
+  await drainMicrotasks(300);
+  // Drain and fire until the run settles: the oracle arms its own bounds on these timers
+  // too, and the combo phase arms one after them.
+  for (let i = 0; i < 8; i += 1) {
+    timers.fireAll();
+    await drainMicrotasks(200);
+  }
+  const verdict = await orFail(running, "the run never settled");
+
+  assert.equal(verdict.refreshed, false);
+  assert.equal(
+    verdict.reason,
+    "object_info_unavailable",
+    "no route produced a payload — that is the fact to act on, not what the combo phase did",
+  );
+  assert.doesNotMatch(
+    verdict.remedy,
+    /exposes no registerNodesFromDefs/,
+    "this frontend HAS registerNodesFromDefs — it was skipped for want of a payload",
+  );
+  assert.doesNotMatch(verdict.remedy, /were fetched/, "nothing was fetched");
+  assert.match(verdict.remedy, /Tried 2 routes/, "…and the routes that were tried survive");
+});
+
+test("#608: a payload that DID arrive still lets the combo phase word its own failure", async () => {
+  // The other side of the same guard: keyed on `defsObtained`, so a run that obtained a
+  // payload and then failed at the combo phase is untouched. Without this the fix would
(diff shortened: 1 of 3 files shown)


artokun:
## Gate: SHIP (exit 0) — round 3

codex was unavailable (account exhausted until 2026-08-19); gated by an independent adversarial review instead, per the standing substitution. Disclosing that is the condition the substitution was accepted under.

Merged as 8105db64. Fixes #1507 (closed manually — the ref was added after the merge).

**This was the third round, and the first two are the reason to trust the third.** Both earlier claims were true everywhere except where the cost actually landed:

| round | claim | where it failed |
|---|---|---|
| 1 | "a merely SLOW client route is overridden by the raw route, which measures ~450ms, so the run still succeeds" | held only when the slowness was **client-transport-specific**. When the *backend* is slow, the raw route inherits the same latency — a new hard-failure band across (3000ms, 6000ms) where the parent returned `refreshed:true` |
| 2 | "an abandoned combo call is a DISCLOSURE (#1193)" | held only when `defs` **exists** — the disclosure needs `comboRebuildCovered` → `reapplyDefsToLiveNodes` → `defs`. In the one state where the fallback costs the combo phase anything *and* returns nothing, it is unavailable by construction |

Each was a general-sounding sentence with a hidden precondition, and in both cases the precondition was false in exactly the state the change created. The round-3 claim is a different shape: it is keyed on `defsObtained` — an input the function is **handed** — not on an assumption about surrounding state, which is why there is no state left for it to be conditional on.

The reviewer tested that directly rather than taking it on the prose. It hunted class 3 for an equal-pair/different-fact input and found all three candidates unreachable:

- `phase: "fetch"` with a payload — `defs = null` on every throwing exit from that phase
- `defsObtained` false with registration or reapply having run — both are gated on `if (defs …)`
- `defsObtained` false with combos actually refreshed — `comboFailed` requires `!comboRan`

It also checked the inverse direction on class 1: `fetchRouteFailures` cannot be non-null on a run that *did* get a payload, since it is only assigned where `fallbackDefs` is falsy. And on class 5 it traced both consumers of `reason` (the `refresh_nodes` reply passthrough and add_node's diagnostic) and confirmed neither branches on `combo_refresh_failed`, so broadening the token changes no control flow.

**The frontend citation is now verified rather than merely sourced.** The round-2 reviewer flagged the `cache: "no-cache"` claim as unverifiable on its rig. Round 3 independently confirmed it: `api--JY_wdaT.js` contains `fetchWithUnifiedRemint(this.apiURL(e),{cache:"no-cache",...t,headers:n},!1)` and `getNodeDefs(){…cache:"no-store"…}`. The comment now names that exact file and says the asset name is content-hashed and the package is not vendored here, so the next reader knows why they may not find it.

Mutations, run rather than reasoned: `noPayload = false` → 2 fail; dropping the fetch exclusion → 9 fail; inverted → the file dies. Full suite 5921 pass / 0 fail, worktree restored clean after the mutation runs.

Issue #1444, opened 2026-08-19
panel_edit_node can inflate node sizes and positions to tens of thousands of pixels


artokun:
### What happened / root cause
While cleaning a Qwen workflow, panel mutations had silently inflated multiple node heights and positions. Examples returned by `panel_edit_node`: LoadImage height 49135.84375; UNETLoader height 41465.515625 with y=90717.390625; CLIPLoader height 42174.546875 with y=132312.90625; KSampler height 46585.40625; Qwen conditioning y=43648.78125 and height 7716.984375. This made the visible canvas extremely tall.

### Steps to reproduce
1. Use `panel_edit_node` to move/resize a normal ComfyUI node in an active workflow.
2. Refresh/reload the canvas state.
3. Inspect the next `panel_edit_node` response; node `size` and/or `pos` may be inflated by tens of thousands of pixels.

### Exact error (scrubbed)
No error is returned. The mutation reports success despite corrupted geometry.

### Fix
Applied locally: yes, workflow-only workaround. Upstream-only: yes.
All nodes were reset to explicit compact positions and sizes, and collapsed to prevent corrupted expanded dimensions from stretching the canvas. The panel must preserve LiteGraph node geometry exactly and validate/reject non-sensical geometry before saving it.

### Environment
Windows 11; ComfyUI 0.33.0; ComfyUI Frontend 1.49.6; RTX 5090 32 GB; comfyui-mcp 0.51.31; panel 0.14.24.

artokun:
Claimed by grok-swarm 2026-08-19T20:03:25Z — driving this in worktree `wt-1444` (branch `fix/1444-edit-node-inflate`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp-panel/pull/1446 (worktree C:/Users/Artokun/code/comfyui-mcp-panel/.worktrees/wt-1444).

PR #1509, opened 2026-08-20
fix(1872): a title-only edit is judged against the node's own snapshot, not a size range

artokun:
Fixes the underlying cause of artokun/comfyui-mcp#1872.

## What the reporter hit

`panel_edit_node({ node_id: 35, title: "…" })` on a collapsed `CreateVideo` whose stored
size was `[225, 0]` came back with:

```
Error: size must be two positive numbers within a reasonable canvas range
```

— an argument-validation message for an argument the call never sent. Passing an explicit
`size: [225, 60]` alongside the title worked, which is what pointed at the size path.

## Mechanism (reproduced, not inferred)

Not the arg schema, and not "collapsed". The MCP boundary only validates `size` when it is
supplied (`panel-tools.ts` `validatePanelEditNodeArgs` / `nodeSize()`), and a collapsed node
with a normal height edits fine. The discriminator is a **zero stored height**, and the
refusal is in `graph_edit_node`'s **no-size** branch — the one added for #1444:

```js
} else if (snap.size.length >= 2) {
  const now = pair(node.size);
  if (!now || now[0] !== snap.size[0] || now[1] !== snap.size[1]) writeSize(node, snap.size);
  if (!sizeSane(pair(node.size))) throw new Error("size must be two positive numbers …");
}
```

That branch exists so a move/title/color cannot let `updateArea` rewrite `size` out of a
shared Rectangle, and it does the right thing up to the last line: it restores the captured
size when it drifts. Its **postcondition**, though, is `sizeSane` — an absolute range check
on geometry the caller never touched. Wrong pair, and it fails in both directions:

1. a node whose *own* stored height is `0` never satisfies `s[1] > 0`, so **every**
   title/color/mode/pin edit on it is refused outright — the reported bug;
2. a drift the restore only half-undid **passes** the range whenever the drifted number
   still looks like a node size (`pos.y + h` is under 16384 for any ordinary canvas), so
   the branch silently accepted an inflated node it was written to catch.

Reproduced against the shipped executor extracted from `web/js/comfyui-mcp-panel.js` (the
same extraction the unit suite uses), with a rect-backed node carrying `size [225, 0]`:
the title-only edit threw the reporter's exact string and left `node.title` unchanged,
while the same node at height 80 accepted it.

## The change

Judge the branch by the property it actually owes the caller: **an untouched size ends where
it started.** Refuse only when the size drifted *and* `writeSize` could not put the captured
value back — and say that, instead of blaming arguments the caller did not supply. A
supplied `size` is still range-validated exactly as before, and nothing is normalized: the
node comes out of a title edit still at `[225, 0]`.

One consequence, deliberate: a node whose size was ALREADY corrupt before the call no
longer blocks a title edit. That corruption predates the call, the tool reports it
truthfully in `edited[].after.size` rather than swallowing it, and #1444's actual
complaint — that `panel_edit_node` INFLATED sizes — is fully covered by restoring the
captured value. Policing pre-existing geometry is not a title edit's job.

## Verification

- New tests in `browser_tests/unit/graph-edit-node.test.mjs`, all against the real extracted
  method:
  - title-only edit lands on a zero-height node and leaves `size` at `[225, 0]`
    (`setSize` is a throwing trap in that fixture — a title edit must never reach it);
  - pos+title on a rect-backed zero-height node restores the shared Rectangle instead of
    refusing (this is #1444's own scenario, now passing at height 0);
  - an untouched size that drifted and will **not** go back is refused, and the message is
    asserted *not* to say "two positive numbers".
- **Mutation check:** reverted `web/js/comfyui-mcp-panel.js` to `origin/main` with the tests
  in place → `pass 22 / fail 3`, exactly the three new tests. Restored → 25/25.
- Full panel unit suite: `5925 tests, 1 fail` — the known manager-install #671 wall-clock
  flake; re-run alone: `125/125`, green.
- `i18n-check`, `i18n-render-check`, `check-tool-vocabulary`, `check-panel-scope`: all OK.
  No `tr()` string added, so no new i18n key.
- Playwright e2e not run: this is a bridge-executor postcondition with no rendered surface
  of its own, so the browser suite would not observe the change. Flagging that explicitly
  rather than implying coverage I do not have.

## Gate

adversarial review instead (`.claude/autopilot/claude-gate.mjs`), per the standing
substitution.

### DIFF
--- a/browser_tests/unit/graph-edit-node.test.mjs
+++ b/browser_tests/unit/graph-edit-node.test.mjs
@@ -762,8 +762,20 @@ test("#1872 refuses when an untouched size drifted and will not go back", () => 
         /two positive numbers/,
         "the caller supplied no size — do not blame their arguments",
       );
+      // updateArea drifted the height to pos.y + h = 580; the failed restore then
+      // left the setter's own 600px minimum behind. Report what was OBSERVED, not
+      // the number our own failed write produced.
+      assert.match(error.message, /\[225, 580\]/, "the message must name the observed drift");
+      assert.doesNotMatch(error.message, /600/, "600 is this guard's failed write, not the drift");
+      assert.doesNotMatch(
+        error.message,
+        /rolled back/,
+        "size is exactly what would not go back — do not promise a rollback of it",
+      );
       return true;
     },
   );
-  assert.deepEqual(node.title, "Node 7");
+  assert.equal(node.title, "Node 7");
+  assert.equal(node.pos[0], 100, "every field the guard CAN restore is restored");
+  assert.equal(node.pos[1], 200);
 });
--- a/browser_tests/unit/graph-edit-node.test.mjs
+++ b/browser_tests/unit/graph-edit-node.test.mjs
@@ -676,3 +676,94 @@ test("#1444 setSize/onResize inflation is rolled back, not reported as success",
   assert.deepEqual([node.size[0], node.size[1]], beforeSize);
   assert.deepEqual(events.filter((e) => e === "before" || e === "after"), ["before", "after"]);
 });
+
+/** A node whose stored body height is ZERO — the shape #1872 reported (a collapsed
+ *  CreateVideo serialized as size [225, 0]). Zero height is the node's OWN stored
+ *  geometry, not something a presentation edit asked for. */
+test("#1872 a title-only edit lands on a node whose stored height is zero", () => {
+  const node = makeNode(35, { pos: [1520, 380], size: [225, 0] });
+  node.flags.collapsed = true;
+  // Real LiteGraph clamping: setSize would raise the height. A title edit must
+  // never reach it, so leave it as the trap it is.
+  node.setSize = () => { throw new Error("setSize must not run for a title-only edit"); };
+  const { fn } = setup([node]);
+
+  const result = fn({ node_id: 35, title: "SCENE 1 - Create Video" });
+
+  assert.equal(node.title, "SCENE 1 - Create Video");
+  assert.deepEqual([node.size[0], node.size[1]], [225, 0], "an untouched size must not be normalized");
+  assert.equal(result.edited[0].after.title, "SCENE 1 - Create Video");
+  assert.deepEqual(result.edited[0].after.size, [225, 0]);
+});
+
+test("#1872 a move on a zero-height node restores the shared Rectangle instead of refusing", () => {
+  const node = makeRectBackedNode(35, { pos: [1520, 380], size: [225, 0] });
+  node.flags.collapsed = true;
+  const graph = {
+    beforeChange: () => {},
+    afterChange: () => {},
+    setDirtyCanvas: () => {},
+    getNodeById: (id) => (id === 35 ? node : null),
+  };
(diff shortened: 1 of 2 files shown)


artokun:
Drive-by from the autopilot run on artokun/comfyui-mcp#1875 (a second report of the same mechanism, now closed as a duplicate of #1872 and pointed here). Two things worth having on this PR — no action needed on the diff itself, which I verified independently.

**1. This branch fixes more than the title-only case, and it is worth saying so.** #1875 reported the `pos` half: a collapsed node reports its old coordinates after a move. Same branch, same postcondition. I extracted the shipped `graph_edit_node` from `origin/main` and from this head and ran #1875's literal steps on a Rectangle-backed node at `[2760,1370]` carrying size `[225, 0]`:

```
=== origin/main ===
  move collapsed node   -> THREW: size must be two positive numbers within a reasonable canvas range
  collapsed:false       -> THREW: size must be two positive numbers within a reasonable canvas range
  move EXPANDED node    -> [2680,1380]                      (control, unaffected)

=== this branch ===
  move collapsed node   -> [2680,1380]   live pos [2680, 1380], size still [225, 0]
  collapsed:false       -> false         live flags.collapsed false
  move EXPANDED node    -> [2680,1380]
```

**2. The impact is wider than "a title-only edit is rejected", and the worst case has no test here.** Live on the rig (ComfyUI 0.33.2 / frontend 1.49.6 / mcp 0.52.26), a scratch `Note` collapsed through the tool itself settles to size `[225, 0]` a moment later, after which `pos`, `title`, `collapsed:false` and legacy `panel_move_node` all throw. So on `main` **a node collapsed via `panel_edit_node` can never be un-collapsed by it** — the recovery path is refused by the same guard that broke the edit. That is the case a user actually gets stuck in, and it is the one your three new tests do not cover (title-only, pos+title, and drifted-unrestorable). A fourth asserting that `collapsed:false` lands on a zero-height node would pin the recovery path directly; it dies on `origin/main` for the same reason the other three do.

Both points are additive. Nothing here contradicts the mechanism or the change as written.

artokun:
Gate record, for accuracy — codex was unavailable (account exhausted), so both rounds were the substituted adversarial review (`.claude/autopilot/claude-gate.mjs`).

- **Round 1 — SHIP (exit 0).** Mechanism, snapshot-vs-range pairing, reachability through `case "graph_edit_node"` + the legacy `graph_set_title`/`graph_move_node` wrappers, and the 3-mutation battery all held. Its only finding was the wording of the new refusal, graded **Below P1** — not a P1, and not a merge block.
- **Round 2 — SHIP (exit 0)**, run against `cf8d9882`. It confirmed the round-1 wording fix landed, then found one residual **P2**: the clause "its size left where the restore attempt landed" is false, because the `catch`'s `restore()` writes size again and re-runs `updateArea`.

I measured that second finding rather than taking it on trust — the fixture's height goes 80 → 580 (the observed drift the message names) → 600 (this guard's failed write) → **800** after `restore()`. No line on that path ever reads 800, so the clause described a state it had not observed.

Corrected in #1513 (message and comment only, no behaviour change), with the measurement pinned as an assertion.

PR #1511, opened 2026-08-20
fix(mcp#1871): a run-to-node stops being vetoed by a branch it excluded

artokun:
Fixes the underlying cause of artokun/comfyui-mcp#1871.

## The report

`panel_run(to_node_id: 43)` was meant to run only the ByteDance output branch. ComfyUI refused the whole run:

```
ComfyUI refused to queue the workflow: Node '...' not found. The custom node may not be installed. (Node ID #56)
```

Nodes 56/57 are Topaz nodes on a **different** output branch. They are not upstream of 43 and would never have executed.

## Why it happens

Read out of ComfyUI 0.33.2 (the reporter's version). `server.py` hands `partial_execution_targets` to exactly one place:

```python
valid = await execution.validate_prompt(prompt_id, prompt, partial_execution_targets)
```

and `execution.validate_prompt` opens with a loop over **every** entry of the posted prompt:

```python
for x in prompt:
    if 'class_type' not in prompt[x]:            # -> return (False, missing_node_type…)
    class_ = nodes.NODE_CLASS_MAPPINGS.get(prompt[x]['class_type'], None)
    if class_ is None:                           # -> return (False, missing_node_type…)
    if OUTPUT_NODE and (partial_execution_list is None or x in partial_execution_list):
        outputs.add(x)
```

The partial list is consulted on that last line — three lines **after** the two early returns. So the scope narrows which outputs become execution roots and narrows nothing about which nodes must resolve to an installed class. One unavailable pack anywhere in a workflow refuses every run-to-node in it, which is exactly the case run-to-node exists for.

That is upstream behaviour and the panel cannot change it. What it can change is which nodes it posts.

## The fix

When ComfyUI refuses a scoped run and the node it **names** lies outside the backward closure of the requested targets, the run gets one more post with the excluded nodes omitted. Same targets, same queue-position mark, same `extra_data` (so a saved image still embeds the full workflow) — only `prompt` is narrowed to the closure. Execution is unchanged by construction: `outputs_to_execute` is the partial targets, and the executor reaches only their dependencies.

**A retry, not the default.** Pruning every scoped run was the first design and it has a real cost, measured in the same file:

```python
for cache in self.caches.all:
    await cache.set_prompt(dynamic_prompt, prompt.keys(), is_changed_cache)
    cache.clean_unused()
```

The default classic cache drops every cached output whose key is not in the prompt it was just handed, so always-pruning would evict the *other* branch's cached results and make the user's next full run recompute a branch it had already rendered. Paying that only where the alternative is a total refusal keeps the accepted path byte-identical to today.

**What licenses a second post** — structured fields only, never message text:

- non-2xx status, **and**
- a top-level `error` object, **and**
- no `prompt_id` (`server.py` mints an id and calls `prompt_queue.put` only inside `if valid[0]:`), **and**
- `error.extra_info.node_id` names a node, **and**
- that node is outside the closure computed from the body we posted.

`details: "Node ID '#56'"` and the message are prose and are never parsed. A missing node **inside** the requested branch is not retried — pruning cannot fix it, and the clear answer is the useful one. `prompt_outputs_failed_validation` and `prompt_no_outputs` carry `extra_info: {}` and never reach a retry. The #944 partial-validation reply (an id *alongside* `node_errors`, for a prompt already running) is excluded by the id check.

**What counts as a dependency** is ComfyUI's own predicate, not an approximation. Both consumers read it from the input value itself: `comfy_execution/graph_utils.is_link` (list, length 2, `obj[0]` a str, `obj[1]` a number) and `execution.validate_inputs` (`isinstance(val, list)` → length 2 → `prompt[val[0]]['class_type']`, an unguarded lookup). The closure matches the broader of the two — a top-level length-2 list whose first element names a node in the prompt — so nothing `validate_inputs` would look up is ever pruned away.

**Disclosed, not silent.** The run result carries `excluded_nodes_omitted` plus a note naming the node ComfyUI refused, what was left out, and the fact that a FULL run will still fail until that pack is installed. A run that succeeds because the panel quietly changed the request is exactly the kind of thing the caller must be told.

## Gate

`node .claude/autopilot/codex-gate.mjs` — **SHIP** (exit 0) on round 3. It found two real defects first, both fixed here:

- **r1 P1** — the closure scanned *recursively* into input values on the theory that over-keeping is harmless. It is not: the retry is declined when the node ComfyUI named is inside the closure, so a coordinate pair like `{config:{coordinates:["56",0]}}` on a node in the branch would have handed the reporter their refusal straight back. Fixed by matching ComfyUI's predicate exactly, with a regression test on that shape.
- **r2 P1** — the disclosure asserts the pruned prompt is the one ComfyUI *accepted*. That was true by control flow, but the invariant lived three early-returns away and one of them returns null for a 400 whose `error` and `node_errors` are both empty. The claim now reads the count of posts ComfyUI accepted.

## Verification

`node --test browser_tests/unit/*.test.mjs` — 5948 pass. The one red is the known `#671 verifyInstalled` wall-clock flake (manager-install.test.mjs:821): it passes 125/125 when that file runs alone, and this branch does not touch it. `tsc --noEmit`, `i18n-check`, `i18n-render-check`, `check-tool-vocabulary`, `check-panel-scope` (with the compiler, via a node_modules junction — a fresh worktree does not inherit one) all clean. `npx playwright test --list` compiles: 78 tests in 33 files.

27 new tests: 21 pure (`browser_tests/unit/partial-run-prune.test.mjs`), 5 integration through the real `dispatchScopedRun` orchestration, 1 wiring test asserting on the panel source that the disclosure reaches the reply and is gated on acceptance.

Five mutations, each watched to fail:

| mutation | result |
| --- | --- |
| delete the retry call in `run-scope-guard.js` | 2 integration tests fail |
| delete the disclosure block in `comfyui-mcp-panel.js` | the wiring test fails |
| stop the closure walking upstream | 12 tests fail |
| drop the length-2 requirement from the link predicate | 1 test fails |
| drop the `verified > 0` acceptance gate | the wiring test fails |

**Not run, and why:** the Playwright e2e suite. ComfyUI serves the panel from `custom_nodes/comfyui-mcp-panel`, a junction to the main checkout — not to this worktree — so a run from here would have exercised the main checkout's working tree and said nothing about this change. This change is not UI-rendering: it alters the JSON `graph_run` returns and the request body a scoped run posts, both of which the unit suite drives through the real orchestration.

**Deliberately not done:** when the pruned post is *also* refused, the second rejection is what the caller is shown (the blocker inside their own branch, which is the actionable one) and the first, superseded refusal appears only in `prunedRetry`. Surfacing it in the rejection prose too would be a second sentence to keep true for no decision it changes.

### DIFF
--- a/browser_tests/unit/partial-run-prune.test.mjs
+++ b/browser_tests/unit/partial-run-prune.test.mjs
@@ -1,0 +1,283 @@
+/**
+ * Unit tests for web/js/lib/partial-run-prune.js — run with `node --test`.
+ *
+ * Guards comfyui-mcp#1871: `panel_run(to_node_id: 43)` was refused because nodes 56/57
+ * on an UNRELATED output branch named a class the server does not have. ComfyUI's
+ * `validate_prompt` checks every posted node's class BEFORE it narrows execution to
+ * `partial_execution_targets`, so a branch the caller excluded could veto a branch they
+ * asked for.
+ *
+ * The two properties these tests pin:
+ *   1. the backward closure is COMPLETE — a run must never lose a node its branch needs
+ *      (the direction that turns a working run into "Required input is missing");
+ *   2. the second post is licensed only by STRUCTURED evidence that the first one queued
+ *      nothing, and only when the node ComfyUI named is one the caller's scope excluded.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import {
+  upstreamClosure,
+  rejectionNodeId,
+  rejectionQueuedNothing,
+  prunedScopedPromptBody,
+  prunedRetryForRejection,
+  prunedRetryNote,
+} from "../../web/js/lib/partial-run-prune.js";
+
+// The reporter's shape: one checkpoint feeding TWO independent output branches. 43 is
+// the ByteDance output they asked for; 56/57 are the Topaz nodes that refused it.
+const TWO_BRANCH = {
+  "1": { class_type: "CheckpointLoaderSimple", inputs: { ckpt_name: "sd.safetensors" } },
+  "2": { class_type: "CLIPTextEncode", inputs: { text: "a cat", clip: ["1", 1] } },
+  "3": { class_type: "KSampler", inputs: { model: ["1", 0], positive: ["2", 0], seed: 42 } },
+  "40": { class_type: "VAEDecode", inputs: { samples: ["3", 0], vae: ["1", 2] } },
+  "43": { class_type: "SaveImage", inputs: { images: ["40", 0] } },
+  "56": { class_type: "TopazUpscale", inputs: { image: ["40", 0] } },
+  "57": { class_type: "SaveImage", inputs: { images: ["56", 0] } },
+};
+
+const bodyFor = (prompt, targets, extra = {}) =>
+  JSON.stringify({
+    prompt,
+    client_id: "abc",
+    number: 1073741822,
+    partial_execution_targets: targets,
+    extra_data: { extra_pnginfo: { workflow: { nodes: ["the whole graph"] } } },
+    ...extra,
+  });
+
+const rejection = (nodeId, classType = "TopazUpscale") => ({
+  error: {
+    type: "missing_node_type",
+    message: `Node '${classType}' not found. The custom node may not be installed.`,
+    details: `Node ID '#${nodeId}'`,
+    extra_info: { node_id: String(nodeId), class_type: classType, node_title: classType },
+  },
+  node_errors: {},
+});
+
+// A Response double with the surface the guard uses (mirrors run-scope-guard.test.mjs).
+function jsonResponse(status, obj) {
+  return {
+    status,
+    clone() {
+      return { json: async () => JSON.parse(JSON.stringify(obj)) };
+    },
+    text: async () => JSON.stringify(obj),
+  };
+}
+
+// ---------------------------------------------------------------------------
+// The closure — completeness first
+// ---------------------------------------------------------------------------
+
+test("#1871 upstreamClosure keeps the target and everything upstream, and nothing else", () => {
+  const keep = upstreamClosure(TWO_BRANCH, ["43"]);
+  assert.deepEqual([...keep].sort(), ["1", "2", "3", "40", "43"]);
+  // The other branch is out — that is the whole fix.
+  assert.equal(keep.has("56"), false);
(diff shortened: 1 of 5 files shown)


artokun:
**Duplicate work alert: there are two open PRs for artokun/comfyui-mcp#1871.**

- **panel#1511** `fix/mcp-1871-partial-run-prune` — created 18:13, draft, CLEAN, +945/-1
- **panel#1512** `fix/1871-run-to-node-missing-branch` — created 18:20, draft, CLEAN, +1145/-76

Both are still drafts, so nothing has landed. **Please merge only one** — two independent prune implementations on the same submit path would conflict, and the second to merge would silently re-do or contradict the first.

For whoever picks: the issue was claimed on the mcp side at **17:39Z** ("dispatched to a fix agent, worktree `wt-1871`"), before either PR existed. That is not a claim of priority, just the timeline — a claim comment only works if it is read before starting, and this is the case it exists to prevent.

Substantive differences I can state about **#1512**, so the choice is on merits rather than timestamps:

- The diagnosis was verified with **three live POSTs on the reporter's stack** (ComfyUI 0.33.2 / frontend 1.49.6): full graph + targets with an uninstalled type on the unrelated branch → **400** with the reporter's error verbatim (`Node ID '#56'`); the same request pruned to the target's closure → **200**; and — the control — full graph + same targets with that branch *installed* → **200**. That third case is what establishes partial execution is not itself broken and nothing we send defeats it.
- It establishes placement rather than assuming it: `panel_run` forwards `{cmd:"graph_run", to_node_id}` and never holds a prompt, so the POST body inside the frontend's `queuePrompt` is the panel's only grip on it. The prune sits on the same seam the scoped-run guard already rewrites, deriving the closure from the body so the deferred-serialization path cannot go stale.
- It also fixes the **#1460 missing-node pre-flight**, whose stated premise ("a run carrying an unregistered type cannot succeed") this change falsifies for scoped runs. Left alone that pre-flight would re-create #1871 one layer earlier.
- 6 of 6 mutations killed. A seventh survived the first draft and changed the code: the original self-check verified its own output, so no input could make it fail; it was replaced with an input-side precondition that a wrong input actually can fail.

I have no visibility into #1511's verification, so this is not an argument that it is worse — only the evidence I can vouch for. If #1511 is further along or better scoped, take it and close #1512; the waste is already spent, and merging both is the only outcome that makes it worse.

PR #1512, opened 2026-08-20
fix(mcp#1871): the panel's own pre-flight stops vetoing a branch the run excluded

artokun:
Closes the remaining half of the underlying cause of **artokun/comfyui-mcp#1871** — "panel_run to_node_id validates unrelated missing-node branches". (Cross-repo, so it will not auto-close; closing that issue by hand on merge.)

> **This PR was rewritten.** It originally carried a different fix for #1871 — trimming every scoped run's POST body to the target's backward closure — which was gated SHIP and then **superseded by #1511**, merged while it was in review. #1511's design is better: it prunes only *after* ComfyUI has actually refused, so it costs nothing on the runs that were never broken. That work is dropped, and what remains is the part #1511 does not reach. The earlier review, including a P2 it found, is preserved in the comments below.

## What #1511 left behind

#1511 removed **ComfyUI's** veto: when the server refuses a run-to-node over a node outside the requested closure, the run gets one more post with that branch left out.

But the server only gets to refuse if we post — and the #1460 missing-node pre-flight refuses **first**, unscoped. So the retry never fires for the case where the pack is missing *entirely* rather than backend-only: no frontend registration means no `class_type` in the serialized prompt, which is precisely this check's trigger. That is also the likelier shape of "a pack you do not have"; the reporter's own case was the narrower one where the frontend JS had loaded and only the server class was absent.

Measured on `main` at `dd7f37ac`, driving the **real extracted pre-flight block** out of `comfyui-mcp-panel.js`:

```
tree is UNSCOPED
  run-to-node 43: REFUSED — NOT queued: 2 nodes in this workflow cannot be executed
                  by the server — TopazVideoAI (node 56), node 57.
```

Node 43's branch is `EmptyImage → PreviewImage`; 56/57 are a separate branch. ComfyUI is never asked, so nothing #1511 added can help. After this change:

```
tree is SCOPED
  run-to-node 43: no refusal (proceeds to ComfyUI)
  full run:       REFUSED — NOT queued: 2 nodes ... TopazVideoAI (node 56), node 57.
```

## The change

The refusal rests on one claim, stated in its own comment: *"a run carrying an unregistered type cannot succeed, so this blocks no working run."* That is true of a full run. It stopped being true of a run-to-node the moment #1511 let the excluded branch be dropped from a retry — such a run **can** succeed, and this check was the only thing left preventing it.

So the scope narrows the refusal and nothing else:

- the result is always a **subset** of the unscoped answer — pinned by a test that checks every returned id against it across six different target sets, because a version that reported closure members instead of filtering would refuse a run for a node that has nothing wrong with it;
- a full run — no targets, `null`, an empty list, a bare string — gets today's answer;
- an **unresolvable** scope (a root that is not a key of this prompt) gets today's answer, because a closure that cannot be computed is a guess and the safe direction is the behaviour that already exists.

`upstreamClosure` is **imported from `partial-run-prune.js`**, not reimplemented. Both places are answering "is this node part of the requested run?", and #1511's version is the more careful one — it deliberately keeps a superset of what the executor walks, matching `validate_inputs`' broader len-2 handling rather than `is_link`'s stricter typing. Two answers to one question is two chances to disagree, and here they are the two ends of the same refusal.

Resolving the run-to-node target moved above the pre-flight so the scope exists when it is read. Resolution is read-only; it only changes which refusal answers first, and *"node N was not found"* is the better reply to a request naming node N than a report about some other node's pack.

## Evidence

- **The ordering is pinned by its own test**, and it needs to be. With the resolve block below the pre-flight, `partialTargets` sits in its temporal dead zone; the `ReferenceError` is swallowed by the pre-flight's own `catch` (it re-throws only `/^NOT queued:/`), and the entire missing-node check silently stops running. A failure with no symptom at all.
- **Mutation-tested, 5 of 5 killed**: the helper ignoring its scope; narrowing a full run; treating an unresolvable closure as "nothing is in scope"; returning the closure instead of the filtered subset; and the call site dropping its second argument. Each fails at least one test that names what it broke.
- The probe above initially reported a false "no refusal" because it did not inject `partialTargets` into the extracted block — the swallowed-ReferenceError trap, hit live. Both the probe and the `#1582` harness now inject it, which is what makes the "no refusal" result mean anything.
- **The load-bearing claim was tested, not assumed.** Removing our refusal is only an improvement if ComfyUI's own refusal is then rescued by #1511's retry. Against the live server and the real merged helper, for the bare-node shape this change newly lets through:

  ```
  1. ComfyUI's answer to the unpruned body: HTTP 400
     {"error": {"type": "missing_node_type",
                "message": "Node 'ID #56' has no class_type. ...",
                "extra_info": {"node_id": "56", "class_type": null, "node_title": null}}}
  2. rejectionNodeId -> "56"   rejectionQueuedNothing -> true
     is that node outside the requested closure? -> true
  3. prunedRetryForRejection -> RETRY PRODUCED  (namedNode=56, removed=["56","57"])
     retry prompt keys: ["43","44"]
  4. the retry POST: HTTP 200  {"prompt_id": "a10e6c67-...", "node_errors": {}}
  ```

  So the rejection names the node in a **structured** field for the no-`class_type` branch too (not only the unresolved-class one), and the run genuinely completes.
- `npm run test:unit`: **5962 pass / 0 fail** (5953 on `origin/main` before this), including the i18n, tool-vocabulary and panel-scope gates. `npx tsc --noEmit` clean.

## Review status — **PENDING**

**This change has not been reviewed yet.** Review has been requested from grok in a comment below, with the load-bearing parts named. It is not gated, and no verdict is claimed for it. Not to be merged while that review is outstanding, and not on the author's own verification alone.

For the record: the *superseded* diff (the unconditional prune, now dropped) did get an independent adversarial pass, whose verdict and P2 are preserved in the comments above — the P2 about ComfyUI cache eviction is why #1511's conditional design is the right one, and is worth keeping in view if anyone reaches for an unconditional prune again.

## Browser suite

Not re-run against this reduced diff. On the previous one it was `21 failed / 57 passed` against `22 failed / 56 passed` for clean `origin/main` on the same rig — **zero failures unique to the branch**. The suite is broadly red here because ComfyUI's `custom_nodes` junction points at the shared checkout, which sits on another branch, so any spec serving panel source from a worktree gets a split module graph and the panel never mounts. e2e is deliberately not in CI for this repo (`ci.yml`: *"playwright.config.ts requires a real ComfyUI at :8188 with models loaded, so it belongs to the pre-release local soak"*). Nothing here is panel-*rendered*; the change removes a refusal.

- **Did not keep the unconditional prune.** It was gated SHIP, but #1511 does the same job without evicting ComfyUI's caches on runs that were never broken. Shipping mine on top would have replaced a cheaper design with a costlier one.
- **Did not add a second closure implementation.** The one #1511 merged is reused as-is.
- **Did not touch the orchestrator.** `panel_run` forwards `{cmd:"graph_run", to_node_id}` and never holds a prompt.
- **Did not repoint the rig's `custom_nodes` junction** to get a greener e2e run. It is shared, and other work is live on that checkout.

### DIFF
--- a/browser_tests/unit/frontend-virtual-nodes.test.mjs
+++ b/browser_tests/unit/frontend-virtual-nodes.test.mjs
@@ -293,7 +293,7 @@ test("#1648 CALL SITE 3: the bounded list still says how many were withheld", ()
 });
 
 test("#1648 CALL SITE 3 WIRING: graph_run hands the refusal this page's registry", () => {
-  const at = PANEL.indexOf("const badIds = unrunnableNodeIds(built);");
+  const at = PANEL.indexOf("const badIds = unrunnableNodeIdsInScope(built, partialTargets);");
   assert.ok(at > 0, "the pre-flight must still be recognisable");
   const end = PANEL.indexOf("} catch (err) {", at);
   assert.ok(end > at, "the pre-flight's try block must still be recognisable");
@@ -305,10 +305,10 @@ test("#1648 CALL SITE 3 WIRING: graph_run hands the refusal this page's registry
 });
 
 test("#1648 CALL SITE 3 WIRING: the registry is diagnosis only — the refusal is still the prompt", () => {
-  const at = PANEL.indexOf("const badIds = unrunnableNodeIds(built);");
+  const at = PANEL.indexOf("const badIds = unrunnableNodeIdsInScope(built, partialTargets);");
   const end = PANEL.indexOf("} catch (err) {", at);
   const block = PANEL.slice(at, end);
-  // The go/no-go remains `unrunnableNodeIds(built)`; the registry appears only inside the
+  // The go/no-go remains `unrunnableNodeIdsInScope(built, partialTargets)`; the registry appears only inside the
   // message construction. A guard keyed on the registry would refuse runs that succeed.
   assert.match(block, /if \(badIds\.length\) \{/);
   assert.equal((block.match(/registered_node_types/g) ?? []).length, 1);
--- a/browser_tests/unit/graph-to-prompt-failure.test.mjs
+++ b/browser_tests/unit/graph-to-prompt-failure.test.mjs
@@ -103,7 +103,7 @@ test("#1582 the run path guards graphToPrompt BEFORE reading offenders", async (
   assert.ok(endOfTry > at, "the pre-flight try block must still be recognisable");
   const block = src.slice(at, endOfTry);
   const guard = block.indexOf("graphToPromptUnusable(built)");
-  const offenders = block.indexOf("unrunnableNodeIds(built)");
+  const offenders = block.indexOf("unrunnableNodeIdsInScope(built, partialTargets)");
   assert.ok(guard > -1, "the unusable-result guard must exist");
   assert.ok(offenders > -1, "the offender check must still exist");
   // ORDER is the fix. Asking for offenders first answers `[]` for an absent result and
@@ -127,7 +127,7 @@ test("#1582 the refusal keeps the prefix its own catch requires", async () => {
 //    the same real-source extraction pattern manager-dialect.test.mjs uses.
 
 /** Pull the pre-flight try/catch out of the monolith and run it with injected deps. */
-async function buildPreflight({ graphToPrompt, nodes = [], viewedNodes, registry = {} }) {
+async function buildPreflight({ graphToPrompt, nodes = [], viewedNodes, registry = {}, partialTargets }) {
   const { readFileSync } = await import("node:fs");
   const src = readFileSync(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url), "utf-8");
   const marker = src.indexOf("      // Inspect the SERIALIZED prompt");
@@ -141,7 +141,12 @@ async function buildPreflight({ graphToPrompt, nodes = [], viewedNodes, registry
     "app",
     "graph",
     "LG",
-    "unrunnableNodeIds",
+    "unrunnableNodeIdsInScope",
+    // comfyui-mcp#1871 - the block now reads the resolved run-to-node scope. It is a
+    // `let` in graph_run, so leaving it out of the factory would make the extracted body
+    // throw a ReferenceError that the pre-flight catch swallows - and the test would then
+    // see "no refusal" from a pre-flight that never ran at all.
+    "partialTargets",
     "describeUnrunnable",
     "missingNodeRunRefusal",
     "graphToPromptUnusable",
@@ -167,7 +172,8 @@ async function buildPreflight({ graphToPrompt, nodes = [], viewedNodes, registry
     { graphToPrompt },
     { _nodes: viewedNodes ?? nodes },
     { registered_node_types: registry },
-    mod.unrunnableNodeIds,
+    mod.unrunnableNodeIdsInScope,
+    partialTargets,
     mod.describeUnrunnable,
     mod.missingNodeRunRefusal,
     mod.graphToPromptUnusable,
@@ -233,6 +239,83 @@ test("#1582 BEHAVIOUR: the #1460 unrunnable-node refusal still fires", async () 
   assert.match(msg, /^NOT queued:/);
   assert.match(msg, /cannot be executed by the server/);
   assert.match(msg, /GoneNode/);
+});
+
+// ── comfyui-mcp#1871. #1511 took ComfyUI's veto of an excluded branch away: when the
+//    SERVER refuses over a node outside the requested closure, the run is re-posted
+//    without that branch. But the server only gets to refuse if we post, and this
+//    pre-flight refuses first. For the case where a pack is missing ENTIRELY there is no
+//    frontend registration, so no class_type, so this check fires and ComfyUI is never
+//    asked — the retry cannot save a run that never left the browser.
+//
+//    Driven through the SAME extracted pre-flight source: this is a claim about what
+//    production refuses, not about what a helper returns.
+
+// Two independent output branches, the reporter's shape: 43 is the branch asked for,
+// 56/57 are the branch whose pack is not installed at all (hence no class_type).
+const twoBranchPrompt = () => ({
+  output: {
+    "43": { class_type: "PreviewImage", inputs: { images: ["44", 0] } },
+    "44": { class_type: "EmptyImage", inputs: { width: 64, height: 64 } },
+    "56": { inputs: { image: ["57", 0] } },
+    "57": { inputs: {} },
+  },
+  workflow: {},
+});
+const twoBranchNodes = [
+  { id: 43, type: "PreviewImage" },
+  { id: 56, type: "TopazVideoAI" },
+];
+
+test("#1871 BEHAVIOUR: a scoped run is NOT refused for an unrunnable node outside its branch", async () => {
+  const msg = await runPreflight({
+    graphToPrompt: async () => twoBranchPrompt(),
+    nodes: twoBranchNodes,
+    registry: { PreviewImage: {} },
+    partialTargets: ["43"],
+  });
+  assert.equal(msg, "__NO_REFUSAL__", "node 56 is not upstream of 43; the run reaches ComfyUI");
+});
+
+test("#1871 BEHAVIOUR: a scoped run IS still refused for an unrunnable node INSIDE its branch", async () => {
+  // The narrowing must not become a blanket exemption for scoped runs: node 56 is now a
+  // dependency of the requested target, so the run genuinely cannot succeed and #1511's
+  // retry could not rescue it either (it declines when the named node is in the closure).
+  const prompt = twoBranchPrompt();
+  prompt.output["43"].inputs.images = ["56", 0];
+  const msg = await runPreflight({
+    graphToPrompt: async () => prompt,
+    nodes: twoBranchNodes,
+    registry: { PreviewImage: {} },
+    partialTargets: ["43"],
+  });
+  assert.match(msg, /^NOT queued:/);
+  assert.match(msg, /TopazVideoAI/);
+});
+
+test("#1871 BEHAVIOUR: a FULL run over the same graph still refuses — the narrowing is scoped-only", async () => {
+  const msg = await runPreflight({
+    graphToPrompt: async () => twoBranchPrompt(),
+    nodes: twoBranchNodes,
+    registry: { PreviewImage: {} },
+    partialTargets: undefined, // no to_node_id ⇒ every node is submitted
+  });
+  assert.match(msg, /^NOT queued:/);
+  assert.match(msg, /TopazVideoAI/);
+});
+
+test("#1871 BEHAVIOUR: an unresolvable scope refuses exactly as an unscoped run does", async () => {
+  // A target that is not a key of this prompt means the closure is a guess. Guessing
+  // toward "let it through" would ship a run that cannot succeed; the safe direction is
+  // the behaviour that already exists.
+  const msg = await runPreflight({
+    graphToPrompt: async () => twoBranchPrompt(),
+    nodes: twoBranchNodes,
+    registry: { PreviewImage: {} },
+    partialTargets: ["not-a-node-in-this-prompt"],
+  });
+  assert.match(msg, /^NOT queued:/);
+  assert.match(msg, /TopazVideoAI/);
 });
 
 // ── ROOT SCOPE (review, P1). graphToPrompt serializes the WHOLE workflow, so naming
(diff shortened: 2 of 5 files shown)


artokun:
**Duplicate work alert: there are two open PRs for artokun/comfyui-mcp#1871.**

- **panel#1511** `fix/mcp-1871-partial-run-prune` — created 18:13, draft, CLEAN, +945/-1
- **panel#1512** `fix/1871-run-to-node-missing-branch` — created 18:20, draft, CLEAN, +1145/-76

Both are still drafts, so nothing has landed. **Please merge only one** — two independent prune implementations on the same submit path would conflict, and the second to merge would silently re-do or contradict the first.

For whoever picks: the issue was claimed on the mcp side at **17:39Z** ("dispatched to a fix agent, worktree `wt-1871`"), before either PR existed. That is not a claim of priority, just the timeline — a claim comment only works if it is read before starting, and this is the case it exists to prevent.

Substantive differences I can state about **#1512**, so the choice is on merits rather than timestamps:

- The diagnosis was verified with **three live POSTs on the reporter's stack** (ComfyUI 0.33.2 / frontend 1.49.6): full graph + targets with an uninstalled type on the unrelated branch → **400** with the reporter's error verbatim (`Node ID '#56'`); the same request pruned to the target's closure → **200**; and — the control — full graph + same targets with that branch *installed* → **200**. That third case is what establishes partial execution is not itself broken and nothing we send defeats it.
- It establishes placement rather than assuming it: `panel_run` forwards `{cmd:"graph_run", to_node_id}` and never holds a prompt, so the POST body inside the frontend's `queuePrompt` is the panel's only grip on it. The prune sits on the same seam the scoped-run guard already rewrites, deriving the closure from the body so the deferred-serialization path cannot go stale.
- It also fixes the **#1460 missing-node pre-flight**, whose stated premise ("a run carrying an unregistered type cannot succeed") this change falsifies for scoped runs. Left alone that pre-flight would re-create #1871 one layer earlier.
- 6 of 6 mutations killed. A seventh survived the first draft and changed the code: the original self-check verified its own output, so no input could make it fail; it was replaced with an input-side precondition that a wrong input actually can fail.

I have no visibility into #1511's verification, so this is not an argument that it is worse — only the evidence I can vouch for. If #1511 is further along or better scoped, take it and close #1512; the waste is already spent, and merging both is the only outcome that makes it worse.

artokun:
Heads-up: we were both on comfyui-mcp#1871. #1511 merged at 18:43Z (`dd7f37ac`), so this branch will now conflict in `run-scope-guard.js` and `comfyui-mcp-panel.js`. Same diagnosis, and your `prompt-closure.js` and my `partial-run-prune.js` compute the same set — posting the difference rather than assuming either of us has it right.

**Where they differ: when the prune is spent.** #1511 prunes only after ComfyUI has already refused, naming a node outside the closure in `error.extra_info.node_id`. This PR prunes every scoped run. The measurement that pushed me to the retry, from `execution.py`:

```python
for cache in self.caches.all:
    await cache.set_prompt(dynamic_prompt, prompt.keys(), is_changed_cache)
    cache.clean_unused()
```

The default classic `HierarchicalCache` drops every cached output whose key is absent from the prompt it was just handed. So a pruned scoped run evicts the *other* branch's cached results, and the user's next full run recomputes a branch it had already rendered — on the big workflows run-to-node is for, that is real GPU minutes. If you have measured that it does not happen (an LRU/RAM cache, or a re-post that restores the keys), that beats my reading and the always-prune is the better design; I could not find a path where it does not.

**Where this PR covers something #1511 does not.** Scoping `missing-node-preflight` to the closure. #1511 left that untouched, so a node with no resolvable `class_type` at all on the *other* branch still trips the panel's own pre-flight and refuses before anything is posted — my retry never gets a chance, because there is no rejection to read. That is the same bug through a second door and it is still open on main. That half looks worth landing on its own, rebased.

Nothing here is a verdict on your PR — you have context I do not, including whatever your gate found. Flagging the collision and the two facts, and leaving the call to you.

artokun:
**Gate round 1 — SHIP** (substituted adversarial review; codex account exhausted, disclosed per the substitution's terms).

```
SHIP

No P0/P1. What I ran, by class:

1. Two states collapsed — backwardClosureIds returns null for unreadable prompt, no roots,
   and root-not-a-key; every one maps to "answer exactly as today" in both callers.
   Fuzzed 190 body×scope combinations: threw=0, pruned=8, declined=182. Holds.
2. Identity across the transformation — number(mark), client_id, partial_execution_targets,
   front, and extra_data.extra_pnginfo.workflow all survive byte-for-byte; a foreign-mark
   post is forwarded as the same object (state.pruned === 0). Holds.
3. Guard on the wrong pair — pruneScopeInBody compares the body's own
   partial_execution_targets against expected before trimming to expected. Adversarial
   pairs all decline; only agreement acts. Right pair.
4. Test pins the fix — 4 mutations, all red, including the panel call site dropping
   partialTargets. Full suite 5943/5943, 0 fail.
5. Fix worse than the bug — closure verified a superset of ComfyUI's own walk; exhaustive
   per-node prune over the real fixture leaves no dangling link. is_link matches
   isPromptLink exactly on the actual ComfyUI 0.33.2.
6. Prose overstates — the central claim is true: validate_prompt scans class_type across
   the whole prompt and returns before partial_execution_list narrows outputs.
7. Reachable — graph_run → dispatchScopedRun → createScopedRunGuard → pruneScopeInBody.

One P2 (not gating, but the comment beside it is wrong): run-scope-guard.js claims "Only
ComfyUI's whole-prompt class_type scan ever looked at the nodes this drops". It isn't.
execution.py:756-758 runs set_prompt(dynprompt, prompt.keys()) + clean_unused() over BOTH
caches. Ran ComfyUI's own HierarchicalCache: unpruned run-to-node → survivors
['43','44','92','93']; pruned → ['43','44'] — for objects (loaded models) as well as
outputs. So every scoped run now evicts the unrelated branches' cached checkpoints and
intermediates, and the next full run reloads and recomputes them. Bounded and cost-only,
not a correctness bug — but the sentence asserting otherwise should go.
```

I verified the P2 independently rather than taking it on faith (`execution.py` 748-760 — `for cache in self.caches.all: await cache.set_prompt(dynamic_prompt, prompt.keys(), ...); cache.clean_unused()`). It is correct, and it is a cost I had asserted did not exist. Commit 278555f5 removes the false sentence and records the measurement in both doc blocks, and separately moves the prune to the top of the dispatch branch (it had split the r4 quota-reservation comment from the `state.inFlight++` it describes) and fences it in a try/catch. All six mutations re-run red afterwards; unit suite 5943 pass / 0 fail; `tsc --noEmit` clean.

Round 2 running against the revised call site; verdict to follow.

artokun:
**Rewritten after #1511 merged.** #1511 landed a fix for this same issue while this PR was in review, using the better design — prune only *after* ComfyUI actually refuses, so nothing is paid on the runs that were never broken. My unconditional prune is dropped rather than layered on top of it.

What is left is the half #1511 cannot reach: the panel's own #1460 pre-flight refuses **before** anything is posted, so ComfyUI never gets to refuse and the retry never fires. That is the case where a pack is missing *entirely* (no frontend registration ⇒ no `class_type`) rather than backend-only. Verified against `main` at dd7f37ac by driving the real extracted pre-flight — `panel_run(to_node_id: 43)` is still refused there, naming nodes 56 and 57 on a branch it excluded.

The superseded work's gate verdict stands above for the record, including its P2 about ComfyUI cache eviction (`set_prompt(dynprompt, prompt.keys())` + `clean_unused()` across `objects` as well as `outputs`) — that finding is precisely why #1511's conditional design is the right one, and it is worth keeping in view if anyone reaches for an unconditional prune again.

artokun:
@grok please review this — and please aim at the three places below rather than the diff as a whole. They are where I think it breaks if it breaks, and two of them are claims I could not fully close myself.

**Context in one line:** #1511 already made ComfyUI's refusal of an excluded branch recoverable (it re-posts once without that branch). This PR removes the panel's *own* pre-flight refusal, which fires earlier and unscoped, so ComfyUI never gets asked and that retry never runs. The whole change is `unrunnableNodeIds(built)` → `unrunnableNodeIdsInScope(built, partialTargets)`, plus moving the target resolution above the pre-flight so the scope exists when it is read.

---

**(a) Can `upstreamClosure` MISS a node the target genuinely depends on?** This is the sharpest one, and note the direction of harm has flipped relative to #1511. There, an over-small closure meant declining a retry. Here, an over-small closure means the pre-flight *stops refusing* for a node the run actually needs — we hand the user ComfyUI's error instead of our own much better one.

The specific worry is the snapshot, not the walk. The pre-flight calls `app.graphToPrompt()` **itself**, separately from the serialization the run later performs. Queue-time link materialisation happens inside the frontend's own `queuePrompt` patch — cg-use-everywhere broadcast links are the documented case, and `run-scope-guard.js`'s `collectVolatileInputs` exists precisely because those inputs are not present in an ordinary serialization. If a dependency edge is materialised at queue time, it is absent from the map the closure walks, so an unrunnable node that IS upstream can look unrelated. Does that reach this code path? Does it matter, given the node still has no `class_type` and ComfyUI will refuse anyway?

Also worth a look: Get/Set bus nodes and other frontend-only types that never appear in `/object_info`. `graphToPrompt` resolves virtual/reroute/primitive nodes away, so I do not think they survive into the map at all — but I have not proven that for every one of them.

**(b) The reordering.** `resolveRunToNodeTarget(rootGraph, viewing, to_node_id)` now runs before the pre-flight. I claim it is read-only; if it is not, I have moved a side effect ahead of a check that can refuse the run. It also changes which refusal answers first when a request is bad in two ways at once (previously "your pack is missing", now "node N was not found") — I think that is an improvement, but it is a user-visible change I made deliberately and would rather have argued than assumed.

There is a booby-trap here worth confirming I actually defused: `partialTargets` is a `let`, and if the resolve block ever slides back below the pre-flight, the read hits its temporal dead zone — and the pre-flight's `catch` re-throws only `/^NOT queued:/`, so the `ReferenceError` is swallowed and the **entire missing-node pre-flight silently stops running**, with every other test still green. I added an ordering test for exactly that. Is the test actually load-bearing, or can the ordering break in a way it does not see?

**(c) The dependency on #1511 being able to rescue what I now let through.** Removing our refusal is only an improvement if ComfyUI's refusal is then recoverable. I verified the specific shape this change newly admits — a node with **no `class_type` at all** — end to end on ComfyUI 0.33.2: the rejection carries `extra_info.node_id: "56"` (so `rejectionNodeId` reads it structurally, and it is a different `validate_prompt` branch from the unresolved-class one #1511 was built against), `prunedRetryForRejection` produces a retry removing 56 and 57, and the retried POST returns 200.

What I did **not** test: `batch_count > 1`, where the first item is rescued and later ones post again; and whether the retry interacts with the batch quota fence in `createScopedRunGuard` (does a retried post consume a slot, and can it trip the overrun refusal?). If the retry is quota-visible, I may have increased how often that path is reached.

---

Two things I would rather you did not spend time on, because they are already settled: the closure implementation itself is #1511's, reused deliberately rather than reimplemented; and the unconditional-prune approach this PR originally carried is dropped — an earlier review found it evicts ComfyUI's caches (`set_prompt(dynprompt, prompt.keys())` + `clean_unused()` across `objects` as well as `outputs`), which is exactly why #1511's conditional design won.

artokun:
**Note for the record: this merged at 19:27:15Z (`da4f7d76`) with the grok review still outstanding.** The request went up at 19:19:51Z and no verdict had come back — the only review on the PR is `copilot-pull-request-reviewer: COMMENTED`.

I am not asking for a revert. The change is well covered on its own terms: 5/5 mutations killed, 5962 unit tests passing, `tsc` clean, CI green, and the Playwright delta against clean `origin/main` was 21 failed/57 passed vs 22 failed/56 passed — zero failures unique to the branch.

@grok — **the review request still stands**, and it is worth doing even though this is merged. If you find something, please file it as a new issue against `web/js/lib/run-scope-guard.js` / `missing-node-preflight.js` rather than replying only here, so it does not get lost behind a closed PR. The three places I asked you to aim at, unchanged:

1. **Can `upstreamClosure` miss a dependency materialised at queue time?** The pre-flight calls `graphToPrompt()` separately from the run's own serialization. The harm direction flipped when this PR narrowed the pre-flight to scope: an over-small closure no longer causes a spurious refusal, it causes us to *skip* refusing for a node the run genuinely needs — so the failure is now silent rather than loud.
2. **The reordering of the target resolution above the pre-flight**, and the TDZ trap in it — getting that wrong disables the entire pre-flight without failing a test.
3. **`batch_count > 1` and the quota fence interacting with #1511's retry**, which is untested.

Flagging the process point because it has a track record here today: five PRs merged ahead of their review verdict on 2026-08-20, and every one of those reviews then reported a real defect — two of which shipped, including a Blind-mode bypass and a P1 that reported a successful `/free` as a failure. A verdict that arrives after the merge button is a review nobody read.

Issue #1495, opened 2026-08-20
Subgraph viewing scope becomes contradictory after panel_exit_subgraph


artokun:
### What happened / root cause
The live-canvas subgraph viewing scope became internally inconsistent across panel calls. A prior panel_exit_subgraph returned scope=root and settled=true, but the next panel_enter_subgraph(root node 5548) failed saying the viewer was still inside subgraph 9002. Immediately calling panel_exit_subgraph then returned note="already at root", after which panel_enter_subgraph(5548) succeeded. This suggests stale or split scope state between validation/routing and the reported viewing state.

### Steps to reproduce
1. Enter subgraph 9002.
2. Call panel_exit_subgraph; observe scope=root, settled=true.
3. On a later turn call panel_enter_subgraph for root node 5548.
4. Observe error claiming current scope is still subgraph 9002.
5. Call panel_exit_subgraph; it reports already at root.
6. Retry panel_enter_subgraph(5548); it succeeds.

### Exact error (scrubbed)
Error: No node with id 5548 in the current graph. Node 5548 is on the ROOT graph, but you are currently inside a subgraph ... Call panel_exit_subgraph, then retry.

Contradictory next response:
{"viewing":{"scope":"root"},"note":"already at root"}

### Fix
Applied locally: no. Upstream-only: yes.
Synchronize the graph mutation router's active subgraph scope with the viewer scope before validating node ownership, or make panel_exit_subgraph await and verify both state stores before returning settled=true. A stale-scope recovery at the beginning of panel_enter_subgraph would also prevent the false rejection.

### Environment
Windows 11
ComfyUI 0.33.1
ComfyUI frontend 1.48.7
Python 3.13.9
PyTorch 2.9.1+cu130
GPU NVIDIA GeForce RTX 5090 31.8 GB
comfyui-mcp 0.52.25
panel 0.14.31

artokun:
Claimed 2026-08-20T14:22Z — dispatched to a fix agent, worktree `wt-1495` (branch `fix/1495-subgraph-scope-after-exit`). PR will be posted here. If this comment is older than 8h with no PR, treat the claim as lapsed and pick it up.

Issue #1298, opened 2026-08-17
Panel mutation referenced a node ID that was absent after graph changes


artokun:
During workflow repair after manual node deletions, panel_remove_node for node 99 returned: No node with id 99 in the current graph, despite the prior graph outline having shown node 99. This is likely stale graph state or a node mutation race and caused an unnecessary failed repair step. Environment: Windows 11, ComfyUI 0.33.0, RTX 3090 24GB, comfyui-mcp 0.51.57, panel 0.14.41.

Related (same root cause): #1187

artokun:
Claimed by grok-swarm 2026-08-19T06:22:48Z — driving this in worktree `/c/Users/Artokun/code/comfyui-mcp-panel/.worktrees/wt-1298` (branch `fix/1298-absent-node-id`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp-panel/pull/1365 — a genuine miss after graph changes now names the current live ids so the next mutation can retarget without another outline. Fixes #1298.

PR #1521, opened 2026-08-20
fix(1501): the locator says "I could not look" instead of "it is not there"

### LINKED ISSUE #1501: node-scope-locator reports "may be from a different workflow" for a walk that threw and for any qualified id
Two pre-existing defects in `web/js/lib/node-scope-locator.js`, both surfaced while fixing #1495 and both the **same shape as that bug**: a diagnostic that cannot find a node reports "it does not exist" when what actually happened is "I could not look."

#1495 fixed the third instance of this in the same file (a remedy that asserted "you are inside a subgraph" without checking). These two were deliberately left out of that diff to keep it to one defect; filing so they are not lost with the PR.

## 1. A walk that THREW is reported as "not found anywhere"

`locateNodeAcrossScopes` (`:80`) ends with:

```js
try {
  return walk(rootGraph, [], 0);
} catch {
  return null; // a diagnostic must never throw
}
```

Not throwing is right. Returning the **same value as a completed search that found nothing** is not: the consumer at `:193` treats `null` as a genuine miss and renders

> `The id may be from a different workflow, or the node was removed.`

So a malformed graph, a throwing getter on a third-party node, or any exception mid-walk is reported to the agent as a positive statement about the node's absence. The walk touches arbitrary third-party node objects (`n?.subgraph`, `n.title`, `n.type`), so this is reachable rather than theoretical.

The fix shape is the one #1495 used: keep failing closed, but make "could not determine" a distinct answer from "searched and absent", and say so.

## 2. A qualified id can never be found

The walk starts with:

```js
const target = Number(nodeId);
if (!Number.isFinite(target)) return null;
```

A **qualified id** — the `<host>:<inner>` form from #1425 — is not a finite number, so the locator returns `null` before searching anything, and the caller emits that same "may be from a different workflow" line. The node exists, is addressable, and the diagnostic says it is from another workflow.

That message is worse than silence here: it points the reader at a workflow-identity problem that does not exist, for an id shape the rest of the system produces deliberately.

## Why together

One file, one consumer, one message, and one root cause — `null` is doing three different jobs (searched-and-absent, could-not-search, cannot-parse-this-id) and the caller can only spell one of them. Fixing either alone leaves the same sentence lying for a different reason.

Not urgent: both produce a misleading diagnostic rather than a wrong mutation, and #1495's fix already removed the most common false remedy from this path.

artokun:
Closes #1501

Two pre-existing defects in `web/js/lib/node-scope-locator.js`, both the same shape #1495 fixed a third instance of: **a diagnostic that cannot find a node reports "it does not exist" when what actually happened is "I could not look."**

`null` was doing three different jobs — searched-every-scope-and-absent, the walk threw part-way, and the id was not a shape the walk could key on — and the only one of the three the consumer at `:193` could spell was the first:

> The id may be from a different workflow, or the node was removed.

## 1. A walk that THREW was reported as "not found anywhere"

`locateNodeAcrossScopes` ended in `catch { return null }`. Not throwing is right; returning the **same value as a completed search that found nothing** is not. The walk reads arbitrary third-party node objects (`n?.subgraph`, `n.title`, `n.type`), so a malformed graph or a throwing getter is reachable rather than theoretical — and it was published to the agent as a positive statement about the node's *absence*.

## 2. A qualified id could never be found

The walk opened with `Number(nodeId)`, so a **qualified id** (`120:104` — the form the read tools deliberately emit once a subgraph is unpacked, comfyui-mcp#1425) came out `NaN` and the locator returned before searching anything. The node existed, was addressable, and the diagnostic pointed the reader at a workflow-identity problem that does not exist.

This is reachable in production precisely where it hurts: a qualified root-level node written to while viewing a subgraph misses `getNodeById`, and used to be blamed on another workflow instead of being told to `panel_exit_subgraph`.

## The fix

- `null` now means **only** "searched and absent" — so #1298's retarget list and #1495's checked scope claim are untouched.
- "Could not look" returns `{ unsearchable: "walk-threw" | "id-shape" }`, and the message says `UNKNOWN` explicitly, stating that this is *not* a claim of removal or of a foreign workflow.
- Ids are compared the way `canonicalNodeId` already resolves them: a qualified id by its string key, everything else numerically, exactly as before. A qualified id is **never parsed** — `parseInt("120:104")` is `120`, a different real node, which would turn a loud miss into a confident pointer at the wrong one.

Still non-throwing, still bounded, still failing closed.

## Verification

Unit suite green (`node --test browser_tests/unit/*.test.mjs`), typecheck clean. 35/35 in `node-scope-locator.test.mjs`, 10 of them new.

**Mutation-verified — each half of the fix was deleted and the right tests went red:**

| mutation | result |
|---|---|
| thrown-walk answer reverted to `return null` | 2 red (`a walk that THREW…`, `a thrown walk says UNKNOWN…`) |
| qualified-id handling reverted to `Number()`/`null` | 4 red (both qualified-id tests, the id-shape test, the malformed-graph test) |
| `isUnsearchable` branch deleted from `describeMissingNode` | 4 red — including the pre-existing `#1298 current-id listing never throws`, so the new branch is load-bearing for the non-throwing guarantee too |
| restored | 35/35 green |

Guarded against over-correction: `#1501 a genuine miss is still a genuine miss` pins that a real absence still gets `not in any other scope either` / `may be from a different workflow` and **not** `UNKNOWN` — the "unknown" answer must not swallow the actionable case.

`manager-install.test.mjs #671` failed once in the full-suite run and passes 125/125 in isolation — the known wall-clock flake, unrelated to this diff.

## Gate

`codex-gate.mjs` → **SHIP** (exit 0). Codex quota has recovered (the exhaustion window ended 2026-08-19 20:43), so this is the real gate, not the substitute.

## What I did not do

**No Playwright run.** This touches one `web/js/lib` diagnostic module with no DOM or rendering surface — nothing the panel *renders* changes, only the text of an error a bridge command throws. The identical footprint shipped in #1499 (same file, same consumer, same message) with unit tests only. No ComfyUI was allocated for this run, and the only listening instance (`:8188`) is shared — its `custom_nodes` junction points at another checkout, so an e2e run from here would have exercised someone else's code and returned a misleading green, and repointing it would have disrupted another agent. Stating this rather than implying coverage I do not have.

**`MAX_DEPTH` truncation left alone.** A walk that stops at the depth ceiling is arguably a third instance of the same "could not look" shape, but real workflows nest a handful deep, an existing test deliberately pins `null` there as a bounded-search invariant, and #1501 names exactly two defects. Flagging rather than widening the diff.

### DIFF
--- a/browser_tests/unit/node-scope-locator.test.mjs
+++ b/browser_tests/unit/node-scope-locator.test.mjs
@@ -4,6 +4,7 @@ import {
   locateNodeAcrossScopes,
   countSubgraphs,
   describeMissingNode,
+  isUnsearchable,
 } from "../../web/js/lib/node-scope-locator.js";
 
 /**
@@ -82,9 +83,14 @@ test("a DEEP acyclic chain is bounded — `seen` never fires when every level is
 test("malformed graphs never throw — a diagnostic must not fail", () => {
   for (const bad of [null, undefined, 42, "x", {}, { _nodes: "nope" }]) {
     assert.doesNotThrow(() => locateNodeAcrossScopes(bad, 1));
+    // These are searchABLE — there is simply nothing in them. A completed search
+    // that found nothing is `null`, and stays `null`.
     assert.equal(locateNodeAcrossScopes(bad, 1), null);
   }
-  assert.equal(locateNodeAcrossScopes(sub([node(1)]), "not-a-number"), null);
+  // …whereas an id the walk cannot key on was never searched at all (#1501).
+  assert.deepEqual(locateNodeAcrossScopes(sub([node(1)]), "not-a-number"), {
+    unsearchable: "id-shape",
+  });
 });
 
 // ── the message ───────────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


artokun:
Superseded by #1520, which merged for #1501 (1fa952f8) while this was in the gate. Closing unmerged — no code from here is needed.

I compared the two before closing rather than assuming: #1520 covers **both** defects and its design is equivalent to this one — a three-state locator result (`{undetermined: "threw"|"unparseable"}` vs. `{unsearchable: ...}` here), qualified-id matching through `canonicalNodeId`, and a message branch that refuses to spell absence when the search did not complete. Its `canonicalNodeId(n.id) === target` comparison is sound across the cases that mattered here: a numeric target still matches the string ids a live ComfyUI actually stores (`Number("5548") === 5548`), a qualified id matches by its string key, and a qualified id is never parsed into the different real node its prefix names (`parseInt("120:104")` → `120`). It also carries the regression guard that matters most — that a *genuine* miss still says the id may be gone, so "inconclusive" does not swallow #1298's actionable retarget path.

One observation from this work that neither fix addresses, left here rather than filed as noise: **a walk truncated by `MAX_DEPTH` is arguably a fourth instance of the same shape.** It stops at depth 12 without having searched the rest, and the caller still reports a completed search. It is remote — real workflows nest a handful deep — and `a DEEP acyclic chain is bounded` deliberately pins `null` there as a bounded-search invariant, so changing it would fight an intentional guarantee rather than fix a defect. Worth knowing if this file is opened for a fifth time.

Gate on this branch was clean (`codex-gate.mjs` → SHIP, exit 0) and CI passed; it is being closed for redundancy, not for a defect.

PR #1525, opened 2026-08-20
fix(1504): a minted prompt_id decides the verdict, not the node_errors beside it

### LINKED ISSUE #1504: panel_run reports refusal even though ComfyUI starts the prompt
panel_run returned a queue refusal with validation errors for downstream VAEDecode nodes, but ComfyUI immediately showed a running prompt. The panel then rejected all graph reads and edits with QUEUE BUSY because that prompt was executing. This produced an outcome mismatch: the caller was told nothing was queued, but a render was actually running.

### Steps to reproduce
1. Open a workflow with several output branches, including branches disabled through panel node modes.
2. Call panel_run when ComfyUI is idle.
3. Observe panel_run return a refusal such as `ComfyUI refused to queue the workflow - VAEDecode ... Required input is missing (samples)`.
4. Immediately inspect the ComfyUI queue; a running prompt is present.
5. Subsequent panel graph calls are rejected with `QUEUE BUSY`.

### Exact error (scrubbed)
`Error: ComfyUI refused to queue the workflow - VAEDecode (node 36): Required input is missing (samples); VAEDecode (node 41): Required input is missing (samples); VAEDecode (node 57): Required input is missing (samples); VAEDecode (node 60): Required input is missing (samples); VAEDecode (node 63): Required input is missing (samples); VAEDecode (node 66): Required input is missing (samples)`
Immediately afterward, queue list reported `running: 1`, prompt_id `b01ba287-500f-418d-b2e1-353c891065f1`, `pending: 0`.

### Fix
Applied locally: no. The issue appears to be an upstream panel/orchestrator queue outcome race; no safe source checkout for the running queue bridge was available in the workspace. Workaround: inspect the queue before retrying or editing, then wait for the running prompt to finish.
upstream-only: yes

### Environment
OS: Windows 11
ComfyUI: 0.28.0
ComfyUI frontend: 1.45.21
GPU: AMD Radeon 780M Graphics, 12.2 GB VRAM
PyTorch: 2.12.0a0+rocm7.13.0a20260313
comfyui-mcp: 0.52.16
comfyui-mcp-panel: 0.15.10

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-luna` · effort `high` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
## What was wrong

`panel_run` answered

```
ComfyUI refused to queue the workflow
- VAEDecode (node 36): Required input is missing (samples)   … ×6
```

while the ComfyUI queue showed `running: 1` for prompt `b01ba287-…` with `pending: 0`.
Every graph read and edit that followed came back `QUEUE BUSY` — blocked by the very
render the caller had just been told did not exist.

## Root cause

Both statements came out of **one** reply.

`execution.validate_prompt` validates each output independently, collects `good_outputs`,
logs `"Output will be ignored"` for the rest, and returns `True` whenever at least one
output survives. `server.py` then takes the `if valid[0]:` branch — `prompt_queue.put(...)`,
then

```python
response = {"prompt_id": prompt_id, "number": number, "node_errors": valid[3]}
return web.json_response(response)     # status 200
```

The reported workflow had several output branches with some disabled through node modes,
so its bypassed `VAEDecode` branches were exactly the outputs ComfyUI dropped from a prompt
it had queued and started.

The panel could not tell the two cases apart, because the frontend funnels both channels
into one field — `app.queuePrompt` calls `recordNodeErrors(res.node_errors)` on the
**resolved (200)** response too, so `app.lastNodeErrors` is populated for a prompt that is
on the GPU right now. Reading only that field, `summarizePromptRejection` returned
`{queued:false, node_errors}` and dropped the prompt id on the floor.

**That last part is why the existing repair never fired.** The orchestrator has handled this
exact shape since mcp v0.50.2 (mcp#944): `prompt_id` + `node_errors` with no top-level
`error` is treated as an accepted partial run and rendered as a loud
`[PARTIAL] ComfyUI ACCEPTED this prompt and is running it, but it dropped N output(s)` note.
It keys on the prompt id — which the panel never sent alongside the errors, so the whole
mechanism was unreachable from here and the reply fell through to the plain refusal.

## The fix

Read the 200 body, which carries both facts as one structured receipt:

- **capture** — a 200 reply's non-empty `node_errors` is forwarded as ACCEPTED drops, on
  both dispatch paths (the unscoped interceptor and the scoped run-to-node guard). This has
  to come from the body, not `lastNodeErrors`, which cannot distinguish a refusal from an
  accepted partial run at all.
- **verdict** — `summarizePromptRejection` takes the prompt ids ComfyUI **minted**. `queued`
  is a flag the panel sets; a prompt id is an identifier ComfyUI issues only after
  `prompt_queue.put(...)`. An id present means work was accepted, so the per-node errors
  beside it name dropped outputs, not a refusal.
- **disclosure** — the accept result reports those dropped outputs verbatim, so the caller
  is told which outputs will produce no file. Silence would be the opposite error: it would
  leave them waiting on a render that was never going to write them.

### What does not regress

- **#358** (a top-level rejection swallowed by `app.queuePrompt`) is a 400, which mints
  nothing — the id list is empty and the verdict is untouched.
- A top-level error arriving **with** a minted id stays a refusal — the two claims
  contradict each other and this is not the place to pick a side — but the result now
  carries the id, which the orchestrator turns into its
  `[UNCERTAIN] … a render may already be in flight, do NOT re-run` answer instead of a flat
  "refused" that invites a duplicate render.

## Verification

**Verified by mutation, not by a green suite.** Eight mutations, each applied to the
working tree and reverted from an in-memory backup (never `git checkout`):

| mutation | result |
| --- | --- |
| verdict rule: a minted id no longer outranks `node_errors` | KILLED |
| accept result: dropped outputs no longer reported | KILLED |
| capture: unscoped 200 `node_errors` never forwarded | KILLED |
| capture: scoped 200 `node_errors` never forwarded | KILLED |
| wiring: verdict no longer receives the minted ids | KILLED |
| wiring: accept no longer receives the dropped outputs | KILLED |
| wiring: unscoped path no longer captures accepted `node_errors` | KILLED |
| wiring: scoped path no longer captures accepted `node_errors` | KILLED |

Because a one-line wiring change is invisible to a helper-level test, the call site is
pinned too: the WIRING test asserts `onAcceptedNodeErrors: captureAcceptedNodeErrors`
appears on **both** dispatch paths and that `acceptedPromptIds` / `droppedOutputs` reach
the verdict and the accept result. Removing any one of them turns it red while every
helper-level test stays green.

Upstream behaviour was read out of the installed ComfyUI (`execution.py`, `server.py`) and
the installed frontend bundle (`comfyui_frontend_package` 1.49.6, `recordNodeErrors` on the
resolved response), not assumed.

`npm run test:unit`: 5966 tests, 5965 pass. The single failure is the known
`manager-install` wall-clock flake (#671) — it passes 125/125 when that file is run alone,
and this change touches nothing it exercises. `check-panel-scope` reports OK (every name in
the monolith resolves).

Closes #1504

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project are documented here. This project adheres to
 ## [Unreleased]
 
 ### Fixed
+- a run ComfyUI accepted while dropping some outputs is reported as queued, with the dropped outputs named, instead of as a refusal (#1504)
 - `panel_list_nodes` with `search` (or `query`) filters the installed-pack list instead of returning every pack (#1496)
 
 ## [0.15.25] - 2026-08-20
--- a/browser_tests/unit/queue-rejection.test.mjs
+++ b/browser_tests/unit/queue-rejection.test.mjs
@@ -252,3 +252,184 @@ test("WIRING: graph_run passes the run-to-node target into the summary", async (
   assert.ok(src.includes("runToNode: runToNodeInfo,"),
     "summarizePromptRejection must receive the run-to-node target");
 });
+
+// ── #1504: a MINTED prompt_id outranks node_errors ─────────────────────────
+//
+// panel_run answered
+//
+//   ComfyUI refused to queue the workflow
+//   - VAEDecode (node 36): Required input is missing (samples)   … ×6
+//
+// and the ComfyUI queue showed `running: 1`, prompt b01ba287-…, `pending: 0` at the
+// same instant. Every following graph read/edit was then rejected with QUEUE BUSY —
+// by the very render the caller had just been told did not exist.
+//
+// Both statements came from ONE reply. ComfyUI validates each output independently
+// (execution.validate_prompt); when some fail but at least one survives it logs
+// "Output will be ignored", returns valid[0]=True, and server.py takes the
+// `if valid[0]:` branch — prompt_queue.put(...), then
+//
+//     web.json_response({"prompt_id": …, "number": …, "node_errors": valid[3]})
+//
+// at status **200**. The reported workflow had several output branches with some
+// disabled through node modes, so its bypassed VAEDecode branches are exactly the
+// ones ComfyUI dropped.
+//
+// The panel could not tell the two apart because the frontend funnels both channels
+// into one place: app.queuePrompt calls recordNodeErrors(res.node_errors) on the
+// RESOLVED (200) response too, so app.lastNodeErrors is populated for a prompt that
+// is on the GPU. The verdict now takes the minted prompt_id — a receipt ComfyUI
+// issues only after queueing — over that ambiguous channel.
+
+// The six bypassed VAEDecode branches from the report, in ComfyUI node_errors shape.
+const DROPPED_VAE_DECODES = Object.fromEntries(
+  [36, 41, 57, 60, 63, 66].map((id) => [
+    String(id),
+    {
+      class_type: "VAEDecode",
+      dependent_outputs: [],
+      errors: [
+        { type: "required_input_missing", message: "Required input is missing", details: "samples" },
+      ],
+    },
+  ]),
+);
+const RUNNING_PROMPT_ID = "b01ba287-500f-418d-b2e1-353c891065f1";
+
+test("#1504: node_errors beside a MINTED prompt_id is an accepted partial run, NOT a refusal", () => {
+  const verdict = summarizePromptRejection({
+    rejection: null, // 200 — nothing was captured on the non-200 rejection channel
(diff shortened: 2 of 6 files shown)


artokun:
Codex gate (read-only): PASS for merge. The accepted-200 prompt body is captured on both unscoped and scoped paths; minted prompt IDs outrank stale node_errors for the verdict, while dropped outputs are disclosed in the accepted result. The new tests pin the production wiring and rejection regressions. Panel CI is green. The known isolated manager-install wall-clock flake is documented by the agent and is unrelated to this diff.

Issue #1282, opened 2026-08-16
panel_set_widget does not refresh dynamic input slots


artokun:
### What happened / root cause
panel_set_widget changed ImageBatchMulti.inputcount from 2 to 5 (and 3), and reported success, but the live node still exposed only image_1 and image_2. The custom node requires its Update inputs callback to rebuild slots; the panel mutation updates the stored widget value without triggering the frontend widget callback.

### Steps to reproduce
1. Add KJNodes ImageBatchMulti through panel_add_node.
2. Call panel_set_widget on inputcount with value 5.
3. Read the node with panel_query_graph fields:detail.
4. Observe widgets.inputcount == 5 but inputs remain image_1, image_2, inputcount.

### Exact error (scrubbed)
No thrown error; this is a silent partial mutation. panel_set_widget returned success while the graph topology was unchanged.

### Fix
applied locally: no
upstream-only: yes
panel_set_widget should invoke the target widget's callback after assigning a value (matching an interactive frontend edit), or add a safe panel widget/button invocation path so dynamic-input nodes can refresh their slots. As a temporary workaround, the workflow was rebuilt with a fixed chain of core ImageBatch nodes.

### Environment
Windows 11
ComfyUI 0.33.1
ComfyUI frontend 1.48.7
Python 3.13.12
PyTorch 2.12.1+cu130
GPU NVIDIA GeForce RTX 4080 16 GB
comfyui-mcp 0.51.17
panel 0.14.16

Issue #1268, opened 2026-08-15
panel_set_widget reports success but custom preprocessor widget keeps its previous value


artokun:
### What happened / root cause
While building a live workflow, panel_set_widget was used to change an AIO preprocessor from DepthAnything to DWPose. The mutation reported success, but panel_query_graph readback showed the widget still held DepthAnything, and the rendered preview was a depth map. The panel appears to accept/acknowledge a custom widget assignment without verifying that LiteGraph persisted the requested normalized option.

### Steps to reproduce
1. Add an installed AIO auxiliary preprocessor node.
2. Set its preprocessor selector to DWPose with panel_set_widget.
3. Read the node back with panel_query_graph fields:detail.
4. Observe that the old DepthAnything value remains even though the setter succeeded.

### Exact error (scrubbed)
No error was returned; this was a silent wrong-success result.

### Fix
Applied locally: no. Upstream-only: yes.
After setting a widget, round-trip the final LiteGraph value and fail clearly when it does not equal the requested normalized option. Return the available enum options when normalization fails. Add coverage for custom/DOM combo widgets. Workaround verified in-session: replace the AIO node with the dedicated DWPreprocessor node, which produced the correct multi-person skeleton.

### Environment
Linux; ComfyUI 0.27.0; ComfyUI frontend 1.45.20; Python 3.14.6; PyTorch 2.12.1+rocm7.2; AMD Radeon RX 9070 XT 16 GB; comfyui-mcp 0.51.56; panel 0.14.42.

artokun:
Claimed by the backlog-drain session as part of a FAMILY (worktree `.claude/worktrees/drain-widgetwrite`). panel#1268 and mcp#1658 were identified by a backlog audit as one root cause: `applyWidgetWrite` (`widget-write.js:1605`) verifies synchronously against `w.value` only, and never reads `node.properties`, the DOM, or `serializeValue`. Assigning one agent to the shared cause rather than two writing two guards. If no PR appears within ~8h, consider the claim lapsed.

artokun:
[...]

Issue #1254, opened 2026-08-15
Sequential panel_add_node calls trip root-shape-mismatch after first successful add


artokun:
### What happened / root cause
While editing a bound live workflow, the first panel_add_node succeeded and increased the canvas from 27 to 28 nodes. The immediately following sequential panel_add_node calls were rejected with root-shape-mismatch because the workflow fence still expected 27 nodes. The panel treated its own just-applied mutation as evidence of a different graph.

### Steps to reproduce
1. Bind to a saved live workflow.
2. In one sequential tool orchestration, await panel_add_node once, then await panel_add_node again.
3. The first succeeds; the second reports workflow 27 vs canvas 28 and refuses.

### Exact error (scrubbed)
[root-shape-mismatch] The live graph is out of sync with the active workflow: the workflow reports 27 node(s) but the live canvas holds 28.

### Fix
applied locally: no  upstream-only: yes
After a successful panel mutation, refresh/update the session's expected live graph shape before validating the next sequential mutation, or key the fence to workflow identity/revision rather than stale node count.

### Environment
Windows 11; ComfyUI 0.33.0; GPU RTX 3090 24GB; comfyui-mcp 0.51.54; panel 0.14.41; ComfyUI-Manager 3.41 legacy.

artokun:
Additional reproduction on current versions (`comfyui-mcp` 0.51.56, panel 0.14.43; Windows 11 / ComfyUI 0.33.1 / frontend 1.48.7): after `panel_open_workflow` returned the normalized-content UNCONFIRMED result without a `workflow_uuid`, the first `panel_add_node` visibly changed the canvas from 27 to 28 nodes while `activeWorkflowCurrentState(...).nodes` stayed at 27 and `isModified` stayed `false`.

The resulting root-shape mismatch rejected subsequent read-only graph calls as well as mutations. More seriously, `panel_save_workflow` returned `{"saved":true}` while persisting the stale 27-node state; reopening then lost the visually added node. Re-opening did not recover the state, while a frontend reload restored normal mutation tracking.

artokun:
Closing as fixed: the fence leg is covered by #1326 (merged, ships in 0.15.2 — the containment disjunct; a panel-applied add is observationally a structural extension, verified pre/post with a repro of the exact two-add sequence), and the save leg by #878 (v0.14.44 — in-place save captures before overwriting, refuses loudly if the capture throws). The no-uuid-anchor residual is #1014's deliberate design constraint.

PR #1527, opened 2026-08-20
fix(1519): a programmatic widget write announces itself to the node, not just to the widget

### LINKED ISSUE #1519: panel_set_widget never fires node.onWidgetChanged, so dynamic-slot nodes (SWF_Subworkflow) silently keep stale slots

`panel_set_widget` writes the value and fires the widget's own `widget.callback`, but it
never fires the NODE-level LiteGraph hook `nodeType.prototype.onWidgetChanged(name, value,
oldValue, widget)`. Custom nodes that rebuild their dynamic slots on a widget change hook
`onWidgetChanged`, not the widget callback — so a programmatic write leaves the node with
the new widget value but the OLD (or empty) input/output slots, silently.

Concretely: `comfyui-subworkflow`'s `SWF_Subworkflow` node builds its `swf_in_*` / `out_*`
slots by fetching `/subworkflow/info` for the selected child workflow. Its
`beforeRegisterNodeDef` patches exactly three hooks — `onConfigure`, `onWidgetChanged`,
`onAdded` (custom_nodes/comfyui-subworkflow/js/subworkflow.js ~L300-400). `panel_add_node`
creates it with an empty `workflow` widget, so `onAdded` fetches nothing; then
`panel_set_widget(node, "workflow", "<child>.json")` sets the value but fires no
`onWidgetChanged`, so the fetch never happens and the node keeps the generic
`out_0..out_7` / no `swf_in_*` shape. Every `panel_connect` to a real slot then fails,
because the slot does not exist.

Grepping the panel bundle, `onWidgetChanged` appears NOWHERE in
`comfyui-agent-panel/web/js/**` — the hook is simply never invoked. This is not specific
to comfyui-subworkflow; any pack whose slot topology is driven by `onWidgetChanged` is
affected the same way.

### Steps to reproduce

1. Install `comfyui-subworkflow` and have any child workflow with `SWF_SubworkflowInput` /
   `SWF_SubworkflowOutput` boundary nodes on disk.
2. `panel_new_workflow`
3. `panel_add_node {class_type: "SWF_Subworkflow"}`
4. `panel_set_widget {node_id: <id>, widget: "workflow", value: "<child>.json"}`  → reports
   `{previous: "", value: "<child>.json"}`, i.e. success.
5. `panel_query_graph {ids:[<id>], fields:"detail"}` → widget is correct, but inputs are
   still only `workflow` + `reload_each_execution`, and outputs are the generic
   `out_0..out_7` with type `*`. No `swf_in_*` slots. Re-querying later does not help; it
   is not a race.
6. `panel_connect` to `swf_in_0` fails — the slot does not exist.

The server side is fine and proves the value is usable:
`GET /subworkflow/info?workflow=<child>.json` returns the correct inputs/outputs.

### Exact error (scrubbed)

No error is raised anywhere — this is a SILENT failure. `panel_set_widget` returns success
and `panel_query_graph` shows a node that looks merely under-configured rather than broken.
That is what makes it expensive: the natural read is "this pack is incompatible with the
panel", when in fact one hook call is missing.

### Workaround that does work

Save the workflow and re-load it from disk:

    panel_save_workflow {name: "..."}
    panel_load_workflow {path: "....json"}

`onConfigure` DOES fire on load, the pack re-fetches `/subworkflow/info`, and the slots
materialise correctly. So the build has to be split into "add every node + set every
widget" -> "save" -> "reload" -> "then wire everything", instead of the normal
add/set/connect flow. It works, but it forces a save+reload in the middle of a build and it
is not discoverable — and on a persisted workflow it walks straight into the
add-node-after-open wedge (comfyui-mcp-panel#1254).

### Fix

<applied locally: no>  <upstream-only: no, but see below>

The fix is one guarded call: after the write is COMMITTED and verified (i.e. not on a path
that rolls back), invoke the node-level hook the same way the frontend does for an
interactive edit:

```js
// after the widget.callback invocation and the write verification, before returning
try {
  node.onWidgetChanged?.(w.name, coerced, previous, w);
} catch (e) {
  // same containment as the widget.callback invocation: a throwing hook must not
  // fail the write, but it must be disclosed in the result.
}
```

I did NOT apply this locally, deliberately, and want to flag why rather than hand you a
patch I could not defend. `web/js/lib/widget-write.js` wraps the single
`widget.callback` invocation in a large amount of rollback machinery — "INVOKED EXACTLY
ONCE per call, by design" (~L176), pre-callback expected-value cloning (~L1522),
prior-value deep clones and structural rollback verification (~L1540-1560), display-proxy
rollback (#477), promoted subgraph inner/outer coherence (#583). Dropping a second hook
invocation in without honouring those semantics risks the worst outcome for this bug's
class: slots rebuilt against a value that was subsequently ROLLED BACK. Getting the
ordering right (fire only after commit; on rollback either suppress it or re-fire with the
restored value) is a call for someone who owns that file's invariants. So: root-caused and
reproducible, with the change above as the intended shape, not a blind patch on a user's
live install mid-session.

Suggested test: any node type registering `onWidgetChanged` in `beforeRegisterNodeDef`
should see it fire for `panel_set_widget` exactly as it does for a canvas edit, and should
NOT see it fire for a write that was rolled back.

### Environment

OS: Windows 11 Pro N 10.0.26200
ComfyUI: 0.33.2 | frontend 1.49.6 | Python 3.13.12 | PyTorch 2.10.0+cu130
GPU: NVIDIA GeForce RTX 5090, 32 GB VRAM (194 GB RAM)
comfyui-mcp: 0.52.27
panel: 0.15.14
Third-party pack involved (not at fault): comfyui-subworkflow

artokun:
Fixes #1519.

## What was wrong

`panel_set_widget` assigned the value, drove the bound property (#1268) and fired the
widget's own `callback` — and stopped there. The NODE-level litegraph hook
`node.onWidgetChanged(name, value, prevValue, widget)` never ran. `onWidgetChanged`
appeared **nowhere** in `web/js/**` before this PR:

```
$ git grep -n onWidgetChanged origin/main -- .
(no output)
```

That hook is how a pack rebuilds slot **topology** from a widget, and it cannot use the
widget callback instead: `beforeRegisterNodeDef` patches the node *type's* prototype,
which is the only hook that survives the widget being rebuilt.

The reported case is `comfyui-subworkflow`'s `SWF_Subworkflow`: it fetches
`/subworkflow/info` from `onWidgetChanged` and builds its `swf_in_*` / `out_*` slots from
the answer. `panel_add_node` creates it with an empty `workflow` widget, so its `onAdded`
fetched nothing; `panel_set_widget` then set the value and fired no node hook, so the
fetch never happened. The write reported success — honestly; the VALUE was verified
present — and every later `panel_connect` failed because the slot did not exist. Nothing
raised an error anywhere, which is what made it expensive: the natural read was "this
pack is incompatible with the panel", and the discovered workaround was a save + reload
mid-build (only `onConfigure` fires on load).

## The mechanism, read from the frontend rather than assumed

comfyui-frontend-package **1.48.7**, `BaseWidget.setValue` (deminified):

```js
this.value = a
this.options?.property && … && node.setProperty(this.options.property, a)
this.callback?.(this.value, canvas, node, pos, e)
node.onWidgetChanged?.(this.name ?? "", a, prev, this)   // <- the missing step
node.graph && node.graph.incrementVersion()
```

and the canvas mouse path, `processNodeWidgets`:

```js
if (prev != w.value) { node.onWidgetChanged?.(w.name, w.value, prev, w); … }
```

`widget-write.js` already performed the assignment, the `setProperty` copy-back and the
callback **in exactly that order**. The node hook was the one step missing.

### This is not a rare path — every node on a real page carries the hook

`installNodeHooksRecursive` (same bundle, `GraphView-*.js`) wraps `onWidgetChanged` on
**every node in the graph** on attach, and again from `onNodeAdded`:

```js
e.onWidgetChanged = Mc(e.onWidgetChanged, function (t, n, r, i) {
  if (!Z.rootGraph) return
  let a = is(Z.rootGraph, e); if (!a) return
  let o = { min: i.options?.min, max: i.options?.max }
  …
  ia().clearWidgetRelatedErrors(a, t, i.name, n, o)
})
```

Two consequences a reviewer should weigh, both deliberate:

1. A panel write now clears widget-related validation errors the way an interactive edit
   does. That is consistent with what the handler already wants — `graph_set_widget`
   clears the sticky `has_errors` red flag after a write (#418) — but it is a real
   behaviour change on ordinary nodes, not just on packs with dynamic slots.
2. The wrapper dereferences the **fourth** argument (`i.options`, `i.name`), so the
   argument shape is load-bearing: handing it `undefined` would make every panel write on
   a real page throw and newly report `widget_changed_hook_failed`. Two tests transcribe
   that wrapper's body and pin the clean outcome.

## Where it fires — the whole of the ordering question the report raised

After the verification verdict, at the success path, where **every failure path has
already rolled back and thrown**. So the hook cannot run for a write that did not stick:
a pack rebuilt against a value that was subsequently restored is worse than the stale
slots this fixes. **None of the rollback machinery is touched** — the new call is
reachable only from the committed path.

Three further decisions, each stated in the source:

- **Fired once**, on the node/widget pair whose callback the write fired — the inner
  target on a definition-scoped promoted write, the rail on an instance-scoped one.
  Announcing the other end would report a change for a widget whose value did not move,
  which is `widget-write.js`'s own reasoning for firing one semantic callback.
- **Handed the VERIFIED value** read back off the widget, not `coerced`. Where a widget's
  own grid normalized the write (#805) the two differ, and handing a pack a value the
  widget does not hold is how a rebuild lands on state nothing agrees with. The reply
  reports that same capture, taken *before* the hook ran, so a hook that re-touches the
  widget cannot put a value in the reply that nothing verified.
- **Not gated on the value having changed**, unlike the frontend's `setValue`. That gate
  does not exist for the callback invocation either, and adding one only here would make
  re-issuing the same `panel_set_widget` — the obvious way a caller recovers from exactly
  this bug — a silent no-op.

It never decides the write's verdict: a throwing hook is disclosed on the success result,
never thrown over it. Same containment the widget callback gets, same shape #1282's
refresh press uses.

**Undo cost, stated rather than waved at.** The hook runs in its own
`beforeChange`/`afterChange` bracket so slot rebuilds join the command's undo history —
and since every node on a real page has a hook, that bracket now opens on every write.
It does not add an undo entry for a hook that changes nothing: litegraph's
`captureCanvasState()` compares the serialized graph against `activeState` and pushes only
on a diff. What it does cost is one extra graph serialize + compare per `panel_set_widget`.

## What the reply reports

Both fields are emitted only on an **observation**, so every node without the hook replies
byte-identically to before:

- `widget_changed_slots` — the hook rebuilt the node's slots synchronously; the input/
  output names after it, so the caller can wire without a re-read.
- `widget_changed_hook_failed` — attempting to invoke the hook threw. The write is
  unaffected and was not rolled back; the state the hook rebuilds may be stale or
  partially rebuilt, and that is stated rather than left for a later `panel_connect` to
  discover.

A hook that ran and changed no slots reports **nothing**: a pack that rebuilds from a
`fetch` (which `SWF_Subworkflow` does) resolves after this returns, so a synchronous
snapshot showing no change means "not yet", not "nothing happened".

## Evidence — verified by mutation, not by a green suite

20 tests in `browser_tests/unit/node-widget-changed.test.mjs`, driving `applyWidgetWrite()`
(the function `graph_set_widget` delegates to) rather than the helper in isolation — the
defect was a *missing call site*, and a helper-level test passes just as happily with the
write path never calling it, which is the state `main` was in.

| mutation | result |
|---|---|
| call site disabled (`null && fireNodeWidgetChanged(…)`) | **16 of 20 red** (the 4 that stay green are the negative controls: no-hook, throwing accessor, helper-in-isolation, verified-reply capture) |
| hook moved ABOVE the rollback branch | **"a write that FAILS verification and rolls back does NOT fire the hook" red** |
| `coerced` passed instead of the verified read | **the #805 normalization test red** |
| reply re-reads the widget AFTER the hook | **"cannot rewrite the VERIFIED reply" red** |

The last one is worth a note: my first draft of that test replaced `node.widgets` inside
the hook and did **not** catch the mutation — the captured widget object was untouched, so
a re-read still returned the verified value. It was rewritten to mutate the widget in
place, and then it caught it.

Suites, on this branch rebased onto `a6a920f5`:

- `npm run test:unit` — **5982/5984**. The one failure is the known #671 manager-install
  wall-clock flake; re-run alone, `manager-install.test.mjs` is **125/125**.
- widget-write-adjacent suites together (`widget-write`, `set-widget-*`,
  `dynamic-inputs-refresh`, `promoted-value-scope`, `widget-bound-property`,
  `rgthree-lora-row`, plus the new file) — **413/413**.
- `npm run typecheck` clean; `check-panel-scope` and `check-tool-vocabulary` OK.

## Playwright: NOT run against this branch, and why

I could not give this a browser gate, and I would rather say so than imply coverage I do
not have.

The only ComfyUI running here is the shared instance on 8188. Its
`custom_nodes/comfyui-mcp-panel` junction points at the main checkout, which sits on an
unrelated branch — so a Playwright run from this worktree would have exercised that
branch's code, not this one. I repointed the junction at this worktree and **measured that
it changed nothing**: `js/lib/node-widget-changed.js` still 404'd and the served
`widget-write.js` still had zero occurrences of the new import, because aiohttp resolved
the static route at ComfyUI startup. The only way through is to restart the shared
instance, and with other sessions active on this machine (23 chrome processes at the time)
that would have killed someone else's in-flight run. The junction was restored to the main
checkout and re-verified serving.

What that leaves unverified in a real browser: that an ordinary `panel_set_widget` against
the frontend's own installed wrapper stays clean end to end. The two transcribed-wrapper
tests are the closest stand-in and they pass, but they are a transcription, not the real
frontend. The specs that would exercise this for real are
`reopen-preserves-node-set-values`, `save-persists-node-set-values`,
`dirty-guards-see-node-writes` and `duplicate-widget-rows-are-visible` — worth a run by
anyone who has a restartable instance.

- **No `graph.incrementVersion()`.** The frontend bumps it next to this hook; the panel
  uses its own `setDirty` / history bracketing, and adding a second dirty mechanism is a
  separate change with its own undo consequences.
- **No summary-line change.** `widget_changed_hook_failed` sits on the result the model
  reads, exactly where #1282 put `dynamic_inputs_refresh_failed`; the human activity line
  is untouched.
- **No await.** The hook fires synchronously at the write boundary. A pack whose rebuild
  is a `fetch` materializes its slots shortly after — the same way it does for an
  interactive edit — so the reporter's `panel_connect` step works on the next round trip,
  not within the same reply.
- **Nothing about #1254** (the add-node-after-open wedge the report mentions as the other
  half of its workaround). Separate issue, untouched.

**Review is pending** — requested from grok in a comment below. This is not gated and not
reviewed yet; no verdict is claimed.

### DIFF
--- a/browser_tests/unit/node-widget-changed.test.mjs
+++ b/browser_tests/unit/node-widget-changed.test.mjs
@@ -534,6 +534,35 @@ test("#1519: a widget with NO options does not break the wrapper's optional read
   assert.deepEqual(seen, [{ min: undefined, widgetName: "text" }]);
 });
 
+test("#1519: a node whose `inputs` is a THROWING accessor still gets its hook fired", () => {
+  // The slot snapshot is instrumentation, not a precondition. A node that defines its
+  // slots as accessors is exactly the kind of node this route exists for, and letting
+  // the BEFORE snapshot abort the route would leave behind the stale slots it fixes —
+  // dressed up as a machinery failure.
+  let fired = 0;
+  const node = {
+    id: 32,
+    type: "N",
+    widgets: [{ name: "w", type: "number", value: 1 }],
+    outputs: [],
+    onWidgetChanged() {
+      fired++;
+    },
+  };
+  Object.defineProperty(node, "inputs", {
+    get() {
+      throw new Error("slots are computed");
+    },
+  });
+
+  const set = applyWidgetWrite(node, "w", 5, HOOKS);
+
+  assert.equal(fired, 1, "the hook must fire even when the slots cannot be read");
+  assert.equal(set.value, 5);
+  assert.equal(set.widget_changed_hook_failed, undefined);
+  assert.equal(set.widget_changed_slots, undefined, "an unreadable snapshot claims nothing either way");
+});
+
 // ---- the helper's own contract ---------------------------------------------
 
 test("#1519: fireNodeWidgetChanged never throws, whatever it is handed", () => {
--- a/browser_tests/unit/node-widget-changed.test.mjs
+++ b/browser_tests/unit/node-widget-changed.test.mjs
@@ -468,6 +468,72 @@ test("#1519: an INSTANCE-SCOPED promoted write announces on the WRAPPER, whose r
   assert.equal(widget, railWidget);
 });
 
+// ---- the hook EVERY node on a real page carries -----------------------------
+
+/**
+ * On a real page this hook is not exotic — the FRONTEND installs one on every node.
+ * `installNodeHooksRecursive` (comfyui-frontend-package 1.48.7, GraphView-*.js) runs
+ * over every node in the graph on attach and again from `onNodeAdded`, wrapping
+ * `onWidgetChanged` to clear widget-related validation errors:
+ *
+ *     e.onWidgetChanged = Mc(e.onWidgetChanged, function (t, n, r, i) {
+ *       if (!Z.rootGraph) return
+ *       let a = is(Z.rootGraph, e); if (!a) return
+ *       let o = { min: i.options?.min, max: i.options?.max }
+ *       let s = wa(Z.rootGraph, e, i)
+ *       s?.sourceExecutionId && ia().clearWidgetRelatedErrors(s.sourceExecutionId, …, n, o)
+ *       ia().clearWidgetRelatedErrors(a, t, i.name, n, o)
+ *     })
+ *
+ * So this change fires on ordinary writes to ordinary nodes, and the argument shape is
+ * load-bearing rather than cosmetic: the wrapper dereferences the FOURTH argument
+ * (`i.options`, `i.name`). Handing it `undefined` there would make every panel write on
+ * a real page throw and newly report `widget_changed_hook_failed`.
+ */
+test("#1519: the frontend's OWN installed hook receives what it dereferences — an ordinary write stays clean", () => {
+  const cleared = [];
+  const node = {
+    id: 30,
+    type: "KSampler",
+    widgets: [{ name: "steps", type: "INT", value: 20, options: { min: 1, max: 100 } }],
+    // The wrapper's body, transcribed from the bundle above.
+    onWidgetChanged(name, value, prevValue, widget) {
+      const bounds = { min: widget.options?.min, max: widget.options?.max };
+      cleared.push({ executionId: `node:${this.id}`, name, widgetName: widget.name, value, bounds });
+    },
+  };
+
+  const set = applyWidgetWrite(node, "steps", 30, HOOKS);
+
+  assert.equal(set.value, 30);
+  assert.equal(
+    set.widget_changed_hook_failed,
+    undefined,
+    "an ordinary write on an ordinary node must not newly carry a hook failure",
+  );
+  assert.equal(set.widget_changed_slots, undefined, "clearing an error moves no slots");
+  assert.deepEqual(cleared, [
+    { executionId: "node:30", name: "steps", widgetName: "steps", value: 30, bounds: { min: 1, max: 100 } },
+  ]);
+});
+
+test("#1519: a widget with NO options does not break the wrapper's optional reads", () => {
+  const seen = [];
+  const node = {
+    id: 31,
+    type: "N",
+    widgets: [{ name: "text", type: "customtext", value: "a" }],
+    onWidgetChanged(name, value, prevValue, widget) {
+      seen.push({ min: widget.options?.min, widgetName: widget.name });
+    },
+  };
+
+  const set = applyWidgetWrite(node, "text", "b", HOOKS);
+
+  assert.equal(set.widget_changed_hook_failed, undefined);
+  assert.deepEqual(seen, [{ min: undefined, widgetName: "text" }]);
+});
+
 // ---- the helper's own contract ---------------------------------------------
 
 test("#1519: fireNodeWidgetChanged never throws, whatever it is handed", () => {
--- a/browser_tests/unit/node-widget-changed.test.mjs
+++ b/browser_tests/unit/node-widget-changed.test.mjs
@@ -1,0 +1,482 @@
+/**
+ * #1519 — `panel_set_widget` never fired the NODE-level litegraph hook
+ * `node.onWidgetChanged(name, value, prevValue, widget)`, so a pack that rebuilds its
+ * slot topology from that hook kept the new widget value and its OLD slots, silently.
+ *
+ * These drive `applyWidgetWrite()` — the same function `graph_set_widget` delegates to
+ * — rather than the helper in isolation, because the whole defect was a MISSING CALL
+ * SITE: a helper-level test of `fireNodeWidgetChanged` passes just as happily with the
+ * write path never calling it, which is exactly the state main was in.
+ *
+ * Run with `node --test`.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { applyWidgetWrite, WidgetWriteError } from "../../web/js/lib/widget-write.js";
+import { fireNodeWidgetChanged, nodeWidgetChangedHook } from "../../web/js/lib/node-widget-changed.js";
+
+const HOOKS = {};
+
+/**
+ * The reported node, reduced to its mechanism: `comfyui-subworkflow`'s SWF_Subworkflow
+ * patches `onWidgetChanged` in `beforeRegisterNodeDef` and builds its `swf_in_*` /
+ * `out_*` slots from the selected child workflow. `panel_add_node` creates it with an
+ * empty `workflow` widget (so its `onAdded` fetches nothing), and the panel write then
+ * set the value while the slots stayed generic.
+ *
+ * The rebuild here is synchronous so the test can observe it; the real pack's is a
+ * `fetch`, which is why the result field is emitted only on an observed change.
+ */
+function makeSubworkflowNode({ onChange } = {}) {
+  const calls = [];
+  const node = {
+    id: 7,
+    type: "SWF_Subworkflow",
+    widgets: [
+      { name: "workflow", type: "combo", options: { values: ["", "child.json"] }, value: "" },
+      { name: "reload_each_execution", type: "toggle", value: false },
+    ],
+    inputs: [{ name: "workflow" }, { name: "reload_each_execution" }],
+    outputs: Array.from({ length: 8 }, (_, i) => ({ name: `out_${i}`, type: "*" })),
+    // Installed on the instance here; on the real pack it is on the node TYPE's
+    // prototype. Either way the lookup this file performs is `node.onWidgetChanged`.
+    onWidgetChanged(name, value, prevValue, widget) {
+      calls.push({ name, value, prevValue, widget, this: this });
+      if (onChange) return onChange.call(this, name, value, prevValue, widget);
+      if (name !== "workflow" || !value) return;
+      // What the pack does once `/subworkflow/info` answers: the boundary nodes of the
+      // selected child become real slots.
+      this.inputs = [
+        { name: "workflow" },
+        { name: "reload_each_execution" },
+        { name: "swf_in_0" },
+        { name: "swf_in_1" },
+      ];
+      this.outputs = [{ name: "out_0", type: "IMAGE" }];
+    },
+  };
+  return { node, calls };
+}
+
+// ---- the defect ------------------------------------------------------------
+
+test("#1519: a programmatic write FIRES the node's onWidgetChanged, so dynamic slots materialise", () => {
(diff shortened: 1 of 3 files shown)


artokun:
@grok please review this one. It is small (one call, one new module, two capture
variables) but it fires third-party code on a path that previously never did, so I would
rather point you at the load-bearing parts than hand you a diff.

**The claim I am least sure of, and the one I would attack first.** The frontend installs
`onWidgetChanged` on **every node** (`installNodeHooksRecursive` in `GraphView-*.js`, run
on attach and from `onNodeAdded`). So this is not a rare pack path — it now runs on every
`panel_set_widget` against every node. I convinced myself the blast radius is acceptable
by reading the wrapper's body and transcribing it into two tests, but I could **not** run
Playwright against this branch (reason and the measurement are in the PR body: repointing
the junction demonstrably did not change what 8188 serves, and restarting the shared
instance would have killed another session's run). So the one thing nobody has observed
in a real browser is: does an ordinary write to an ordinary node stay clean, or does every
reply now carry `widget_changed_hook_failed`? If you can reach a restartable ComfyUI, that
is the check worth spending it on.

**Second, the placement.** `web/js/lib/widget-write.js` — the fire sits after the
verification verdict and after `if (failure) { … throw }`, so it is unreachable from any
rollback path. That is the ordering the reporter explicitly asked for and I pinned it by
mutation (moving the call above the failure branch turns the rollback test red). What I
want checked is whether there is a path out of `applyWidgetWrite` that reaches the return
**without** the write having stuck — i.e. is `if (failure) { … throw }` genuinely the only
exit between the verification block and the return? I read it as yes; I would like a
second reading, because if it is not, the failure mode is a pack rebuilding its slots
against a value that was restored.

**Third, the promoted-write choice.** I fire ONCE, on `valueNode`/`valueWidget` — the
inner node on a definition-scoped promoted write, the wrapper's rail on an instance-scoped
one — mirroring the single semantic callback the file already fires. The frontend, on a
canvas edit of a promoted control, runs the projection's `setValue`, which fires the hook
on the **wrapper**. So on the definition-scoped path my choice and the frontend's diverge.
I think mirroring the callback is the coherent choice (announcing both ends would report a
change for a widget whose value did not move), but this is a judgement call and it is the
one I would most like a second opinion on.

**Fourth, an argument choice.** I pass the VERIFIED value read back off the widget, not
`coerced`. The frontend's `setValue` passes the assigned value; `processNodeWidgets` passes
the live one. They differ only when a widget normalized the write (#805). I chose the live
value so the hook and the reply cannot disagree — but I picked one of two frontend
precedents, and if you think `coerced` is the better match I would take that.

**Not worth your time** (already pinned by mutation, each named in the PR table): that the
call site is actually reached, that a throwing hook does not fail the write, that the reply
reports the pre-hook capture, and the disclosure wording.

artokun:
Codex independent merge gate: PASS for eligible P2 #1519.

Verified the production path in the literal diff: applyWidgetWrite invokes the node-level hook after the write has passed verification and rollback paths have exited, with the frontend argument order and the concrete node/widget receiver. Hook failures are contained and disclosed, synchronous slot changes are reported, and promoted writes announce only the widget whose value moved. The tests drive applyWidgetWrite rather than only testing the helper, including rollback, normalization, promoted scope, hook failure, and frontend-wrapper argument shape.

CI is fresh green: Lint pack completed successfully. No P0/P1 or enhancement scope is present.

artokun:
Self-audit found and fixed one defect before review, recording it so it does not have to be re-found:

`slotSnapshot` read `node.inputs` / `node.outputs` **outside** the try that walks them. A node defining its slots as accessors — exactly the kind of node this route exists for — would have thrown from the BEFORE snapshot, aborting before the hook ever fired, and reported a machinery failure while leaving the stale slots behind. That is the original bug's silent-staleness class wearing a disclosure. Both reads are now in their own try; an unreadable snapshot yields empty lists that compare equal, so it claims nothing either way.

Pinned by mutation: restoring the one-line form turns exactly one test red — *"a node whose `inputs` is a THROWING accessor still gets its hook fired"* — and nothing else.

Counts are now **21 tests** in the new file, **416/416** across the widget-write-adjacent suites, typecheck clean, CI green. The mutation table in the PR body was measured at 20 tests and is otherwise unchanged.

Also worth flagging from that audit, since it bounds what the change can do: between the verification verdict and the new call site there are exactly three exits, all `throw`, all inside `if (failure) { … }` — no `return` at all. So the hook is unreachable from every rollback path by construction, not by convention.

artokun:
Follow-up Codex gate: PASS remains valid for the final head 795c129.

The final hardening commit was reviewed: slot snapshot reads are independently contained, so an accessor that throws cannot prevent the node hook from firing; the new regression test covers that case. The required lint context is now recognized and passing.

Issue #1523, opened 2026-08-20
panel_add_node rejects valid loaded subgraph UUID and reports unrelated failed pack


artokun:
### What happened / root cause
On a live canvas containing the official core `Image Segmentation (SAM3)` subgraph (node type UUID `6e7ab3ea-96aa-470f-9b94-3d9d0e01f481`), `panel_add_node` with that exact class_type rejects it as an unknown backend node. The error then appends a misleading diagnosis that `comfyui-reactor-node` failed to import, even though ReActor is unrelated to the requested subgraph UUID and the existing subgraph executes from its loaded definition.

### Steps to reproduce
1. Load `blueprints/Image Segmentation (SAM3).json` with `panel_load_workflow`.
2. Confirm the root node exists and `panel_query_graph` reports `is_subgraph:true` with type UUID `6e7ab3ea-96aa-470f-9b94-3d9d0e01f481`.
3. Call `panel_add_node({class_type:"6e7ab3ea-96aa-470f-9b94-3d9d0e01f481"})`.

### Exact error
`Unknown node type "6e7ab3ea-96aa-470f-9b94-3d9d0e01f481" — the ComfyUI backend does not provide it ... ComfyUI reported that this pack FAILED TO IMPORT at startup — comfyui-reactor-node.`

### Fix
Applied locally: no. Upstream-only: yes. `panel_add_node` should recognize subgraph definitions already loaded in the live workflow (or explicitly direct callers to copy/paste for subgraphs). Its missing-type diagnostic must not attribute an arbitrary unrelated failed custom node pack unless ownership mapping links that pack to the requested class_type.

### Environment
Windows 11; ComfyUI 0.33.3; frontend 1.49.6; RTX 5070 12 GB; comfyui-mcp 0.52.27; panel 0.15.11.

---
Filed from the ComfyUI panel by **codex** · model _unreported_ — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #512

artokun:
Claimed by grok-swarm 2026-08-20T19:28:33Z — driving this in worktree `wt-1523` (branch `fix/1523-add-node-loaded-subgraph`). I will post the PR here. If this comment is older than 3h with no PR, treat as abandoned.

artokun:
PR: https://github.com/artokun/comfyui-mcp-panel/pull/1526

[...]

Issue #1496, opened 2026-08-20
panel_list_nodes rejects search parameter


artokun:
### What happened
Calling panel_list_nodes with {"search":"SeedVR2"} returned: `MCP error -32602: Input validation error: Invalid arguments for tool panel_list_nodes: Unrecognized key: "search"`.

### Steps to reproduce
1. Open a connected panel canvas.
2. Call panel_list_nodes with a search string.

### Expected
A documented/discoverable node listing search should accept a search/filter parameter, or the tool description/schema should state the accepted parameter.

### Environment
Windows 11; ComfyUI 0.33.2; frontend 1.49.6; RTX 5070 12GB; comfyui-mcp 0.52.24; panel 0.15.11.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-terra` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Claimed by grok-swarm 2026-08-20T18:52:54Z — driving this in worktree `wt-1496` (branch `fix/1496-list-nodes-search`). I will post the PR here. If this comment is older than 3h with no PR, treat as abandoned.

artokun:
Closing as **no code needed in this repository** — the rejection comes from a schema this repo does not own, and the tool is behaving as declared.

### Where the rejection actually happens

`panel_list_nodes` is defined in **comfyui-mcp**, not the panel. On `origin/main` (v0.52.27), `src/orchestrator/panel-tools.ts:14793`:

```ts
def(
  "panel_list_nodes",
  "List the custom-node packs currently installed in the user's ComfyUI (via the built-in Manager). Read-only.",
  {},                                                    // <- the zod shape: no parameters
  async (_args, ctx) => ctx.call({ cmd: "nodes_list" }, 20000),
),
```

That `{}` is the whole argument schema. `MCP error -32602: ... Unrecognized key: "search"` is that strict object rejecting the call **inside the MCP server, before anything is sent over the bridge**.

The panel's only involvement is the bridge handler `async nodes_list()` at `web/js/comfyui-mcp-panel.js:20937` (origin/main, panel 0.15.24), which takes no arguments and is never reached on a rejected call. Grepping panel source (vendor bundles excluded) for `Unrecognized key`, `Input validation error`, and `inputSchema` returns **zero hits** — the panel declares no MCP tool schemas and cannot emit or influence this error. No change in this repo can affect the reported symptom.

### Not a version lag

[...]

PR #1528, opened 2026-08-20
release: v0.15.26 — run-scope honesty, locator and group footprint fixes

artokun:
Cuts **0.15.26** from `main`.

Publishing is triggered by a push to `main` touching `pyproject.toml`, so **merging this publishes to the Comfy Registry**.

## Carries

- `panel_add_node` of a subgraph UUID already loaded in the live workflow no longer rejects it as an unknown backend node or names an unrelated failed pack (#1523)
- a run ComfyUI accepted while dropping some outputs is reported as queued, with the dropped outputs named, instead of as a refusal (#1504)
- `panel_list_nodes` with `search` (or `query`) filters the installed-pack list instead of returning every pack (#1496)
- a run-to-node no longer has its own branch vetoed by an uninstalled pack on a branch it excluded (mcp#1871)
- the panel's own pre-flight stops refusing before ComfyUI is asked, so that recovery can run (mcp#1871)
- a group is built from the same footprint its members are judged by, so requested collapsed nodes are included (mcp#1877)
- a locator that could not look no longer reports the node as being from another workflow (#1501)
- a refusal claims nothing about a final size it never read (#1872)

**Five of those eight had no CHANGELOG entry** — their commits do not touch `CHANGELOG.md` (`da4f7d76`, `1fa952f8`, `7ef4872e`, `dd7f37ac`, `aaafb55f`). I wrote entries rather than let them ship invisibly. If any wording misstates what a fix does, correct it here rather than after the release.

## Version sites

`scripts/set-version.mjs 0.15.26` updated all three, and I checked each rather than trusting the script's exit code — CI asserts they match, so a drift here is a red build:

- `pyproject.toml` → `version = "0.15.26"`
- `web/js/comfyui-mcp-panel.js` → `const PANEL_VERSION = "0.15.26";`
- `web/changelog.json` → 226 releases, `0.15.26` present with its `Fixed` section

Exactly **one** `## [0.15.26]` heading in `CHANGELOG.md`, and `## [Unreleased]` is left empty above it. Worth stating explicitly: `set-version.mjs` warns that a pre-numbered heading skips its promotion step and can leave a second heading if the script is re-run for the same version — that is how 0.14.31, 0.14.32 and 0.14.33 each ended up duplicated. This heading was written by hand, so **do not re-run `set-version.mjs 0.15.26` against this branch**.

## Note on provenance

I cut this because 0.15.25 shipped at 19:0xZ and eight fixes have accumulated since with no release cut. If another session already had a 0.15.26 in flight, say so and I will close this rather than race it.


artokun:
Codex release gate: PASS for panel v0.15.26.

Verified the literal release diff:
- pyproject.toml, the bundled PANEL_VERSION, and the changelog entry all move from 0.15.25 to 0.15.26.
- CHANGELOG has one new 0.15.26 Fixed section above the prior release; the carried entries are bug fixes, with no P0/P1 or enhancement item in the release manifest.
- The release PR changes only release metadata/changelog files; the required lint pack is fresh green.

This is the authorized release boundary between backlog swarms.

Issue #1515, opened 2026-08-20
Graph writes refused after in-place pack update: tab advertises no version; ENV line reports panel 0.7.3 while disk is 0.15.24


artokun:
### What happened / root cause

All `panel_*` graph WRITES were refused mid-task with the "does not enforce per-command workflow targeting / this tab advertised NO panel version" error. Reads (`panel_graph_outline`, `panel_query_graph`, `panel_enter_subgraph`) worked throughout.

The error text attributes this to browser HTTP caching ("The panel's module URLs carry no cache-busting key, so an ordinary reload can serve the stale file again"). In this session that does not appear to be the actual trigger, and two things look worth fixing:

**1. The real trigger was an in-place pack update, not an HTTP cache.** From `/internal/logs`:

- `11:11:54` — ComfyUI 0.33.2 starts; `[comfyui-mcp-panel] agent panel routes registered`. Browser tab opened around here.
- `11:37:49` — `Updating: <pack>` … `After restarting ComfyUI, please refresh the browser.` (ComfyUI-Manager updated the pack in place.)
- No ComfyUI restart and no tab reload after that point.

So the tab had simply been open since *before* the pack was updated. `__init__.py` already sets `Cache-Control: no-store` on `/extensions/` assets (and notes ComfyUI 0.31.1+ does too), so an ordinary reload should pick up new JS fine — the tab just never reloaded. The current message sends the user to `Cmd+Shift+R` and to consider `install_comfyui(action:'panel', panel_action:'update')`, neither of which names the cause that actually applied. Suggest adding "the panel pack was updated in place while this tab was open" as the *leading* listed cause, since ComfyUI-Manager's own post-update line already tells users to refresh.

**2. Version reporting is inconsistent across three surfaces.** In the same session:

- Agent ENVIRONMENT line: `panel 0.7.3`
- `pyproject.toml` `version` and `PANEL_VERSION` in `web/js/comfyui-mcp-panel.js`: `0.15.24`
- Write-fence check: tab advertised **no version at all**

`0.7.3` vs `0.15.24` is a large gap and it is unclear which surface produced it. If the ENV line is reading the connected tab's advertised version, it should not disagree with the fence's "advertised nothing"; if it is reading disk, it disagrees with `pyproject.toml`. Either way an agent reasoning about panel capability gets three different answers.

[...]

PR #1530, opened 2026-08-20
fix(1516): a timestamp-less transcript keeps the order it was written in

### LINKED ISSUE #1516: Hard-refresh of ComfyUI tab replays the user's earlier chat messages in the panel

After a cache-bypassing reload of the ComfyUI tab (Cmd+Shift+R on macOS — done to pick up a freshly in-place-updated panel bundle, see #1515), the user reported: "refreshing the page seems to have repeated my initial messages to you."

What the agent observed from its side: NO duplicate user prompts arrived in the agent's context. The first post-reload turn carried the normal workflow-binding banner ("this message was sent from a different workflow than the previous one — you are now operating on 'image_krea2_turbo_t2i'") and a single new user message. So the replay appears to be a PANEL-UI transcript-rendering symptom (earlier user messages re-rendered / duplicated in the chat pane on reconnect), not a re-submission to the agent — though I can't rule out that they were re-sent and de-duplicated before reaching me.

Likely area: session transcript rehydration on reconnect after reload — the panel replays the journaled history into the chat pane and either (a) appends it to a DOM that already had it from localStorage/sessionStorage, or (b) replays the user-side entries twice (once from local cache, once from the orchestrator's journal).

Context that may matter: the reload crossed a panel JS version boundary — the tab was running a pre-update bundle (advertised no version; the ENV line said 0.7.3) and came back on 0.15.24. If the transcript cache schema changed between those versions, a migration/dedupe mismatch would explain a one-time replay.

### Steps to reproduce

1. Open ComfyUI with the panel; hold a conversation of a few turns (here: ~4 user turns).
2. Update the panel pack in place via ComfyUI-Manager (no ComfyUI restart).
3. Hard-refresh the ComfyUI tab (Cmd+Shift+R).
4. Observe the panel chat: the user's earlier messages appear again (per user report).

Not yet confirmed whether a plain reload on an unchanged bundle reproduces it, or whether it's specific to the cross-version reload.

### Exact error (scrubbed)

No error text — UI symptom only. User's words: "Strange, refreshing the page seems to have repeated my initial messages to you."

### Fix

applied locally: no
upstream-only: yes

Suggested: on reconnect, dedupe transcript entries by (message id / sequence number) before rendering, or clear the locally cached transcript when the orchestrator journal is authoritative. If the cached transcript schema is versioned, bump + migrate across the 0.7.x → 0.15.x boundary.

### Environment

- OS: macOS (Darwin 25.6.0), Apple Silicon, 128 GB unified
- ComfyUI: 0.33.2
- ComfyUI FRONTEND: 1.49.6
- Python 3.13.12 / PyTorch 2.13.0 / GPU: mps
- comfyui-mcp: 0.52.27
- comfyui-mcp-panel: 0.15.24 (after reload; tab was on a pre-update bundle before)
- Backend: Claude
- Related: #1515 (write fence refused on the stale pre-update tab; this reload was the remedy)

---
Filed from the ComfyUI panel by **claude** · model `claude-fable-5[1m]` · effort `xhigh` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #98

artokun:
Fixes #1516

## What the user saw

After a cache-bypassing reload that crossed a panel bundle boundary (the tab had
been open since before an in-place ComfyUI-Manager update — see #1515), the user
reported: *"refreshing the page seems to have repeated my initial messages to you."*
The agent confirmed nothing was re-sent: no duplicate prompts reached its context.
So the issue filed this as a transcript-**rendering** symptom, and that is exactly
what it is.

## Mechanism

A pre-0.11.0 panel — everything before IndexedDB and the atomic snapshot key
existed, which is the era of a tab that advertises **no** `panel_version` — wrote a
**bare array** of threads under `comfyui-mcp.panel.threads`, and its messages
carried no `id`, no `createdAt` and no `ts`.

On the next mount the current panel normalizes those records: it floors
`createdAt` at `1` (`normalizeMessage`) and mints a synthetic
`legacy-<stableHash(threadId:ordinal:canonicalJson(message))>` id. Restore then
merges its own mount-time read with the copy `ChatHistoryStore.load()` resolves
to (`comfyui-mcp-panel.js`, `historyRestoreReady`):

```js
const merged = mergeHistorySnapshots({ threads, meta: historyMeta }, loaded)
```

`mergeThreadMessages` sorts that union:

```js
finiteTs(a.createdAt || a.ts) - finiteTs(b.createdAt || b.ts) ||
String(a.id).localeCompare(String(b.id))
```

With every message pinned at `createdAt: 1` the timestamp comparison is a dead
heat **for the whole thread**, so the tiebreak decides — and for these records the
tiebreak is a hash of the message's own text. The conversation is dealt back in
content-hash order.

## Evidence — reproduced, not reasoned

Seeded the exact pre-0.11 shape at document start and loaded the real panel in
Chromium (`browser_tests/legacy-transcript-order.spec.ts`). What the sidebar
painted, verbatim from the run, **before** the fix:

```
stored   u1  a1  u2  a2  u3  a3
painted  a3  u1  u2  a1  a2  u3
```

Rendered chat pane before:

```
legacy reply 2 / legacy first 1516 / legacy second 1516 / legacy reply 0 / legacy reply 1 / legacy third 1516
```

after:

```
legacy first 1516 / legacy reply 0 / legacy second 1516 / legacy reply 1 / legacy third 1516 / legacy reply 2
```

The user's first two prompts land **below the agent's last reply**. Scrolled to the
bottom of a fresh reload, that reads as your earlier messages showing up again.

## The fix

One tiebreak in `mergeThreadMessages`: rank by the position the message held in
the source arrays — the order those records were actually written in — before
falling back to the id. Timestamped messages are untouched, because their
comparison resolves first, so cross-tab causal interleaving still decides by time.

## Verification

- `npm run test:unit` — **6004 pass / 0 fail** (4 new).
- Playwright, `--workers=1` — the new spec passes against a real ComfyUI + panel.
- **Mutation-verified in three directions**, not just "the suite is green":
  - delete the rank tiebreak → the two order tests fail, the two control tests stay green, **and the browser spec fails** (`Expected - 2 / Received + 2`);
  - move the rank tiebreak *above* the clock → only `timestamped messages still interleave by TIME` fails, proving that guard is load-bearing and not decorative;
  - restore → all four green again.

- **The ordinal-shift identity hazard.** The synthetic legacy id embeds the
  message's ordinal, so two copies of one thread that disagree on ordinals mint
  different ids for the same message and the merge **duplicates** it. I reproduced
  that at module level (a 6-message thread merged with its own 4-message tail
  yields 10). It is already fenced in production by
  `mergeUnderCanonicalCheckpoint` — a flagged pre-v3 snapshot is quarantined
  wholesale against a fenced canonical, and replaced as a unit against an
  unfenced one — and I could not construct a production path that reaches the
  unguarded merge with shifted ordinals. Widening the fix to cover it would mean
  changing legacy message identity, which is a much larger change than this
  issue's evidence supports. Flagged here rather than silently dropped.
- **Nothing agent-facing.** No change to `armVisibleTranscriptReplay` /
  `buildReplayTranscript`; the issue's own agent-side evidence says nothing was
  re-sent, and my reading agrees — the panel has no server frame that replays a
  transcript into the chat pane.

## Unrelated observation from the harness (not fixed here)

The FIRST message sent on an unsaved canvas is refused by the #1095 route guard:
grounding auto-saves the workflow, the tab identity flips `tmp:` → `wf:` mid-dispatch,
and `sendUserMessage` returns false. Reproduced with frame logging; the second send
goes through. This currently breaks every e2e spec whose first action is
`sendMessage` (`pending-queue.spec.ts` fails on unmodified `main` on this box for
exactly this reason). Not this PR's scope, and this spec is deliberately agent-free
so it does not depend on it — but it is worth its own issue.

## Review

**Review is PENDING.** Not gated, not reviewed, no verdict claimed.

### DIFF
--- a/browser_tests/legacy-transcript-order.spec.ts
+++ b/browser_tests/legacy-transcript-order.spec.ts
@@ -1,0 +1,112 @@
+/**
+ * #1516 — after a hard refresh that crossed a panel version boundary, the user
+ * reported "refreshing the page seems to have repeated my initial messages to
+ * you". Nothing was re-sent to the agent; the CHAT PANE re-painted the stored
+ * transcript out of order, dropping the earliest prompts below the agent's last
+ * reply.
+ *
+ * A pre-0.11.0 panel (before IndexedDB and the atomic snapshot key existed)
+ * wrote a bare array of threads under `comfyui-mcp.panel.threads`, and its
+ * messages carried no id and no timestamp. The restore path merges its own
+ * mount-time read of that array with the copy ChatHistoryStore.load() resolves
+ * to, and the merge sorted timestamp-less records by their content-hash id.
+ *
+ * Measured against this exact seed before the fix:
+ *
+ *     stored   u1 a1 u2 a2 u3 a3
+ *     painted  a3 u1 u2 a1 a2 u3
+ *
+ * This spec is deliberately agent-free and never sends a message: the defect is
+ * in hydration, and seeding at document start is the only way to reproduce a
+ * bundle-version boundary the harness cannot otherwise cross.
+ */
+import { test, expect } from './fixtures/panelTest'
+import { routeWorktreeSource } from './fixtures/worktreeSource'
+
+test.beforeEach(async ({ context }) => {
+  await routeWorktreeSource(context)
+})
+
+const LEGACY_THREAD_ID = '11111111-2222-3333-4444-555555555555'
+const USER_TEXTS = ['legacy first 1516', 'legacy second 1516', 'legacy third 1516']
+
+test('#1516: a pre-0.11 transcript hydrates in the order it was written', async ({
+  page,
+  panel,
+  context
+}) => {
+  // Seeded at document start, so the panel's synchronous restore reads exactly
+  // what the older bundle left behind — no id, no createdAt, no ts, and a BARE
+  // ARRAY under the pre-v3 key.
+  await context.addInitScript(
+    ({ threadId, texts }) => {
+      const msgs: Array<Record<string, string>> = []
+      texts.forEach((text: string, i: number) => {
+        msgs.push({ role: 'user', text })
+        msgs.push({ role: 'agent', text: `legacy reply ${i}` })
+      })
+      localStorage.removeItem('comfyui-mcp.panel.historySnapshot')
+      localStorage.removeItem('comfyui-mcp.panel.historyMeta')
+      localStorage.setItem(
+        'comfyui-mcp.panel.threads',
+        JSON.stringify([
+          { id: threadId, ts: Date.now() - 60_000, workflowKey: 'panel:global', msgs }
+        ])
+      )
+      sessionStorage.setItem('comfyui-mcp.panel.currentThreadId', threadId)
+    },
+    { threadId: LEGACY_THREAD_ID, texts: USER_TEXTS }
+  )
+
+  await panel.goto()
+  await panel.openSidebar()
+  await expect(panel.userBubble(USER_TEXTS[2])).toBeVisible()
+
+  // Settle the async durable hydration: the synchronous paint-only pass reads the
+  // shadow, and the merge that used to shuffle the transcript runs after
+  // IndexedDB and settings resolve.
+  await expect
+    .poll(async () =>
+      page.evaluate(() =>
+        [...document.querySelectorAll('.cmcp-root .cmcp-bubble')].map((el) =>
+          (el as HTMLElement).innerText.trim().split('\n')[0]
+        )
+      )
+    )
+    .toEqual([
+      'legacy first 1516',
+      'legacy reply 0',
+      'legacy second 1516',
+      'legacy reply 1',
+      'legacy third 1516',
+      'legacy reply 2'
+    ])
+
+  // Each prompt appears ONCE. The reporter read the reshuffle as a repeat, so
+  // assert the count too rather than only the order.
+  for (const text of USER_TEXTS) {
+    await expect(panel.userBubbles.filter({ hasText: text })).toHaveCount(1)
+  }
+
+  // What was painted is what got persisted: a second reload must not shuffle it
+  // back, and the durable record must not keep the scrambled order either.
+  await expect
+    .poll(async () =>
+      page.evaluate(() => {
+        const raw = localStorage.getItem('comfyui-mcp.panel.historySnapshot')
+        if (!raw) return null
+        const thread = JSON.parse(raw).threads.find(
+          (t: { id: string }) => t.id === '11111111-2222-3333-4444-555555555555'
+        )
+        return thread ? thread.msgs.map((m: { text: string }) => m.text) : null
+      })
+    )
+    .toEqual([
+      'legacy first 1516',
+      'legacy reply 0',
+      'legacy second 1516',
+      'legacy reply 1',
+      'legacy third 1516',
+      'legacy reply 2'
+    ])
+})
--- a/browser_tests/unit/legacy-transcript-order.test.mjs
+++ b/browser_tests/unit/legacy-transcript-order.test.mjs
@@ -21,6 +21,8 @@
  */
 import assert from 'node:assert/strict'
 import test from 'node:test'
+import { readFileSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
 
 import { mergeHistorySnapshots } from '../../web/js/lib/chat-history-store.js'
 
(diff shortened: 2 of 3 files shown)


artokun:
@grok please review — and here is where the load is, so you don't have to hunt for it.

**The one-line change is in `web/js/lib/chat-history-store.js`, `mergeThreadMessages`.**
I insert a position rank as a sort tiebreak *between* the timestamp comparison and
the id comparison.

Three things I am least sure of, in the order I'd attack them:

1. **Does the rank map ever outrank a real clock?** The whole safety argument is
   that `finiteTs(a.createdAt || a.ts) - finiteTs(b.createdAt || b.ts)` resolves
   FIRST, so two tabs writing concurrently still merge causally by time and only
   timestamp-less (pre-v3) records ever reach my tiebreak. I pinned that with
   `#1516: timestamped messages still interleave by TIME, not by position` and
   mutation-checked it (hoisting the rank above the clock turns exactly that test
   red). But the guard is only as good as my claim that **every** non-legacy
   message really carries a usable `createdAt`. If there is a writer that records a
   message with `createdAt: 0`/absent while its siblings have real times, my change
   reorders a live transcript, and that would be a regression I introduced rather
   than the bug I fixed. `normalizeMessage` floors to `1`, so such a message ties
   with legacy ones — is that reachable?

2. **Is the rank stable across the round trip, or does it just move the shuffle
   one reload later?** `rank` is rebuilt per call from `[...oldMessages,
   ...newMessages]` — nothing is persisted. My argument that this converges is:
   the merged order is what gets written, so the next mount reads an
   already-correct array and re-ranks from it. I tested three round trips
   (`the order survives the round trip a reload actually makes`). Three is not
   a proof. Is there an input where the fixed point oscillates — in particular a
   message present ONLY in `newer` (it gets a rank after every `older` message,
   so with no timestamps it is appended to the end rather than placed where it
   belongs)?

3. **Scope: I deliberately left the duplication half alone.** The synthetic
   legacy id is `legacy-<hash(threadId:ordinal:canonicalJson(message))>`, so two
   copies of one thread that disagree on ordinals mint different ids for the same
   message and the merge duplicates it — I reproduced that at module level (a
   6-message thread merged with its own 4-message tail yields 10 messages). I
   argue it is unreachable in production because `mergeUnderCanonicalCheckpoint`
   quarantines a flagged pre-v3 snapshot wholesale against a fenced canonical and
   replaces it as a unit against an unfenced one, and because the panel's own
   unguarded `mergeHistorySnapshots({ threads, meta }, loaded)` derives both sides
   from the same localStorage read at the same ordinals. **If that reachability
   argument is wrong, the issue's literal wording ("repeated my initial messages")
   is the duplication and not the reordering, and this PR fixes the wrong half.**
   That is the finding I most want you to try to produce.

Evidence I actually ran (not requests I made): the reordering is reproduced in a
real browser against a real ComfyUI, and deleting the fix turns both the named
unit tests and the browser spec red. Full unit suite 6004/0. The Playwright run is
still going; I will post its result before merging and I will not merge on my own
verification.

artokun:
## Independent review (second agent, not the author of this PR)

I was dispatched on #1516 in parallel and established the mechanism **before** reading this
PR, so treat the first section as corroboration and the second as a finding.

### Corroboration — same mechanism, reached independently

Ran the pristine `origin/main` module in Node against the exact pre-0.11 shape (bare array,
no `id`, no `createdAt`, no `ts`):

```
STORED           ["u1","a1","u2","a2","u3","a3"]
normalizeThread  ["u1","a1","u2","a2","u3","a3"]   createdAt: [1,1,1,1,1,1]
merge(1 snap)    ["u1","a1","u2","a2","u3","a3"]
merge(2 snaps)   ["a2","u3","a3","a1","u2","u1"]   n=6
```

Same conclusion as the PR: a **re-order**, not a duplication and not a re-submission, and
`n` never grows. I also confirmed the two hypotheses the issue offered are both dead:

* *"appended to a DOM that already had it"* — `paintThread` opens with `resetFeed()`, which
  removes every child of `log`. Neither of the two restore paints can append to a populated
  pane.
* *"re-sent and de-duplicated"* — the only re-send is `armVisibleTranscriptReplay()` →
  `client.armContext(replay)`, which arms **context**, paints nothing, and only fires on
  `foreignSession`. Matches the reporter's agent-side observation.

Also checked the PR's "deliberately did not do" claim about the ordinal-shift hazard:
`mergeUnderCanonicalCheckpoint` does fence it in both directions (whole-snapshot quarantine
when `canonicalFenced`, replace-as-a-unit when not). That claim holds.

### Finding — the rank tiebreak does not survive a truncated shadow

A **global** rank over `[...oldMessages, ...newMessages]` can only preserve order when
`older` contributes a prefix of the combined sequence. When `older` is a **tail**, the whole
head of the conversation sorts after it — the reported symptom, in the same place.

Measured against **this PR's head** (`chat-history-store.js` from
`fix/1516-transcript-replay-on-reload`), post-migration records: ids minted, `createdAt` still
floored at `1`:

```
A identical       ["u1","a1","u2","a2","u3","a3"]   ok
B tail + full     ["a2","u3","a3","u1","a1","u2"]   want u1,a1,u2,a2,u3,a3
C full + tail     ["a2","u3","a3","u1","a1","u2"]   want u1,a1,u2,a2,u3,a3
D interleaved     ["u1","u2","u3","a1","a2"]        want u1,a1,u2,a2,u3
```

**This is reachable, not hypothetical.** `_writeLocalSnapshot` truncates the localStorage
startup shadow to the tail of each ordinary thread —

```js
msgs: thread.msgs.slice(-LOCAL_SHADOW_MESSAGES),   // 200
```

— while the canonical IndexedDB copy holds up to `DEFAULT_MAX_MESSAGES` (5000). The restore
path then merges exactly those two:

```js
const merged = mergeHistorySnapshots({ threads, meta: historyMeta }, loaded)
```

So a **pre-v3-originated thread longer than 200 messages** — one that already migrated into
canonical, so it is `legacyShadow: false` and not exempt from the truncation — reproduces
#1516 after this fix. Every one of its messages is pinned at `createdAt: 1` permanently, so
the rank tiebreak governs every merge it will ever take part in.

Narrower than what #1516 reports (that thread was ~8 messages), so I would not call this a
NO-SHIP for the reported defect. But the unit test is named *"the order survives the round
trip a reload actually makes"*, and for a >200-message legacy thread it does not — the test
claims a broader invariant than the change delivers.

### On the obvious one-token fix — it does not work

Seeding the rank from the other side, `[...newMessages, ...oldMessages]`, fixes B, C and D
exactly, and then fails the mirror case. Measured, both variants, same input shape:

```
E: full(older) x tail(newer)
  this PR's head            ["u1","a1","u2","a2","u3","a3"]   ok
  [...newMessages, ...old]  ["a2","u3","a3","u1","a1","u2"]   broken
```

`older`/`newer` is decided by `updatedAt`, not by argument position, and the shadow is
rewritten on every persist — so which copy is "newer" is not something the merge can lean on.
The two orderings are exact mirror images of each other.

The shape that actually holds is an **order-preserving two-sequence merge**: rank each message
by its position **within its own array**, and splice messages unique to one array after their
nearest common predecessor, rather than ranking the concatenation. That is bigger than a
one-line tiebreak, so it may be right to land this PR for the reported defect and take the
truncated-shadow case separately — but the choice should be explicit rather than implied by a
test name.

### What I did not do

- Did not run the full Playwright suite. This is not my branch and staging it would have
  collided with the author working the same worktree; the measurements above are against this
  PR's own module, which is the stronger evidence for this particular claim anyway.
- Did not open a competing PR or touch this branch.
- No verdict claimed. This is one agent's read, not a gate — the grok review requested above
  is still the review of record.

PR #1531, opened 2026-08-20
fix(1516): the archive import carried the same broken tiebreak

artokun:
Follow-up to #1530 (merged as `1465054b`). The underlying cause of #1516 is fixed on
main; this closes the **sibling call site** that carried the identical defect and
which #1530 did not reach.

## What #1530 fixed, and what it missed

#1530 fixed `mergeThreadMessages`: a pre-v3 transcript has no per-message
timestamp, `normalizeMessage` floors every message at `createdAt: 1`, the time
comparison ties for the whole thread, and the tiebreak — the message id, which for
these records is `legacy-<hash of the message's own text>` — dealt the conversation
back in content-hash order.

`ChatHistoryStore.importPayload` sorted messages with its **own copy of the same
two-term rule**:

```js
thread.msgs = [...messages.values()].sort(
  (left, right) =>
    finiteTs(left.createdAt || left.ts) - finiteTs(right.createdAt || right.ts) ||
    String(left.id).localeCompare(String(right.id)),
)
```

Same tie, same tiebreak, same scramble. And it matters more here than in the merge,
because **an import is a write**: the scrambled order is persisted as the
conversation's real one, so a repaired thread can be re-broken by importing an
archive of itself.

Reachable, not theoretical: `exportPayload` runs every thread through
`portableThread` → `normalizeThread`, which is exactly what floors a legacy message
at `createdAt: 1`. So exporting a pre-0.11 transcript produces an archive carrying
the shape that ties, and importing it scrambles it.

## The change

One comparator (`compareMessageOrder`) and one rank builder (`messagePositions`)
now serve **both** call sites, so they cannot drift apart again. No behaviour change
for timestamped messages at either site: the time comparison still resolves before
the rank is consulted.

## Verification

- `browser_tests/unit/legacy-transcript-order.test.mjs` gains
  `#1516: importing an exported pre-v3 transcript keeps its order`, driving the real
  `exportPayload` → `importPayload` round trip.
- **Mutation-verified three ways** against the committed state:
  - restore the id tiebreak at the **import site only** → the new import test fails, the other six stay green (this is what proves the new test tests the import, not the merge);
  - drop the rank from the shared comparator → 3 fail / 4 pass;
  - hoist the rank **above** the clock → only `timestamped messages still interleave by TIME` fails, so that guard is load-bearing rather than decorative.
- `npm run test:unit` — **6043 pass / 0 fail**. `npm run typecheck` — clean.

## Review

**Review is PENDING.** Not gated, not reviewed, no verdict claimed. #1530 was merged
before the review I requested on it came back; I would rather this one waited.

### DIFF
--- a/browser_tests/unit/legacy-transcript-order.test.mjs
+++ b/browser_tests/unit/legacy-transcript-order.test.mjs
@@ -24,7 +24,7 @@ import test from 'node:test'
 import { readFileSync } from 'node:fs'
 import { fileURLToPath } from 'node:url'
 
-import { mergeHistorySnapshots } from '../../web/js/lib/chat-history-store.js'
+import { ChatHistoryStore, mergeHistorySnapshots } from '../../web/js/lib/chat-history-store.js'
 
 const THREAD_ID = '11111111-2222-3333-4444-555555555555'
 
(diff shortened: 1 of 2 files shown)


artokun:
@grok please review — the load-bearing claim here is a **reachability** one, not the code change.

The change itself is mechanical: two call sites that each had their own copy of the same `time || id` comparator now share one `compareMessageOrder`. What I actually need checked is the claim that the import site was **reachable with timestamp-less messages at all**. My argument:

1. `exportPayload` maps each thread through `portableThread`, which calls `normalizeThread`;
2. `normalizeThread` -> `normalizeMessages` -> `normalizeMessage` floors a message with no `createdAt` and no `ts` at `createdAt: 1` and mints `legacy-<stableHash(threadId:ordinal:canonicalJson(message))>`;
3. `portableMessage` strips `revision`/`createdRevision` but keeps `createdAt` and `id`;
4. so the archive carries createdAt:1 across the whole thread, the time comparison ties, and the id tiebreak is a content hash.

If any step of that is wrong — in particular if something between export and import re-stamps a real time onto these records — then this PR fixes nothing and should be closed rather than merged. I drove the real `exportPayload` -> `importPayload` round trip and it did scramble before the change, so I believe step 4 holds empirically; what I cannot rule out by reading is whether a real user's archive can actually contain such records, or whether every archive in the wild predates or postdates the window.

Two smaller things I'd want a second pair of eyes on:

- `messagePositions(collected)` is built **before** `collected.sort(...)` reorders it in place. That ordering dependency is now load-bearing and is only stated in a comment. If you think a future edit will trip over it, say so and I will copy the array instead.
- At the import site the rank comes from the Map's insertion order — local messages first, then archive additions. For a colliding thread id where the archive's new messages are timestamp-less, that appends them after every local message rather than interleaving. I believe that is the only defensible answer (there is nothing to interleave *on*), but it is a choice, not a derivation.

Also, for the record: #1530 was merged before the review I requested on it returned. The merge is correct as far as my own evidence goes — reproduced in a real browser, mutation-verified in three directions, 6043/0 unit — but it has not had an independent read, so a finding against **that** commit is still worth filing even though it has shipped.

PR #1538, opened 2026-08-20
fix(1535): a no-name save of a .app.json workflow writes that file, not a plain .json fork

### LINKED ISSUE #1535: No-name save of active .app.json forks to plain .json
panel_save_workflow({}) did not save the active .app.json workflow in place. After panel_save_workflow({name:"11 ANIME - ANIMA Turbo.app"}) correctly created workflows/11 ANIME - ANIMA Turbo.app.json, the graph was reopened, panel_configure_app_mode wrote linearMode=true with four inputs and one output, and panel_save_workflow({}) then reported saved_as:true and created workflows/11 ANIME - ANIMA Turbo.json. The active workflow display filename omits the .app suffix, and the in-place save path appears to reconstruct a plain .json target from that display name instead of retaining the routing key/path.

### Steps to reproduce
1. Start from a saved .json workflow.
2. panel_save_workflow({name:"Example.app"}) -> workflows/Example.app.json.
3. panel_open_workflow("Example.app.json").
4. panel_configure_app_mode({default_mode:"app", inputs:[...], outputs:[...]}).
5. panel_save_workflow({}) with no name.
6. Observe saved_as:true to workflows/Example.json; Example.app.json remains unchanged.

### Exact result
The tool reported workflow_uuid changed and routing_key "wf:workflows/11 ANIME - ANIMA Turbo.json". Disk readback showed the plain JSON had linearMode=true, 4 inputs, 1 output, while the intended .app.json still had linearMode=false, 0 inputs, 0 outputs.

### Fix
applied locally: no
upstream-only: yes
Precise change needed: for no-name saves, retain the active workflow's full routing key/path including the .app suffix. Do not rebuild the filename from the suffix-stripped display title. Treat it as in-place and never report saved_as when the caller omitted name on an existing .app.json workflow.

Temporary verified workaround: validate the mistakenly-created plain JSON, copy it over the exact .app.json target, read back extra.linearMode/linearData, then remove only the generated plain JSON.

### Environment
Windows 11
ComfyUI 0.33.2
ComfyUI frontend 1.49.6
NVIDIA RTX 5070 Laptop GPU, 8 GB VRAM
comfyui-mcp 0.52.30
panel 0.15.12

---

artokun:
Closes #1535.

## What the caller saw

`panel_save_workflow({})` — a no-name, save-in-place call — on a workflow living at
`workflows/X.app.json` created a **new** `workflows/X.json` holding the caller's App Mode
configuration, reported `saved_as: true`, and never wrote the file the caller was editing.
The reporter's `.app.json` still read `linearMode:false, 0 inputs, 0 outputs` while the
orphan `.json` held `linearMode:true, 4 inputs, 1 output`.

## Mechanism, established by execution

Reproduced in a real browser against the reporter's own stack (ComfyUI 0.33.2, frontend
1.49.6) before any code was written. On unmodified `origin/main` the new spec fails with:

```
Expected: "wf:workflows/cmcp-e2e-1535-1787261586640.app.json"
Received: "wf:workflows/cmcp-e2e-1535-1787261586640.json"
```

A second probe dumped the frontend's own identity at each step of the report:

```
after-open      path=workflows/<stem>.app.json  filename=<stem>  initialMode=null
after-configure path=workflows/<stem>.app.json  filename=<stem>  initialMode=null  liveLinearMode=true
saved           saved_as=true  routing_key=wf:workflows/<stem>.json  copied_from=<stem>
```

`saveActiveWorkflow` resolves its destination by rebuilding `filename + modeExtension`,
and that round-trip is lossy for exactly one shape:

- the frontend's `getFilenameDetails` strips the **compound** `.app.json` suffix, so a
  workflow at `X.app.json` reports `filename` `"X"` — the `.app` survives only in `path`;
- `initialMode` is populated from the **file's** `extra.linearMode` at load, and
  `graph_configure_app_mode` writes `extra.linearMode` on the live root and never touches
  `initialMode`. For a workflow configured into App Mode after it was opened it stays
  unset (or `"graph"`).

Both halves then say "plain", the target comes out `workflows/X.json`, that reads as a
relocation, and the save routes itself down the Save-As **copy** path.

## The fix

For a **no-name** save, the target is the workflow's own path when that path is exactly
the `.app.json` sibling the reconstruction dropped. The path is the identity that
round-trips to the file on disk, so it wins.

The in-place branch really does write that path — this is the load-bearing claim and it
was measured, not read. `saveInPlace` calls `svc.saveWorkflow(wf)`, and `svc` here is the
workflow **store** (`extensionManager.workflow` exposes `saveAs` / `renameWorkflow` /
`openWorkflows`, not the service's `saveWorkflowAs`). The store's `saveWorkflow` is
`wf.save()` → `UserFile.save({force:true})`, a write to `this.path`. Driven on the rig to
`path` `workflows/X.app.json` with `initialMode` unset and `extra.linearMode` true:

```
PROBE direct={"before":{"path":"workflows/<stem>.app.json","filename":"<stem>","initialMode":null},
              "err":null,"afterPath":"workflows/<stem>.app.json","afterInitialMode":null}
PROBE appFile={"status":200,"linearMode":true}
PROBE forkFile={"status":404,"linearMode":null}
```

It is the **service's** `saveWorkflow` that recomputes
`directory + appendWorkflowJsonExt(filename, isApp)` and `renameWorkflow()`s the source to
it. Nothing on this path calls that.

### Deliberately one direction

The mirror case — an on-disk `Foo.json` whose content declares app mode, so the
mode-derived target is `Foo.app.json` (the existing `P0-b` test) — is **not** changed and
still goes down the non-destructive copy route. That direction produces a file whose
extension agrees with its own content and leaves the source intact; this one produced a
`.json` holding `linearMode:true` — an extension contradicting its content — while
abandoning the consistent file that already existed. Only the lossy direction is repaired.

## Evidence

**Deleted the fix, watched named tests fail.** Five mutations, each applied to real bytes
and verified to have applied (the file is CRLF; an LF-only anchor reads as "survived"
without ever landing):

| # | mutation | result |
|---|---|---|
| 1 | **call site**: `finalTargetPath` back to the plain reconstruction | 3 named tests fail |
| 2 | **helper**: `pathIsAppSuffixedSiblingOf` always false | 3 named tests fail |
| 3 | drop the `!desired` guard (a named save would read the source suffix) | 1 named test fails |
| 4 | drop the `!wasUnsaved` guard | 1 named test fails |
| 5 | match any `.app.json` path instead of this workflow's OWN sibling path | 1 named test fails |

Mutations 3–5 **survived** the first pass — the three boundary tests exist because of
that, not as restatements of the fix.

- Live save test: `browser_tests/noname-save-keeps-app-json.spec.ts` drives the reporter's
  five steps against a throwaway workflow on the real ComfyUI and asserts the routing key,
  the absence of `saved_as`, `extra.linearMode` in the `.app.json` **on disk**, and a
  404 at the `.json` path. Red on `origin/main`, green here.
- Full panel unit suite + gates (`npm run test:unit`): **6058 pass, 0 fail**.
- Full browser suite `npx playwright test --workers=1`: running; result posted below.

## Review pending

Not gated and not reviewed. `codex-gate.mjs` was deliberately not run. grok review is
requested in a comment on this PR naming the risky parts; **do not merge until it comes
back.**

## Deliberately not done

- The mirror direction (`Foo.json` + app-mode content) is unchanged, so the reporter's
  orphan `Example.json` will still copy to `Example.app.json` on its next no-name save.
  Changing that is a product decision about whether the panel should follow ComfyUI's own
  `.json` → `.app.json` migration, not a bug fix.
- The `P0` refusal for a persisted workflow with an **empty** filename is untouched.
- `svc.saveWorkflow` was **not** replaced or re-pointed, and `wf.initialMode` is never
  mutated — the fix changes only which path the save aims at.

### DIFF
--- a/browser_tests/noname-save-keeps-app-json.spec.ts
+++ b/browser_tests/noname-save-keeps-app-json.spec.ts
@@ -1,0 +1,110 @@
+/**
+ * #1535 — a NO-NAME `panel_save_workflow({})` on an active `.app.json` workflow must
+ * write THAT file, not fork a plain `.json` beside it.
+ *
+ * The reported sequence, run here against the real ComfyUI frontend:
+ *   save under a ".app" name  → workflows/<stem>.app.json
+ *   open it                   → the frontend reports filename "<stem>" (getFilenameDetails
+ *                               strips the compound ".app.json" suffix) while
+ *                               `initialMode` is derived from the FILE's extra.linearMode
+ *   graph_configure_app_mode  → writes extra.linearMode = true on the live root, and does
+ *                               NOT touch `initialMode`
+ *   save with no name         → the save layer rebuilt the target as
+ *                               "<stem>" + modeExt(initialMode) = "<stem>.json", saw a
+ *                               relocation, and routed to the Save-As COPY path. A NEW
+ *                               `workflows/<stem>.json` appeared holding the app-mode
+ *                               configuration, `saved_as: true` was reported, and the file
+ *                               the user was actually editing was never written.
+ *
+ * That is silent data divergence: the caller is told the save succeeded and keeps working
+ * in a file that does not have its edits.
+ *
+ * The assertions below are on the OBSERVED EFFECT — the routing key of the reply and the
+ * bytes on disk — not on which branch the code took.
+ */
+import { test, expect } from './fixtures/panelTest'
+import { routeWorktreeSource } from './fixtures/worktreeSource'
+
+test.beforeEach(async ({ context }) => {
+  await routeWorktreeSource(context)
+})
+
+/** Read a workflow file back through ComfyUI's own userdata API. `null` = absent. */
+async function readWorkflowFile(page: import('@playwright/test').Page, userdataPath: string) {
+  return await page.evaluate(async (path) => {
+    const api = (window as any).comfyAPI?.api?.api
+    if (typeof api?.fetchApi !== 'function') return { status: -1, body: null }
+    const res = await api.fetchApi(`/userdata/${encodeURIComponent(path)}`)
+    if (!res?.ok) return { status: res?.status ?? -1, body: null }
+    let body: any = null
+    try {
+      body = JSON.parse(await res.text())
+    } catch {
+      body = null
+    }
+    return { status: res.status, body }
+  }, userdataPath)
+}
+
+test('a no-name save of an active .app.json workflow writes that file, not a plain .json fork', async ({
+  page,
+  panel,
+  mockBridge
+}) => {
+  await panel.goto()
+  await panel.setBridgeUrl(mockBridge.url)
+  await panel.openSidebar()
+  await panel.connect()
+
+  const stem = `cmcp-e2e-1535-${Date.now()}`
+  const appPath = `workflows/${stem}.app.json`
+  const forkPath = `workflows/${stem}.json`
+
+  // Step 2 of the report: a save under a ".app" name lands the file at "<stem>.app.json".
+  const created = await mockBridge.command('workflow_save', { name: `${stem}.app` })
+  expect(created.ok, `the setup save must succeed: ${created.error ?? ''}`).toBe(true)
+  expect(
+    (await readWorkflowFile(page, appPath)).status,
+    'the setup must actually produce the .app.json file the report starts from'
+  ).toBe(200)
+
+  // Step 3: reopen it, so the workflow's identity comes from the PATH the way the report's
+  // does — this is what makes `filename` "<stem>" with the ".app" suffix living only in the
+  // path.
+  const opened = await mockBridge.command('workflow_open', { path: appPath })
+  expect(opened.ok, `reopening the .app.json must succeed: ${opened.error ?? ''}`).toBe(true)
+
+  // Step 4: configure app mode. This writes extra.linearMode on the live root.
+  const configured = await mockBridge.command('graph_configure_app_mode', { default_mode: 'app' })
+  expect(configured.ok, `configure_app_mode must succeed: ${configured.error ?? ''}`).toBe(true)
+  expect(configured.result?.linearMode, 'app mode must be on for the save to carry it').toBe(true)
+
+  // Step 5: THE CALL UNDER TEST — save with no name at all.
+  const saved = await mockBridge.command('workflow_save', {})
+  expect(saved.ok, `the no-name save must succeed: ${saved.error ?? ''}`).toBe(true)
+
+  // THE ASSERTIONS. A no-name save is an in-place save of the file the caller is editing.
+  expect(
+    saved.result?.routing_key,
+    'the no-name save must report the file it was editing, not a plain-.json fork'
+  ).toBe(`wf:${appPath}`)
+  expect(
+    saved.result?.saved_as,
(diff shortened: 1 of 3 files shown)


artokun:
@grok please review this one — it is a data-integrity fix on the save path, and I would rather you look at where the risk actually is than hunt for it.

**The load-bearing claim, and the one thing that would sink this if wrong:** that `svc` in `saveInPlace` is the workflow **STORE**, not the workflow **service** — so `svc.saveWorkflow(wf)` is `wf.save()` → `UserFile.save({force:true})`, a write to `this.path`, and never a relocation. The whole fix rests on it: I make `finalTargetPath` equal `currentPath` for one case, which classifies the save as in-place, and if that primitive DID relocate the way the service's `saveWorkflow` does (`renameWorkflow(e, directory + appendWorkflowJsonExt(filename, isApp))`) this would MOVE and consume the user's `.app.json` — issue #226, the worst outcome in this file.

I established it two ways on the live rig (ComfyUI 0.33.2 / frontend 1.49.6): `Object.keys(extensionManager.workflow)` contains `saveAs`, `renameWorkflow`, `openWorkflows`, `getWorkflowByPath` and has **no** `saveWorkflowAs`/`exportWorkflow` (store, not service); and calling `svc.saveWorkflow(wf)` directly with `path` `workflows/X.app.json`, `initialMode` unset and `extra.linearMode` true wrote X.app.json, left `wf.path` unchanged, and produced no X.json. What I have NOT established is whether any older or newer frontend puts the *service* on `extensionManager.workflow`. If you know of one, this is unsafe as written.

**Second-riskiest: the shared unit double `makeService` models `saveWorkflow` as RELOCATING** (`browser_tests/unit/workflow-save.test.mjs`), which contradicts the measurement above. I deliberately did not touch it — my new tests use `makeFaithfulService`, whose `saveWorkflow` writes `wf.path`. So the suite now contains two different models of the same API. Tell me if you think the strict one should win and the fix should be reshaped to satisfy it.

**Third: the scope boundaries.** The rule is `!desired && !wasUnsaved && pathIsAppSuffixedSiblingOf(wf, currentName)`. Each guard has a test that dies without it (mutations 3, 4 and 5 in the PR body all SURVIVED my first pass — the three boundary tests exist because of that). The judgement calls I would most like a second opinion on:
- `!wasUnsaved` leaves the #215 drift case (a file that IS on disk but reads `isTemporary`) still forking. I chose the conservative fail-safe — the collision-checked copy route — over fixing it, because a genuinely never-persisted tab taking the in-place branch would `force`-overwrite whatever occupies that path.
- full-path equality rather than "the path ends in .app.json". A tab whose `filename` has drifted from its `path` therefore does NOT take the in-place branch. Is that the right side to err on?

**Fourth: the deliberate asymmetry.** The mirror case — an on-disk `Foo.json` whose content declares app mode — is unchanged and still copies to `Foo.app.json` (the existing P0-b test). My argument is that direction yields a file whose extension agrees with its content, while the broken one yielded a `.json` holding `linearMode:true` and abandoned the consistent file. It is the same user-visible shape though, so if you think it should be symmetric, say so — it is a product call and I did not want to make it under the freeze.

Full unit suite green (6058/0). Browser suite `--workers=1` running; I will post the result. Not merging until you answer.

PR #1542, opened 2026-08-20
fix(1540): the default saveWorkflow test double writes the tab's own path

### LINKED ISSUE #1540: test harness: workflow-save.test.mjs carries two contradictory models of extensionManager.workflow.saveWorkflow
## What is wrong

`browser_tests/unit/workflow-save.test.mjs` contains **two contradictory models of the same
frontend API**, and the one used by the most tests is the wrong one.

The panel's `saveInPlace` (`web/js/lib/workflow-save.js`) writes through `svc.saveWorkflow(wf)`,
where `svc` is `app.extensionManager.workflow`. The three doubles in the test file disagree about
what that call does:

| double | call sites | models `saveWorkflow(wf)` as |
|---|---|---|
| `makeService` (line 29) | 20 | recomputes `dir + stripExt(filename) + extFor(wf)` and **`renameWorkflow`s** the source when it differs |
| `makeFaithfulService` (line 457) | 14 | writes `wf.path` |
| `makeStore147Service` (line 1009) | 19 | writes `wf.path` |

`makeService` even documents its model in a comment: *"Mirrors ComfyUI's saveWorkflow: it recomputes
the expected path from the workflow's mode-derived extension, and if that differs from the current
path it RENAMES (moves) the file before saving."*

## Why it is wrong — measured, not reasoned

Measured live on the rig during #1535 (ComfyUI 0.33.2, frontend 1.49.6):

**1. `extensionManager.workflow` is the workflow STORE, not the workflow SERVICE.**
`Object.keys()` on it returns `activeWorkflow`, `openWorkflows`, `persistedWorkflows`,
`getWorkflowByPath`, `createNewTemporary`, `attachWorkflow`, `saveAs`, `renameWorkflow`,
`openWorkflow`, `closeWorkflow`, `deleteWorkflow`, `syncWorkflows`, `loadWorkflows`, plus Pinia's
`$id`/`$patch`/`$subscribe`. It has **no** `saveWorkflowAs`, no `exportWorkflow`, no
`loadDefaultWorkflow` — those are the service's.

**2. The store's `saveWorkflow` does not relocate.** In the shipped bundle it is
`async e => { isBusy = true; try { await e.save(); const t = detachWorkflow(e); attachWorkflow(e, t) } finally { isBusy = false } }`,
and `ComfyWorkflow.save()` is `super.save({ force: true })` — `UserFile.save`, a write to `this.path`.

**3. Driven directly in the exact state where the two models diverge** (path
`workflows/X.app.json`, `initialMode` unset, `extra.linearMode` true — the mode-derived target
would be `X.json`):

```
PROBE direct={"before":{"path":"workflows/<stem>.app.json","filename":"<stem>","initialMode":null},
              "err":null,"afterPath":"workflows/<stem>.app.json","afterInitialMode":null}
PROBE appFile={"status":200,"linearMode":true}     <- wf.path was written
PROBE forkFile={"status":404,"linearMode":null}    <- nothing created at the mode-derived path
```

`wf.path` unchanged, no rename, no second file. It is the **service's** `saveWorkflow` that does
`renameWorkflow(e, directory + appendWorkflowJsonExt(filename, isApp))`, and nothing in the panel
calls that.

## Why it matters

This is a harness that can steer a fix into the wrong shape. While fixing #1535 the relocating
double made two behaviours look mandatory that production does not require, and only the live
measurement settled it. The next agent to touch this file will meet the same fork with no
measurement in hand, and `makeService` — the model that is wrong — has the most tests behind it,
so it looks authoritative.

It is not urgent and nothing is currently broken: `makeService`'s model is strictly *stricter*
than production, so tests written against it assert the panel never triggers a relocation that
production could not perform anyway. The cost is comprehension, not coverage.

## Suggested resolution

Not obvious enough to just do — recording the measurement so whoever picks it up does not have to
re-derive it. Options as I see them:

1. **Rename it.** `makeService` -> `makeRelocatingServiceService` (or similar) and say in its
   comment that it models `useWorkflowService.saveWorkflow`, which the panel does **not** call,
   kept as a belt-and-braces guard in case a frontend ever puts the service on
   `extensionManager.workflow`. Cheapest, keeps all existing coverage, removes the ambiguity.
2. **Correct it** to write `wf.path` and move the relocation guard onto an explicitly-named
   service double. More faithful, but it weakens ~20 tests whose `renameWorkflow` assertions
   become vacuous.

Option 1 looks right to me, but it is a judgement call about what those 20 tests are for.

## Caveat on the measurement

Established on frontend **1.49.6** only. I did not check whether an older or newer frontend puts
the *service* on `extensionManager.workflow`. If one does, `makeService` is right for that build
and the fix is to model both explicitly rather than to pick one.

## Context

Found while fixing #1535 (merged as e430fc99). Called out in the review request on #1538 as the
second-riskiest thing in that change; filing it here so it is not lost on a closed PR.

artokun:
Fixes #1540.

## What was wrong

`browser_tests/unit/workflow-save.test.mjs` modeled `extensionManager.workflow.saveWorkflow` two ways. `makeService` (the default, used by most tests) recomputed `dir + stripExt(filename) + extFor(wf)` and `renameWorkflow`'d the source when it differed. `makeFaithfulService` and `makeStore147Service` wrote `wf.path`.

The relocating model is `useWorkflowService.saveWorkflow`. The panel's `saveInPlace` calls `svc.saveWorkflow(wf)` on the workflow **store** (`app.extensionManager.workflow`). Measured on frontend 1.49.6: the store's saveWorkflow is `await e.save()` → `UserFile.save({force:true})`, a write to `this.path`. No rename, no second file.

That fork made two behaviours look mandatory that production does not require, and it forced the #1535 in-place cases onto `makeFaithfulService` because `makeService` would have relocated the `.app.json` to a mode-derived `.json`.

## The fix

- `makeService.saveWorkflow` now shares a `storeSaveWorkflow` helper with `makeFaithfulService`: write `wf.path`.
- The #1535 in-place cases run on `makeService`.
- A new `#1540` test drives `saveActiveWorkflow` through that default double in the exact state the two models disagreed on (`path` already `.app.json`, `filename` the stem, `initialMode` unset). Exact call list is `[['saveWorkflow', 'workflows/X.app.json']]`. A relocating saveWorkflow would have pushed `renameWorkflow` first and then saved the mode-derived `.json`.

Production save routing is unchanged.

## Evidence

```
node --test browser_tests/unit/workflow-save.test.mjs
# 95 pass, 0 fail
```

### DIFF
--- a/browser_tests/unit/workflow-save.test.mjs
+++ b/browser_tests/unit/workflow-save.test.mjs
@@ -11,9 +11,20 @@ import {
   pathSeparatorNameError
 } from '../../web/js/lib/workflow-save.js'
 
-// A minimal ComfyUI workflow-service double that records what was called and
-// simulates the on-disk file set, so we can prove Save-As never consumes the
-// source file (issue #226).
+// A minimal ComfyUI workflow-STORE double (`app.extensionManager.workflow`) that
+// records what was called and simulates the on-disk file set, so we can prove
+// Save-As never consumes the source file (issue #226).
+//
+// STORE vs SERVICE (#1540). `saveInPlace` writes through `svc.saveWorkflow(wf)`,
+// and `svc` is the store, not `useWorkflowService`. Measured on frontend 1.49.6:
+// the store's saveWorkflow is `await e.save()` → `UserFile.save({force:true})`, a
+// write to `this.path`. It does NOT recompute a mode-derived path or rename.
+// Relocating saveWorkflow (`directory + appendWorkflowJsonExt(filename, isApp)`,
+// then `renameWorkflow`) belongs to the service, which the panel does not call.
+// Do not put that behavior back on this double; if a frontend ever exposes the
+// service at `extensionManager.workflow`, model it under an explicitly-named
+// helper rather than silently forking the store.
+//
 // Production-accurate extension derivation (mirrors ComfyUI formatUtil):
 // app-mode workflows persist as "<base>.app.json", everything else as
 // "<base>.json".
@@ -24,6 +35,14 @@ const stripExt = (name) => {
   if (lower.endsWith('.app.json')) return s.slice(0, -'.app.json'.length)
   if (lower.endsWith('.json')) return s.slice(0, -'.json'.length)
   return s
+}
+
+/** Store saveWorkflow: write `wf.path`. Shared by `makeService` and
+ *  `makeFaithfulService` so those two cannot disagree about the call the panel
+ *  makes. `makeStore147Service` writes the same path (through a content Map). */
+function storeSaveWorkflow(svc, wf) {
+  svc.calls.push(['saveWorkflow', wf.path])
+  svc.disk.add(wf.path) // overwrite / create in place — UserFile.save writes this.path
 }
 
 function makeService({ files = [], active } = {}) {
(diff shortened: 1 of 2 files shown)


artokun:
@grok — review requested. Please attack the test-model claim before merge: (1) whether `app.extensionManager.workflow.saveWorkflow` is actually the store write-to-wf.path shape for the supported frontend versions, while the relocating model belongs only to `useWorkflowService`; (2) whether changing the shared default `makeService` double accidentally weakens existing Save-As/source-preservation tests or changes any unrelated model; (3) whether the new #1540 test would still fail if production save routing regressed, and whether the call-list assertion observes the exact production path. No verdict is claimed until your review and CI are complete.

Issue #1447, opened 2026-08-19
panel_add_node reports unrelated ReActor import failure for unknown node type


artokun:
### What happened / root cause
`panel_add_node` for unknown class_type `VideoToImages` returned an error claiming `comfyui-reactor-node` failed to import at startup, even though `ReActorFaceSwap` had just been added successfully in the same live workflow. The missing node type and named pack are unrelated, making the diagnostic misleading.

### Steps to reproduce
1. Create a new workflow.
2. Add `ReActorFaceSwap` successfully.
3. Call `panel_add_node` with an unavailable type such as `VideoToImages`.
4. Observe the error appends an unrelated ReActor startup-import warning.

### Exact error (scrubbed)
`Unknown node type \"VideoToImages\" ... ComfyUI reported that this pack FAILED TO IMPORT at startup — comfyui-reactor-node.`

### Fix
Applied locally: no. Upstream-only: yes. The add-node diagnostic should only attach import-failure context when it owns the requested class_type, or clearly label it as unrelated/stale.

### Environment
Windows 11; ComfyUI 0.33.2; RTX 5070 12GB; comfyui-mcp 0.52.7; panel 0.15.1.

artokun:
Claimed by grok-swarm 2026-08-19T21:03:23Z — driving this in worktree `wt-1447` (branch `fix/1447-reactor-unknown-type`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp-panel/pull/1450 (worktree C:/Users/Artokun/code/comfyui-mcp-panel/.worktrees/wt-1447).

artokun:
Re-engaged by grok-swarm 2026-08-20T00:11:32Z — PR #1450 last commit 21:13Z (>3h), lint-red (`writePoint is not defined`). Driving rebase in existing `wt-1447`.

artokun:
Rebased https://github.com/artokun/comfyui-mcp-panel/pull/1450 onto origin/main (v0.15.15) so `c2bacd7c` injects shipped `writePoint` into the graph_edit_node eval harness — that was the lint-red `#1443 graph_edit_node then ack: writePoint is not defined`.

- `node --test browser_tests/unit/pack-import-failures.test.mjs` — 23/23
- `node scripts/check-tool-vocabulary.mjs` — OK
- `#1443 graph_edit_node then ack` — pass
[...]

PR #1545, opened 2026-08-20
fix(1544): name a failed pack as the cause only when Manager's node map proves it owns the type

### LINKED ISSUE #1544: panel_add_node attributes unknown type to unrelated failed pack
panel_add_node returned a misleading causal diagnosis for an unknown node type. Requesting class_type PreviewVideo correctly found that the type is unavailable, but the error then claimed a completely unrelated failed custom-node pack (coldinfire_fal_privacy) explained the missing type. The diagnostic appears to attach any startup import failure to any unknown class_type without proving ownership.

### Steps to reproduce
1. Have any unrelated custom-node pack fail import at startup.
2. Call panel_add_node with an unknown/unregistered class_type such as PreviewVideo.
3. Observe the error names the unrelated failed pack as the reason.

### Exact error (scrubbed)
Error: Unknown node type "PreviewVideo" - the ComfyUI backend does not provide it ... ComfyUI reported that this pack FAILED TO IMPORT at startup - coldinfire_fal_privacy.

### Fix
Precise change needed: in panel_add_node's unknown-type error builder, only mention a failed-import pack when Manager/object-info metadata maps the requested class_type to that exact pack. If ownership cannot be established, return the generic unknown/uninstalled/failed-import message and list failed packs separately as unrelated diagnostics, not as the cause.

### Environment
Windows 11; ComfyUI 0.33.3; frontend 1.49.6; RTX 4080 16 GB; comfyui-mcp 0.51.17; panel 0.15.23.

artokun:
Closes #1544.

## The bug, reproduced by execution

`panel_add_node PreviewVideo` was refused with `coldinfire_fal_privacy FAILED TO IMPORT at startup` attached. Running the shipped `importFailureNote` on `origin/main` against a live ComfyUI (4498 types, `PreviewVideo` genuinely absent) reproduces the reported string exactly:

> …Check the exact class_type via create_workflow (action:"node_info") **BEFORE INSTALLING ANYTHING: ComfyUI reported that this pack FAILED TO IMPORT at startup — coldinfire_fal_privacy.** … This does not prove it provides "PreviewVideo" — the failure may be unrelated.

The hedge was already there. It was the **last sentence**, behind a lead that named a pack — so it read as an accusation with a footnote. The reporter acted on the lead, which is what leads are for.

## Why it could not do better before — and why that changed

The module's own docblock recorded the constraint:

> Establishing that needs the pack's NODE_CLASS_MAPPINGS, which is not readable from the browser

**That is no longer true, and it was the reason every caller settled for a hedge.** ComfyUI-Manager's `/customnode/getmappings` is keyed `pack → [[classNames…], meta]`, and the panel *already fetches that exact payload* for `panel_search_nodes` — `parseNodeMappings` reads `val[1]` (the meta) and throws `val[0]` (the class list) away. The ownership evidence was one array index from being in hand.

`packsProvidingType` reads it. Ownership now decides the framing:

| evidence | message |
|---|---|
| catalogue ties a failed pack to this `class_type` | `BEFORE INSTALLING ANYTHING: "X" is provided by <pack>…` — #775's causal advice, unchanged |
| catalogue searched, no link | `SEPARATE ISSUE, not the cause of this missing type: …` |
| no usable catalogue came back (`{}`, a proxy's sign-in page) | same, but *ownership could not be checked* |
| map unreadable (Manager off/legacy/unreachable) | same, but *ownership was not checked* |

Those last three are distinguished deliberately: "we did not check", "nothing came back to check against" and "we checked and found nothing" send a reader to different next steps.

The reported case now reads:

> …**SEPARATE ISSUE, not the cause of this missing type:** ComfyUI reported that this pack FAILED TO IMPORT at startup — coldinfire_fal_privacy. Nothing ties it to "PreviewVideo" — ComfyUI-Manager's node map does not link it to that type — so do not reinstall it expecting "PreviewVideo" to appear. That failure is still worth fixing on its own…

## Two defects this PR found in its own code

Both came out of the risk list I posted for review, not out of the diff. Both are fixed here.

**1. A 200 from Manager is not a catalogue** (`dcb7168d`). The unproven branch chose its wording on `typeof nodeMap === "object"`. But #808 already measured that Manager answers `{}` when it built its list from none of channel, cache or bundled copy, and a captive proxy answers 200 with a sign-in page — both objects. So an empty catalogue printed *"the node map does not link it to that type"*: a check that never ran, reported as one that ran and found nothing. `nodeMapPackCount` now counts entries that actually carry a class list.

**2. Ownership was matched on a key that merges distinct packs** (`8112fdb2`). `packKey` strips every non-alphanumeric, so `ComfyUI-OmniSVG` (A043-studios) and `ComfyUI_OmniSVG` (smthemex) — different repos, different authors — collapsed to one key, as did `ComfyUI-QwenVL` / `ComfyUI-Qwen-VL` / `ComfyUI_QwenVL`.

| owner key | keys claimed by >1 entry | …of those, entries **disagreeing** about class names |
|---|---|---|
| `packKey` (strips separators) | 79 | **72** |
| `ownerKey` (separators kept) | — | **3** |

That would have named the **wrong** pack as a proven cause — this issue's bug with more confidence. Ownership now matches on `ownerKey` (case-folded, separators kept); `dropLivePackImportFailures` keeps `packKey`, where loose matching is correct. **The tightening costs nothing:** installed-folder hit rate is 59/75 under both. For the 3 keys that still collide, a key owns a type only when **every** entry filed under it provides it — aliases agree and still promote; two projects sharing a key promote nothing.

## The claim I am least sure of

**`packsProvidingType` is positive-evidence-only, and that asymmetry is the whole safety argument.** The stronger rule — *"the catalogue knows this pack, its class list omits the type, therefore it is not the owner"* — would let the note actively exonerate a pack and **drop** `coldinfire_fal_privacy` outright rather than demote it.

I measured it before trusting it: of the 59 installed packs the catalogue recognises on this machine, **4 share no class at all** with their own `/object_info` entries, because the catalogue lags a pack's current release. That is ~7% of real import failures silently discarded — #775's fault with the sign flipped. Not implemented; a test pins that it stays unimplemented. **If that reasoning is wrong, this PR under-reports real causes.**

Also heuristic, and worth attacking: the folder↔catalogue-key match (last path segment, `.git` stripped). It ties `(IMPORT FAILED): …/ComfyUI-LTXVideo` to `https://github.com/lightricks/ComfyUI-LTXVideo.git`. A pack cloned into a renamed folder will not match — failing toward "unproven", the safe direction, but a genuinely-owning pack can go un-promoted. **59/75 (79%)** measured hit rate.

Third: only the documented MAP shape is read. `parseNodeMappings` also accepts an ARRAY of pack objects, but never reads a class list from one and I observed no array payload carrying one — so rather than invent a field name, an array yields no owners and the caller says ownership was unestablished.

## Cost

The catalogue is ~1.4 MB. `relevantPackImportFailures` is exported so the resolver fetches it **only when a surviving import failure actually needs adjudicating** — 0 fetches when nothing failed, when #1447's live-drop emptied the list, or for a #1523 subgraph UUID; exactly 1 otherwise. Pinned by test.

## Verified by mutation, not by a green suite

Ten mutations across three rounds; **every one went RED, each caught by its own named test**, with the baseline green before and after each:

| mutation | caught by |
|---|---|
| restore the causal-first framing for an unproven pack | `#1544 the reported refusal no longer blames an unrelated failed pack` |
| `packsProvidingType` always returns no owners | `#1544 ownership PROVEN still gets #775's causal note` |
| **panel stops passing `readNodeMap`** (fix ships inert) | `#1544 WIRING: the add-node guard is handed the Manager node map` |
| fetch the node map unconditionally | `#1544 the ~1.4MB node map is read ONLY when a failure needs adjudicating` |
| claim the map was checked when never read | `#1544 no map means ownership was NOT CHECKED, and says so` |
| disown a failure the map does not list | `#1544 POSITIVE evidence only — a map that omits the pack never drops it` |
| treat any object as a consulted catalogue | `#1544 a 200 that is not a catalogue is NOT a completed ownership check` |
| count entries that carry no class list | `#1544 nodeMapPackCount counts what is checkable, not what is present` |
| match ownership on the loose `packKey` again | `#1544 two DIFFERENT packs that normalise alike never prove ownership` |
| drop unanimity among entries sharing a key | `#1544 entries under ONE key must agree before that key owns a type` |

The wiring mutation matters most: the injection is one line in a 2 MB monolith, invisible to every helper-level test, and without that pin the whole check could ship dead. It was caught by that test **and only** that test.

An ordering assertion ("not the cause" must appear *before* the pack name) is the other deliberate one — the pre-#1544 note contained the same hedge words and still misled, so presence alone is not a regression guard.

## Test results

- **Unit: 6078 pass, 0 fail, 1 todo** (`npm run test:unit`, full suite). 16 new tests; #1447's two superseded wording pins updated to assert the *stronger* front-loaded claim plus ordering.
- `npm run typecheck` clean; `check:scope` OK (every name in the monolith resolves — the gate that catches a bad injection).
- **Real-data end-to-end**, against this machine's live ComfyUI and its 5583-entry Manager catalogue: the reported `PreviewVideo` refusal names no cause, and **400/401 sampled packs are still proven owner of their own class** — the collision tightening did not cost promotion.
- **Browser: 63 passed / 18 failed** (`npx playwright test --workers=1`, against a ComfyUI serving this branch). The 18 are **pre-existing**: the same 7 spec files run against unmodified `origin/main` (`9d332968`) on the same instance give a **byte-identical list of 18 failures**. They are MockBridge conversation/persistence/streaming specs, unrelated to this change; no assertion was loosened. Fixing that rig belongs to its own issue.

- **Did not list the packs that *do* provide the type.** The catalogue knows `PreviewVideo` comes from `llm-toolkit`, `comfyui-nodes-erpk` and `vertex-ai-comfyui-nodes`, and naming them would be genuinely useful — but the issue asked for the misattribution to stop, not for install suggestions, and the freeze says bugs only.
- **Did not change the workflow-LOAD path.** With no requested type there is nothing to establish ownership against, so `api-workflow-load.js` keeps #775's exact wording and still reports every failure.
- **Did not drop unproven failures** — see the measurement above.
- **Did not touch `parseNodeMappings`** or `packKey`, though both now share a concern with the new code; changing a live search path or #1447's matcher was more risk than this bug is worth.
- In the proven-ownership branch, only the owner is named; other failures are omitted from *that* message rather than re-listed, which would put the ambiguity straight back.

## Review status

**Review pending — not gated.** `codex-gate.mjs` was deliberately not run (it currently answers from the owner's weakest model, so a SHIP from it would wear an authority it does not have). grok review requested in the comments, with the risky parts named. I have not merged and will not merge on my own verification alone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+- panel_add_node names a failed custom-node pack as the reason a type is missing only when ComfyUI-Manager's node map says that pack provides it; an unproven failure is reported as a separate issue rather than the cause (#1544)
+
 ## [0.15.30] - 2026-08-20
 
 ### Fixed
--- a/browser_tests/unit/pack-import-failures.test.mjs
+++ b/browser_tests/unit/pack-import-failures.test.mjs
@@ -535,9 +535,9 @@ test("#1544 POSITIVE evidence only — a map that omits the pack never drops it"
 test("#1544 packsProvidingType reads the class list, and claims nothing else", () => {
   assert.deepEqual(
     [...packsProvidingType(NODE_MAP, "PreviewVideo")].sort(),
-    ["llmtoolkit", "vertexaicomfyuinodes"],
+    ["llm-toolkit", "vertex-ai-comfyui-nodes"],
   );
-  assert.deepEqual([...packsProvidingType(NODE_MAP, "LTXVAddGuide")], ["comfyuiltxvideo"]);
+  assert.deepEqual([...packsProvidingType(NODE_MAP, "LTXVAddGuide")], ["comfyui-ltxvideo"]);
   // Nothing to go on ⇒ no owners, which is the branch that handles not knowing.
   assert.equal(packsProvidingType(NODE_MAP, "NoSuchType").size, 0);
   assert.equal(packsProvidingType(null, "PreviewVideo").size, 0);
@@ -670,4 +670,58 @@ test("#1544 nodeMapPackCount counts what is checkable, not what is present", () 
   assert.equal(nodeMapPackCount(null), 0);
   assert.equal(nodeMapPackCount("<html>"), 0);
   assert.equal(nodeMapPackCount([["A"]]), 0, "an array payload carries no class list we can read");
+});
+
+test("#1544 two DIFFERENT packs that normalise alike never prove ownership", () => {
+  // Real collision from the live catalogue: ComfyUI-OmniSVG (A043-studios) and
+  // ComfyUI_OmniSVG (smthemex) are separate projects by separate authors. Under the
+  // #1447 `packKey` normalisation — which strips every non-alphanumeric — they are
+  // one key, and a failed A043 install would have been named as the proven owner of
+  // a class only smthemex provides. That is #1544's own bug restated with MORE
+  // confidence, which is strictly worse than the hedge it replaced.
+  //
+  // Measured on the live 5583-entry map: `packKey` leaves 79 keys claimed by more
+  // than one entry, 72 of which disagree about class names. Keeping separators takes
+  // that to 3, at no cost to the installed-folder hit rate (59/75 either way).
+  const COLLIDING = {
+    "https://github.com/A043-studios/ComfyUI-OmniSVG": [
+      ["OmniSVG Model Loader"],
+      { title_aux: "ComfyUI-OmniSVG" },
+    ],
+    "https://github.com/smthemex/ComfyUI_OmniSVG": [
+      ["SVG Saver"],
+      { title_aux: "ComfyUI_OmniSVG" },
+    ],
+  };
+  // Each is proven owner of its OWN class, and of nothing else.
+  assert.deepEqual([...packsProvidingType(COLLIDING, "OmniSVG Model Loader")], ["comfyui-omnisvg"]);
+  assert.deepEqual([...packsProvidingType(COLLIDING, "SVG Saver")], ["comfyui_omnisvg"]);
+
+  // The hyphen pack failed; the underscore pack owns the requested class. Nothing
+  // may present the hyphen pack as the cause.
+  const note = importFailureNote(["ComfyUI-OmniSVG"], {
+    forType: "SVG Saver",
+    liveDefs: CORE_ONLY_DEFS,
+    nodeMap: COLLIDING,
+  });
+  assert.match(note, /SEPARATE ISSUE/);
+  assert.doesNotMatch(note, /is provided by/);
+});
+
+test("#1544 entries under ONE key must agree before that key owns a type", () => {
+  // The 3 collisions that survive separator-preserving keys. Aliases of a single
+  // pack list the same classes and still promote; two projects filed under one key
+  // that disagree promote nothing.
+  const ALIASES = {
+    "https://github.com/x/Same-Pack": [["ClassA"], {}],
+    "https://github.com/x/Same-Pack.git": [["ClassA"], {}],
+  };
+  assert.deepEqual([...packsProvidingType(ALIASES, "ClassA")], ["same-pack"], "agreeing aliases promote");
+
+  const DISAGREE = {
+    "https://github.com/one/Dup-Pack": [["ClassA"], {}],
+    "https://github.com/two/Dup-Pack": [["ClassB"], {}],
+  };
+  assert.equal(packsProvidingType(DISAGREE, "ClassA").size, 0, "one entry omits it ⇒ not proven");
+  assert.equal(packsProvidingType(DISAGREE, "ClassB").size, 0);
 });
--- a/browser_tests/unit/pack-import-failures.test.mjs
+++ b/browser_tests/unit/pack-import-failures.test.mjs
@@ -725,3 +725,30 @@ test("#1544 entries under ONE key must agree before that key owns a type", () =>
   assert.equal(packsProvidingType(DISAGREE, "ClassA").size, 0, "one entry omits it ⇒ not proven");
   assert.equal(packsProvidingType(DISAGREE, "ClassB").size, 0);
 });
+
+test("#1544 a map HIT promotes through the shipped guard, not just the helper", async () => {
+  // Review gap: the other resolver test asserts SEPARATE ISSUE, which is ALSO the
+  // no-map wording — so discarding the fetched map (`await readNodeMap()` without
+  // assigning it) would still have passed every test here. The cost test proves the
+  // injector is CALLED and the wiring test proves the panel PASSES it; neither
+  // proves the fetched payload reaches importFailureNote. This does: the map owns
+  // the type, so the refusal must carry the causal wording.
+  const { assertAddNodeResolvableRefreshing } = await import(
+    "../../web/js/lib/node-resolve.js"
+  );
+  const err = await assertAddNodeResolvableRefreshing({}, "LTXVAddGuide", {
+    getFreshObjectInfo: async () => CORE_ONLY_DEFS,
+    wasTypeEverDefined: () => false,
+    readImportFailures: async () => ["ComfyUI-LTXVideo"],
+    readNodeMap: async () => NODE_MAP,
+  }).then(
+    () => null,
+    (e) => e,
+  );
+  assert.ok(err, "an absent type must still be refused");
+  assert.match(err.message, /Unknown node type "LTXVAddGuide"/);
+  assert.match(err.message, /BEFORE INSTALLING ANYTHING/);
+  assert.match(err.message, /"LTXVAddGuide" is provided by ComfyUI-LTXVideo/);
+  assert.match(err.message, /Ownership is from ComfyUI-Manager's node map/);
+  assert.doesNotMatch(err.message, /SEPARATE ISSUE/);
+});
--- a/browser_tests/unit/pack-import-failures.test.mjs
+++ b/browser_tests/unit/pack-import-failures.test.mjs
@@ -24,6 +24,7 @@ import {
   readPackImportFailures,
   dropLivePackImportFailures,
   packsProvidingType,
+  nodeMapPackCount,
 } from "../../web/js/lib/pack-import-failures.js";
 
 const HERE = dirname(fileURLToPath(import.meta.url));
@@ -635,4 +636,38 @@ test("#1544 WIRING: the add-node guard is handed the Manager node map", () => {
     /readNodeMap: \(\) => managerGet\("customnode\/getmappings\?mode=cache"\)/,
     "panel_add_node must fetch the ownership map, or #1544 ships as dead code",
   );
+});
+
+test("#1544 a 200 that is not a catalogue is NOT a completed ownership check", () => {
+  // #808's finding, one layer down. Manager answers `{}` when it assembled its list
+  // from none of channel, cache or bundled copy; a captive proxy answers 200 with a
+  // sign-in page. Both are objects. Reporting "the node map does not link it to that
+  // type" about either asserts a check that never ran — the same unearned claim
+  // #1544 exists to stop.
+  const forEmpty = (nodeMap) =>
+    importFailureNote(["coldinfire_fal_privacy"], {
+      forType: "PreviewVideo",
+      liveDefs: CORE_ONLY_DEFS,
+      nodeMap,
+    });
+
+  for (const payload of [{}, { error: "sign in" }, { "some-pack": { title: "no class list" } }]) {
+    const note = forEmpty(payload);
+    assert.match(note, /returned no usable node catalogue, so ownership could not be checked/);
+    assert.doesNotMatch(note, /does not link it to that type/);
+    assert.match(note, /SEPARATE ISSUE/, "still non-causal either way");
+  }
+
+  // A populated catalogue that genuinely lacks the link keeps the stronger wording.
+  assert.match(forEmpty(NODE_MAP), /does not link it to that type/);
+});
+
+test("#1544 nodeMapPackCount counts what is checkable, not what is present", () => {
+  assert.equal(nodeMapPackCount(NODE_MAP), 3);
+  assert.equal(nodeMapPackCount({}), 0);
+  assert.equal(nodeMapPackCount({ error: "sign in" }), 0);
+  assert.equal(nodeMapPackCount({ pack: [["A"], {}], broken: { no: "classes" } }), 1);
+  assert.equal(nodeMapPackCount(null), 0);
+  assert.equal(nodeMapPackCount("<html>"), 0);
+  assert.equal(nodeMapPackCount([["A"]]), 0, "an array payload carries no class list we can read");
 });
--- a/browser_tests/unit/pack-import-failures.test.mjs
+++ b/browser_tests/unit/pack-import-failures.test.mjs
@@ -23,6 +23,7 @@ import {
   importFailureNote,
   readPackImportFailures,
   dropLivePackImportFailures,
+  packsProvidingType,
 } from "../../web/js/lib/pack-import-failures.js";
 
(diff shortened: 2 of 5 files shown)


artokun:
@grok please review this one — and start at the reasoning, not the diff. The code is small; the argument underneath it is what can be wrong.

**1. The asymmetry is the whole safety case. Try to break it.**
`packsProvidingType` only ever *promotes* a failure to "the cause". I deliberately did **not** implement the converse ("the map knows this pack, its class list omits the type ⇒ not the owner"), because that would let the note drop `coldinfire_fal_privacy` entirely instead of demoting it. My justification is one measurement on one machine: of the 59 installed packs the live 5583-entry map recognised, **4 shared no class at all** with their own `/object_info` entries. I concluded ~7% false-disown and refused the inference.

Attack that: is 4/59 on a single rig enough to justify the weaker rule, or am I leaving a better answer on the table out of caution? And is the *reverse* risk real — does demoting-instead-of-dropping now bury a genuine cause in cases where the old note happened to be right?

**2. The folder↔map-key match is a heuristic, and it is load-bearing for the good branch.**
`packKeyFromNodeMapKey` takes the last path segment, strips `.git`, normalises via the existing `packKey`. That is what ties `(IMPORT FAILED): …/ComfyUI-LTXVideo` to `https://github.com/lightricks/ComfyUI-LTXVideo.git`. Measured hit rate against real installed folders: **79% (59/75)**. Where does this collide — two packs whose repo names collide after normalisation (`packKey` strips *all* non-alphanumerics, so `my-pack` and `my_pack` and `MyPack` are one key)? A false *positive* here is worse than a false negative: it would name the wrong pack as the proven cause, with more confidence than the bug I am fixing.

**3. The state I am least sure of: Manager reachable but answering something odd.**
I handle "map read" and "map unreadable" as distinct messages. But `managerGet` can return a 200 with a body that is not a catalogue at all (a proxy's sign-in HTML, `{}` from an offline Manager — #808 documents both). That lands in `packsProvidingType` as an object with no owners, so the note says *"the node map does not link it to that type"* — asserting a check that effectively did not happen. Is that a real dishonesty bug? #808 already built `catalogueSize` for exactly this distinction and I did not reuse it. I think this is the most likely genuine defect in the PR.

**4. Cost gate.**
The ~1.4 MB map is fetched only when `relevantPackImportFailures` is non-empty. Is there a refusal path where that still fires often enough to matter, or where `readNodeMap` could be called twice?

Not asking for a rubber stamp — if #3 is a defect I will fix it and re-request. Evidence for everything above (mutation table, measurements, e2e baseline comparison) is in the PR body.

artokun:
Self-correction before anyone spent time on it: **point 3 in my review request was a real defect, and I have fixed it in `dcb7168d`.**

The unproven branch chose its wording on `typeof nodeMap === "object"`. But #808 already established that a 200 from Manager is not a catalogue — it answers `{}` when it built its list from none of channel, cache or bundled copy, and a captive proxy answers 200 with a sign-in page. Both are objects, so an empty catalogue printed *"ComfyUI-Manager's node map does not link it to that type"*: a check that never ran, reported as one that ran and found nothing. That is precisely the #1544 fault one layer down, inside the fix for it.

`nodeMapPackCount` now counts entries that actually carry a class list, and there are three distinct endings: catalogue searched and no link / no usable catalogue came back / map could not be read. All three stay non-causal.

Both mutations went RED, each caught by its named test. Full unit suite **6076 pass, 0 fail**; typecheck and `check:scope` clean.

**Points 1, 2 and 4 are still open and still the ones I want attacked** — especially (1) whether 4/59 on a single rig justifies refusing the negative inference, and (2) whether `packKey` stripping all non-alphanumerics can make two distinct packs collide into one key, which would name the *wrong* pack as a proven cause.

artokun:
**Point 2 was also a real defect — and materially worse than I estimated. Fixed in `8112fdb2`.**

I asked whether `packKey` stripping all non-alphanumerics could collide two distinct packs. It can, and it is not a corner case. Measured against the live 5583-entry catalogue:

| owner key | keys claimed by >1 entry | …of those, entries **disagreeing** about class names |
|---|---|---|
| `packKey` (strips separators) | 79 | **72** |
| separator-preserving | — | **3** |

`ComfyUI-OmniSVG` (A043-studios) and `ComfyUI_OmniSVG` (smthemex) are different repos by different authors and collapsed to one key; so did `ComfyUI-QwenVL` / `ComfyUI-Qwen-VL` / `ComfyUI_QwenVL`. A failed install of one would have been named the **proven owner** of a class only the other provides — #1544's bug restated with more confidence, which is strictly worse than the hedge it replaced.

Two changes:
1. Ownership now matches on `ownerKey` — case-folded, **separators kept**. `dropLivePackImportFailures` keeps `packKey`, where loose matching is correct (it matches a folder against itself across drift). **The tightening costs nothing:** installed-folder hit rate is **59/75 under both**, because Manager clones into a folder named after the repo, separators and all.
2. For the 3 keys that still collide, a key owns a type only when **every** catalogue entry filed under it provides it. Aliases of one pack agree and still promote; two projects sharing a key promote nothing.

Re-verified end to end against the real map and real `/object_info`: the reported `PreviewVideo` refusal names no cause, and **400/401 sampled packs are still proven owner of their own class** — promotion survives the tightening. Both mutations RED, each caught by its named test. Unit suite **6078 pass, 0 fail**.

That is two genuine defects found in this PR's own code (this one and the empty-catalogue case in `dcb7168d`), both from the list I flagged rather than from the diff. **Point 1 is the one still standing and still the one I most want challenged:** refusing the negative inference on the strength of 4/59 measured on a single rig. If that call is wrong, this PR under-reports real causes.

artokun (commented):
COMMENT INTENDED: APPROVE

Reviewed `8112fdb2` (after `dcb7168d` and the ownerKey/unanimity follow-up). Same GitHub user, so this cannot be an Approve event.

## The four questions, on the current SHA

**1. Asymmetry — keep it.** The converse ("map knows this pack, omits the type ⇒ drop") is still wrong. Catalogue lag turning a real import failure into silence is #775 with the sign flipped. Demoting-instead-of-dropping does not bury a genuine cause: a map hit still gets the causal note; a stale/omitted owner stays disclosed as SEPARATE ISSUE plus "still worth fixing". 4/59 on one rig is enough to refuse the stronger rule because the failure mode is silent data loss of a real diagnostic, not a slightly worse sentence. Do not implement the converse.

**2. Folder↔map-key — closed on this SHA.** `packKey` (strip every non-alphanumeric) was a real false-positive path on the SHA you asked me to start from: `ComfyUI-OmniSVG` vs `ComfyUI_OmniSVG` collapse and a failed hyphen folder can be named as proven owner of a class only the underscore pack provides. That is #1544 with more confidence. `8112fdb2` keeps `packKey` for #1447 (folder vs itself) and matches ownership on `ownerKey` (case-fold, separators kept) plus unanimity for the 3 leftover last-segment collisions. The OmniSVG pin and the Dup-Pack disagreement pin are the right tests. A renamed clone folder still fails toward unproven. Good.

**3. Empty 200 — closed on `dcb7168d`.** You were right that this was the most likely genuine defect. `{}`, `{ error: "sign in" }`, and a pack object with no class-list array must not print "the node map does not link it". `nodeMapPackCount` plus the three-way why (`does not link` / `no usable node catalogue` / `could not be read`) is the #808 distinction applied here. HTML string falls through to "could not be read". Keep that.

**4. Cost gate — holds.** `readNodeMap` runs only when `relevantPackImportFailures` is non-empty; empty log, #1447 live-drop, and #1523 UUID are 0 fetches; a real surviving failure is 1. Inner `catch` leaves ownership unestablished if Manager throws. Residual: the GET is not charged against the add-node command budget, so a slow Manager on an already-expensive refusal can still lose the relay window (#1180/#1192). Not a reason to block; worth an AbortSignal on a follow-up if you see timeouts.

## Also checked

- #1447 live-drop still applies; #1523 UUID still yields no note; no-`forType` workflow-load wording unchanged.
- Proven-ownership branch naming only the owner is OK because the unproven branch still lists every surviving failure.
- Tests drive `importFailureNote` / `packsProvidingType` / `relevantPackImportFailures` / `nodeMapPackCount` and pin the panel injector string. One gap: the `assertAddNodeResolvableRefreshing` e2e only asserts SEPARATE ISSUE, which is also the no-map wording, so dropping `noteOpts.nodeMap` after a successful fetch would still pass. The cost test proves the injector is *called* and the wiring test proves the panel *passes* it; nothing proves a map *hit* promotes through the shipped guard. Add a resolver run where the injected map owns the type and assert `BEFORE INSTALLING ANYTHING` / `is provided by`. Suggestion, not a blocker.

I would merge this as a bugfix once CI is green on `8112fdb2`. Not merging from here (same account).

artokun:
Thanks — the test gap you found was real, and it was a hole in my **mutation set**, not just the assertions. Fixed in `749e4452`.

You were exactly right about why it was invisible: `SEPARATE ISSUE` is also the no-map wording, so

```diff
-  noteOpts.nodeMap = await readNodeMap();
+  await readNodeMap();
```

fetched the 1.4 MB catalogue, threw it away, and stayed green across all 40 tests. The cost test proved the injector was *called*; the wiring test proved the panel *passed* it; nothing proved the payload *arrived*. I had mutated the wiring and the gate but never the assignment between them.

The new test runs the shipped `assertAddNodeResolvableRefreshing` with a catalogue that owns the requested type and requires `BEFORE INSTALLING ANYTHING` / `is provided by` / `Ownership is from ComfyUI-Manager's node map`, and refuses `SEPARATE ISSUE`. Re-running that discard mutation now: **RED, caught by that test and only that test.**

On your residual: agreed that the `readNodeMap` GET is not charged against the add-node command budget, so a slow Manager can eat into the relay window (#1180/#1192). I have **not** added an AbortSignal here — it is a real concern but a different one from the misattribution this PR fixes, and threading a budget-derived signal through the resolver is the kind of change that deserves its own issue and its own evidence rather than riding along at the end of this one. Noted in the PR body as deliberately-not-done; happy to file it.

Full state on `749e4452`: unit **6079 pass, 0 fail, 1 todo**; typecheck and `check:scope` clean; 11 mutations across four rounds, every one RED and caught by its named test.

(One honest note on flakiness: a mid-run `npm run test:unit` reported a single manager-install failure — `unverified` vs `installed` — while the Playwright suite was running concurrently on this machine. Re-run with the machine quiet: 0 fail. That is the known load-sensitive flake, not a regression, and I am not counting it as green-by-retry — the clean run is the one I am reporting.)

artokun (commented):
COMMENT INTENDED: APPROVE

Re-review of `749e4452` (HEAD). Prior COMMENT INTENDED on `8112fdb2` still stands; this SHA only closes the hole that review named.

## What changed after the last review

Production code is unchanged. The new test drives the shipped `assertAddNodeResolvableRefreshing` with a map that *owns* `LTXVAddGuide` and requires the causal wording (`BEFORE INSTALLING ANYTHING` / `is provided by ComfyUI-LTXVideo` / no `SEPARATE ISSUE`).

That is the right pin. The other resolver test asserted SEPARATE ISSUE, which is also the no-map wording, so this mutation was invisible:

```
-  noteOpts.nodeMap = await readNodeMap();
+  await readNodeMap();
```

Cost-test (injector called) + wiring-test (panel passes `readNodeMap`) do not prove the payload reached `importFailureNote`. This one does: the assignment in `node-resolve.js` (`noteOpts.nodeMap = await readNodeMap()`) is now load-bearing.

## Rechecked on HEAD

- Panel injector still `readNodeMap: () => managerGet("customnode/getmappings?mode=cache")`. `managerGet` returns `JSON.parse` of the body — the MAP shape `packsProvidingType` reads. Same GET search already uses.
- Asymmetry, `ownerKey`+unanimity, empty-200 three-way why, and the cost gate are untouched.
- Matching still uses the log folder name in the sentence (`ComfyUI-LTXVideo`), not the normalised key — the new assertion is that string.

Same GitHub user, so this cannot be an Approve event. I would merge as a bugfix on `749e4452` once CI is green. Not merging from here.

PR #1552, opened 2026-08-21
fix(mcp#1940): serverDeclaresEmptyComboOptions reads the V2 combo shape

artokun:
Panel-side root cause for **artokun/comfyui-mcp#1940** — "panel_set_widget cannot set ANY promoted COMBO on a subgraph node".

Review is **pending**; nothing here has been reviewed yet.

## The report's framing is a red herring

The issue diagnoses this as a promoted-widget problem: the subgraph node's type is a UUID absent from `/object_info`, so the validator resolves an empty option list and fails closed. It asks for the option list to be resolved from the INNER node instead.

That mechanism does not hold up. The panel already walks the promotion chain to the ultimate concrete node and deliberately refuses to authorize a virtual subgraph type (`set-widget.js:333-340`) — "a virtual intermediate subgraph type is never in /object_info and must be traversed, not authorized". And `CustomCombo` is not missing from `/object_info` at all: it is present, along with `UNETLoader` and `CLIPLoader`, in all 4498 types on this rig.

## The actual cause: a combo spec shape the gate cannot see

`serverDeclaresEmptyComboOptions` is the gate on #507's last-resort "an empty option list is not knowable, so take the value as written". It read the option list off `spec[0]` (`input-asset.js:182-184`):

```js
if (!Array.isArray(spec)) return false;
const options = spec[0];
return Array.isArray(options) && options.length === 0;
```

`spec[0]` is the option array only in the **V1** shape `[[opt, ...], config]`. The now-common **V2** shape puts the list under the config object and leaves the literal type string `"COMBO"` at `spec[0]`. The reported node, verbatim from this machine's `/object_info`:

```json
"choice": ["COMBO", { "multiselect": false, "options": [] }]
```

So `Array.isArray(spec[0])` is false and every V2 combo was filed as "not a combo". The gate's own docblock said it fails closed on "a non-combo spec (a type string like `INT`/`STRING`)" — `"COMBO"` **is** a type string, so the bug was written down in the comment without being seen.

With the gate always false, the accept branch at `set-widget.js:967` is structurally unreachable, the ladder falls through, and the user gets the retryable "may simply be stale" wording that no refresh can ever clear. That is the reported string exactly, including the `after refreshing combo options` frame.

**This has nothing to do with subgraphs.** It reproduces on a bare `CustomCombo` node with no subgraph anywhere.

### Measured on a live ComfyUI 0.33 `/object_info`

| | V1 `[[...], {}]` | V2 `["COMBO", {options}]` |
|---|---|---|
| combo inputs | 3335 | 676 |
| server-declared **empty** | 30 | 11 |
| recognised by the gate **before** | 30 / 30 | **0 / 11** |
| recognised **after** | 30 / 30 | 11 / 11 |

The 11: `CustomCombo.choice`, `HypernetworkLoader.hypernetwork_name`, `PhotoMakerLoader.photomaker_model_name`, `AudioEncoderLoader.audio_encoder_name`, `OpticalFlowLoader.model_name`, `LoadMoGeModel.model_name`, `LoadDA3Model.model_name`, `LoadBackgroundRemovalModel.bg_removal_name`, `LoadTrainingDataset.folder_name`, `GeminiInputFiles.file`, `OpenAIInputFiles.file`.

## The fix — adopt the reader that already existed

`authoritativeComboValues` in `asset-staleness.js` already handles V1 and V2, and was measured for precisely this reason ("V1 ... is now the MINORITY", V2 "silently skipped before"). `serverDeclaresEmptyComboOptions` lives in `input-asset.js` and never adopted it — the same per-call-site adoption gap as #1621.

`input-asset.js` is a **leaf** that `asset-staleness.js` imports FROM, so it cannot import back without a cycle. The reader moves to the leaf and `asset-staleness.js` re-exports it: one canonical implementation, every existing importer untouched, no cycle.

**Strictness is unchanged where it matters.** A remote V2 (`{ remote: { route } }`) and a dynamic V3 both still yield `null`, not `[]` — an UNREAD list is never mistaken for a server-declared-empty one and keeps failing closed, exactly as a non-empty list does.

## Evidence

- **Executed, not read.** The real module against the real `/object_info`: V2 goes 0/11 → 11/11, V1 holds 30/30, `UNETLoader.unet_name` (14 real options) and `KSampler.seed` (not a combo) stay `false`.
- **Delete the fix → named tests fail.** Reverting the two-line body: `mcp#1940 e2e: a V2 combo whose SERVER option list is empty becomes writable` and `mcp#1940: serverDeclaresEmptyComboOptions reads the V2 shape...` both fail (119 → 117 pass, 2 fail).
- **The e2e drives the production path**, not the helper — it calls `runSetWidget` (what `graph_set_widget` delegates to) and asserts the value reached the live widget, so a helper-only fix would not satisfy it.
- **Loosening is also caught.** Mutating the reader so a remote V2 returns `[]` instead of `null` fails the refuses-to-guess assertion.
- Full panel unit suite **6089 pass / 0 fail / 1 todo** (6090); `check-panel-scope` OK; `tsc --noEmit` clean.
- Playwright `--workers=1` running; result posted below before merge.

- **The issue's requested fix** — resolving combo options from the inner node for promoted widgets. The evidence says that path is not what fails, and building it would add a second resolution mechanism next to the one already in `followPromotionToConcrete`.
- **The "Secondary" wording change.** With this fix the reported case no longer reaches that message. The cases still reaching it (remote V2 whose list has not landed, a shape newer than this code) genuinely *can* be resolved by waiting or refreshing, so "may simply be stale" is not wrong for them. Rewording it now would be guessing at a message I can no longer reproduce.
- **The `unet_name` / `clip_name` half of the report.** `UNETLoader.unet_name` is V1 with 14 real options on this rig, so it cannot fail through this path. The report bundles two symptoms under one error string; this PR fixes the `choice` one and I am **not** claiming it fixes the other. That is why it does not auto-close mcp#1940.
- **Anything in the `comfyui-mcp` repo.** grok-swarm holds an active claim on the orchestrator half of #1940; this branch does not touch it.

### DIFF
--- a/browser_tests/unit/asset-staleness.test.mjs
+++ b/browser_tests/unit/asset-staleness.test.mjs
@@ -1611,3 +1611,39 @@ test("#1172 WIRING: the disclosure survives the `refreshed: true` branch (#981's
   // verdict what #1133 deliberately permits via the write path.
   assert.doesNotMatch(code, /empty_combo_lists[\s\S]{0,200}?refreshed = false/, "disclosure, not failure");
 });
+
+test("mcp#1940 review: emptyComboListsOnGraph discloses V2 empties, and still never discloses an UNREAD list", () => {
+  // The sibling of the write-path bug, found by review of its fix. isEmptyComboSpec tested
+  // Array.isArray(spec[0]), true only for V1, so a graph carrying a V2 empty combo
+  // disclosed nothing at all.
+  const defsByType = {
+    CustomCombo: { input: { required: { choice: ["COMBO", { multiselect: false, options: [] }] } } },
+    LegacyEmpty: { input: { required: { pick: [[], {}] } } },
+    V2Full: { input: { required: { pick: ["COMBO", { options: ["a", "b"] }] } } },
+    V2Remote: { input: { required: { pick: ["COMBO", { remote: { route: "/internal/files/output" } }] } } },
+    V3Dyn: { input: { required: { pick: ["COMFY_DYNAMICCOMBO_V3", { options: [{ key: "png" }] }] } } },
+  };
+  const rootGraph = {
+    _nodes: [
+      { type: "CustomCombo" },
+      { type: "LegacyEmpty" },
+      { type: "V2Full" },
+      { type: "V2Remote" },
+      { type: "V3Dyn" },
+    ],
+  };
+  const found = emptyComboListsOnGraph(rootGraph, defsByType);
+  assert.deepEqual(found, [
+    { type: "CustomCombo", widget: "choice" },
+    { type: "LegacyEmpty", widget: "pick" },
+  ]);
+});
+
+test("mcp#1940 review: a REMOTE V2 is unread even when it also carries an empty options array", () => {
+  // The invariant the whole change rests on: unread must never read as empty. A remote
+  // list arrives from a separate fetch, so options:[] alongside it is "not here yet".
+  assert.equal(authoritativeComboValues(["COMBO", { remote: { route: "/r" }, options: [] }]), null);
+  assert.equal(authoritativeComboValues(["COMBO", { remote: { route: "/r" } }]), null);
+  // Without a remote marker an empty list is still a real, server-published answer.
+  assert.deepEqual(authoritativeComboValues(["COMBO", { options: [] }]), []);
+});
--- a/browser_tests/unit/node-resolve.test.mjs
+++ b/browser_tests/unit/node-resolve.test.mjs
@@ -2728,3 +2728,110 @@ test("#1523 add_node: a throwing isLoadedSubgraphType fails closed rather than a
   assert.doesNotMatch(err.message, /graph unreadable/);
   assert.doesNotMatch(err.message, /comfyui-reactor-node/);
 });
+
+// ── mcp#1940 ─────────────────────────────────────────────────────────────────
+// The V2 combo spec. Everything above declares combos in the V1 shape
+// `[[opt, ...], config]`, where the option array IS `spec[0]`. A live ComfyUI 0.33
+// /object_info publishes 676 inputs in the V2 shape instead —
+// `["COMBO", { options: [...] }]` — with the literal type string "COMBO" at
+// `spec[0]` and the list under the config object. `serverDeclaresEmptyComboOptions`
+// tested `Array.isArray(spec[0])`, so every V2 combo was filed as "not a combo" and
+// the #507 accept could never be reached: measured 0 of 11 server-declared-empty V2
+// inputs recognised, against 30 of 30 V1.
+//
+// The reported node is verbatim from that rig:
+//     "choice": ["COMBO", { "multiselect": false, "options": [] }]
+// which made `CustomCombo.choice` permanently unwritable, blamed on a STALE list that
+// no refresh could ever change.
+function customComboFixture(liveOptions = []) {
+  const reg = loadedRegistry(["CustomCombo"]);
+  const widget = { name: "choice", type: "combo", options: { values: liveOptions }, value: "" };
+  const node = { id: 816, type: "CustomCombo", widgets: [widget], constructor: reg["CustomCombo"] };
+  return { reg, node, widget };
+}
+// /object_info in which CustomCombo's `choice` is declared in the V2 shape.
+function customComboObjectInfo(config) {
+  const info = objectInfo(["CustomCombo"]);
+  info["CustomCombo"] = { input: { required: { choice: ["COMBO", config] } } };
+  return info;
+}
+
+test("mcp#1940 e2e: a V2 combo whose SERVER option list is empty becomes writable", async () => {
+  const { reg, node, widget } = customComboFixture([]);
+  const res = await runSetWidget(node, "choice", "Default", {
+    registry: reg,
+    getRegistry: () => reg,
+    getFreshObjectInfo: async () => customComboObjectInfo({ multiselect: false, options: [] }),
+    wasTypeEverDefined: () => true,
+    refreshCombos: refreshFromServer,
+    ...HOOKS,
+  });
+  assert.equal(res.set.value, "Default");
+  assert.equal(widget.value, "Default", "the write reached the live widget");
+  assert.equal(res.empty_option_list, true, "reported honestly as an unvalidatable empty list");
+});
+
+test("mcp#1940 e2e: a V2 combo with a REAL server list is still validated STRICTLY", async () => {
+  // The fix must not turn "V2" into a blanket accept. A V2 spec that publishes a real
+  // list is refreshed and enforced exactly like V1: a member lands, a non-member is
+  // refused and nothing is written.
+  const config = { multiselect: false, options: ["Quality", "Default", "Turbo"] };
+  const ok = customComboFixture([]);
+  const res = await runSetWidget(ok.node, "choice", "Turbo", {
+    registry: ok.reg,
+    getRegistry: () => ok.reg,
+    getFreshObjectInfo: async () => customComboObjectInfo(config),
+    wasTypeEverDefined: () => true,
+    refreshCombos: refreshFromServer,
+    ...HOOKS,
+  });
+  assert.equal(ok.widget.value, "Turbo");
+  assert.notEqual(res.empty_option_list, true, "a real list is not the empty-list path");
+
+  const bad = customComboFixture([]);
+  await assert.rejects(
+    () =>
+      runSetWidget(bad.node, "choice", "Nonexistent", {
+        registry: bad.reg,
+        getRegistry: () => bad.reg,
+        getFreshObjectInfo: async () => customComboObjectInfo(config),
+        wasTypeEverDefined: () => true,
+        refreshCombos: refreshFromServer,
+        ...HOOKS,
+      }),
+    /not a valid option|refused/i,
+  );
+  assert.equal(bad.widget.value, "", "an off-list value is still refused (#240 intact)");
+});
+
+test("mcp#1940: serverDeclaresEmptyComboOptions reads the V2 shape, and still refuses to guess", () => {
+  const defs = {
+    T: {
+      input: {
+        required: {
+          v2empty: ["COMBO", { multiselect: false, options: [] }],
+          v2full: ["COMBO", { options: ["a", "b"] }],
+          // The list is a SEPARATE fetch that has not landed. Unread is not empty.
+          v2remote: ["COMBO", { remote: { route: "/internal/files/output" } }],
+          // The keys select SUB-INPUTS to materialize; they are not an option list.
+          v3dynamic: ["COMFY_DYNAMICCOMBO_V3", { options: [{ key: "png" }] }],
+          v1empty: [[], {}],
+          v1full: [["a"], {}],
+          notacombo: ["STRING", {}],
+        },
+        optional: { v2optEmpty: ["COMBO", { options: [] }] },
+      },
+    },
+  };
+  assert.equal(serverDeclaresEmptyComboOptions(defs, "T", "v2empty"), true);
+  assert.equal(serverDeclaresEmptyComboOptions(defs, "T", "v2optEmpty"), true, "optional inputs count too");
+  assert.equal(serverDeclaresEmptyComboOptions(defs, "T", "v2full"), false);
+  // Both of these are "could not read the list", which must fail CLOSED — never be
+  // mistaken for the server declaring the list empty.
+  assert.equal(serverDeclaresEmptyComboOptions(defs, "T", "v2remote"), false, "an unlanded remote list is not an empty one");
+  assert.equal(serverDeclaresEmptyComboOptions(defs, "T", "v3dynamic"), false, "dynamic V3 keys are not an option list");
+  // V1 and non-combo behaviour is unchanged.
+  assert.equal(serverDeclaresEmptyComboOptions(defs, "T", "v1empty"), true);
+  assert.equal(serverDeclaresEmptyComboOptions(defs, "T", "v1full"), false);
+  assert.equal(serverDeclaresEmptyComboOptions(defs, "T", "notacombo"), false);
+});
--- a/browser_tests/unit/node-resolve.test.mjs
+++ b/browser_tests/unit/node-resolve.test.mjs
@@ -2835,3 +2835,60 @@ test("mcp#1940: serverDeclaresEmptyComboOptions reads the V2 shape, and still re
   assert.equal(serverDeclaresEmptyComboOptions(defs, "T", "v1full"), false);
   assert.equal(serverDeclaresEmptyComboOptions(defs, "T", "notacombo"), false);
 });
+
+// The REPORTED shape end-to-end: a subgraph instance promoting a V2 empty COMBO.
+// mcp#1940 was filed as a promoted-widget bug — "cannot set ANY promoted COMBO on a
+// subgraph node" — so a fix for the combo SHAPE has to be shown to clear the actual
+// reported scenario, not just a bare node. The promotion resolves to the concrete
+// inner CustomCombo (`followPromotionToConcrete`), which is the type whose def the
+// empty-list gate reads; before the fix that read said "not a combo" and the write
+// was refused with a stale-list message no refresh could clear.
+function promotedCustomComboFixture() {
+  const inner = {
+    id: 795,
+    type: "CustomCombo",
+    // Empty live list — nothing for the parent rail to project either.
+    widgets: [{ name: "choice", type: "combo", options: { values: [] }, value: "" }],
+    constructor: { nodeData: { input: { required: {} } } },
+  };
+  const subgraph = { _nodes: [inner], getNodeById: (id) => (String(id) === "795" ? inner : null) };
+  const railWidget = { name: "mode", type: "combo", options: { values: [] }, value: "" };
+  const parent = {
+    id: 816,
+    // The synthetic subgraph UUID from the report — never in /object_info, and never
+    // looked up: the promotion is traversed to the inner type instead.
+    type: "6c697765-ebd7-4e1e-8af0-d84d620be471",
+    subgraph,
+    inputs: [{ name: "mode", _widget: railWidget, _subgraphSlot: { name: "mode" } }],
+    widgets: [railWidget],
+  };
+  const resolveSource = (_n, si) =>
+    si?.name === "mode" ? { sourceNodeId: "795", sourceWidgetName: "choice" } : null;
+  return { parent, inner, resolveSource };
+}
+
+test("mcp#1940 e2e: a PROMOTED V2 empty combo on a subgraph instance is settable from the parent", async () => {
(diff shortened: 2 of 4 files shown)


artokun:
@grok please review. Naming the load-bearing claims rather than asking for a general pass — here is where I think this is most likely to be wrong:

**1. The strictness claim is the whole safety argument, and it rests on `null` vs `[]`.**
`serverDeclaresEmptyComboOptions` is the gate on #507's "take the value as written" — a false TRUE means an off-list value gets written to a combo, which is the #240 violation this gate exists to prevent. I widened what it can read. I claim the widening is safe because `authoritativeComboValues` returns `null` (not `[]`) for a remote V2 and a dynamic V3, so "could not read the list" never becomes "the server says the list is empty". If there is a spec shape where it returns `[]` for a list that is merely *unread*, this fix writes unvalidated values to combos and should not ship. That is the finding I most want someone hunting for.

**2. I moved a function between modules to avoid an import cycle.**
`authoritativeComboValues` moves from `asset-staleness.js` to `input-asset.js` (a leaf), re-exported from its old home. I believe every importer is unaffected because the re-export preserves the old path, and the only consumers are `asset-staleness.js` itself and its test. If the panel bundle loads these differently than `node --test` does, a re-export through a module that imports the same file could behave differently in the browser than in unit tests — and my unit-test evidence would not see it. Panel JS loads as an ES module, so this is a real question, not a hypothetical.

**3. My e2e may prove less than I claim.**
I assert the new `runSetWidget` test exercises the production write path and would not be satisfied by a helper-only fix. I verified that by deleting the fix and watching it fail. What I did NOT verify is that the panel's real `graph_set_widget` reaches `runSetWidget` with the same options I pass in the fixture — in particular `refreshCombos` and `getFreshObjectInfo`. If the fixture's wiring is more generous than production's, the e2e is green for a path users do not take.

**4. I contradicted the issue reporter and could be wrong about it.**
The issue says the cause is promoted-widget/subgraph resolution against a UUID type. I claim that is a red herring and the real cause is combo schema shape, reproducible on a bare node. Evidence: `CustomCombo` IS in `/object_info`, and `set-widget.js:333-340` already traverses virtual subgraph types rather than authorizing them. If the promoted path *does* independently fail, this PR fixes a real bug but not the reporter's, and the issue must stay open — which is why I deliberately did not auto-close it.

**5. Scope I knowingly left.** `UNETLoader.unet_name` is V1 with 14 real options, so it cannot fail through this path, yet the reporter says it produced the same error. I have no explanation for that half and did not attempt one. If you think it shares a cause with the `choice` case, say so — that would mean my root-cause claim is incomplete.

Suite state: 6089/6090 unit pass, `check-panel-scope` OK, `tsc --noEmit` clean, Playwright `--workers=1` in flight (result posted here before any merge). Not merging until this comes back.

artokun:
Answering my own risk **#3** above (does the e2e exercise a path production actually takes), since it is checkable without waiting for review.

There is exactly one production caller of `runSetWidget` — `web/js/comfyui-mcp-panel.js:14692` — and it passes a single options object built at `14398-14689`. Its top-level keys:

```
registry, getRegistry, readObjectInfo, getFreshObjectInfo, refetchObjectInfoLive,
describeObjectInfoFailure, wasTypeEverDefined, schemaProvenance, resolveSource,
canvas, beforeChange, afterChange, setDirty, prepareWriteTarget,
assertTargetStillCurrent, refreshCombos, confirmServerAsset
```

All three of the options my fixture depends on — `getFreshObjectInfo`, `refreshCombos`, `wasTypeEverDefined` — are **present in production**, so the empty-combo ladder the e2e drives is the one a real `graph_set_widget` enters. The fixture also uses the real `refreshComboOptionsFromDefs` as `refreshCombos` (via the file's existing `refreshFromServer`), not a stand-in, so it inherits production semantics including the deliberate never-clobber-a-function-source rule.

**The honest gap that remains:** production passes four options the fixture does not (`prepareWriteTarget`, `assertTargetStillCurrent`, `confirmServerAsset`, `schemaProvenance`). Those are write-target and asset guards, not option-list logic, so I do not expect them to change this verdict — but the direction of the gap is that the fixture is *more permissive* than production, which means the e2e could pass while a real write is still refused by a guard upstream of the combo check. It cannot fail the other way (a green e2e hiding a broken fix), which is the direction that matters for this change.

Risks **#1, #2, #4 and #5** are unchanged and still want a reviewer.

artokun:
## Update — I reproduced the reporter's exact refusal, and the promoted rail is NOT a second cause

This answers risk **#4** above (that I contradicted the reporter and might be wrong), and it is the strongest evidence on this PR. Pushed as `fc168da3`.

I built the reported scenario end-to-end against `runSetWidget`: a `SubgraphNode` whose type is the report's own UUID `6c697765-ebd7-4e1e-8af0-d84d620be471` on node **816**, promoting an inner `CustomCombo`'s `choice` as `mode`, with an empty live list and the V2 server spec.

**With the fix**, the write lands on the inner widget and reports `promoted_from.inner_node_id: 795`.

**Reverting the fix** fails it with the reported string verbatim:

```
panel_set_widget refused "mode" on node 816 (6c697765-ebd7-4e1e-8af0-d84d620be471)
after refreshing combo options: Combo widget "choice" has an EMPTY option list;
the server's option list may simply be stale - refreshing it before deciding.
```

Side by side with the issue's "Exact error" block: same node id, same UUID, same `after refreshing combo options` frame, same message. (Mine addresses the display label `mode`; the reporter tried both that and `choice`.)

So the combo **shape** was the whole cause. The promoted rail resolves correctly and is not independently broken — the subgraph framing in the report is incidental. I no longer hold this as an expectation; it is measured.

### One harness trap worth recording

My first version of the fixture set `wasTypeEverDefined: () => true` and failed with something entirely unrelated:

> the outer subgraph node type was defined by the ComfyUI backend earlier this session but is ABSENT from the current /object_info — its backend was removed (pack uninstalled/disabled)

A subgraph type was **never** a backend type, so answering `true` makes its absence from `/object_info` look like an uninstalled pack (#458) and the write is refused before the combo path is reached. That is a stale-fixture failure, not a product bug — flagging it because a red result there is easy to misread as "the promoted path is broken too", which is exactly the wrong conclusion.

### Status of the risks I raised

- **#4 — resolved.** Measured, not argued.
- **#3 — resolved** by the wiring evidence above.
- **#1 (`null` vs `[]` strictness), #2 (the cross-module move under the browser's ES-module loader), #5 (the unexplained `unet_name` half)** are unchanged and still want a reviewer. #1 is still the one that would sink this.

Playwright `--workers=1` is at 38/81; posting the result here before any merge regardless of what it says.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `fc168da3e450177725e6073f76747edd8c4902fa`. Same GitHub user, so this cannot be an Approve event. Answering the five named risks against shipped `authoritativeComboValues` / `serverDeclaresEmptyComboOptions` and the `#507` accept at `set-widget.js:965-978`.

**1. `null` vs `[]` — the reader does not turn unread into empty.** The body is a MOVE of the already-shipped `#1193` function, not a rewrite:

```js
if (!Array.isArray(spec)) return null;
if (Array.isArray(spec[0])) return spec[0];
if (spec[0] === "COMBO" && Array.isArray(spec[1]?.options)) return spec[1].options;
return null;
```

- Remote V2 `{ remote: { route } }` (no `options` array) → `null` → gate false.
- Dynamic V3 `["COMFY_DYNAMICCOMBO_V3", { options: [{ key }] }]` → `spec[0] !== "COMBO"` → `null` → gate false.
- Non-combo `["STRING", {}]` / missing spec / missing type → `null`/`false`.
- V2 with a real list → membership, not `#507`. The e2e pins an off-list `Nonexistent` is refused and the widget is unchanged.

`options: []` on a V2 COMBO is the same statement V1 `[[], {}]` already authorized: the server published a readable empty list. That is `#507`/`#1133` (zero files on disk, or a client-populated combo). CustomCombo.choice is that shape verbatim. The 11 empty V2 inputs are a different population from the 1 remote V2 `#1193` measured; remote is the unread case and still fails closed.

The `#507` accept is still double-gated (`latest.emptyOptions` AND the server oracle) and still last after refresh + upload confirm. A function-valued live source that happens to return `[]` while `/object_info` publishes a real list still cannot pass the oracle — that is the codex-round-2 hole this gate exists to keep shut.

**2. The module move is ordinary ESM.** `input-asset.js` has zero imports (leaf). `asset-staleness.js` imports the binding and re-exports it. No cycle. Browser native ESM and `node --test` share live bindings; a re-export of an imported binding is not a second copy. Existing `#1193` tests still import `authoritativeComboValues` from `asset-staleness.js`; `serverDeclaresEmptyComboOptions` already imported from `input-asset.js`. Production `set-widget.js` does not go through the re-export.

**3. e2e path — production `runSetWidget`, real `refreshComboOptionsFromDefs`.** Same wiring as the `#507` StarNodes tests (`refreshFromServer` → `refreshComboOptionsFromDefs`). The fixture is more permissive than production (`prepareWriteTarget` / `assertTargetStillCurrent` / `confirmServerAsset` / `schemaProvenance` omitted) — that can hide an *upstream* refusal, not a broken empty-list accept. The delete-the-fix failure on the named tests is the right direction.

**4. Promoted rail is not a second cause.** `followPromotionToConcrete` already refuses to authorize the UUID (`set-widget.js:333-340`). The promoted e2e writes `mode` on the report's UUID/`816`, lands on inner `795.choice`, and reports `promoted_from.inner_node_id: 795`. `wasTypeEverDefined` answering only `CustomCombo` is the correct oracle: `true` for the UUID is the `#458` uninstalled-pack trap, not this bug.

**5. `unet_name` / `clip_name` — correctly not claimed.** `UNETLoader.unet_name` is V1 with a real list on that rig, so this gate stays false. If that half still refuses, it is a different path. Do not auto-close mcp#1940.

Residual, not blocking: `isEmptyComboSpec` in `asset-staleness.js` still tests `Array.isArray(spec[0]) && length === 0`, so `emptyComboListsOnGraph` will not disclose V2 empties. That is the `#1172` note, not the write path; the canonical reader now lives one module away and this sibling did not adopt it. Same class of leftover as the bug this PR fixes — follow up, do not hold the write.

Residual, not blocking: the reader does not fail closed on `{ remote, options: [] }` together. Measured remote V2 has no `options` array; a `if (spec[1]?.remote) return null` would be defense in depth, not required by the live shapes.

Do not close mcp#1940 from this PR. Do not make unread (`null`) look like empty (`[]`). Playwright still in flight per the PR thread — treat this as unit + `runSetWidget` e2e until that result is posted.

artokun:
## Browser gate: I could NOT obtain coverage for this branch, and I am not claiming any

I ran `npx playwright test --workers=1` as required and it is producing failures — but they are **not evidence about this PR in either direction**, and the reason matters more than the count.

`playwright.config.ts` targets `PLAYWRIGHT_BASE_URL || http://localhost:8188`. The panel JS the browser loads comes from **that ComfyUI's** `custom_nodes/comfyui-mcp-panel`, which is a different checkout from this worktree. I verified what is actually under test by fetching the file the browser gets:

```
$ curl -s http://127.0.0.1:8188/extensions/comfyui-mcp-panel/js/lib/input-asset.js | grep -n "const options = spec\[0\]"
183:    const options = spec[0];

$ ... | grep -c "authoritativeComboValues"
0
```

**The served panel still contains the pre-fix line and zero occurrences of the new symbol.** The code under test is `fix/1280-video-card-poster-and-two-stage-visibility` at `d5025cfd`, not this branch. So the failures — `autostart-mcp`, `chat-history-v2`, `conversation-persistence`, all failing at `.cmcp-root never mounted` — are failures of that commit and cannot be attributed to this change, which is not present in the bytes the browser loaded.

That also means the run gives me **no positive signal**. I am explicitly not claiming browser coverage for this PR.

I did not fix the harness by re-pointing the shared ComfyUI's `custom_nodes` checkout at my commit. That is a live working tree on someone else's branch, driven by other agents on this machine, and commandeering it to green my own PR is not a trade I get to make unilaterally. If you want browser verification before this lands, that is the step — and it needs to be done by whoever owns that instance.

Worth noting the harness already diagnoses itself: the mount error text says "ComfyUI must be serving the SAME commit — its custom_nodes copy provides the ./lib/* imports, and a version mismatch there fails the import silently."

### What I *can* stand behind

- Full panel unit suite **6090 pass / 0 fail / 1 todo** (6091), run in this worktree against this code.
- `check-panel-scope` OK (the gate that catches a name resolving in a sibling function), `check-tool-vocabulary` OK, `tsc --noEmit` clean.
- Both mutation directions caught, including the reporter's verbatim refusal string.
- **Risk #2 (the cross-module move under the browser's ES-module loader) partially addressed by execution**: there is no bundler — `web/js` is served as raw ES modules — and loading all three modules through their real specifiers gives `asset-staleness.authoritativeComboValues === input-asset.authoritativeComboValues` (`true`, same function object), with `set-widget.js` importing cleanly. That is the failure mode I was worried about, checked directly rather than argued. It is still not the same as the panel mounting in a real browser.

artokun:
Both residuals from the review are **fixed**, not deferred — pushed as `8d8166dd`. Taking them in this PR rather than as follow-up because deferring a per-call-site adoption gap is literally how the original bug survived; leaving a second one open while the lesson is on the screen would repeat it.

### Residual 1 — `isEmptyComboSpec` (the disclosure path). Fixed.

Confirmed exactly as described: `Array.isArray(spec[0]) && length === 0`, with a docblock still asserting "A combo's spec is `[[opt, ...], config?]`". So `emptyComboListsOnGraph` (#1172) silently disclosed nothing for a graph carrying a `CustomCombo` or `HypernetworkLoader`. It now reads through `authoritativeComboValues`.

New test pins the whole matrix at once — V1 empty and V2 empty are disclosed; V2-with-a-real-list, remote V2 and dynamic V3 are not:

```js
assert.deepEqual(found, [
  { type: "CustomCombo", widget: "choice" },
  { type: "LegacyEmpty",  widget: "pick"   },
]);
```

Reverting `isEmptyComboSpec` to the `spec[0]` read fails it.

### Residual 2 — `{ remote, options: [] }` together. Fixed.

You called this defense in depth rather than a live shape, and that is right — measured remote V2 carries no `options` array. I took it anyway, because it guards the exact invariant you identified as load-bearing in risk #1: that combination is the one case where "not here yet" would become "the server says empty", which is precisely what would let the #507 accept write an unvalidated value.

```js
if (spec[0] === "COMBO" && spec[1]?.remote) return null;
```

Dropping that line fails the new unread-is-not-empty test.

### State

- Suite **6092 pass / 0 fail / 1 todo** (6093); `check-panel-scope` OK; `check-tool-vocabulary` OK; `tsc --noEmit` clean.
- Every fix in this PR is now pinned in **both** directions — delete it and a named test fails, loosen it and a different named test fails.

### Still outstanding, and I am not merging past it

**Playwright gives no signal on this branch.** I verified the ComfyUI the suite drives is serving the pre-fix bytes (`const options = spec[0]` at line 183, zero occurrences of `authoritativeComboValues`), so its failures belong to `fix/1280` @ `d5025cfd` and its passes prove nothing about this code. Detail in the comment above. I did not re-point that shared checkout at my commit to manufacture a green.

So: unit + `runSetWidget` e2e + rig measurement, exactly as your note says to treat it — **no browser coverage claimed**.

Two things I am deliberately leaving to you rather than doing myself:

1. **Merging.** Your review is `COMMENT INTENDED: APPROVE` and notes it cannot be an Approve event because it is the same GitHub user. I am not going to treat my own account's comment as the independent sign-off and merge on it. If you are content with unit-level coverage and no browser verification, say so and I will merge; otherwise this wants either a real browser run or a second pair of eyes.
2. **mcp#1940 stays open** — agreed, and for the reason you gave: the `unet_name` / `clip_name` half is unexplained by this gate. I have also asked artokun/comfyui-mcp#1943 to drop its `Fixes #1940`, since by its own body the write is still refused there.

artokun:
**Playwright final: 20 failed / 61 passed (13.4m), `--workers=1`.** Recording the number for completeness, but it is a measurement of the **wrong commit** and I am not treating it as a result for this PR in either direction.

The suite drives `localhost:8188`, and I confirmed by HTTP that the ComfyUI there serves the pre-fix bytes — `const options = spec[0]` at line 183, zero occurrences of `authoritativeComboValues`. So those 20 failures belong to `fix/1280-video-card-poster-and-two-stage-visibility` @ `d5025cfd`, and the 61 passes are not coverage of this branch. The dominant failure mode is `.cmcp-root never mounted`, whose own error text names the cause: "ComfyUI must be serving the SAME commit".

I could have produced a green by re-pointing that shared checkout at my commit. I did not: it is a live working tree on someone else's branch, driven by other agents on this machine.

**So the honest coverage statement for this PR is: unit + `runSetWidget` e2e + live-`/object_info` measurement, and NO browser verification.** That is a real gap for a panel-visible change and I would rather name it than paper over it.

Worth fixing at the harness level separately — a worktree run currently produces a confident-looking pass/fail count about whatever commit the shared instance happens to serve, which is misleading in both directions. Cheap guard for anyone running this: fetch the file the browser gets and grep for your own change before believing the result.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `8d8166dd1e4a2c527ad09d47027b46fecb13448a`. Fresh review — prior APPROVE was on `fc168da3` and does not cover this SHA. Same GitHub user, so this cannot be an Approve event.

Both residuals from `fc168da3` are closed in the shipped readers, not deferred.

**1. `isEmptyComboSpec` now shares the canonical reader.** `emptyComboListsOnGraph` (#1172) was the same V1-only `Array.isArray(spec[0])` blindness one module away from the write-path fix. It now is:

```js
const values = authoritativeComboValues(spec);
return Array.isArray(values) && values.length === 0;
```

The new test pins the matrix: V1 empty + V2 empty (`CustomCombo.choice`) disclosed; V2-with-a-list, remote V2, and dynamic V3 not disclosed. Reverting the body to `spec[0]` fails it. Write and disclosure now cannot drift without a named test dying.

**2. Unread never becomes empty, including `{ remote, options: [] }`.** Guard is *before* the V2 options read:

```js
if (spec[0] === "COMBO" && spec[1]?.remote) return null;
if (spec[0] === "COMBO" && Array.isArray(spec[1]?.options)) return spec[1].options;
```

So `serverDeclaresEmptyComboOptions` cannot authorize a `#507` blind write on a remote list that happens to carry a placeholder `options: []`. Measured remote V2 still has no `options` array; this is the defense-in-depth case that would have broken the safety argument. Dropping the line fails the new test.

V1 `[[opts], config]` is still decided first — a hypothetical `[[], { remote }]` would still read as empty. Not a live shape; do not reorder unless one shows up.

`null` vs `[]` still holds for the original five risks. `#507` remains double-gated (`latest.emptyOptions` AND the server oracle) and last after refresh + upload confirm. Do not auto-close mcp#1940 (`unet_name` / `clip_name` unexplained). Playwright is honestly no-signal (shared 8188 serving `fix/1280` @ `d5025cfd`, pre-fix `spec[0]` still at line 183) — treat as unit + `runSetWidget` e2e, as on the previous SHA.

Not merging from here.

Issue #1562, opened 2026-08-21
After ComfyUI restart, panel tab stays disconnected; panel_refresh_nodes 3s budget fails on 25MB object_info


artokun:
### What happened / root cause
Two related recovery failures on a large custom-node install after installing a missing node pack and restarting ComfyUI.

1) panel_restart_comfyui dispatched the reboot, saw ComfyUI go down, then reported "not healthy within 127s" (confirmed_cycle:false). ComfyUI DID come back (health check succeeded: 0.33.0, frontend 1.49.6, queue idle). But the Agent panel socket never reconnected: subsequent panel_* calls fail with "Connected: none — the ComfyUI panel tab is not connected". panel_set_workflow_target(mode:current) only deferred. The install itself succeeded (GACLove/ComfyUI-VFI present, /object_info/RIFEInterpolation returns the node).

2) panel_refresh_nodes cannot re-pull defs on this machine: GET /object_info is ~25,104,088 bytes and takes ~20.8s, but refresh splits a ~3s budget (~1499ms /object_info share of 2999ms) so every attempt returns refreshed:false, reason:object_info_fetch_failed.

Likely root cause: restart health timeout is too short for a huge custom_nodes tree (import times dump is enormous), and refresh_nodes timeout is not sized for a 25MB object_info payload. After restart the frontend tab drops the panel websocket and never auto-reattaches.

### Steps to reproduce
1. ComfyUI with a very large custom_nodes install (this host: 200+ packs, object_info ~25MB / ~21s).
2. Install a missing pack via panel_install_node (here: https://github.com/GACLove/ComfyUI-VFI).
3. Call panel_restart_comfyui.
4. Observe: restart tool times out at ~127s even though the server later becomes healthy; panel_* then reports Connected:none.
5. Call panel_refresh_nodes: fails with object_info_fetch_failed (3s budget vs 21s fetch).

### Exact error (scrubbed)
panel_restart_comfyui: rebooting:true, ready:false, confirmed_cycle:false, recovered_ms:127254, probes:603, saw_down:true — "has not become healthy within 127s"

panel_refresh_nodes: refreshed:false, reason:object_info_fetch_failed, "GET /object_info did not answer within its 1499ms share of the 2999ms budget"

panel_add_node / panel_list_workflows / panel_set_todo: "no connected tab ... Connected: none — the ComfyUI panel tab is not connected"

[...]

PR #1566, opened 2026-08-21
fix(1562): panel_refresh_nodes spends the window it holds, and stops blaming a server it never saw fail

artokun:
Addresses the **`panel_refresh_nodes` half** of artokun/comfyui-mcp-panel#1562. The reconnect half of that report is **not** in this PR and is filed on its own — see "What this deliberately does not do".

> **Review is PENDING.** grok has been asked, in a comment here and on the issue, with the risky parts named. This is not gated and no clean verdict exists yet.

---

### Update — rebased, and a P1 found by the fallback gate and fixed

* **Rebased onto `main` @ `a1844f68`** after #1561 (the #1560 fix) landed and made this PR `CONFLICTING`. The only overlap was the oracle import block; nothing #1561 shipped is modified. The two `chore(1562)` commits — which existed solely to keep that import on one line while #1560's branch was open — are gone.
* **`claude-gate.mjs`, the repo's fallback adversarial gate, returned NO-SHIP with a real P1**, disclosed in full in a comment below: route 1's ending was classified as an abandonment even when an earlier attempt in the same retry loop had FAILED for real, so the verdict could contradict its own evidence clause and tell the user of a genuinely-restarting backend that a retry would not help. Fixed in `603f2b37` (`&& !sawRealError` — the loop's own precedence rule, applied across attempts as well as across routes) and covered **behaviourally** against the shipped fetch phase rather than by asserting source text.
* **Re-verified on `603f2b37`:** `npm run test:unit` **6162 pass / 0 fail / 1 todo**; **9 / 9 mutations killed** (the 8 below re-run on this head, plus the new one), tree clean after each; CI green.
* **The browser suite has not been re-run since the rebase** — both ComfyUI slots on this machine are held by another agent's run, and this session was not allocated an instance. The earlier `--workers=1` run (19 failed / 62 passed, byte-identical failure sets on both trees) was on the pre-rebase head. Stated rather than implied.
* The mutation table further down says **14/14** and was true of the pre-rebase head; the current head's set is the **9/9** above.

## The bug, reproduced by execution before a line was written

The reporter's install serves `GET /object_info` at **25,104,088 bytes / 20.84 s** while `/system_stats` and `/object_info/<Type>` answer instantly. Every `panel_refresh_nodes` returned:

```
refreshed:false  reason:object_info_fetch_failed
"GET /object_info did not answer within its 1499ms share of the 2999ms budget"
```

…against a command whose own window is **25,000 ms**. It had spent 7.5 s and refused with 17.5 s unused.

I drove the **SHIPPED fetch phase** — the real `boundedGetNodeDefs`, the real `fetchNodeDefsWithRetry` schedule, the real `object-info-oracle.js`, the real `nodeDefsBudgetLeft` arithmetic and the real constants read out of the panel source — against a **REAL** http server answering a **REAL** 25,000,581-byte `/object_info` in 20.80 s:

```
GET /object_info (raw)    : 25000581 bytes, 20.80s
NODE_DEFS_RUN_BUDGET_MS   : 9000
NODE_DEFS_FETCH_SHARE     : 0.666… → route 1 bound 6000ms

elapsed ms                : 7509
whole /object_info reqs   : 2
defs obtained             : NONE
route failures:
  - api.getNodeDefs() failed: api.getNodeDefs() did not answer within this refresh's remaining budget
  - GET /object_info did not answer within its 1492ms share of the 2985ms budget
```

The reported sentence, byte for byte, and **two** whole-document downloads for a document neither of them could finish.

## THE TWO BUDGETS WERE THE WRONG WAY ROUND

`REFRESH_NODES_COMMAND_BUDGET_MS` (25,000) is how long the command **waits**. `NODE_DEFS_RUN_BUDGET_MS` (9,000) is how long the run it starts may **spend**, and that default is sized — by its own note — for a refresh that happens as a **sub-step** of `graph_add_node`, where time has to be held back for registration and the widget wait.

Because the run's allowance was the *smaller* of the two, **the run always died before the join could abandon**. So `REFRESH_JOIN_ABANDONED` → `refresh_still_running` — the retryable verdict #1404 built for exactly *"a big install … without any concurrency at all"*, whose whole point is that the run is **not cancelled** and the retry **joins it** — was **structurally unreachable on the installs it was written for**. The caller got `object_info_fetch_failed` instead, which tears the run down, so every retry started another doomed 9 s run. The tool could not succeed on that machine, ever.

## What suffices — measured, not picked

Same harness, same 20.80 s document, sweeping only the run budget:

| run budget | route 1 bound | whole-document GETs | outcome |
|---|---|---|---|
| **9,000 ms** (shipped) | 6,000 ms | 2 | FAILED at 7.5 s — the reported sentence |
| **25,000 ms** (= the command's own window) | 16,666 ms | 2 | **STILL FAILED** at 20.9 s |
| 32,000 ms | 21,333 ms | 1 | obtained 20,020 types at 20.86 s |
| **37,500 ms** | 25,000 ms | 1 | obtained 20,020 types at 20.98 s |

The 25,000 row is the reason this is not "hand the run the command's window": the **fetch share is two thirds** of whatever the run gets, and two thirds of the window is not the window. So the number is **derived from the property that has to hold**, not fitted to one install:

```
the fetch phase alone must be able to spend everything the command will wait for
  →  runBudget × NODE_DEFS_FETCH_SHARE  ≥  REFRESH_NODES_COMMAND_BUDGET_MS
```

`REFRESH_NODES_RUN_BUDGET_MS = Math.ceil(REFRESH_NODES_COMMAND_BUDGET_MS / NODE_DEFS_FETCH_SHARE)` = 37,500, and **that inequality is asserted by test**, so raising either input cannot silently break it again.

**What the reporter's install actually gets — corrected in round 2, because the first wording was optimistic.** The fetch lands at ~21 s, inside the 25 s join. The combo phase then runs, and on that install it re-fetches the same 25 MB document, so it is abandoned at its own bound rather than answering — which puts the RUN at roughly 30 s, past the 25 s join. So the honest expectation is: the first `panel_refresh_nodes` answers **`refresh_still_running`** (retryable, and the run is not cancelled), and **the refresh itself completes** — the definitions are registered, which is the outcome the missing node depends on. Before this change the run was torn down at 9 s having obtained nothing, every retry started another doomed run, and `refreshed:true` was not the thing that was unreachable: **any** completed refresh was. What this PR makes reachable is the work landing at all, and a verdict that is true rather than one that blames a healthy server.

## What it costs, stated rather than left to be discovered

* **A healthy install: nothing.** Every one of these numbers is a ceiling. Measured with a fast backend under both budgets: **312 ms vs 322 ms**, one request, identical verdict. (The ~0.5 s figures quoted in the new comment — 456 ms warm, 1,062 ms cold — are **#1180's** measurements on this rig, not mine; attributed as such in the source.)
* **A run that outlives its command.** A `panel_refresh_nodes` against a backend that never answers now leaves one whole-document download in flight for up to ~12,500 ms after the reply, holding the coalescer's single slot. That is exactly the work the retry the reply asks for will join — the trade #1351 already made deliberately for the join — but it is real.
* **`graph_add_node` during that window** meets a longer-lived run and can refuse with its worded, retryable busy message where it would previously have waited out a shorter one. On an install this budget changes anything for, that add was failing its own refresh already.
* **Only the explicit command passes it.** Reconnect, install-completed and download-triggered refreshes keep `NODE_DEFS_RUN_BUDGET_MS`, so #1180's composition property — two serialized runs inside the bridge's read default — is untouched for every caller that is not this one.

## Second half of the same report: the refusal named a cause it never established

The `object_info_fetch_failed` remedy read:

> The backend may still be restarting — retry once it answers, and if it never does, **check that the ComfyUI server process is still running.**

The reporter's server returned 25,104,088 bytes in 20.84 s while that sentence was printed. **Every route had been ABANDONED AT ITS BOUND; not one of them had failed.** That is #982's defect again, and it is the same class this repo keeps recording: a report asserting an outcome it did not observe.

So when — and only when — every whole-document route ended by being abandoned at its bound, the remedy now says that, says it is **not** evidence the server is down, says a plain retry meets the same bound, and names the two things that actually change the answer (reload the tab, whose page load fetches the schema under no such bound; or reduce the pack count).

**Decided from TAGS, never from prose (#1223):** route 1 from the fetch loop's own `lastAttemptTimedOut`, route 2 from the oracle's `TRANSPORT_OUTCOME.NO_ANSWER`. A route that genuinely **failed**, or that **answered unusably** (a 500), outranks the other's silence and keeps the old, correct wording. `null` — no whole-document route was reached at all — falls to the wording that claims less, by `=== true` rather than truthiness.

## Mutation evidence — 14 mutations, 14 killed

Applied to a **committed** tree and reverted with `git checkout` (baseline 0 failing, 0 cancelled; worktree verified clean after each), over `refresh-nodes-run-budget`, `refresh-nodes-command-budget`, `node-def-refresh` and `single-node-def`:

| # | mutation | verdict | a named test it kills |
|---|---|---|---|
| M1 | **delete the FIX** — the call site stops stating the run's allowance | KILLED (2) | `#1562: refresh_nodes hands the RUN its own allowance, not just the join` |
| M2 | the **CALL SITE** keeps the option but names the sub-step default | KILLED (14) | `#1562: refresh_nodes hands the RUN its own allowance…` |
| M3 | the coalescer stops **FORWARDING** it to the run | KILLED (4) | `#1562: the coalescer forwards runBudgetMs on the NOTHING-IN-FLIGHT path` |
| M4 | the **TRAILING** (forced) path drops it | KILLED (1) | `#1562: the coalescer forwards runBudgetMs on the TRAILING (forced) path` |
| M5 | the **PAYLOAD** path drops it | KILLED (1) | `#1562: the coalescer forwards runBudgetMs on the PAYLOAD path` |
| M6 | the run **IGNORES** the allowance it is handed | KILLED (2) | `#1562: a stated run budget is used; anything unusable falls back to the default` |
| M7 | the run **VALIDATES NOTHING** — Infinity/NaN restore the unbounded run | KILLED (2) | same |
| M8 | the **DERIVATION** shrinks to the command window (the value measurement showed still fails) | KILLED (2) | `#1562: the run's FETCH SHARE alone covers everything the command will wait for` |
| M9 | route 1's ending **assumed** rather than read from the timeout flag | KILLED (2) | `#1562: a route that genuinely FAILED outranks the other's silence` |
| M10 | route 2's **TAG** dropped — silence claimed on route 1's evidence alone | KILLED (2) | `#1562: a route that ANSWERED unusably outranks the other's silence too` |
| M11 | route 2's ending decided by matching its **PROSE** instead of its tag (#1223) | KILLED (1) | `#1562: the fetch phase classifies route endings by TAG, not by message text` |
| M12 | the honest remedy **deleted** — every fetch failure blames the server process again | KILLED (2) | `#1562: BOTH whole-document routes abandoned at their bound is not reported as a dead server` |
| M13 | the remedy fires on an **UNESTABLISHED** fact (`null` reads as abandoned) | KILLED (1) | `#1562: a route that genuinely FAILED keeps the old, correct remedy` |
| M14 | the verdict stops **accepting** the input | KILLED (16) | `#1562: every route ABANDONED AT ITS BOUND is not reported as a server that is down` |

**M5 survived the first round, and that is the finding worth recording.** The payload-branch test released the held run *before* issuing the payload call, so it took the nothing-in-flight branch and never touched the code it named — a test on the wrong branch reads exactly like a passing one. Fixed in its own commit, along with lifting the `runDeadline` expression *inside* the test so a mutation fails **by name** rather than failing the whole file (a file-level throw counts as a kill and names nothing a reader can act on).

## Two source-pinning assertions were UPDATED, not loosened

Both pin code this change deliberately moves, and both keep the property they exist for:

* `#1180: a whole refresh RUN, not each phase, is what fits the budget` pinned the literal `let runDeadline = monotonicNow() + NODE_DEFS_RUN_BUDGET_MS;`. It now pins the new expression **and** additionally asserts there is exactly **one** `let runDeadline =` in the panel — strictly stronger than what it asserted before.
* `#1404: the shipped call site passes the budget` pinned `{ force: true, joinMs: … }`. It now requires **both** `joinMs` **and** `runBudgetMs` by name.

Neither assertion was deleted, skipped, or weakened. M1/M2/M6/M7 all still die to them.

## Suites

* `npm run test:unit` (i18n × 2, tool-vocabulary, panel-scope, `node --test`): **6130 tests / 6129 pass / 0 fail / 1 todo**. Baseline on clean `origin/main` @ `75728263`: **6116 / 6115 / 0 fail / 1 todo**. +14 new, nothing pre-existing disturbed.
* `check-panel-scope` OK, `check-tool-vocabulary` OK.
* Playwright — table below.

## What this deliberately does not do

* **It does not make `panel_refresh_nodes` succeed on an install slower than the command's whole window.** Nothing inside a 30,000 ms relay window can pay for a whole-document read that takes longer than that, and this PR does not pretend otherwise — it makes the panel spend the window it actually holds, and when even that is not enough it says so truthfully instead of naming a cause it did not establish.
* **It does not touch #1560/#1561's `scoped-object-info.js`, and does not reuse it.** #1561 states plainly that `panel_refresh_nodes` is unhelped by the type-scoped route because it re-registers the **whole** schema — an install-wide question no per-class route can answer — and that is correct. Nothing here reads, writes, or depends on that file, and no function or test file is shared. This branch is cut from `origin/main` @ `75728263`, the same base #1561 measured against.

  **There IS one textual conflict, and it is declared rather than left to be discovered.** `git merge-tree origin/main <#1561 head> <this head>` conflicts on exactly **one hunk**: the `./lib/object-info-oracle.js` import line in `comfyui-mcp-panel.js`. I add `TRANSPORT_OUTCOME` to it; #1561 rewrites the `./lib/object-info-snapshot.js` import immediately below it. I already reduced my side to a **single changed line** (an earlier version expanded the import across multiple lines and made the conflict bigger for no reason). Whichever of the two merges second resolves it by keeping the other's snapshot import and adding `TRANSPORT_OUTCOME` to the oracle import — nothing semantic is in dispute. Flagged on #1561 as well, so neither of us diverges quietly.
* **It does not fix the reconnect half of #1562**, which is the more serious one. That is filed as **artokun/comfyui-mcp#1996** with what I could establish (the health check polls `GET <loopback boot base>/system_stats`, certifies only on `healthy`-after-`down`, and gave up at the **120 s** `COMFYUI_PANEL_REBOOT_BUDGET_S` default against a 200-pack boot — while its own file declares 240 s safe; and because the readiness budget expired, `awaitPostRestartReachable`, the wait for the tab to re-register, was **never entered at all**) and, marked as **not established**, what I could not: why the tab's socket never re-helloed. I had no 200-pack install and no live browser tab, and I have not guessed a mechanism into the record.
* **It does not raise `NODE_DEFS_RUN_BUDGET_MS`.** That constant is load-bearing for #1180's composition property and every other caller keeps it.
* **It does not add a disclosure field to the reply.** The two verdicts a slow install now reaches (`refreshed:true`, `refresh_still_running`) already exist and already carry their own wording.

---

## Playwright — measured against a clean `origin/main`, not asserted

The panel is served to the browser from ComfyUI's `custom_nodes`, so running the suite out of a worktree proves nothing unless ComfyUI is actually serving that worktree. I did **not** repoint the developer's shared junction. Two isolated stacks instead, both freshly created:

| | fix tree | baseline |
|---|---|---|
| ComfyUI | `--base-directory .../ap-1562-cc --port 8462 --enable-cors-header --cpu` | `.../ap-1562-base --port 8562` |
| `custom_nodes/comfyui-mcp-panel` | junction → `wt-1562-cc` (this branch) | junction → detached `origin/main` @ `75728263` |
| served panel contains `REFRESH_NODES_RUN_BUDGET_MS` | **yes** (verified over HTTP) | **no** (verified over HTTP) |
| served `refresh-coalesce.js` forwards `runOpts` | **yes** (verified over HTTP) | n/a |
| `npx playwright test --workers=1` | **19 failed / 62 passed (8.3 m)** | **19 failed / 62 passed (8.2 m)** |

**The two failure sets are byte-identical** — `diff` of the 19 `file:line:col` locations is empty. Zero new failures. They are 7 `conversation-persistence`, 4 `workflow-chat-identity`, 3 `chat-history-v2`, and one each of `agent-drive`, `external-orchestrator`, `hidden-tab`, `pending-queue`, `streaming-render` — storage/IndexedDB and streaming, none of them near this budget. The same 19 #1561 measured on the same base commit. **No assertion was loosened and no spec was skipped.**

Nothing was written into the developer's own ComfyUI or its `custom_nodes`.

**Head accounting, so nobody has to wonder.** The suite ran against `7cd2e82d`. The only commit after it is `0f8419a0`, which is **comment-only** (it attributes two quoted measurements to #1180 instead of letting them read as mine). `npm run test:unit` (6130/6129/0 fail), `check-panel-scope`, and the full 14/14 mutation harness were all re-run **on `0f8419a0`**, not inherited from the earlier head.

**Coverage I do NOT have:** no e2e spec exercises a backend whose whole `/object_info` outlives the refresh budget — that needs a 25 MB schema and a 20 s server. That behaviour is covered by the node-test reproduction driving the shipped fetch phase against a real slow 25 MB endpoint, not by the browser suite.

### DIFF
--- a/browser_tests/unit/_panel-constants.mjs
+++ b/browser_tests/unit/_panel-constants.mjs
@@ -89,6 +89,12 @@ export const REFRESH_NODES_EXECUTOR_DEPS = Object.freeze({
   REFRESH_JOIN_ABANDONED,
   NODE_DEF_REFRESH_REASONS,
   REFRESH_NODES_COMMAND_BUDGET_MS,
+  // #1562 — the RUN allowance the executor now hands the coalescer. Adding it here is what
+  // keeps the three harnesses that rebuild `refresh_nodes` working; leaving it out throws
+  // ReferenceError in all three at once, which is the signal this collection exists for.
+  get REFRESH_NODES_RUN_BUDGET_MS() {
+    return REFRESH_NODES_RUN_BUDGET_MS;
+  },
 });
 
 /** #1413 — the whole-command deadline `graph_set_widget` takes on its first line. */
@@ -118,6 +124,34 @@ export const NODE_DEFS_FETCH_SHARE = (() => {
   const m = PANEL_SRC.match(/const NODE_DEFS_FETCH_SHARE = (\d+) \/ (\d+);/);
   if (!m) throw new Error("the panel no longer states NODE_DEFS_FETCH_SHARE as a ratio");
   return Number(m[1]) / Number(m[2]);
+})();
+
+/**
+ * #1562 — the run budget `refresh_nodes` derives, EVALUATED FROM THE PANEL'S OWN
+ * EXPRESSION.
+ *
+ * Not `readPanelNumber`: the panel does not state a literal, it states a derivation
+ * (`REFRESH_NODES_COMMAND_BUDGET_MS / NODE_DEFS_FETCH_SHARE`), and the property the tests
+ * assert is about that derivation. Restating it here would make the harness agree with
+ * itself — the exact trap this file's header records — so the expression is lifted out of
+ * the source and evaluated against the two inputs, both of which are themselves read from
+ * the panel.
+ */
+export const REFRESH_NODES_RUN_BUDGET_MS = (() => {
+  const m = PANEL_SRC.match(/const REFRESH_NODES_RUN_BUDGET_MS = ([^;]*);/);
+  if (!m) {
+    throw new Error("REFRESH_NODES_RUN_BUDGET_MS is no longer findable in the panel source");
+  }
+  // eslint-disable-next-line no-new-func
+  const value = new Function(
+    "REFRESH_NODES_COMMAND_BUDGET_MS",
+    "NODE_DEFS_FETCH_SHARE",
+    `return ${m[1]};`,
+  )(REFRESH_NODES_COMMAND_BUDGET_MS, NODE_DEFS_FETCH_SHARE);
+  if (!Number.isFinite(value) || value <= 0) {
+    throw new Error(`the panel's REFRESH_NODES_RUN_BUDGET_MS expression evaluated to ${value}`);
+  }
+  return value;
 })();
 
 /**
--- a/browser_tests/unit/node-def-refresh.test.mjs
+++ b/browser_tests/unit/node-def-refresh.test.mjs
@@ -16,7 +16,7 @@ import { fileURLToPath } from "node:url";
 import { dirname, join } from "node:path";
 import { fetchNodeDefsWithRetry, OBJECT_INFO_RETRY_DELAYS_MS } from "../../web/js/lib/object-info-retry.js";
 import { withTimeout } from "../../web/js/lib/bounded-step.js";
-import { fetchWholeObjectInfo } from "../../web/js/lib/object-info-oracle.js";
+import { fetchWholeObjectInfo, TRANSPORT_OUTCOME } from "../../web/js/lib/object-info-oracle.js";
 
 // #1180 — READ from the panel, never restated. Shared with the other harnesses that
 // rebuild shipped functions, so one copy of a constant cannot drift from another.
@@ -448,6 +448,10 @@ function buildRegisterComfyNodeDefs({
     // body throws ReferenceError without them.
     "objectInfoSnapshot",
     "backendReconnectEpoch",
+    // #1562 — the oracle's per-route OUTCOME vocabulary. The fetch phase reads route 2's
+    // ending from the TAG rather than from its sentence (#1223), so the extracted body
+    // throws ReferenceError without it.
+    "TRANSPORT_OUTCOME",
     `// #1180 — defined HERE, from the api this scope already has, so each factory site
      // gets its own stub without threading a helper through every call.
      const boundedGetNodeDefs = async (ms = NODE_DEFS_FETCH_TIMEOUT_MS) => {
@@ -501,6 +505,7 @@ function buildRegisterComfyNodeDefs({
     // re-records only a payload it fetched itself.
     snapshotSpy,
     7,
+    TRANSPORT_OUTCOME,
   );
 }
 
@@ -1790,4 +1795,77 @@ test("#608: the no-payload guard is keyed on defsObtained, not on the phase name
   });
   assert.equal(fetchFailed.reason, "object_info_fetch_failed");
   assert.match(fetchFailed.detail, /Failed to fetch/);
+});
+
+// ---------------------------------------------------------------------------
+// #1562 — an ABANDONED route is not a FAILED one, and the remedy must not confuse them.
+//
+// The reporter's `GET /object_info` returned 25,104,088 bytes in 20.84 s from a ComfyUI
+// that was up throughout, while this verdict told them to "check that the ComfyUI server
+// process is still running". Every route had been abandoned at its bound; not one of them
+// had failed. These drive the SHIPPED run — the real oracle on injected timers — because
+// the distinction is made inside the fetch phase from the oracle's TAGS, and a test of
+// `describeNodeDefRefresh` alone cannot see whether the panel computes the input correctly.
+// ---------------------------------------------------------------------------
+
+/** Run the shipped register with the real oracle on virtual timers, and settle it. */
+async function runWithHangingRoutes(apiValue) {
+  const timers = virtualTimers();
+  const { registerComfyNodeDefs } = buildRegisterComfyNodeDefs({
+    appValue: FULL_APP,
+    apiValue,
+    timers,
+    fetchWholeObjectInfo: (opts) => fetchWholeObjectInfo({ ...opts, timers }),
+  });
+  const running = registerComfyNodeDefs(undefined);
+  await drainMicrotasks(300);
+  for (let i = 0; i < 8; i += 1) {
+    timers.fireAll();
+    await drainMicrotasks(200);
+  }
+  return orFail(running, "the run never settled");
+}
+
+test("#1562: BOTH whole-document routes abandoned at their bound is not reported as a dead server", async () => {
+  const verdict = await runWithHangingRoutes({
+    getNodeDefs: () => new Promise(() => {}), // bounded → abandoned, never failed
+    fetchApi: () => new Promise(() => {}), // the oracle's NO_ANSWER, same shape
+  });
+  assert.equal(verdict.refreshed, false);
+  assert.equal(verdict.reason, "object_info_fetch_failed");
+  assert.doesNotMatch(
+    verdict.remedy,
+    /check that the ComfyUI server process is still running/,
+    "nothing here established that the server is down — both routes merely ran out of time",
+  );
+  assert.match(verdict.remedy, /ABANDONED AT ITS BOUND/);
+  assert.match(verdict.remedy, /plain retry meets the same bound/);
+  assert.match(verdict.remedy, /Tried 2 routes/, "…and #608's evidence clause survives");
+});
+
+test("#1562: a route that genuinely FAILED outranks the other's silence", async () => {
+  // Route 1 fails for real while route 2 is merely slow. That IS evidence about the
+  // server, so the old remedy is the right one and the new branch must not fire.
+  const verdict = await runWithHangingRoutes({
+    getNodeDefs: async () => {
+      throw new TypeError("Failed to fetch");
+    },
+    fetchApi: () => new Promise(() => {}),
+  });
+  assert.equal(verdict.reason, "object_info_fetch_failed");
+  assert.match(verdict.remedy, /check that the ComfyUI server process is still running/);
+  assert.doesNotMatch(verdict.remedy, /ABANDONED AT ITS BOUND/);
+});
+
+test("#1562: a route that ANSWERED unusably outranks the other's silence too", async () => {
+  // Route 1 is abandoned, route 2 answers 500. A server that answers is a server that is
+  // up, but it did not answer THIS question — and "answered badly" is not "ran out of
+  // time", so the abandonment finding must not be claimed.
+  const verdict = await runWithHangingRoutes({
+    getNodeDefs: () => new Promise(() => {}),
+    fetchApi: async () => ({ ok: false, status: 500, json: async () => ({}) }),
+  });
+  assert.equal(verdict.reason, "object_info_fetch_failed");
+  assert.doesNotMatch(verdict.remedy, /ABANDONED AT ITS BOUND/);
+  assert.match(verdict.remedy, /GET \/object_info was not OK \(status 500\)/);
 });
--- a/browser_tests/unit/node-def-refresh.test.mjs
+++ b/browser_tests/unit/node-def-refresh.test.mjs
@@ -1869,3 +1869,40 @@ test("#1562: a route that ANSWERED unusably outranks the other's silence too", a
   assert.doesNotMatch(verdict.remedy, /ABANDONED AT ITS BOUND/);
   assert.match(verdict.remedy, /GET \/object_info was not OK \(status 500\)/);
 });
+
+test("#1562: a real failure EARLIER in route 1's retry loop outranks the stall that ended it", async () => {
+  // THE SEQUENCE A POST-RESTART REFRESH ACTUALLY MEETS, and the one the first version of
+  // this branch got wrong. ComfyUI is genuinely down when attempt 1 runs (`Failed to
+  // fetch`, connection refused); by the retry it has bound the port but is still importing
+  // packs, so attempt 2 stalls past its bound and the loop ends with BOTH flags set —
+  // `sawRealError` AND `lastAttemptTimedOut`. The loop already ranks these ("A REAL ERROR
+  // OUTRANKS A SYNTHESIZED TIMEOUT": it rethrows `lastRealError`, not the stall), and the
+  // classification has to rank them the same way. Reading that ending as an abandonment
+  // produced a verdict headed "none of them FAILED" whose own evidence clause read
+  // "api.getNodeDefs() failed: Failed to fetch", and told a reader whose backend was still
+  // booting that a plain retry would not help — which is the one thing that does help there.
+  let attempt = 0;
+  const verdict = await runWithHangingRoutes({
+    getNodeDefs: async () => {
+      attempt += 1;
+      if (attempt === 1) throw new TypeError("Failed to fetch");
+      return new Promise(() => {}); // bounded → NODE_DEFS_NO_ANSWER on the retry
+    },
+    fetchApi: () => new Promise(() => {}), // route 2 merely stalls
+  });
+  assert.ok(attempt >= 2, `the retry loop must have run twice for this to be the case under test (ran ${attempt})`);
+  assert.equal(verdict.reason, "object_info_fetch_failed");
+  assert.match(
+    verdict.remedy,
+    /check that the ComfyUI server process is still running/,
+    "route 1 FAILED for real before it stalled, so the remedy that names a possibly-down " +
+      "server is the correct one — and a plain retry is what fixes a booting backend",
+  );
+  assert.doesNotMatch(
+    verdict.remedy,
+    /ABANDONED AT ITS BOUND/,
+    "a route that failed and then stalled did not merely run out of time",
+  );
+  // The self-contradiction this guards: the heading and the evidence clause disagreeing.
+  assert.match(verdict.remedy, /Failed to fetch/, "#608's evidence clause still names what route 1 did");
+});
--- a/browser_tests/unit/node-def-refresh.test.mjs
+++ b/browser_tests/unit/node-def-refresh.test.mjs
@@ -27,6 +27,7 @@ import {
   NODE_DEFS_FETCH_TIMEOUT_MS,
   NODE_DEFS_NO_ANSWER,
   NODE_DEFS_RUN_BUDGET_MS,
+  REFRESH_NODES_RUN_BUDGET_MS,
   REFRESH_NODES_EXECUTOR_DEPS,
   monotonicNow,
   nodeDefsBudgetLeft,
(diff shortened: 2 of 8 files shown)


artokun:
@grok — review request on **#1566** (panel#1562, the `panel_refresh_nodes` half). Review is **pending**; nothing here is gated.

Naming where the risk actually is, rather than handing you a diff:

**1. The derivation, and whether it is the right property.**
`REFRESH_NODES_RUN_BUDGET_MS = Math.ceil(REFRESH_NODES_COMMAND_BUDGET_MS / NODE_DEFS_FETCH_SHARE)` = 37,500. The claim it rests on is *"the fetch phase alone must be able to spend everything the command will wait for"*. I measured that 25,000 (the command's own window) **still fails** on the reported install, which is why it is not simply that number — but 37,500 exceeds the 25,000 ms join, so **the run now deliberately OUTLIVES the reply by up to ~12,500 ms**, holding the coalescer's single slot. Is that the right shape, and is the boundary safe? `37500 * (2/3) === 25000` exactly in IEEE754 here, and `nodeDefsBudgetLeft` then floors `(deadline - now) * share`, so route 1's real bound is a few ms under the join. I believe that is immaterial; tell me if it is not.

**2. The coalescer contract change is the widest-blast-radius part.**
`refresh-coalesce.js` now takes `opts.runBudgetMs` and forwards it as `runRegister(preloadedDefs, runOpts)` on all three start paths. The **trailing (forced) run is SHARED**, and I made it keep the allowance of the caller that QUEUED it — mirroring what `joinMs` already documents. A second forced caller with a different budget gets the queued run's. Is first-queuer-wins right, or can a short-budget caller now pin a long run for someone who needed a short one (or vice versa)?

**3. `fetchAbandonedAtBound` — where it is computed, and its fail direction.**
Route 1's ending comes from the retry loop's own `lastAttemptTimedOut`; route 2's from the oracle's `TRANSPORT_OUTCOME.NO_ANSWER` tag (never from prose — #1223). It starts `null` when no whole-document route was reached, and the verdict gates on `=== true`, so `null` falls to the wording that claims LESS. It is assigned unconditionally after route 1's try/catch, inside `if (!defs && typeof api?.getNodeDefs === "function")`, and cleared to `false` when the fallback answers. **The state I am least sure of** is whether any path can reach `describeNodeDefRefresh` with `phase === "fetch"` and a value that does not describe that run — I convinced myself none can, and I would rather be told I am wrong than believe myself.

**4. The cost claim, which is the part a reader will inherit.**
I claim a healthy install pays nothing (measured 312 ms vs 322 ms, one request, identical verdict) and that the only real costs are the outliving run and a `graph_add_node` that meets it and takes its worded busy refusal. If there is a fifth cost I have not named, that is the finding I most want.

**5. Two source-pinning assertions were UPDATED** — `#1180`'s `runDeadline` pin and `#1404`'s call-site pin. Both pin code this change moves. I believe I strengthened both (the `runDeadline` one now also asserts there is exactly ONE such statement; the call-site one now requires `joinMs` AND `runBudgetMs` by name). Please check I did not quietly loosen either — that is precisely the failure this repo warns about.

**6. Overlap with #1561, declared rather than discovered.** Both branches change **one adjacent import line** in `comfyui-mcp-panel.js` (`git merge-tree` conflicts on exactly that hunk and nothing else). I already reduced my side to a single-line edit for it. No other file, function, or line is shared, and this PR does not read, write, or depend on `scoped-object-info.js` — #1561's own body says `panel_refresh_nodes` is unhelped by the type-scoped route, and it is right: this path re-registers the WHOLE schema.

Evidence in the body: the reported message reproduced byte-for-byte against a real 25 MB `/object_info`, a budget sweep showing which values suffice, 14 mutations / 14 killed (one of which survived the first round and is written up), unit 6130/6129 vs a 6116/6115 baseline, and a Playwright baseline-vs-fix comparison against two isolated ComfyUI instances.

artokun:
**State: OPEN and GREEN, HELD ON AN ABSENT REVIEWER. Not merging.**

* CI: **success** on `0f8419a0` (the head everything below was measured on).
* `npm run test:unit`: **6130 / 6129 pass / 0 fail / 1 todo**, against a **6116 / 6115 / 0 fail** baseline on clean `origin/main` @ `75728263`.
* Mutations: **14 of 14 killed**, re-run on this head rather than inherited, worktree verified clean after each.
* Playwright `--workers=1`: **19 failed / 62 passed** on both trees, **byte-identical failure sets**, zero new failures.
* `mergeable: MERGEABLE` against `main`.

I have asked for grok's review above and named where the load-bearing claims are. **No review has come back** — `repos/artokun/comfyui-mcp-panel/pulls/1566/reviews` holds one entry, and it is Copilot saying it hit its quota limit, which is not a review.

My own evidence is not a substitute for that, and the derivation in point 1 of the review request plus the coalescer contract change in point 2 are exactly the kind of thing an author reads as obviously right. So this sits here until a reviewer answers. No `autopilot:parked` label — this is not parked work, it is finished work waiting on a person.

artokun:
**Rebased onto current `main` (`a1844f68`), and every number below was re-measured on the new head `8e4b6876` — not inherited.**

`main` moved under this branch when #1561 (the #1560 fix) landed, and the two touched the same import block, so the PR went `CONFLICTING`. Resolved and re-verified:

* **The conflict was the oracle import and nothing else.** #1561 added `noBackendAnswerEstablished` and the `scoped-object-info` import; this branch added `TRANSPORT_OUTCOME`. The resolution keeps all three. No behaviour of either change was touched — `git diff` against `origin/main` for `object-info-oracle.js`, `object-info-snapshot.js`, `scoped-object-info.js` and `set-widget.js` is **empty**, so nothing #1561 shipped is modified here.
* **The two `chore(1562)` commits are gone**, deliberately. They existed only to keep the oracle import on one line so #1560's *open* branch would not conflict over formatting. #1560 has merged, so they had no remaining purpose; the import is back in its normal multi-line form and the branch is 4 commits.

Re-verification on `8e4b6876`:

| | result |
|---|---|
| `npm run test:unit` | **6161 pass / 0 fail / 1 todo** (6162 tests) |
| Mutation re-run | **8 / 8 killed** — worktree verified clean after each |
| CI | `Lint pack` **SUCCESS**, `mergeStateStatus: CLEAN` |

The 8 mutations, each applied to the shipped source and reverted after:

| mutation | verdict |
|---|---|
| M1 — the `refresh_nodes` **call site** drops `runBudgetMs` | KILLED |
| M2 — the coalescer stops forwarding `runOpts` to `runRegister` | KILLED |
| M3 — `registerComfyNodeDefs` ignores the stated allowance | KILLED |
| M4 — route 1 never reports abandonment | KILLED |
| M5 — the abandonment finding survives a fallback route that ANSWERED | KILLED |
| M6 — route 2's ending is not required to be an abandonment | KILLED |
| M7 — the verdict keeps the old process-is-down remedy for an abandonment | KILLED |
| M8 — the verdict claims abandonment when it was never established (`null`) | KILLED |

M1 is the one that matters most: a one-line wiring change at a call site is invisible to a helper-level test, and killing it is what proves the assertion is on the call site rather than on `makeRefreshCoalescer` in isolation.

**Browser suite — stated plainly rather than implied.** The full Playwright run reported earlier on this PR (`--workers=1`, 19 failed / 62 passed, byte-identical failure sets on both trees) was made on the pre-rebase head. It has **not** been re-run since the rebase: both ComfyUI slots on this machine (8188, 8189) are held by another agent's run, and this session was not allocated an instance and is not permitted to start one. What changed since that run is the import-block resolution and the removal of two formatting-only commits — no executable path in this diff differs — but I am not going to call that browser coverage.

**Review is still pending.** The verdict from the fallback gate, if it returns before a reviewer does, will be posted here in full and labelled as the substitution it is.

artokun:
## Substituted review — `claude-gate.mjs` returned **NO-SHIP**, and it was right

Disclosed in full, and labelled as the substitution it is. grok has not answered here or on any of the other open PRs today; Copilot's only entry on this PR is a quota-limit message. The repo's fallback gate — a **fresh adversarial session that never saw this implementation, handed only the diff and told to refute** — is what I ran instead. It is weaker than codex in one specific way (it does not execute as reliably), and it is **not** a substitute for a human or for grok. This one did execute, and it found a **P1**.

### The finding, verbatim

> **[P1]** `web/js/comfyui-mcp-panel.js:1581` — `fetchAbandonedAtBound = clientRouteThrew && lastAttemptTimedOut === true` ignores `sawRealError` (in scope, 11 lines up), so route 1's ending is classified as an abandonment even when an earlier attempt in the same retry loop FAILED for real. The loop's own rule — *"A REAL ERROR OUTRANKS A SYNTHESIZED TIMEOUT"* — is enforced across routes but not across attempts within route 1.
>
> Concrete state, executed against the shipped `registerComfyNodeDefs`: post-restart, `api.getNodeDefs()` attempt 1 rejects `Failed to fetch` (connection refused, server genuinely down), the retry stalls past its bound (server bound the port but is still importing packs), and the `/object_info` fallback stalls → `NO_ANSWER`. Produced remedy:
>
> > "…every transport was ABANDONED AT ITS BOUND and **none of them FAILED**… A plain retry meets the same bound… Tried 2 routes: **api.getNodeDefs() failed: Failed to fetch**; …"

### Confirmed, not taken on trust

I re-derived it from the loop before changing anything. `if (sawRealError) throw lastRealError;` is right there in the retry body — the loop *already* ranks a real error above a synthesized timeout, and rethrows the real one. So on that sequence the loop ends with `clientRouteThrew`, `lastAttemptTimedOut` **and** `sawRealError` all true, `clientRouteError` is the `TypeError`, and the old expression called it an abandonment.

Two failures at once, and both are exactly what this PR set out to fix:

* the verdict **contradicts its own evidence clause** — headed "none of them FAILED" over a line reading `api.getNodeDefs() failed: Failed to fetch`;
* it tells a reader whose backend is genuinely still booting that **a plain retry will not help**, when a plain retry is the one thing that does help there. The old wording — *"the backend may still be restarting — retry once it answers"* — was **correct** for that input. On that state my change was worse than the bug.

### The fix — `603f2b37`

```js
fetchAbandonedAtBound = clientRouteThrew && lastAttemptTimedOut === true && !sawRealError;
```

The loop's own precedence rule, applied to the classification as well as to the thrown value: an abandonment is what ends a route that **never failed** — across ATTEMPTS as well as across ROUTES.

### Why the existing tests could not see it, and what now does

The gate was right about that too: the tests asserted the **source text** of that line and unit-tested `describeNodeDefRefresh` with **hand-supplied** flags, so neither could observe a mixed real-error/stall sequence. The new test is **behavioural** — it drives the shipped `registerComfyNodeDefs` through the same extraction harness the #608/#1223 tests use, with a `getNodeDefs` that **fails once and then hangs** while route 2 stalls, and asserts the possibly-down-server remedy survives:

```
✔ #1562: a real failure EARLIER in route 1's retry loop outranks the stall that ended it
```

The source-text assertion was updated rather than deleted — it now pins `&& !sawRealError` and says in its message that the behavioural coverage lives in the other file.

### Re-verified on `603f2b37`

| | result |
|---|---|
| `npm run test:unit` | **6162 pass / 0 fail / 1 todo** (6163 tests) |
| Mutations | **9 / 9 killed**, tree verified clean after each |

M9 is the new one — deleting `&& !sawRealError` and watching the named test fail. The prior eight were **re-run on this head**, not inherited, because a narrow re-fix disabling an earlier guard is a failure mode this repo has recorded before; M4 (route 1 never reports abandonment) and M5 (the finding survives a fallback that ANSWERED) both still die.

One process note worth recording: the first mutation pass restored `comfyui-mcp-panel.js` with `git checkout --`, which silently reverted the **uncommitted** fix under the test run. Caught by re-running the mutation immediately (anchor missing), re-applied, and committed *before* any further mutation run — the numbers above are all from the committed head.

**Review is still pending.** A fallback-gate NO-SHIP that I then fixed is evidence the change got adversarial attention; it is not a clean independent verdict on the result. A second gate run against `603f2b37` is in flight and will be posted here in full whatever it says, and grok's review is still the one I would rather have — @grok, the new clause is the thing to attack: it is a **third** flag consulted in a classification whose other two inputs were already subtle, and if there is a sequence where `sawRealError` is set but route 1 really did only run out of time, the wording regresses to the one this PR is fixing.

artokun:
## Fallback gate, second pass on `603f2b37`: **NO-SHIP** again — two more P1s, both executed, both real. Fixed on `6e756191`

Disclosed in full, same substitution as before (`claude-gate.mjs` — a fresh session that never saw the implementation, handed the diff and told to refute, and told to run things rather than read them). **This is not a grok review and this PR is not gated.**

### Finding 1 — the larger budget reached a phase it was never measured for

> **[P1]** The new run budget (37,500 ms) is longer than the command's join (25,000 ms), and the combo phase's bound is *whatever is left of the run budget*. So the combo phase alone can now outlast the join, flipping a previously successful `panel_refresh_nodes` into a non-refresh.

It executed it — shipped `refresh_nodes` executor, real coalescer, real `registerComfyNodeDefs`, healthy fast `/object_info`, varying only `refreshComboInNodes()` latency:

| combo latency | before this issue | uncapped 37,500 ms run |
|---|---|---|
| 15,000 ms | `refreshed:true` | `refreshed:true` |
| 26,000 ms | `refreshed:true` at 9 s | **`refreshed:FALSE`** — `refresh_still_running` |
| never answers | `refreshed:true` at 9 s | **`refreshed:FALSE`** — `refresh_still_running` |

That is a case that **worked** being broken by the fix for a case that did not — the exact shape I had just fixed in round 1, one phase over. It also takes #1193's `combo_refresh_confirmed:false` disclosure with it: unreachable through this command whenever the combo call exceeds 25 s.

**The fix is not to clamp the run inside the join** — that is the 25,000 row the derivation already measured as still failing. The extension exists for the **fetch** phase, so it now reaches that phase and no further:

```js
const comboBudgetLeft = nodeDefsBudgetLeft(runDeadline);
comboWaitMs = Math.min(comboBudgetLeft, NODE_DEFS_RUN_BUDGET_MS);
```

A **no-op for every caller that states no budget** — what is left of a 9,000 ms run is never more than 9,000 ms — and that is asserted, byte-for-byte, on the default path's refusal text.

### Finding 2 — a sentence the change made arithmetically impossible

> **[P1]** The combo-abandoned message hardcodes `NODE_DEFS_RUN_BUDGET_MS` as "this refresh's budget". Executed: `refreshComboInNodes() did not answer within the 37500ms this refresh had left of its 9000ms budget`.

37,500 ms left of a 9,000 ms budget. #1161's defect — naming a number that was never spent — reintroduced by this change, and it reaches the tool reply's `detail` as soon as finding 1 is fixed. The run's allowance is now **named** (`runBudgetMs`) and the sentence quotes it; the new third case (the phase holding its ceiling while the run still had time) reads:

```
refreshComboInNodes() did not answer within the 9000ms this phase is capped at
(this refresh had 37499ms left of its 37500ms budget)
```

so #1193's requirement — that "starved to a sliver by a slow fetch" and "handed the whole budget and hung anyway" stay distinguishable — now covers three cases instead of two.

### Coverage, behavioural rather than textual

Three new tests over the **shipped** `registerComfyNodeDefs`: the bound a stated budget actually yields (`≤ NODE_DEFS_RUN_BUDGET_MS`), the sentence it produces, and the default path's wording unchanged. The two source-shape harnesses that pinned the old expressions (`refresh-nodes-run-budget`, `single-node-def`) are updated in the same commit — a source pin that silently stops matching is a test that has quietly stopped testing.

| | result on `6e756191` |
|---|---|
| `npm run test:unit` | **6165 pass / 0 fail / 1 todo** (6166 tests) |
| Mutations | **11 / 11 killed**, tree clean after each |

The new ones: **M10** combo phase uncapped → the bound test fails; **M11** refusal quotes the default constant again → the sentence test fails; **M12** the run ignores the caller's allowance → the budget test fails. The round-1 set (M1, M2, M4–M9) was **re-run on this head** — M3's anchor is gone because that expression was refactored, and M12 replaces it.

### One claim in the PR body was optimistic, and is corrected

The body said the reporter's install would now answer `refreshed:true`. Working through finding 1 showed why that is not the honest expectation: the fetch lands at ~21 s, then the combo phase — which **re-fetches the same 25 MB document** — is abandoned at its bound, putting the run at ~30 s, past the 25 s join. So the first call answers **`refresh_still_running`**, and the refresh itself **completes**: the definitions are registered, which is what the missing node depends on. Before this change nothing landed at all, ever, on that machine. The body now says that.

@grok — if you get to this, the thing to attack is the new `Math.min`. It bounds the phase by a constant named for the *whole run*, which is a slightly odd thing for a phase to be capped by; I chose it because it is exactly the allowance this phase has always had and because it is caller-independent (the trailing run in the coalescer is SHARED, so keying a phase bound on one caller's join would be wrong). If there is a better ceiling, or a case where 9,000 ms is now too little where it was not before, that is the defect.

artokun:
## Fallback gate, third pass on `6e756191`: **SHIP**

Same substituted gate, third fresh session. Verbatim verdict:

> **SHIP** — Nothing material: 16 mutations run against the fix, 14 killed by named tests, and the two survivors are message-text/strictness only with no reachable behavioural difference. Full suite 6165 pass / 0 fail; worktree restored clean.

It walked all seven taxonomy classes by execution rather than by reading. The parts worth quoting, because they are the claims I was least sure of:

* **Class 3 (a guard comparing the wrong pair)** — *"Named the pairs out loud. `Math.min(comboBudgetLeft, **NODE_DEFS_RUN_BUDGET_MS**)` — not `runBudgetMs`. I mutated it to the wrong-pair version; killed. Same for `!sawRealError` (compares across attempts, not just the last one) — killed. This was the likeliest trap and it's guarded."* Those are exactly the two things it made me fix in rounds 1 and 2, checked from the other direction.
* **Class 6** — it re-derived `37500 × 2/3 = 25000 ≥ 25000` itself, and independently verified the claim I put in the source comment: that `Math.min` is a no-op for callers stating no budget *even across the local-work clock-stop* (`runDeadline += monotonicNow() - localWorkStartedAt` cancels out, so `comboBudgetLeft ≤ 9000` always). That is the one line in this change that a reader has to take on trust, and it no longer rests on my arithmetic alone.
* **Class 7** — reachability traced from the production caller to the second parameter, and *"deleting the call-site line fails both a source test and a behavioural one over the real coalescer."*
* **Class 5** — the run outliving its command by up to ~12.5 s is *"disclosed, bounded, worded/retryable — and the cap is what stops the leak."*

Its two mutation survivors are the non-capped combo message reverting to the default constant, and `=== true` → truthy; it verified both are unreachable behaviourally on the shipped path (`ADV1` prints `7500ms … of its 37500ms budget` correctly). I am leaving them: adding assertions for states the code cannot reach is how a suite acquires tests that pass while proving nothing.

**One non-blocking note it raised, recorded rather than fixed.** With route 1 stalled *and* a proxy answering 504, the remedy heads "none of them FAILED" above an evidence clause reading "a proxy in front of ComfyUI reporting the backend did not answer it (status 504)". That pairing comes from **#1223's** deliberate gateway ≡ no-answer mapping, not from this change, and the gate argued itself out of calling it blocking: a genuinely dead backend behind nginx **502s instantly**, which fails route 1 for real and takes the old remedy. Re-deciding #1223's mapping is a separate, reviewable change and I am not smuggling it in here.

---

## Merging, and exactly what that verdict is and is not

**What it is:** three adversarial passes by fresh sessions that never saw the implementation, told to refute and to execute. Two returned **NO-SHIP with real, executed P1s** — one of which (`!sawRealError`) made this change *worse than the bug it fixes* on a booting backend, and one of which (the uncapped combo phase) broke a case that previously worked. Both are fixed and pinned behaviourally. Combined with 11 mutations of my own on the final head and a green 6165-test suite, this had genuine adversarial attention.

**What it is not:** a grok review. grok has posted nothing on this PR, on this issue, or on any of the other open PRs in this repo today — `/pulls/1566/reviews` holds one entry and it is Copilot reporting its quota limit. I asked twice, on the PR and on the issue, naming the load-bearing claims each time. The substitution is disclosed here, as the condition under which it is acceptable at all.

**Coverage I do not have:** the browser suite has not been run since the rebase. Both ComfyUI slots on this machine are held by another agent's run, this session was not allocated an instance and is not permitted to start one. The earlier `--workers=1` run (19 failed / 62 passed, byte-identical failure sets on both trees) was on the pre-rebase head. This change alters tool-reply wording and two timeout numbers on a path the panel e2e suite drives through a **mock** orchestrator, so a run would have been a regression check rather than coverage of the fix — but I am not calling something coverage that I did not run.

If grok's review lands after this merges and finds something, it gets a follow-up PR with the same evidence discipline, not a silent close.

PR #1568, opened 2026-08-21
fix(745): the live availability scan reads V2/V3 combos, not just the V1 shape

### LINKED ISSUE #745: panel_get_errors omits missing model on newly added loader node
After adding a LoraLoaderModelOnly node on the live canvas while the local lora dropdown was empty, panel_set_widget accepted an unavailable basename with empty_option_list:true. panel_graph_outline and panel_query_graph showed the new node and value, but panel_get_errors still returned only the four missing-model nodes that existed when the workflow was loaded. The newly added missing LoRA node was omitted.

Likely cause: the missing-model/error scan is not recomputed for nodes or combo values added after initial workflow load.

### Steps to reproduce
1. Start with a workflow containing missing loader values and call panel_get_errors.
2. panel_add_node LoraLoaderModelOnly.
3. panel_set_widget lora_name to an unavailable basename while options are empty.
4. Confirm the node/value with panel_query_graph.
5. Call panel_get_errors; the new missing-model node is absent.

### Exact observed result
New node 12 had lora_name=krea2_realism_lora.safetensors, active, and wired into the model path. panel_get_errors returned errored_count:4 and only nodes 1-4; node 12 was missing from nodes[] and missing_models[].

### Fix
Applied locally: no. Upstream-only for this session because applying a panel patch/reload was outside the user's scoped canvas edit. Recompute missing-model reasons from the current live graph after panel mutations, or invalidate/rescan the missing-asset cache when add_node/set_widget changes a loader combo.

### Environment
OS: macOS/Darwin; ComfyUI 0.29.0; MPS 16 GB; comfyui-mcp 0.49.8; panel 0.11.41.

Related (same root cause): #356

artokun:
Closes #745.

## It reproduced — but not for the reason the title says

#745 was fixed once (#798 / `def53f0`, shipped 0.11.45) and **reopened** on 0.15.19: the live scan omits combos serialized as `["COMBO", {"options": [...]}]`, reported against `Load3D`.

The original cause is gone, established by execution rather than by reading. `scanComboAvailability` runs on every `panel_get_errors` call, over the live `nodes` array, from a fresh per-class `/object_info/<class>` fetch — no result cached from load, no graph snapshot taken at load. **So this shares no cause with the ChangeTracker drift in #1563/#1564 (#1567, since merged).**

What was stale is the **parser**. `parseClassCombos` read the option list off `def[0]`, true only of the V1 shape `[[...allowed], {opts}]`. A V3-schema node (`IO.Combo.Input` under `@comfytype(io_type="COMBO")`) leaves the string `"COMBO"` at `def[0]` and puts the list under the config, so every one was filed as "not a combo" and the widget was skipped **entirely** — no finding, and no `unchecked_nodes` entry either. Same per-call-site adoption gap as #1552, one module further out: `authoritativeComboValues` was already canonical in two neighbouring modules and had never been adopted here.

Measured on live ComfyUI 0.33.2 (853 types, 652 combo inputs): the `def[0]` read recognised **61**, all V1. The canonical reader recognises **528** (61 V1 + 467 V2), 27 of them server-declared empty. The scan was blind to 91% of the combos it was asked to judge.

## Judging values the way the server judges them

Reading V2 lists opened a seam on the **value** side, and four defects lived in it — one found by self-review, three by gate review. Every one was introduced by this branch: all 44 numeric options on the live server sit on V2 combos the old parser could not see, so none was reachable before.

The server settles the rule. `execution.py` validates a combo with `invalid_vals = [val] if val not in combo_options else []`, and Python's `in` is `==`. Every expectation below is the real 0.33.2 interpreter's answer, not a guess:

| | | | |
|---|---|---|---|
| `10 in [6,8,10]` | True — accept | `True in [False]` | False — reject |
| `10.0 in [6,8,10]` | True — accept | `False in [False]` | True — accept |
| `"10" in [6,8,10]` | **False — reject** | `True in [1,2]` | **True — ACCEPT** |
| `99 in [6,8,10]` | False — reject | `False in [0,1]` | **True — ACCEPT** |

1. **Integer lists collapsed to empty** (self-review). `filter(v => typeof v === "string")` turned `[6,8,10]` into `[]`, which in this module means "the server enumerates nothing" — `option_count: 0`, `kind: "missing_asset"`. 15 live inputs publish pure-int lists, all V2. The list is now stored **as declared**.
2. **Coerced comparison** (P1). `comboOffers` matched `"10"` against option `10`. The server rejects that, so it was a false negative — and the frontend stringifying combo values on queue ([ComfyUI_frontend#14641](https://github.com/Comfy-Org/ComfyUI_frontend/issues/14641)) is exactly how a canvas acquires one, which makes reporting it the useful behaviour.
3. **Numeric values skipped** (P1). The guard read `typeof value !== "string"`, so `duration: 99` on `[6,8,10]` returned `{"unavailable":[],"unknown":[]}` while the server rejects it.
4. **Boolean values skipped** (P1). `options: [false]` with a widget holding `true` returned clean. The naive fix would have created a *false positive*: Python's `bool` subclasses `int`, so `True == 1` and the server **accepts** `true` on `[1, 2]`.

The comparison now lives in `serverConsidersEqual`, implementing Python `==` for the primitives a combo can carry: identical primitives match, a boolean matches a number numerically, and **a string never matches either**.

**Verified by exhaustive agreement, not by example.** A 132-row truth table was generated by running `value in options` in the real interpreter across bool/int/float/str/empty-list combinations and transported over JSON — the same collapse the `/object_info` wire performs, so it is the comparison that actually governs. `comboOffers` agrees **132 of 132**. The discriminating rows are pinned as a test.

## Both directions, on the live payload

257 classes, 474 combo widgets (15 numeric), values in the shape each side actually produces:

| | `origin/main` | this PR |
|---|---|---|
| **A** — server-declared values | 0 false positives | **0 false positives** |
| **B** — poisoned, same type | 34 of 474 (440 missed) | **474 of 474** |
| **C** — numeric values **stringified** (the #14641 trap) | 0 of 15 | **15 of 15** |

Direction A is the control that must not move, and it does not. Closed over real HTTP too — live ComfyUI → the panel's own `fetchSingleNodeDef` → `scanComboAvailability`, the chain `panel_get_errors` uses at `comfyui-mcp-panel.js:16448`: **6 of 6** cases match the backend's own verdict, and the V1 control (`CheckpointLoaderSimple`) is identical either side, so nothing already being judged changed behaviour.

## Mutation evidence

Each mutation **asserted to have applied** (byte delta printed) before running; each kills a *different* named test:

| mutation | fails |
|---|---|
| reader reverted to `def[0]` | 4, incl. the reporter's case |
| stop propagating the V2 **config** | annotated-abstention + `comboConfigsOf` |
| restore the non-string **filter** | int-collapse + stringified-value |
| re-introduce string **coercion** | stringified-value + type-faithful |
| value guard back to **string-only** | numeric-value + `0`-is-real |
| **call site**: skip numbers past `comboOffers` | the same two |
| guard reverted so **booleans** are skipped | boolean + bool/int |
| **bool/int rule dropped** | bool/int + truth table |
| bool/int rule widened to coerce **strings** | **both earlier P1 tests** + truth table |
| drop the empty-string guard | non-primitive-skipped |
| loosen so a **remote** list reads as `[]` | the UNREAD-V2 test |

The third-from-last matters most: the newest helper cannot be used to silently undo the earlier findings.

## Boundaries, stated rather than left to be discovered

- **Remote V2 (1 input) and dynamic V3 (123) stay unjudged and undisclosed.** They yield `null`, not `[]`, so an unread list is never mistaken for an empty one. Disclosing them as `unchecked` would put a permanent entry on ordinary canvases — the #1284 flooding failure.
- **Array values are not judged.** The server validates a *multiselect* combo element-wise (`is_multiselect and isinstance(val, list)`) and otherwise tests the whole list as one value. Modelling both needs a new reporting shape, and **0 of 468 live V2 combos declare `multiselect`** — so it would be machinery written against a shape I cannot verify. Skipped explicitly, not silently.
- **`null`/`undefined` and objects stay skipped.** Reporting every unset combo widget would be noise on canvases that are fine.
- **`empty_option_list: true` on `panel_set_widget` is untouched** — per the original thread the backend is the authority, not a possibly-unpopulated combo, and nothing measured contradicts that.
- **No Playwright spec for `panel_get_errors`.** That surface has ten unit files and zero specs; the reachability question is answered above on real server bytes.
- **`declaresUnrecognizedUploadKind` was deleted** once #1571 landed `file_upload` in `UPLOAD_CONFIG_FLAGS` — the canonical #1357 path now arms for `Load3D`, so the local workaround this branch carried became a second mechanism for a flag no ComfyUI emits.

## Gates

Unit **6194 pass / 0 fail / 1 todo**; `check-panel-scope` OK; `check-tool-vocabulary` OK; `tsc --noEmit` clean. The lone suite failure across runs was manager-install #671's known wall-clock flake under load — 125/125 in isolation.

Playwright `--workers=1` against a real ComfyUI 0.33.2 with the panel junctioned in (served file verified to carry — and for the baseline, to *not* carry — the fix):

| | result |
|---|---|
| `origin/main` baseline | 18 failed / 63 passed (81 specs) |
| this branch | **18 failed / 64 passed** (82 specs) |

The **identical** 18 failures in the identical seven spec files — all chat/session-persistence, pre-existing and untouched by this diff — and the spec main newly added passes. No assertion was loosened.

**Review: codex gate SHIP** on `614cacc3` (round 3, no findings), against `origin/main` at `03da405d`, with worktree/remote/PR heads confirmed identical before the run.


artokun:
@grok please review. This widens what `panel_get_errors` is willing to accuse a node of, so the risk is concentrated in a few specific places rather than spread over the diff. Naming them so you can go straight at them:

**1. The load-bearing claim, and the one I am least sure of.** Reading V2 combos takes 425 widgets that were previously *never judged* and starts judging them. My false-positive proof is a canvas built from the live `/object_info` where every widget holds a value the server itself lists, and it reports 0 findings — but that only covers values that are literally members of the option list. The case I cannot enumerate is a value that is legitimately resolvable while **not** being a member: annotated filepaths are the family I found and handled, but if there is a second such family on a V2 combo, my change turns it into a false `missing_asset` and the old code was accidentally protecting it by skipping V2 wholesale. That is the failure mode worth attacking.

**2. The abstention I added, which is the part most likely to be wrong in the other direction.** `declaresUnrecognizedUploadKind` + `annotated` makes the scan go quiet. I justified it with a measurement — 0 of 652 option strings on the live server carry an `[output]/[input]/[temp]` annotation — but that is one server's node set. If a pack ships annotated options, this abstains where it should report, and it fails *silent-ish* (an `unknown` entry, which readers skim). Check whether the condition is narrower than I think, in particular that a subfolder-only value is still judged so the reporter's own `3d/absent.glb` keeps reporting.

**3. Config propagation is a real behavioural dependency, not bookkeeping.** `configs.set(name, def[1] …)` for V2 is what arms the #1357 upload abstention. If it were dropped the module still compiles and 27 of 29 tests still pass, while LoadImage-style abstentions quietly stop arming for V2 inputs. I pinned it with a mutation, but a second opinion on whether `def[1]` is the right config object for every V2 shape would help.

**4. Where I chose NOT to act, in case you disagree.** Remote V2 and dynamic V3 yield `null` and stay both unjudged *and* undisclosed. The issue text complains about values being "neither reported nor disclosed in `unchecked_nodes`", so an argument exists that they should be disclosed. I judged that 113 types carry a dynamic combo and a permanent `unknown` entry on ordinary canvases repeats the #1284 error-surface flooding. If you think the honesty argument wins, say so.

Not gated and not reviewed — no codex run on this. Unit 6122/0, tsc clean, check-panel-scope OK; Playwright `--workers=1` is running against a 19-failed/62-passed baseline and I will post the per-file comparison.

artokun:
Pushed `ea147dd6` — an adversarial re-read of my own diff against the live `/object_info` found a false positive that **this branch introduced**, and the first commit shipped it. Recording it here because it changes what is worth reviewing.

`parseClassCombos` stored `values.filter((v) => typeof v === "string")`. Harmless for V1 — every V1 list on this server is all-strings — but a combo may publish **numbers**, and filtering those out turns a real list into `[]`. `[]` is not neutral in this module: it means "the server enumerates nothing", which yields `option_count: 0` and `kind: "missing_asset"` — a claim that nothing of the kind is installed.

Measured: **15 inputs publish pure-int lists** (`LtxvApiTextToVideo.duration [6,8,…]`, `MinimaxHailuoVideoNode.duration [6,10]`, `HitPawGeneralImageEnhance.upscale_factor [1,2,4]`, …) and **all fifteen are V2**. Zero were reachable before this branch, so every one would have been a false `missing_asset` caused by this change and nothing else — the exact failure mode #774 refused to risk, arriving through a side door.

Two narrow fixes: the list is stored **as declared**, and membership goes through `comboOffers`, which compares by value so a combo publishing the number `10` still matches a widget holding the string `"10"` (which is what a workflow round-trip or a `panel_set_widget` write hands back). Only primitives coerce; an object never does.

Re-measured on the same 257-class canvas with values set to `String(option)`:

| | origin/main | this PR |
|---|---|---|
| server-offered values | 0 false positives | **0 false positives** |
| poisoned values | 34 of 474 (440 missed) | **474 of 474** |
| findings claiming `option_count: 0` on a healthy list | 0 | **0** |

Both pinned by mutation: restoring the filter fails both new tests, reverting `comboOffers` to `includes()` fails the stringified-int test.

Two things this says about the review I asked for. My risk item **#1** — "a value that is legitimately resolvable while not being a member" — was real, and the family was not the one I guessed; it was type coercion, not annotation. So that item is worth more scrutiny, not less. And the pattern is the one this fix is *about*: a reader that was correct for the shape it was written against, quietly wrong for the shape that replaced it.

artokun:
Closing the loop over **real HTTP** rather than fixtures, since reachability is the thing green unit tests cannot show. This drives a live ComfyUI 0.33.2 on :8188 through the panel's own `fetchSingleNodeDef` into `scanComboAvailability` — the exact chain `panel_get_errors` uses at `comfyui-mcp-panel.js:16448`. No fixture anywhere in the path.

| case | `origin/main` | this PR |
|---|---|---|
| `Load3D.model_file = "3d/absent.glb"` — the reporter's case | **omitted entirely** | `invalid_value`, option_count 1 |
| `Load3D.model_file = "none"` — a value the server offers | none | **none** |
| `Load3D.model_file = "chair.glb [output]"` — annotated | silently dropped | **unchecked, with a reason** |
| `LtxvApiTextToVideo.duration = "10"` — stringified int the server offers | none | **none** |
| `LtxvApiTextToVideo.duration = "99"` — int list, genuinely off-list | **omitted entirely** | `invalid_value`, option_count **8** |
| `CheckpointLoaderSimple.ckpt_name = "__nope__"` — V1 control | `missing_asset` | `missing_asset` |

Three things worth reading off it. The V1 control is byte-identical before and after, so nothing already being judged changed behaviour. `option_count: 8` on the int list is the collapse-to-empty defect not happening. And both "the server offers this" rows report nothing on both sides — the false-positive direction holds on real server bytes, not just on my canvas.

The panel was verified to be serving *this branch* before the run (`/extensions/comfyui-mcp-panel/js/lib/live-combo-availability.js` carries `comboOffers`), so this is the code that ships, not the code in my worktree.

**What I did not do here:** I did not add a Playwright spec for `panel_get_errors`. That surface has no e2e spec in this repo — it is covered by ten unit files and zero specs — so adding one would mean building a new e2e harness for a tool reply rather than using an existing rail, inside a fix for an omission. The reachability question it would answer is answered above on real bytes. Say the word if you would rather have the spec.

artokun:
@grok re-requesting on the current head (`324c50d0`) — the head moved twice since my first request and one of those moves fixed a false positive I had shipped, so the earlier request was against a diff that no longer exists.

The PR body now carries the full evidence. Where to aim, in priority order:

1. **`comboOffers` coercion.** This is the newest and least-exercised code. It decides that a combo publishing the number `10` offers a widget holding `"10"`. If that coercion is too generous anywhere, the failure is *silent* — a genuinely-wrong value stops being reported, i.e. this PR's own bug class in reverse. I restricted it to string/number/boolean and excluded objects, but the boolean arm is speculative; I found no boolean combo on the live server.
2. **The int-list defect I shipped in the first commit.** Worth reading as evidence about my judgement here, not just as a fixed bug: I reasoned my way to "filtering non-strings is harmless" and it was false for 15 live inputs, all V2, all newly reachable. If there is a third coercion family I have not thought of, that is where it will be.
3. **`declaresUnrecognizedUploadKind`.** Makes the scan go *quiet* on an annotated value. Justified by "0 of 652 option strings carry an annotation" — one server's node set. A pack shipping annotated options would make this abstain where it should report.
4. **The scope call in "what I deliberately did not do".** Remote V2 and dynamic V3 stay both unjudged and undisclosed. Argue me out of it if you think the honesty rule wins over the #1284 flooding risk.

Status while it waits: unit 6125/0, tsc clean, check-panel-scope OK, Playwright measured against a real ComfyUI both ways — `origin/main` 18 failed / 63 passed, this branch 19 failed / 62 passed, the single delta re-run alone 3x at 10/10 (load flake).

**This PR is HELD, not stalled.** It is green and mergeable, and I am deliberately not merging it on my own verification — the brief is explicit that a self-review is not a review, and no non-Copilot review has landed on any open panel PR today. It stays open until a reviewer answers. Not labelled `autopilot:parked`: this is a ready fix awaiting review, not a terminal decision.

artokun:
Heads-up on overlap, from #1569 / #1571 (adding `file_upload` to `UPLOAD_CONFIG_FLAGS`). **No file conflict** — #1571 touches `web/js/lib/input-asset.js` only, this PR touches `live-combo-availability.js` and its own test. Two interactions, both self-resolving in either merge order, but worth knowing about:

1. `declaresUnrecognizedUploadKind` matches `/_upload$/` on any flag `uploadConfigOf` does not know — which today includes `file_upload`, so this PR's annotated-value abstention currently covers Load3D. Once #1571 lands, `uploadConfigOf` answers truthy for `file_upload`, that predicate returns false for it, and the workaround goes inert for Load3D while staying armed for a genuinely unknown future flag — which is exactly the intent. #1571 deliberately does **not** delete the workaround; whether it eventually goes is this PR's call.
2. Going the other way: the read-path half of #1571 **cannot reach a live server until this PR lands.** `parseClassCombos` on main recognises the V1 shape only, and live 0.33.2 publishes `Load3D.model_file` as `["COMBO", {multiselect:false, options:["none"], file_upload:true}]` — so today the scan does not see it as a combo at all and never judges it. Measured on a 0.33.2 booted for that change. #1571's get_errors tests therefore drive the V1 shape (which ComfyUI really did ship for Load3D from `bdf39379` until the V3 conversion in `440268d3`), and say so in a comment; the V2 reading is this PR's to prove.

Nothing to change here — just so neither of us is surprised at merge time.

artokun:
Direct Codex merge gate: NO-SHIP (P1 x2). web/js/lib/live-combo-availability.js:213 coerces numeric options to strings, so options [6,8,10] with widget.value "10" returns clean even though ComfyUI documents this stringification as a queued-prompt type bug. At line 429 numeric widget values are skipped, so options [6,8,10] with widget.value 99 returns unavailable=[] and unknown=[] in execution. Preserve numeric semantics, add regressions for both states, rerun CI, and request a fresh direct gate.

artokun:
Both P1s accepted and fixed in `96a4f70a`. Reproduced in execution first, then settled against the server rather than by reasoning from the widget type.

**Both reproduced exactly as reported**, on the head codex reviewed:

```
options [6,8,10], widget.value "10"  -> {"unavailable":[],"unknown":[]}
options [6,8,10], widget.value 99    -> {"unavailable":[],"unknown":[]}
```

**The server is the authority, and it agrees with codex.** `execution.py` validates a combo with `invalid_vals = [val] if val not in combo_options else []`, and Python's `in` is `==`. Run against this machine's ComfyUI 0.33.2 interpreter:

| value | `in [6,8,10]` | backend |
|---|---|---|
| `10` | True | accepted |
| `10.0` | True | accepted |
| `"10"` | **False** | **rejected** — `value_not_in_list` |
| `99` | False | rejected |

So `"10"` is a real defect, not a spelling of a good value — and the linked #14641 (the frontend stringifying combo values on queue) is exactly how a canvas acquires one. My coercion was hiding the failure the user meets at queue time, with nothing else to warn them. `comboOffers` is now `o === value`, which reproduces Python `==` exactly for the only two types a live option list contains (measured: string 2630, number 44, **zero booleans** — so the `True == 1` quirk has nothing to model).

**P1 #2 was the more damaging one, and you were right about why.** The guard read `typeof value !== "string"`. Numbers are now judged; `null`/`undefined`/objects/`""` still skipped, and `0` explicitly not dropped as falsy. Only a string routes through the annotated-filepath parser.

**You also named the root correctly.** This branch taught the scan to *read* numeric option lists; both findings are the same seam on the *value* side. Neither was reachable before this branch — all 44 numeric options sit on V2 combos the old parser could not see — so both were defects this PR introduced, not pre-existing ones it exposed.

Re-measured on the live payload, 257 classes / 474 combo widgets (15 numeric):

| | `origin/main` | this PR |
|---|---|---|
| A — server-declared values | 0 false positives | **0 false positives** |
| B — poisoned, same type | 34 of 474 | **474 of 474** |
| C — numeric values **stringified** (the #14641 trap) | 0 of 15 reported | **15 of 15 reported** |

Direction A is the control your fix must not break; it holds at zero.

Mutations, each **asserted to have applied** (byte delta printed) before running — including the call site:

| mutation | fails |
|---|---|
| re-introduce the string coercion | stringified-value + type-faithful tests |
| value guard back to string-only | numeric-value + `0`-is-real tests |
| **call site**: skip numbers past `comboOffers` | the same two |
| drop the empty-string guard | non-primitive-skipped test |

One thing outside your findings: **#1571 merged**, so `file_upload` is now in `UPLOAD_CONFIG_FLAGS` and the canonical #1357 upload path arms for `Load3D`. The `declaresUnrecognizedUploadKind` workaround this branch carried is therefore dead code and is **deleted**. The reporter's-case test now injects the `confirmServerAsset` probe production always injects, because a subfolder value on a recognised upload input is settled by `/view`, not by the combo — the old fixture asserted against a shape production never has.

Unit 6190 pass / 0 fail / 1 todo; tsc clean; check-panel-scope OK. The single suite failure was manager-install #671's known wall-clock flake under load — 125/125 twice in isolation.

artokun:
Gate round 2 returned **NO-SHIP** on one P1; fixed in `47ac19b8` (now `614cacc3` after merging current main). Verdict verbatim:

> [P1] `live-combo-availability.js:414` — boolean combo values are skipped despite boolean options being accepted — with `options: [false]` and widget value `true`, the scan returns clean while the server rejects the value.

Correct, and it is the **third instance of one seam**: this branch taught the scan to READ V2 option lists, and each round has closed one more type on the VALUE side — numbers stringified, numbers skipped, now booleans skipped. Worth stating plainly since it is a pattern in my own work, not three unrelated bugs.

**The naive fix would have introduced a false positive**, which is why I measured before writing. Python's `bool` is a subclass of `int`, so `True == 1`. From this machine's ComfyUI 0.33.2 interpreter:

| | | | | |
|---|---|---|---|---|
| `True in [False]` | **False** — reject | | `True in [1, 2]` | **True — ACCEPT** |
| `False in [False]` | True — accept | | `False in [1, 2]` | False — reject |
| `True in [True, False]` | True — accept | | `False in [0, 1]` | **True — ACCEPT** |
| `1 in [True, False]` | True — accept | | `'true' in [True]` | False — reject |

A bare `===` reports `true` on options `[1, 2]` as unavailable while the server runs it fine. So the compare moved into `serverConsidersEqual`, implementing Python `==` for the primitives a combo can carry: identical primitives match, a boolean matches a number numerically, and **a string never matches either** — the rule your round-1 P1 established, preserved through the refactor.

**Verified by exhaustive agreement, not by example.** A 132-row truth table was generated by running `value in options` in the real interpreter across bool/int/float/str/empty-list combinations and transported over JSON — the same collapse `/object_info` performs, so it is the comparison that actually governs. `comboOffers` agrees **132 of 132**. The discriminating rows are pinned as a test.

Live controls unchanged — no live combo declares boolean options (measured: string 2630, number 44, **boolean 0**), so this closes the *shape*, not a sighting, and cannot move the numbers:

| | |
|---|---|
| A — server-declared values | **0 false positives** |
| B — poisoned, same type | **474 of 474** |
| C — numeric values stringified | **15 of 15** |

Real-HTTP agreement against the running server: **6 of 6** cases match the backend's own verdict.

Mutations, each asserted applied (byte delta) — the third is the one I care about most:

| mutation | fails |
|---|---|
| guard reverted so booleans are skipped | boolean + bool/int tests |
| bool/int rule dropped | bool/int + truth table |
| bool/int rule widened to coerce **strings** | **both round-1 P1 tests** + truth table |

That last one matters: the new helper cannot be used to silently undo your earlier finding.

Also merged current main (`03da405d`), which brings in #1567, #1570 and #1574 — my module is untouched by it (`git diff` on the file is empty). Unit 6194 pass / 0 fail / 1 todo; tsc clean; scope + vocabulary OK. The lone suite failure is manager-install #671's known wall-clock flake — 125/125 in isolation.

Re-gating on `614cacc3`, and re-running Playwright against the new base since main moved and added a spec.

artokun:
**Gate round 3: SHIP** on `614cacc3`, no findings.

```
node .claude/autopilot/codex-gate.mjs --worktree .../wt-745-cc --base origin/main --name 1568r3
SHIP
```

Worktree HEAD, remote branch and PR head were all confirmed identical (`614cacc3`) before the gate ran, and `origin/main` is unchanged at `03da405d` since I merged it — so this verdict is against the main that is actually there, not a stale one.

Three review rounds, and the honest summary is that **every finding was mine and they were all one seam**. Teaching the scan to READ V2 option lists opened the value side, and it took four passes to close it:

| round | defect | direction |
|---|---|---|
| self-review | integer lists collapsed to `[]` → `missing_asset` | false **positive** |
| codex P1 | `"10"` coerced to match option `10` | false **negative** |
| codex P1 | numeric values skipped entirely | false **negative** |
| codex P1 | boolean values skipped entirely | false **negative** |

That progression is worth recording rather than tidying away: I found the false-positive direction myself and was blind to the false-negative direction three times running, on the same seam, in the same file. The thing that finally closed it was not more care — it was replacing my judgement with the interpreter's, via a **132-row truth table generated from the real Python** and asserted 132/132 in JS.

Boundaries are stated in the body rather than left to be discovered: array/multiselect values are **not** judged (the server validates those element-wise, but 0 of 468 live V2 combos declare `multiselect`, so building it would mean writing machinery against a shape I cannot verify), and `null`/objects stay skipped.

Merging once the Playwright run finishes and I have compared it per-file — main added a spec since my last run, so the baseline moved from 81 to 82.

PR #1570, opened 2026-08-21
fix(1565): graph_run answers inside the window its reply is relayed in

### LINKED ISSUE #1565: panel_run times out on "graph_run" (20s) repeatedly while panel_query_graph reads on the same tab succeed instantly

While iterating on a graph in the sidebar panel, `panel_run` began failing with a 20s `graph_run` timeout — **twice in a row**, including an explicit `retry_of` retry. The notable part is the asymmetry: **read** commands to the *same* panel tab kept working perfectly throughout, with sub-second responses.

Sequence in one turn:

1. `panel_set_widget(node_id=328, widget="expand", value=-2)` → **succeeded**
2. `panel_run(to_node_id=327)` → **timed out** on `graph_run` after 20000 ms
3. `queue(action:"list")` (headless MCP, direct to ComfyUI) → `{running: 0, pending: 0}` — so nothing had actually been queued
4. `panel_run(to_node_id=327, retry_of="505b...")` → **timed out again**, identical error
5. `panel_query_graph(ids=[328,325], fields="compact")` → **succeeded immediately**, returned correct current state

So the tab was demonstrably alive and answering the panel on the read path while `graph_run` got no reply at all. The error text suggests a backgrounded/frozen tab, but that explanation doesn't fit a tab that is concurrently answering `panel_query_graph` and `panel_set_widget` without delay.

Root cause not established from outside the frontend. It looks like `graph_run` specifically depends on something (the frontend queue loop / `app.queuePrompt` chain) that can stall independently of the command channel the read/mutate tools use — so the tab reports healthy on one path and silently never acknowledges on the other.

### Steps to reproduce

Not deterministically reproduced. Observed after ~20 minutes of rapid iteration on a 69-node workflow with many user-initiated Queue presses interleaved with agent `panel_run` calls. Prior `panel_run` calls in the same session (prompts `b4c944c4…`, `e414846e…`, `41b3ded1…`) all queued and completed normally; the failure started immediately after.

### Exact error (scrubbed)

```
Error: Panel tab wf:a3b8bab3-...:workflows/BatchArs.json did not reply to "graph_run"
within 20000 ms — the ComfyUI tab may be backgrounded or frozen. This command MUTATES
and was already delivered to the tab, so it may have been applied despite the missing
reply — check the current state before re-issuing it; a blind retry can apply it twice

To retry this exact mutation, re-issue identical args plus retry_of:"505b6249-..."
```

Suggested improvement regardless of root cause: when `graph_run` times out, the panel could probe the same tab with a cheap read before returning. If the read answers, the "backgrounded or frozen tab" wording is actively misleading and sends the agent (and user) down the wrong diagnostic path — which is what happened here.

### Fix

applied locally: no
upstream-only: yes

No local patch attempted. The stall is inside the frontend/orchestrator queue path, and the orchestrator process only reloads on a full process restart, so a local patch could not have unblocked this user in-session anyway.

### Also observed (same session, may be a separate issue)

Every successful `panel_run` in this session returned `scope_applied_by: "request_body_repair"` with the note asking that this be reported. Including the requested details verbatim:

- Body keys the frontend produced, with **no** `partial_execution_targets`: `client_id`, `extra_data`, `number`, `prompt`
- Queue chain line, as requested: *"app.queuePrompt is shadowed by an own property; api.queuePrompt is shadowed by an own property; its prototype's source mentions partialExecutionTargets."*
- The note says frontend **1.48.7** carries the scope natively via the positional argument shape. This machine is on frontend **1.49.6** and takes the fallback path — so 1.49.6 is a build where the positional shape does *not* pass through, which may narrow the regression window.

Whether this is related to the `graph_run` timeout above is unknown; flagging it because the run path is common to both and the scope note explicitly requests these two lines.

### Environment

- OS: Windows 11 Enterprise 10.0.22621
- ComfyUI: 0.33.0
- ComfyUI frontend: **1.49.6**
- Python 3.13.11 / PyTorch 2.13.0+cu132 / CUDA 13.2
- GPU: NVIDIA GeForce RTX 5090, 31.8 GB VRAM (29.4 GB free at time of failure)
- RAM free: 128 GB
- Queue at time of failure: 0 running, 0 pending
- comfyui-mcp: **0.52.45**
- comfyui-mcp-panel: **0.15.32**
- Backend: Claude

---
Filed from the ComfyUI panel by **claude** · model `opus` · effort `high` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #1175

artokun:
Fixes #1565

## The report

`panel_run` timed out on `graph_run` after 20,000 ms, **twice in a row** including an explicit `retry_of` retry, while `panel_query_graph` reads to the **same panel tab** answered sub-second throughout. That asymmetry rules out a dead socket, a lost `hello` and a wedged bridge, so this is not a generic timeout — something makes the *run* path specifically block while the read path is fine.

## Reproduced, by execution, before anything was written

Driving the **shipped** `dispatchScopedRun` (the orchestration `graph_run` runs) against a frontend modelled on the reporter's machine — a busy queue processor (they had user Queue presses interleaved with agent runs) that **drops the scope in every `app.queuePrompt` argument shape**, which is what their `scope_applied_by: "request_body_repair"` on every successful run means:

| busy processor drain | `dispatchScopedRun` elapsed | outcome | POST /prompt after the caller's 20s deadline | read on the same objects |
|---|---|---|---|---|
| 4,800 ms/item | 19,255 ms | dispatched | 0 | 0.055 ms |
| 4,950 ms/item | 19,856 ms | dispatched | 0 | 0.041 ms |
| **4,990 ms/item** | **20,003 ms** | dispatched | **1** | 0.043 ms |
| `app.queuePrompt` never settles | **never replied** (raced out at 22 s) | — | — | 0.056 ms |

## What the run path awaits that the read path does not

1. **`app.graphToPrompt()`** — serializing the whole workflow, every extension's own `serializeValue` included. **No bound of any kind.**
2. **`app.queuePrompt()`** — the frontend's queue processor. **No bound of any kind**, and on the reporter's machine *both* `app.queuePrompt` and `api.queuePrompt` are shadowed by an extension's own property (they reported the chain line verbatim). A wrapper that never settles held `graph_run` open **forever**.
3. **the scoped-dispatch observation wait — once per argument shape**, each attempt starting a **fresh** `verifyTimeoutMs`. Four shapes x 5,000 ms = **20,000 ms: exactly the window `panel_run` relays this command in**, with nothing left for the reply itself.

`panel_query_graph` takes none of them: it reads live LiteGraph objects. That is the whole of the reported asymmetry.

`graph_run` was also the only mutating panel command with **no command budget at all**, and the only one relayed in 20,000 ms rather than 30,000 — `nodes_install`, `graph_add_node`, `graph_set_widget`, `graph_refresh_nodes` and `workflow_save` all have one. Same shape as #1192 and #671.

## Did the prompt actually run while the caller timed out?

**Yes, on the boundary — and that is worse than the reported bug.** At a 4,990 ms drain the run reported `dispatched` at 20,003 ms and the POST left at 20,003 ms, i.e. *after* the caller had already been told the tab "may be backgrounded or frozen". A `retry_of` retry of that "failed" run renders the branch **twice**. After the fix the same scenario answers in 3,762 ms with `serverPostsEver: 0` — the abandoned item meets the fence and is refused, so nothing queues behind the caller's back.

(In the reporter's own session nothing was queued — their `queue(action:"list")` was empty — which is the sub-20s side of the same boundary.)

## The fix

**One deadline for the whole command**, taken on `graph_run`'s first line and threaded into the dispatch, so the bounds *compose* instead of summing.

- `RUN_COMMAND_BUDGET_MS = 15000` — the same 5,000 ms of slack the other command budgets leave against their own window, for composing the reply and the websocket hop. Mirrored, not derived: the relay constant lives in the other repo.
- `RUN_SERIALIZE_TIMEOUT_MS = 8000` bounds **one** `app.graphToPrompt()`; the command budget caps the sum whatever it says.
- `dispatchScopedRun` takes `budget` and gives each attempt a **fair share of what is left**, never more than `verifyTimeoutMs`. Fair share rather than first-come, because on the reporter's build the **last** argument shape is the only one that carries the scope — an allocation that starved the later attempts would refuse the runs that work today.
- `await app.queuePrompt(...)` is bounded on **both** the scoped and the full-graph paths, falling into the **existing** give-up / unknown-outcome paths.

**No budget means no bound.** `withTimeout` treats `ms <= 0` as no bound, so any caller that passes no budget keeps today's behaviour byte for byte.

## The full-graph run (review P1)

The first cut bounded only the scoped path and disclosed the unscoped one as deliberately out of scope, on the argument that abandoning an unbounded queue call risks claiming `queued` with no receipt. **Review returned that as a P1 and was right.** Measured against the shipped body:

- `graph_run({})` never replied; the caller timed out at **20,007 ms**, retried on the advice of a message blaming a "backgrounded or frozen" tab, and the retry queued at **20,022 ms** while the FIRST run's post landed at **21,011 ms** — **two full-graph renders from one user intent**, on the more common path.
- The receipt objection was also real: `buildQueueAcceptResult` with no ids answers a bare `{queued: true, batch_count: N}`.

It is resolved by routing into **`queued_unknown`** — with `indeterminate_count` and retry guidance, and `queued` *omitted* rather than set false — the honest shape an earlier round on this cluster had already built for exactly this state. One rule for the run path instead of two, and no receipt design left to file. `createRunFetchInterceptor` now carries live `state = { posted, inFlight }` so the reply states what actually left the panel instead of inferring it.

A defect that fix introduced, caught by measuring it: the first cut answered `queued: false` when nothing had left the panel. `posted === 0` is an observation about one instant while `app.queuePrompt` is still running — on the 21 s-drain frontend the panel answered at 1.5 s with `posted === 0` and the post left at 21 s anyway, **moving** the duplicate render rather than removing it. An abandoned run therefore never says `queued: false`.

The pending item is **not** cancelled, unlike the scoped path: cancellation there is licensed by an ownership tag plus a unique queue mark, and an unscoped item is byte-identical to what a user's own Queue press leaves. Cancelling it could kill **their** render, so it is disclosed instead.

A healthy full run is byte-identical: `{queued: true, batch_count: 1, prompt_id: …}` in 15 ms, budget never consulted.

## Why abandoning a queue call is safe here

Only because the give-up path already **CLOSES the guard** (`retireGuard` defaults false; the fence is only ever stood down when handing over to another attempt of the same run). A post the abandoned `queuePrompt` emits later carries this run's unique mark, meets the closed guard, and is **refused rather than dispatched scopeless** — measured: `serverPostsEver: 0`, and pinned by a test that installs a *second* run's guard on top. The bound also preserves a **rejection**: `withTimeout` folds a rejection into its timeout fallback, which would have turned a frontend that *threw* into one that "timed out", so the call settles into `{value}`/`{error}` first and the caller re-throws.

## Mutation evidence (fix deleted, named tests watched)

**Round 1 — the scoped path**

| mutation | tests that died |
|---|---|
| **M1 CALL SITE** — drop `budget,` from the `dispatchScopedRun` options | `CALL SITE: graph_run hands the dispatch its own command budget`, `CALL SITE: … never settles` |
| **M2 CALL SITE** — `graph_run` takes no budget | `CALL SITE: … own command budget`, `CALL SITE: … on the monotonic clock` |
| **M3 LIB** — per-attempt wait back to a fresh `verifyTimeoutMs` | `four argument shapes on a busy frontend cannot outlast ONE command budget` |
| **M4 LIB** — `app.queuePrompt` back to an unbounded await | `an app.queuePrompt that NEVER SETTLES is bounded…` |
| **M4b LIB** — falsy-throw check back to truthiness | `even a FALSY thrown value still throws…` |
| **M5 LIB** — fingerprint serialization back to an unbounded await | `an app.graphToPrompt that NEVER SETTLES is bounded, and nothing is queued` |
| **M6 CALL SITE** — pre-flight serialization takes its own fresh clock | `CALL SITE: … on the monotonic clock` |
| **M7 CONSTANT** — budget raised past the relay window | `the shipped run budget leaves room inside the 20,000 ms window…` |
| **M8 / M8b LIB** — abandoned attempt RETIRES, or RESTORES raw `fetchApi` | `P0: a run abandoned at its bound still FENCES its own late post…` |
| **M9 CALL SITE** — drop `Promise.resolve(...)` from the bounded pre-flight | `a SYNCHRONOUS graphToPrompt still reaches the pre-flight…` |
| **M10 LIB** — `boundedBy` always unbounded | both NEVER-SETTLES tests |
| **M5b LIB** — falsy-throw check on the fingerprint | **SURVIVED — equivalent mutant, reported not hidden** (see below) |

**Round 2 — the full-graph path**

| mutation | tests that died |
|---|---|
| **N1 CALL SITE** — unscoped `queuePrompt` back to an unbounded await | `P1: a full run whose queue call never settles ANSWERS inside its budget` |
| **N2 CALL SITE** — abandoned run falls through to the normal accept path | `P1: never answers a bare queued:true`; `P1: queued:false is never earned`; `P1: … real ids` |
| **N3 CALL SITE** — abandoned run with nothing posted claims `queued:false` | `P1: never answers a bare queued:true`; `P1: queued:false is never earned` |
| **N4 LIB** — interceptor stops counting what left the panel | `P1: the run interceptor COUNTS what left the panel` |
| **N5 LIB** — request classification moves back AFTER the fetch | `P1: an unreadable request cannot stop a fetch that previously went out` |

**M5b survived and is not being hidden.** On the fingerprint path the two forms are **observably identical**, so no test can separate them: verified by calling `canonicalizePrompt(undefined)` → `null` and `describeFingerprintFailure(undefined)` → `""`, both branches end at the same `unverifiable` refusal with an empty cause. The change is defensive symmetry with the `queuePrompt` bound, where it IS observable (M4b).

Every mutation was applied against a **committed** fix and reverted afterwards, with `git status --porcelain` verified clean between runs. Anchors that failed to match (CRLF, and a later edit) were re-anchored and re-run rather than read as "survived".

## Existing tests that had to follow the code

Four source anchors named literals the bounds replaced — `graph-to-prompt-failure`, `missing-node-preflight`, `scope-fingerprint-cause`, and `scoped-batch-seed`'s `#988` ordering guard. Each was re-anchored on the new form **and additionally asserts the call is still the bounded one**, so an anchor cannot drift back to an unbounded dispatch while the test stays green. `graph-to-prompt-failure.test.mjs` also needed `withTimeout` / `budget` / `RUN_SERIALIZE_TIMEOUT_MS` injected into its `new Function` harness: without them the extracted body throws a `ReferenceError` that the pre-flight's own catch swallows, and every one of those tests would read "no refusal" from a pre-flight that never ran. **No assertion was loosened.**

- **Did not raise the 20,000 ms relay budget.** A constant bump would have hidden exactly the block measured above.
- **Did not implement the reporter's suggested "probe the tab with a cheap read before returning the timeout".** That wording lives in the orchestrator repo (`ui-bridge.ts`), not here, and with the panel now answering inside the window it no longer fires for this cause.
- **Did not touch the `scope_applied_by: "request_body_repair"` / frontend-1.49.6 observation** they flagged as possibly separate. It is a real signal about that build and belongs in its own issue; it is not what blocked the run.
- **Did not cancel the unscoped pending queue item** — no ownership identity exists to do it safely; see above.
- No live-rig verification: this rig has no ComfyUI frontend 1.49.6 with the reporter's extension set. Everything above is measured against the shipped modules.

## Verification

- **unit**: `npm run test:unit` — **6196 passed / 0 failed** on the merged head (i18n, i18n-render, tool-vocabulary and panel-scope gates all OK).
- **gate**: `codex-gate.mjs --base origin/main --name 1570` → **SHIP**.
- **mutation**: seventeen mutations across both rounds, sixteen killed by named tests; the one survivor (M5b) is an observably equivalent mutant, reported above.
- **browser**: `npx playwright test --workers=1` — branch **17 failed / 64 passed** vs a clean `origin/main` baseline of **18 failed / 63 passed** measured back to back; the branch's failing set is a **strict subset** (zero regressions). Two limitations stated plainly in the evidence comment: the Tier-1 suite never exercises `graph_run` at all, and on this rig the pack junction serves the main checkout rather than the worktree — so no browser coverage is claimed for the fix itself.
- **main moved mid-round** (`75728263` → `869362ec`, three PRs, all touching the same 39k-line panel file). Merged in as `7ddf421f`, no conflicts, suite re-run green on the merged result, and both the scoped and unscoped measurements re-taken on it: the duplicate-render case answers at 15,007 ms with `queued_unknown`, and the scoped path still refuses at 3,761 ms with `serverPostsEver: 0`.

Full evidence and the exact spec-set diff are in the comments below.

### DIFF
--- a/browser_tests/unit/graph-to-prompt-failure.test.mjs
+++ b/browser_tests/unit/graph-to-prompt-failure.test.mjs
@@ -94,6 +94,18 @@ test("#1582 a long type list is bounded", () => {
 //    a few lines inside a 30k-line file that a refactor could drop with every unit test
 //    still green.
 
+test("#1565: a SYNCHRONOUS graphToPrompt still reaches the pre-flight — the bound must not remove a guard", async () => {
+  // An extension may replace graphToPrompt with a plain function returning the prompt
+  // object. `await` accepted that; a bare `.then` on a non-thenable throws, and the
+  // pre-flight's own catch would swallow it and skip a guard that used to run.
+  const msg = await runPreflight({
+    graphToPrompt: () => ({ output: { 1: { class_type: undefined, inputs: {} } }, workflow: {} }),
+    nodes: [{ id: 1, type: "LCKreaSampler", constructor: {} }],
+  });
+  assert.notEqual(msg, "__NO_REFUSAL__", "the pre-flight must still refuse an unrunnable node");
+  assert.match(msg, /^NOT queued:/);
+});
+
 test("#1582 the run path guards graphToPrompt BEFORE reading offenders", async () => {
   const { readFileSync } = await import("node:fs");
   const src = readFileSync(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url), "utf-8");
--- a/browser_tests/unit/graph-to-prompt-failure.test.mjs
+++ b/browser_tests/unit/graph-to-prompt-failure.test.mjs
@@ -16,6 +16,10 @@
 // `unrunnableNodeIds(undefined)` answers `[]` — no offenders — because a result that does not
 // exist has no unrunnable entries in it. Absence of evidence, read as evidence of absence.
 import test from "node:test";
+// #1565 — the pre-flight serialization is bounded by graph_run's command budget now;
+// the harness below drives the SHIPPED block, so it needs the same two collaborators.
+import { withTimeout } from "../../web/js/lib/bounded-step.js";
+import { makeCommandBudget } from "../../web/js/lib/command-budget.js";
 import assert from "node:assert/strict";
 
 import {
@@ -93,8 +97,16 @@ test("#1582 a long type list is bounded", () => {
 test("#1582 the run path guards graphToPrompt BEFORE reading offenders", async () => {
   const { readFileSync } = await import("node:fs");
   const src = readFileSync(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url), "utf-8");
-  const at = src.indexOf("const built = await app.graphToPrompt();");
+  // #1565 — the call is bounded by the command budget now, so the recognisable line is the
+  // bounded step rather than a bare await. Same block, same assertions below it, plus a
+  // check that it is still graphToPrompt being bounded.
+  const at = src.indexOf("const preflightBuild = await withTimeout(");
   assert.ok(at > 0, "the pre-flight's graphToPrompt call must still be recognisable");
+  assert.match(
+    src.slice(at, src.indexOf("const built = preflightBuild.value;", at)),
+    /app\.graphToPrompt\(\)/,
+    "and it must still be the pre-flight that serializes the prompt",
+  );
   // Bounded by the pre-flight's OWN catch, not a byte count. A fixed 1800 truncated
   // `unrunnableNodeIds(built)` the moment the guard's comment grew — which is the third
   // time today a fixed window in this repo reported missing wiring that was present
@@ -162,6 +174,14 @@ async function buildPreflight({ graphToPrompt, nodes = [], viewedNodes, registry
     // without this the extracted body throws a ReferenceError that the pre-flight catch
     // swallows, and the test sees no refusal at all.
     "window",
+    // #1565 — the serialization is BOUNDED by graph_run's command budget now. All three
+    // names are consts in graph_run, so leaving any of them out makes the extracted body
+    // throw a ReferenceError the pre-flight catch swallows — the test would then read "no
+    // refusal" from a pre-flight that never ran, which is the silent failure this file
+    // exists to stop, one layer down.
+    "withTimeout",
+    "budget",
+    "RUN_SERIALIZE_TIMEOUT_MS",
     `return async function preflight() {\n${body}\n};`,
   );
   // The VIEWED graph is deliberately a DIFFERENT object from the root when the caller
@@ -181,6 +201,11 @@ async function buildPreflight({ graphToPrompt, nodes = [], viewedNodes, registry
     { _nodes: nodes },
     mod.unresolvedNodeTypes,
     { LiteGraph: { registered_node_types: registry } },
+    withTimeout,
+    // A REAL budget with room to spare: this file is about what the pre-flight REFUSES,
+    // and an already-spent budget would answer every case with a skipped pre-flight.
+    makeCommandBudget(30000),
+    8000,
   );
 }
 
--- a/browser_tests/unit/missing-node-preflight.test.mjs
+++ b/browser_tests/unit/missing-node-preflight.test.mjs
@@ -121,8 +121,10 @@ test("#1460 a long list is bounded but says how many were withheld", () => {
 test("#1460 WIRING: the preflight reads the SERIALIZED prompt, not the canvas", () => {
   const src = readFileSync(join(ROOT, "web/js/comfyui-mcp-panel.js"), "utf8");
   assert.match(src, /import \{[\s\S]*?unrunnableNodeIdsInScope,[\s\S]*?\} from "\.\/lib\/missing-node-preflight\.js";/);
-  const at = src.indexOf("const built = await app.graphToPrompt();");
+  // #1565 — bounded by graph_run's command budget now; the block is the same one.
+  const at = src.indexOf("const preflightBuild = await withTimeout(");
   assert.ok(at > 0, "the preflight must serialize the prompt itself");
+  assert.match(src.slice(at, at + 400), /app\.graphToPrompt\(\)/, "and it is graphToPrompt it bounds");
   // Bounded by the pre-flight's OWN catch rather than a byte count. A fixed 600 excluded
   // `unrunnableNodeIds(built)` as soon as #1582 added a guard above it — reporting missing
   // wiring while the wiring was fine. The window exists to keep this assertion scoped to
@@ -212,7 +214,7 @@ test("#1871 WIRING: the run-to-node target is RESOLVED before the pre-flight rea
 
 test("#1460 WIRING: it runs BEFORE the prompt is queued", () => {
   const src = readFileSync(join(ROOT, "web/js/comfyui-mcp-panel.js"), "utf8");
-  const check = src.indexOf("const built = await app.graphToPrompt();");
+  const check = src.indexOf("const preflightBuild = await withTimeout(");
   const queue = src.indexOf("app.queuePrompt", check);
   assert.ok(check > 0 && queue > check, "the check must precede queuePrompt");
 });
--- a/browser_tests/unit/run-command-budget.test.mjs
+++ b/browser_tests/unit/run-command-budget.test.mjs
@@ -556,3 +556,206 @@ test("#1565 CALL SITE: graph_run takes its budget from RUN_COMMAND_BUDGET_MS on 
     "the pre-flight serialization draws from the command budget, not its own fresh clock",
   );
 });
+
+// ---------------------------------------------------------------------------
+// 5. THE UNSCOPED FULL RUN — `graph_run({})`, no to_node_id.
+//
+// The first cut of #1565 bounded only the scoped path and disclosed the unscoped one as
+// deliberately out of scope, on the argument that abandoning an unbounded queue call
+// risks claiming `queued` with no receipt. A review returned that as a P1. MEASURED
+// against the shipped body, both halves of the argument turned out to be true and only
+// one of them survives:
+//
+//   - the harm is REAL and identical to the scoped one. `graph_run({})` against a
+//     frontend whose queue processor never answers never replied; the caller timed out
+//     at 20,007 ms, retried on the advice of a message blaming a "backgrounded or
+//     frozen" tab, and the retry queued at 20,022 ms while the FIRST run's post landed
+//     at 21,011 ms — TWO full-graph renders from one user intent, on the MORE COMMON
+//     path;
+//   - and the receipt objection is real: `buildQueueAcceptResult` with no ids answers a
+//     bare `{queued: true, batch_count: N}` — a definite positive with nothing behind it.
+//
+// What resolves it is that the honest vocabulary already existed. `queued_unknown` was
+// built for exactly this epistemic state on the scoped path, so the bound routes into it
+// rather than into a `queued` the run cannot back.
+// ---------------------------------------------------------------------------
+
+/**
+ * A frontend for the UNSCOPED path.
+ * @param {"late"|"never"|"silent"|"postThenHang"} mode
+ *   late         posts and settles after drainMs (healthy)
+ *   never        never settles; posts drainMs later (the busy processor)
+ *   silent       never settles, never posts
+ *   postThenHang posts and captures a prompt_id, then never settles (partial batch)
+ */
+function makeUnscopedFrontend({ apiTarget, mode, drainMs = 5 }) {
+  const app = {
+    queueItems: [],
+    graph: { _nodes: [] },
+    graphToPrompt: async () => ({ output: OUR_OUTPUT, workflow: {} }),
+    queuePrompt: async () => {
+      const post = () =>
+        apiTarget.fetchApi("/prompt", {
+          method: "POST",
+          headers: { "Content-Type": "application/json" },
+          body: JSON.stringify({ prompt: OUR_OUTPUT, client_id: "x" }),
+        });
+      if (mode === "silent") return new Promise(() => {});
+      if (mode === "postThenHang") {
+        await post();
+        return new Promise(() => {});
+      }
+      if (mode === "never") {
+        const t = setTimeout(post, drainMs);
+        if (typeof t.unref === "function") t.unref();
+        return new Promise(() => {});
+      }
+      await post();
+      return true;
+    },
+  };
+  return app;
+}
+
+test("#1565 P1: a full run whose queue call never settles ANSWERS inside its budget — it does not hang to the relay timeout", async () => {
+  const stop = keepAlive();
+  try {
+    const apiTarget = { fetchApi: makeServer() };
+    const app = makeUnscopedFrontend({ apiTarget, mode: "never", drainMs: 60000 });
+    const built = realGraphRun({ app, apiTarget, budgetMs: 600, serializeMs: 400 });
+    const started = Date.now();
+    const answer = await Promise.race([
+      built.graph_run({}).then((r) => ({ replied: r })),
+      new Promise((r) => setTimeout(() => r({ hung: true }), 5000)),
+    ]);
+    assert.ok(
+      !answer.hung,
+      "graph_run({}) never replied — the unbounded queue call is the reported hang, on the " +
+        "MORE COMMON path: the caller then times out and retries, and the late post still lands",
+    );
+    assert.ok(Date.now() - started <= 5000);
+  } finally {
+    stop();
+  }
+});
+
+test("#1565 P1: an abandoned full run NEVER answers a bare queued:true — the receipt objection, honoured", async () => {
+  const stop = keepAlive();
+  try {
+    const apiTarget = { fetchApi: makeServer() };
+    const app = makeUnscopedFrontend({ apiTarget, mode: "silent" });
+    const built = realGraphRun({ app, apiTarget, budgetMs: 600, serializeMs: 400 });
+    const res = await built.graph_run({});
+    // buildQueueAcceptResult with no ids answers {queued:true, batch_count:N}. Routing an
+    // abandoned run through it would assert a render nobody observed.
+    assert.notEqual(res.queued, true, "a run with no receipt must never claim it queued");
+    assert.equal(res.queued_unknown, true, "it takes the honest shape the scoped path already uses");
+    assert.equal("queued" in res, false, "and OMITS `queued` — there is no boolean that means unknown");
+    assert.match(String(res.retry_guidance), /blind retry renders the whole graph twice/);
+  } finally {
+    stop();
+  }
+});
+
+test("#1565 P1: `queued:false` is never earned by an abandoned run — a bounded observation cannot license an unbounded negative", async () => {
+  const stop = keepAlive();
+  try {
+    const apiTarget = { fetchApi: makeServer() };
+    const server = apiTarget.fetchApi;
+    // Nothing has left the panel when the budget expires — but the queue call is STILL
+    // RUNNING and posts afterwards. Measured on the real body: answered at 1.5 s with
+    // posted === 0, and the post left at 21 s anyway. Telling the caller "nothing was
+    // queued, safe to re-issue" there moves the duplicate render rather than removing it.
+    const app = makeUnscopedFrontend({ apiTarget, mode: "never", drainMs: 400 });
+    const built = realGraphRun({ app, apiTarget, budgetMs: 250, serializeMs: 150 });
+    const res = await built.graph_run({});
+    assert.equal(server.calls.length, 0, "nothing had left the panel at the moment of the reply");
+    assert.notEqual(
+      res.queued,
+      false,
+      "`queued:false` says the run can be re-issued safely; the frontend can still post it",
+    );
+    assert.equal(res.queued_unknown, true);
+    assert.match(String(res.retry_guidance), /STILL RUNNING/);
+    // And the late post does arrive, which is exactly why the negative was not honest.
+    await new Promise((r) => setTimeout(r, 700));
+    assert.equal(server.calls.length, 1, "the abandoned queue call posted after the reply");
+  } finally {
+    stop();
+  }
+});
+
+test("#1565 P1: prompts that DID queue before the budget expired are reported with their real ids", async () => {
+  const stop = keepAlive();
+  try {
+    const apiTarget = { fetchApi: makeServer() };
+    const app = makeUnscopedFrontend({ apiTarget, mode: "postThenHang" });
+    const built = realGraphRun({ app, apiTarget, budgetMs: 600, serializeMs: 400 });
+    const res = await built.graph_run({});
+    // Reporting this as a failure would send the caller to re-run work that is rendering.
+    assert.equal(res.queued, true);
+    assert.equal(res.partially_queued, true);
+    assert.deepEqual(res.queued_prompt_ids, ["srv-1"]);
+    assert.equal(res.complete, false);
+    assert.match(String(res.incomplete_reason), /command budget/);
+  } finally {
+    stop();
+  }
+});
+
+test("#1565 P1: a HEALTHY full run is untouched — same accept result, no budget in sight", async () => {
+  const stop = keepAlive();
+  try {
+    const apiTarget = { fetchApi: makeServer() };
+    const app = makeUnscopedFrontend({ apiTarget, mode: "late", drainMs: 5 });
+    const built = realGraphRun({ app, apiTarget, budgetMs: 15000, serializeMs: 8000 });
+    const res = await built.graph_run({});
+    assert.deepEqual(res, { queued: true, batch_count: 1, prompt_id: "srv-1" });
+  } finally {
+    stop();
+  }
+});
+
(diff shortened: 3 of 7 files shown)


artokun:
@grok please review — and please go at these five specific claims rather than the diff as a whole. Named in descending order of how much I would like to be wrong about them.

**1. Abandoning `await app.queuePrompt(...)` is safe because the give-up path CLOSES the guard.** This is the load-bearing safety claim of the whole change and it is the one that, if wrong, ships a full-graph execution. The argument: `retireGuard` defaults to `false`, and the only exit that sets it true is the hand-over to another attempt *of the same run* — so a bounded-out attempt falls through to `finally { guard.close() }`, and a post the abandoned `queuePrompt` emits later carries this run's unique mark, hits the `state.closed` branch, and gets `SCOPE_DROPPED_RESPONSE()` instead of being forwarded. I measured `serverPostsEver: 0` for that scenario. **What I cannot rule out from the code alone: a path where the bound fires while `apiTarget.fetchApi` has already been overwritten by a *concurrent* run's guard**, so that our closed guard is no longer in the chain the late post travels. The chain-composition notes in `dispatchScopedRun` (r8) say a guard is never unhooked, only retired — I believe that makes it impossible, but I did not construct a two-concurrent-runs test for the new bounded exit, and that is the gap I am least comfortable with.

**2. The fair-share allocation, `shareFor(attemptsLeft)`.** Each attempt gets `min(verifyTimeoutMs, remaining / attemptsLeft)` rather than a fresh `verifyTimeoutMs`. The *hard* guarantee comes from `budget.bounded()` capping by what is left; the share is only a fairness heuristic. The behavioural consequence I chose deliberately: a frontend whose busy processor drains between ~3,750 ms and 5,000 ms per item used to walk all four argument shapes and blow the 20,000 ms window; it now gives up on attempt 1 with a worded refusal. I argue that is strictly better — at 20 s the caller has already been told "unknown outcome" and a retry double-renders — but it IS a real behaviour change for a slow-but-working frontend, and I would like a second opinion on the direction, not just the arithmetic.

**3. `boundedStep`'s `{value}`/`{error}` settling.** `withTimeout` folds a rejection into its timeout fallback, so wrapping `app.queuePrompt` naively would have relabelled a frontend that THREW as one that "timed out". I settle into a discriminated object first and re-throw. Test: *"a queuePrompt that THROWS still throws"*. Please check I have not created the mirror error somewhere — e.g. a resolved value that is legitimately `undefined`, or a `TIMED_OUT` Symbol that could collide.

**4. `RUN_COMMAND_BUDGET_MS = 15000` against a 20,000 ms relay window.** The other command budgets are 25,000 against 30,000, so I kept the same 5,000 ms of absolute slack rather than the same ratio. If the ratio is the right invariant the number should be ~16,600. I do not think 1.6 s matters here, but it is a number I picked rather than measured.

**5. The four existing test anchors I moved.** `graph-to-prompt-failure.test.mjs` / `missing-node-preflight.test.mjs` / `scope-fingerprint-cause.test.mjs` all indexed on the literal `const built = await app.graphToPrompt();`, which the bound replaced. I re-anchored on the bounded step and added an assertion that it is still `graphToPrompt` being bounded — please confirm I re-anchored rather than loosened, especially the `#1582` ordering test (`guard < offenders`), whose block boundary is still found via `src.indexOf("} catch (err) {", at)` and now starts from a different `at`. I also had to inject `withTimeout` / `budget` / `RUN_SERIALIZE_TIMEOUT_MS` into `buildPreflight`'s `new Function` factory: without them the extracted body throws a `ReferenceError` that the pre-flight's own catch swallows, and every one of those tests would read "no refusal" from a pre-flight that never ran. That is the exact silent-pass shape the file exists to prevent, so it is worth a second pair of eyes that the injection is faithful.

Reproduction, measurements and the seven-mutation table are in the PR body. Nothing here has been reviewed or gated.

artokun:
## Verification evidence — final head `1781717f`

### Unit

`npm run test:unit` — **6127 passed / 0 failed** (1 skipped). `i18n-check`, `i18n-render-check`, `check-tool-vocabulary` and `check-panel-scope` all OK.

### Browser suite — what it can and cannot say about this change

`npx playwright test --workers=1`, this machine, back to back against the same live ComfyUI:

| tree | result |
|---|---|
| clean `origin/main` (`75728263`, its own worktree) | **18 failed / 63 passed** (7.5 min) |
| this branch (`fe258967`) | **17 failed / 64 passed** (7.4 min) |

The branch's failing set is a **strict subset** of the baseline's — zero regressions, one baseline failure (`workflow-chat-identity.spec.ts:379`) did not reproduce. Full comparison:

```
in BRANCH but not BASELINE (regressions):  (none)
in BASELINE but not BRANCH:               workflow-chat-identity.spec.ts:379 an abandoned turn cannot leak output into a conversation opened mid-turn
```

All 17/18 are the known chat-history / conversation-persistence / streaming / hidden-tab / external-orchestrator / workflow-chat-identity failures. None touch the run path.

**Two limitations, stated rather than implied:**

1. **The Tier-1 browser suite never exercises `graph_run` at all.** Grepping the specs for `graph_run` / `queuePrompt` / `dispatchScopedRun` returns only `auto-layout.spec.ts` and `connect-matcher.spec.ts`, and in both the hit is `graph_connect`'s own `to_node_id` argument, not a run. So this suite cannot validate or invalidate the change; it can only show the change breaks nothing else, which is what the table above shows.

2. **On this rig the pack junction serves the MAIN CHECKOUT, not the worktree.** `ComfyUI/custom_nodes/comfyui-mcp-panel -> C:/Users/Artokun/code/comfyui-mcp-panel` (currently `d5025cfd`). The specs run from whichever worktree invokes Playwright, but the panel JS under test is whatever that junction points at — so **neither run above served this branch's panel code**, and I am not claiming browser coverage for the fix. I did not repoint a shared install's junction to manufacture that coverage.

3. The final head could not be re-run: the ComfyUI the suite is pinned to (`localhost:8188`) went down mid-session. A live instance exists on `:8211` but it is a different install serving another agent's base-directory, so I did not point the suite at it or restart anything on the shared rig. Given (1), a third run would not have said anything about this change.

The meaningful gate here is the unit suite, which drives the **shipped** `graph_run` body and the **real** `dispatchScopedRun`.

### Mutation table (final head), fix deleted one piece at a time

| mutation | named tests that died |
|---|---|
| **M1** CALL SITE: drop `budget,` from the `dispatchScopedRun` options | `CALL SITE: graph_run hands the dispatch its own command budget`; `CALL SITE: graph_run answers inside its budget when the frontend queue call never settles` |
| **M2** CALL SITE: `graph_run` takes no budget | `CALL SITE: graph_run hands the dispatch its own command budget`; `CALL SITE: … on the monotonic clock` |
| **M3** LIB: per-attempt wait back to a fresh `verifyTimeoutMs` | `four argument shapes on a busy frontend cannot outlast ONE command budget` |
| **M4** LIB: `app.queuePrompt` back to an unbounded await | `an app.queuePrompt that NEVER SETTLES is bounded by the command budget` |
| **M4b** LIB: falsy-throw check back to truthiness (queuePrompt) | `even a FALSY thrown value still throws…` |
| **M5** LIB: fingerprint serialization back to an unbounded await | `an app.graphToPrompt that NEVER SETTLES is bounded, and nothing is queued` |
| **M6** CALL SITE: pre-flight serialization takes its own fresh clock | `CALL SITE: graph_run takes its budget from RUN_COMMAND_BUDGET_MS on the monotonic clock` |
| **M7** CONSTANT: budget raised past the relay window | `the shipped run budget leaves room inside the 20,000 ms window…` |
| **M8** LIB: an abandoned attempt RETIRES (transparent) instead of CLOSING | `P0: a run abandoned at its bound still FENCES its own late post…`; `four argument shapes…` |
| **M8b** LIB: an abandoned attempt RESTORES raw `fetchApi` | same two |
| **M9** CALL SITE: drop `Promise.resolve(...)` from the bounded pre-flight | `a SYNCHRONOUS graphToPrompt still reaches the pre-flight…` |
| **M10** LIB: `boundedBy` always unbounded (budget ignored) | `an app.queuePrompt that NEVER SETTLES…`; `an app.graphToPrompt that NEVER SETTLES…` |
| **M5b** LIB: falsy-throw check back to truthiness (fingerprint) | **SURVIVED — reported, not hidden.** See below. |

**M5b survived and I am not going to pretend otherwise.** On the fingerprint path the two forms are **observably identical**, so no test can separate them: with the truthiness form a `throw undefined` leaves `built.error` falsy, execution continues to `canonicalizePrompt(undefined)` which returns `null` (verified by calling it), so `contentHash` is null and the run refuses `unverifiable`; with `"error" in` the value is re-thrown, the catch binds a falsy cause, `describeFingerprintFailure(undefined)` returns `""` (also verified), and the run refuses `unverifiable` with the same empty cause. The change is defensive symmetry with the `queuePrompt` bound (where it IS observable — M4b), not a behaviour I can pin.

Every mutation was applied against a **committed** fix and reverted afterwards, with `git status --porcelain` verified clean between runs. One run timed out with the mutation still applied; the file was restored from HEAD and re-verified before continuing. M1's and M4's first anchors did not match (CRLF, and a later edit) — both re-anchored and re-run rather than read as "survived".

### Review

**Still pending.** No review has been posted on this PR. The only entry on the reviews API is `copilot-pull-request-reviewer[bot]` saying it "reached their quota limit", which is not a review. No review has been posted on any open PR in this repo since 07:08Z today, so this is held on an absent reviewer rather than merged on the author's own verification.

artokun:
## HELD on an absent reviewer — not merged

State: **open, ready for review, CI green** (`Lint pack` SUCCESS), mergeable, and `origin/main` has not moved since the checks ran (`75728263`).

**Not merging.** No review has been posted on this PR. The only entry on `/pulls/1570/reviews` is `copilot-pull-request-reviewer[bot]` reporting a quota limit, which is not a review. Nothing has reviewed any open PR in this repo today either — `1561`, `1566`, `1567` and `1570` all carry exactly one Copilot quota notice and nothing else — so this is an absent reviewer, not a slow one, and merging on the author's own verification is exactly the thing that put a live P1 into 0.51.4.

**What would unblock it:** one independent review of the five claims named in the request comment above. The one I most want a second opinion on is the first: that abandoning `await app.queuePrompt(...)` is safe *only* because the give-up path closes the guard rather than restoring raw `fetchApi`. I have since proved that by mutation rather than by reading — making the abandoned attempt `retire()` instead of `close()`, and making it restore raw `fetchApi`, both kill the named P0 test — but a mutation proves my test matches my code, never that I reasoned about the right thing.

The change is inert on any frontend that answers promptly: all four argument shapes still run in milliseconds, and the budget never bites. It only ever fires where the command was previously going to blow its own reply window.

artokun:
Direct Codex merge gate: NO-SHIP (P1). web/js/comfyui-mcp-panel.js:15986 does not apply the command budget to normal full-graph runs: app.queuePrompt remains unbounded. graph_run({}) with a queue call that never settles can hang until relay timeout and a retry may duplicate a late queue. Add the missing budget enforcement and regression coverage, rerun CI, and request a fresh direct gate.

artokun:
## [P1] `comfyui-mcp-panel.js:15986` — accepted, fixed, and the objection I raised was measured rather than argued

The review is right, and it was right against my own evidence. I had scoped the unscoped full run out with a disclosure rather than a measurement. Both halves of that disclosure turned out to be true; only one of them survives contact with the rig.

### 1. The harm is real, and identical on the more common path

Driving the **shipped** `graph_run` body with `graph_run({})` against a frontend whose queue processor never answers:

```
callerOutcome:         timedOut
atTimeoutMs:           20007        <- the relay window, nothing replied
retryQueued:           true
retryPromptId:         srv-1
postsReachingComfyUI:  2
postTimes (from t0):   [20022, 21011]
```

The caller waited out the 20,000 ms window, was told the tab "may be backgrounded or frozen", retried — and the retry queued at 20,022 ms while the **first** run's post landed at 21,011 ms. **Two full-graph renders from one user intent.** Exactly the harm proved for the scoped path, on the path users hit more often. The gate's reading of my own evidence was correct.

### 2. My receipt objection was also real

```json
{ "whatTheNaiveBoundWouldReport": { "queued": true, "batch_count": 1 } }
```

`buildQueueAcceptResult` with no ids answers a bare positive with nothing behind it. So a naive bound genuinely would have traded a hang for a lie.

### 3. What resolves it: the honest vocabulary already existed

`queued_unknown` — with `indeterminate_count` and retry guidance, and `queued` **omitted** rather than set false — was built by an earlier round on this same cluster for precisely this epistemic state, and was reachable only from the scoped path. No receipt design was needed; the bound routes into it. **One rule for the run path instead of two**, which is what the review asked for. No new issue to file — the design it would have described is already shipped.

The counts are observed, not inferred: `createRunFetchInterceptor` now carries live `state = { posted, inFlight }`, so the reply states what actually left the panel.

### 4. A defect this fix introduced, caught by measuring it

The first cut answered `queued: false` when nothing had left the panel — "nothing was submitted, safe to re-issue". That is a **claim about the future** made from a **bounded observation**: `app.queuePrompt` is still running. On the 21 s-drain frontend the panel answered at 1.5 s with `posted === 0`, and the post left at 21 s anyway. A caller told it was safe to re-issue re-issues, and the duplicate render comes back — moved, not removed.

An abandoned run therefore never says `queued: false`; that answer is only reachable when `app.queuePrompt` actually settled. Pinned by `` `queued:false` is never earned by an abandoned run ``, which asserts the late post really does arrive.

### 5. Measured after the fix

| scenario | answered | reply |
|---|---|---|
| queue call never settles, posts at 21 s | **15,007 ms** (inside the window) | `queued_unknown`, "STILL RUNNING… a blind retry renders the whole graph twice" |
| queue call never settles, nothing ever posts | 1,506 ms | `queued_unknown` |
| posts + captures an id, then hangs | 1,502 ms | `queued: true, partially_queued, queued_prompt_ids: ["srv-1"]` |
| **healthy frontend** | **15 ms** | `{queued: true, batch_count: 1, prompt_id: "srv-1"}` — byte-identical, budget never consulted |

### Why the pending item is not cancelled, unlike the scoped path

Cancellation there is licensed by an ownership tag **plus** this run's unique queue mark. An unscoped full run's pending item carries `number: 0` and no scope — byte-identical to what a user's own Queue press leaves. Removing it could cancel **their** render, so it is left alone and disclosed instead. Stated as a limit rather than papered over.

### Mutation evidence for this round

| mutation | named tests that died |
|---|---|
| **N1** the unscoped `queuePrompt` back to an unbounded await | `P1: a full run whose queue call never settles ANSWERS inside its budget` |
| **N2** the abandoned run falls through to the normal accept path | `P1: never answers a bare queued:true`; `P1: queued:false is never earned`; `P1: prompts that DID queue … real ids` |
| **N3** an abandoned run with nothing posted claims `queued:false` | `P1: never answers a bare queued:true`; `P1: queued:false is never earned` |
| **N4** the interceptor stops counting what left the panel | `P1: the run interceptor COUNTS what left the panel` |
| **N5** request classification moves back AFTER the fetch | `P1: an unreadable request cannot stop a fetch that previously went out` |

All five killed. Worktree verified clean after each.

One more stale anchor followed the code: `scoped-batch-seed.test.mjs`'s `#988` ordering guard indexed on the literal `await app.queuePrompt(0, batch, undefined)`. Re-anchored on the call, **plus** a new assertion that it is still the bounded form — so the anchor cannot drift back to an unbounded dispatch while the ordering test stays green.

### Gate + suite

- `codex-gate.mjs --worktree … --base origin/main --name 1570` → **SHIP** (on `4c23c66c`).
- `npm run test:unit` → **6196 passed / 0 failed** on the merged head.
- `origin/main` moved during the round (`75728263` → `869362ec`, three PRs, all touching the same 39k-line panel file). Merged in as `7ddf421f`, no conflicts, suite re-run green on the merged result, and both the scoped and unscoped measurements re-taken on it: the duplicate-render case answers at 15,007 ms with `queued_unknown`, and the scoped path still refuses at 3,761 ms with `serverPostsEver: 0`.

**M5b stays disclosed as an equivalent mutant** — the fingerprint's falsy-throw check has no observable difference (verified by calling `canonicalizePrompt(undefined)` → `null` and `describeFingerprintFailure(undefined)` → `""`), so no test can separate the two forms and none pretends to.

PR #1574, opened 2026-08-21
fix(1573): three refusal/comment inaccuracies from #1561 now state what the code observed

### LINKED ISSUE #1573: Three refusal/comment inaccuracies left by #1561's type-scoped route (no behaviour change)
Three comment/message inaccuracies left behind by #1561 (merged `a1844f68`). All were raised by
that PR's fourth adversarial gate pass, all were classified sub-threshold and non-blocking, and
all were **deliberately not fixed there** — the PR was already at its third round of fixes, and
moving the head a fifth time would have meant merging a SHA the gate had not seen.

None of these changes behaviour. Every one is in the refusal direction, so none can cause a
wrongful write. Filing so they are tracked rather than lost.

**1. `web/js/lib/scoped-object-info.js` — a false claim in the module header.**

The `ownKeys` note says anything that ranges over the map — "`Object.keys(defs).length`, a
serializer" — sees "a small, honest object rather than a throw". `Object.keys` and spread do.
**`JSON.stringify(defs)` does not: it throws**, because it looks up `toJSON`, which is not in
`covered`, and the scope trap refuses it.

No production reader serializes `freshDefs`, so this is a false comment rather than a fault. It
matters anyway: in this repo a comment gets quoted back as an established measurement by the
next reader, so a header that says "a serializer is fine" will eventually license one.

Worth deciding at the same time whether `toJSON`/`toString`/`Symbol.toPrimitive` should simply be
**in scope** (they are not class types, and the module already passes symbols through), which
would make the comment true instead of correcting it downward.

**2. `web/js/lib/set-widget.js` — "A type-scoped read was tried too".**

`describeObjectInfoFailureWithScope` appends this whenever `scopedIneligibility` is set, including
the case where the read was **not licensed** and therefore nothing was issued. The clause after
the dash does say why, so the sentence self-corrects — but the lead-in asserts an attempt that did
not happen, which is the #982 shape.

**3. `web/js/lib/set-widget.js` — "the schema … was not fetched live".**

Reachable now with a type-scoped map, which **was** fetched live — about fewer types, but not
staler ones. The wording predates #1560 and was accurate for every provenance that existed then.

---

Suggested together in one small change, since all three are one-liners in two files, plus a test
that asserts the OUTPUT rather than the absence of a phrase — #1561 shipped a broken `sanitizeName`
green precisely because its test only checked that a newline was gone.

artokun:
Closes #1573.

Three inaccuracies #1561's own fourth adversarial gate pass filed against its type-scoped route and deliberately left unfixed there — that PR was at its third round of fixes, and moving the head again would have meant merging a SHA the gate had not seen. #1561 has since merged as `a1844f68`.

**All three are in the refusal direction. This changes what the code says about itself and nothing about what it does.** Review is **pending** — requested from grok in a comment below; not gated, no verdict claimed.

## Re-verified against merged main before touching anything

They were filed pre-merge. Each was reproduced by execution against `origin/main` (`56fd8eb0`, which contains `a1844f68`) — **all three still held**, none was skipped.

| # | Reproduced on merged main |
|---|---|
| 1 | `JSON.stringify(defs)` on a two-type scoped map **threw**: `Cannot verify node type "toJSON"…`. `String(defs)` threw on `toString`. |
| 2 | With the panel's own unlicensed reason, the refusal said a read `was tried too` while `perClassCalls()` was `[]` — **not one request issued**. |
| 3 | Scoped map declares the list empty, whole-map re-ask publishes a real list, refusal asserts the schema `was not fetched live` while `/object_info/StarOllamaPromptHelper` sat in the issued-request list. |

## What each now says, and why that is accurate

**1. `scoped-object-info.js` — the `ownKeys` note (comment only).**

It claimed anything ranging over the map — "`Object.keys(defs).length`, a serializer" — sees a small honest object rather than a throw. Enumeration does. A serializer never did: `JSON.stringify` looks up `toJSON` and `String` looks up `toString`, neither is a class type in `covered`, so the `get` trap refuses both.

The note now **records the measurement instead of the premise** — what was measured, on which build, and what stays a trade: node v24.16.0, two-type scoped map, `Object.keys` gives 2 names, `getOwnPropertyNames` / `entries` / `for…in` / spread all succeed, `Reflect.ownKeys` gives 3 (types + brand); `JSON.stringify` and `String()` throw. It says why this is a documented edge and not a fault — every consumer reads the map **by key** (`freshBackendDefinesType`, `assertTypeAgainstFreshBackend`, `serverDeclaresEmptyComboOptions`, `uploadInputConfig`, `refreshCombos`, `isTypeScopedObjectInfo`), a repo-wide search finds no serialization of a defs map in either repo, and the direction is fail-closed regardless.

The issue asked whether `toJSON` / `toString` / `Symbol.toPrimitive` should simply be **in scope**, which would make the original comment true instead of correcting it downward. That is a **behaviour change**, so it is deliberately **not decided here** and the note says so.

**2. `set-widget.js` — "A type-scoped read was tried too".**

Appended for *every* non-empty `reason`, including the ones that issue nothing (unlicensed, no type resolved, over the type cap, no `fetchApi`, no time left). The clause after the dash self-corrects; the lead-in does not, and it is the sentence a stuck caller reads. #982's shape.

This site knows only that the route produced **no usable map**. It does not know whether a request left the panel. So the lead-in now reports the outcome and lets the reason report the cause:

> A type-scoped /object_info read **did not stand in for the whole-schema routes either** — {reason}.

True on every reason, and the ones that *did* issue requests still say so, in the reason, by naming the deadline they missed.

**3. `set-widget.js` — "the schema … was not fetched live".**

Reachable and false since #1560. The re-ask gate is `provenance !== "live"`, and `"scoped"` is not `"live"` — but a type-scoped map **is** fetched live, per class, moments earlier. It is about **fewer types, never staler ones**. What the branch actually establishes is that the schema which said empty has been replaced and the replacement no longer says it, so that is all it claims now:

> The schema that appeared to declare this input's option list empty **has since been REPLACED**, and the live re-read NO LONGER declares it empty — either it now publishes a real option list for this input, or it no longer describes this input at all.

`staleDeclaredEmpty` renamed to `priorDeclaredEmpty`: the false premise lived in the identifier that fed the message, and an identifier is prose every reader sees.

## No predicate changed

- `if (!scopedIneligibility) return base;` — untouched.
- `serverDeclaresEmptyComboOptions(authDefs, authTarget?.type, comboDefInput)` — **byte-identical expression**; only the const name moved.
- `if (reAsked && priorDeclaredEmpty)` — same two operands, renamed.
- The proxy traps in `scoped-object-info.js` are **not touched at all** — that file's diff is comment-only.

## Consumers of the changed strings

`git grep` of each literal across **both** repos, tests included:

- `"was tried too"` — **one** consumer: `browser_tests/unit/set-widget-scoped-object-info.test.mjs:263`. Updated **deliberately, not loosened**: it now asserts the new sentence *and* that `perClassCalls()` equals `["/object_info/SmartResolution"]`, so the wording and the world are checked against each other rather than the wording alone. Its title asserted the old claim too and was corrected.
- `"was not fetched live"` — **zero** consumers.
- `"a small, honest object"` — **zero** consumers.
- `staleDeclaredEmpty` — two sites, both in `set-widget.js`.
- Nothing in `comfyui-mcp` references any of them.

## Verification

`npm run test:unit` — **6169 tests, 6168 pass, 0 fail, 1 pre-existing todo**. `check-panel-scope` **OK** (needed `npm install` in the fresh worktree first — it correctly refused to run without tsc rather than printing OK). `npm run typecheck` clean. `node --check` on both touched files.

Deleting a comment fails no test, so the bar was applied where it bites — three mutations, each killing a **named** test:

| Mutation | Result |
|---|---|
| Restore `A type-scoped /object_info read was tried too` | FAILS 2 tests — the unlicensed test **and** the timeout test |
| Restore `empty was not fetched live` | FAILS `#1573: a LIVE type-scoped map that declared the list empty is never reported as 'not fetched live'` |
| Put `toJSON` / `toString` **in scope** (the open behaviour question) | FAILS `#1573: ENUMERATION of the scoped map does not throw — a SERIALIZER does` |

The third is the point of that test: it is coupled to the header by measurement, so whoever decides to widen the scope fails an assertion right beside the paragraph that says the question is open.

Tests assert the **OUTPUT** — the exact refusal sentence paired with the issued-request list — which is what the issue asked for, since #1561 shipped a broken `sanitizeName` green precisely because its test only checked that a newline was gone.

## Deliberately not done

- **No Playwright run.** Not that it passed — it was **not run**. These strings are `panel_set_widget` refusal text returned over the bridge plus two source comments; none of them renders in the sidebar, and the browser suite asserts none of them. Claiming e2e coverage here would be claiming coverage I do not have.
- **`toJSON` / `toString` / `Symbol.toPrimitive` left out of scope.** Behaviour change, its own PR, its own evidence.
- **`"the live re-read"` left as-is** in item 3's message. A re-read can be `reconnected` or `retired` (issued live, superseded in flight), so it is arguably a mild over-claim — but the read *is* issued live, no measurement shows a defect, and changing it would be a wording change with nothing behind it. Named here rather than quietly fixed.
- **#1223's over-specific claim** one sentence earlier in the same refusal — pre-existing, called out by #1561, still untouched.
- **No predicate, guard or trap adjusted** anywhere.

## The standard applied to this PR's own diff

Worth stating plainly, because it is the point of #1573 rather than a footnote: **the first version of this change committed the exact defect it exists to close.** My new `ownKeys` note claimed "a repo-wide search finds no `JSON.stringify` or string coercion of a defs map in **either repo**" — and I had only run that search on the panel repo. I grepped `comfyui-mcp` for the three changed *literals*, never for a serializer. So a comment I wrote to replace an unverified premise asserted a check I had not run.

Caught it on an adversarial re-read of my own diff, fixed it before any review started, and disclosed the head move rather than folding it in quietly. The note now states the scope it actually has — a search of `web/js` — and the structural reason that is the whole scope there is: this map is built in the panel and consumed in the panel, it never crosses the bridge, so no orchestrator reader can reach it to serialize.

That is the failure mode #1561's gate was pointing at. It is not rare, it does not announce itself, and a comment is where it survives longest.

### DIFF
--- a/browser_tests/unit/set-widget-scoped-object-info.test.mjs
+++ b/browser_tests/unit/set-widget-scoped-object-info.test.mjs
@@ -86,10 +86,15 @@ function largeInstall({ defined = [], perClass } = {}) {
  * The panel's own wiring for the scoped route, reproduced verbatim in shape: the SILENCE
  * LICENCE first, then the bounded type-scoped read, then the "scoped" provenance stamp.
  */
-function panelStyleScopedRoute(fetchApi, outcomes, onScoped) {
+function panelStyleScopedRoute(
+  fetchApi,
+  outcomes,
+  onScoped,
+  unlicensedReason = "a whole-schema route ANSWERED rather than going silent",
+) {
   return async (types) => {
     if (!noBackendAnswerEstablished(outcomes)) {
-      return { defs: null, covered: [], reason: "a whole-schema route ANSWERED rather than going silent" };
+      return { defs: null, covered: [], reason: unlicensedReason };
     }
     const scoped = await fetchTypeScopedObjectInfo(types, { fetchApi, deadlineMs: SCOPED_OBJECT_INFO_DEADLINE_MS });
     if (scoped.defs && typeof onScoped === "function") onScoped(scoped);
@@ -241,7 +246,7 @@ test("#1560 B: an EVER-SEEN type now absent per class is still diagnosed as a RE
   assert.equal(node.widgets[0].value, 20, "no mutation");
 });
 
-test("#1560 B: the per-class route ALSO going silent refuses exactly as today, and SAYS a third route was tried", async () => {
+test("#1560 B: the per-class route ALSO going silent refuses exactly as today, and SAYS what the third route did", async () => {
   const reg = loadedRegistry(["SmartResolution"]);
   const node = regNode(1976, "SmartResolution", [{ name: "value", type: "INT", value: 512 }]);
   const backend = largeInstall({ perClass: () => new Promise(() => {}) }); // never settles
@@ -260,8 +265,67 @@ test("#1560 B: the per-class route ALSO going silent refuses exactly as today, a
     (err) =>
       err instanceof Error &&
       /no usable \/object_info schema was obtained/.test(err.message) &&
-      /A type-scoped \/object_info read was tried too — .*did not all answer within/.test(err.message),
+      // #1573 — the lead-in reports the OUTCOME; the reason reports the cause. On THIS path
+      // requests really were issued, and the reason says so by naming the deadline they
+      // missed. The assertion below pins that the requests happened, so the wording and the
+      // world are checked against each other rather than the wording alone.
+      /A type-scoped \/object_info read did not stand in for the whole-schema routes either — .*did not all answer within/.test(
+        err.message,
+      ),
   );
+  assert.deepEqual(
+    backend.perClassCalls(),
+    ["/object_info/SmartResolution"],
+    "the request the reason blames a deadline for really was issued",
+  );
+  assert.equal(node.widgets[0].value, 512, "no mutation");
+});
+
+test("#1573: an UNLICENSED scoped route issues NOTHING, and the refusal does not claim a read was tried", async () => {
+  // #1561's own fourth gate pass: `describeObjectInfoFailureWithScope` appended "A
+  // type-scoped /object_info read was tried too" for EVERY non-empty reason, including the
+  // ones where the route was never licensed and no request left the panel. The clause after
+  // the dash self-corrected, but the lead-in asserted an attempt that did not happen — the
+  // #982 shape, in the one sentence a stuck caller reads.
+  //
+  // Asserts the OUTPUT, not the absence of a phrase: the exact sentence the caller gets,
+  // paired with the request list that proves what actually happened.
+  const src = readFileSync(PANEL_JS, "utf8");
+  const UNLICENSED_REASON =
+    "no whole-schema route was both CONTACTED and SILENT, and a type-scoped read " +
+    "may only stand in for one that was — it must never overrule what a route did " +
+    "establish, and it is not a substitute for a route nobody ran";
+  assert.ok(
+    src.includes("no whole-schema route was both CONTACTED and SILENT"),
+    "the reason under test is the panel's own, not one invented here",
+  );
+
+  const reg = loadedRegistry(["SmartResolution"]);
+  const node = regNode(1976, "SmartResolution", [{ name: "value", type: "INT", value: 512 }]);
+  const backend = largeInstall({ defined: ["SmartResolution"] });
+  await assert.rejects(
+    () =>
+      runSetWidget(node, "value", 768, {
+        registry: reg,
+        getRegistry: () => reg,
+        getFreshObjectInfo: async () => null,
+        describeObjectInfoFailure: () => " Tried 2 routes: a; b.",
+        // The panel's unlicensed early return, verbatim: no request is issued at all.
+        fetchScopedObjectInfo: panelStyleScopedRoute(backend.fetchApi, ANSWERED_OUTCOMES, undefined, UNLICENSED_REASON),
+        wasTypeEverDefined: () => false,
+        ...HOOKS,
+      }),
+    (err) => {
+      assert.ok(
+        err.message.includes(
+          `A type-scoped /object_info read did not stand in for the whole-schema routes either — ${UNLICENSED_REASON}.`,
+        ),
+        `the refusal must state the outcome and the panel's own reason, verbatim. Got: ${err.message}`,
+      );
+      return true;
+    },
+  );
+  assert.deepEqual(backend.perClassCalls(), [], "not one per-class request was issued — so none may be claimed");
   assert.equal(node.widgets[0].value, 512, "no mutation");
 });
 
@@ -534,6 +598,45 @@ test("#1560: the scoped map is a well-behaved object — branding, freezing and 
   assert.doesNotThrow(() => Object.freeze(defs), "object-info-cache.js freezes its payload; this must survive the same");
   assert.deepEqual(Object.keys(defs), ["KSampler"]);
   assert.equal(defs[Symbol.iterator], undefined, "an unrelated symbol is not a class type and passes through");
+});
+
+test("#1573: ENUMERATION of the scoped map does not throw — a SERIALIZER does, and the header now says so", async () => {
+  // #1561's header claimed that "anything that ranges over the map (`Object.keys(defs).length`,
+  // a serializer) sees a small, honest object rather than a throw". The first half is true;
+  // the serializer half never was. `JSON.stringify` looks up `toJSON` and `String` looks up
+  // `toString` — neither is a class type in `covered`, so the scope trap refuses both.
+  //
+  // This pins the MEASUREMENT the header now records, in both directions, so the header
+  // cannot drift back into a claim nobody re-checked. It is not a change of behaviour: the
+  // traps are untouched and this test passes against the merged head too.
+  const backend = largeInstall({ defined: ["KSampler", "SubgraphA"] });
+  const { defs } = await fetchTypeScopedObjectInfo(["KSampler", "SubgraphA"], { fetchApi: backend.fetchApi });
+
+  assert.deepEqual(Object.keys(defs), ["KSampler", "SubgraphA"]);
+  assert.equal(Object.keys(defs).length, 2, "the example the header names by name");
+  assert.deepEqual(Object.getOwnPropertyNames(defs), ["KSampler", "SubgraphA"]);
+  assert.equal(Object.entries(defs).length, 2);
+  assert.deepEqual(Object.keys({ ...defs }), ["KSampler", "SubgraphA"], "spread copies, it does not refuse");
+  const forIn = [];
+  for (const k in defs) forIn.push(k);
+  assert.deepEqual(forIn, ["KSampler", "SubgraphA"]);
+  assert.equal(Reflect.ownKeys(defs).length, 3, "the two types plus the brand symbol");
+
+  // The other direction, stated rather than glossed. Both are FAIL-CLOSED — a throw refuses
+  // a write, it can never forge one — and no production reader serializes this map.
+  assert.throws(
+    () => JSON.stringify(defs),
+    (err) => err instanceof Error && /Cannot verify node type "toJSON"/.test(err.message),
+    "JSON.stringify looks up toJSON, which is not a covered class type",
+  );
+  assert.throws(
+    () => String(defs),
+    (err) => err instanceof Error && /Cannot verify node type "toString"/.test(err.message),
+    "and a string coercion looks up toString",
+  );
+  // Deliberately coupled to the header rather than to a phrase in it: whoever decides that
+  // `toJSON`/`toString` belong IN SCOPE will fail these two assertions, and the paragraph
+  // that says the question is still open is right above the trap they will be editing.
 });
 
 test("#1560: a hostile class name is FLATTENED before it reaches a refusal a caller reads", async () => {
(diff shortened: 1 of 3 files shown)


artokun:
@grok — review requested. This is P3 and comment/message-only, so rather than hand you a diff, here is exactly where I think it can be wrong.

**What changed, in one line each.** (1) `scoped-object-info.js`'s `ownKeys` note retracts a false claim that a serializer works on the scoped map, and records the measurement instead. (2) `set-widget.js`'s scoped-route lead-in stops saying `A type-scoped /object_info read was tried too` and says `did not stand in for the whole-schema routes either`. (3) `set-widget.js`'s re-ask refusal stops saying the empty-declaring schema `was not fetched live` and says it `has since been REPLACED`; `staleDeclaredEmpty` renamed `priorDeclaredEmpty`.

**The load-bearing claim, and the one I am least sure of, is #3's reachability.** I assert a type-scoped map can reach `if (reAsked && priorDeclaredEmpty)` — which is what makes `was not fetched live` false rather than merely awkward. My argument: the re-ask gate is `provenance !== "live"` and `"scoped"` is not `"live"`, so a scoped map enters; `refetchObjectInfoLive` returns a WHOLE map so `provenanceOf` no longer answers `"scoped"`; the `provenance === "scoped"` refusal above is therefore skipped; and if the whole map publishes a real list the empty-shape test is false and the branch fires. I did not stop at reading that — I drove the real `runSetWidget` and got the old refusal out with `/object_info/StarOllamaPromptHelper` in the issued-request list. **If there is a guard that makes this unreachable in production and my fixture only simulates the path, the whole of #3 is a fix to dead prose and I would rather know.**

**Second, the rename.** `staleDeclaredEmpty` to `priorDeclaredEmpty` is the only edit that touches a line with a predicate on it. I claim the expression is byte-identical and only the binding's name moved — `web/js/lib/set-widget.js` around L1273 and L1470. That is exactly the kind of "obviously fine" edit an author reads past, so please diff those two lines specifically rather than trusting the claim.

**Third, the new wording of #2 is a broadening, and broadenings can under-claim.** `did not stand in for the whole-schema routes either` is true for all six reasons `fetchTypeScopedObjectInfo` can return, including the four that issue no request. But it drops the information that on the timeout/failure paths a request genuinely WAS issued. I decided the reason string carries that (it names the deadline it missed) and pinned it with an assertion on `perClassCalls()`. If you think a caller loses something actionable there, say so — the alternative is a structured "was a request issued" field on the return shape, which is a behaviour change I deliberately did not make.

**Fourth, what I chose NOT to fix, in case you disagree with the line I drew.** The issue offered a second resolution for #1: put `toJSON`/`toString`/`Symbol.toPrimitive` IN SCOPE, which would make the original comment true instead of correcting it downward. I refused that as a behaviour change and left the question explicitly open in the header, with a test that fails if someone widens the scope. I also left `"the live re-read"` alone in #3's message even though a re-read can be `reconnected`/`retired` — issued live, superseded in flight — because no measurement showed a defect there and I did not want a wording change with nothing behind it.

**Verification, so you know what is and is not covered.** `npm run test:unit` 6169 tests / 6168 pass / 0 fail; `check-panel-scope` OK; `typecheck` clean. Three mutations each kill a named test (restore either old message; put `toJSON`/`toString` in scope). **Playwright was NOT run** — none of these strings renders in the sidebar and the browser suite asserts none of them, so I am claiming no e2e coverage rather than implying it.

No predicate changed and no trap was touched; if you find one that did, that is a NO-SHIP and I will take it.

artokun:
Head moved to `123e03d6` **before** any review started (only the Copilot quota-limit entry existed, which is not a review). One change, and it is the same defect class this PR is about: the `ownKeys` note said "a repo-wide search finds no `JSON.stringify` or string coercion of a defs map in either repo", and I had only searched the panel repo for that — I grepped comfyui-mcp for the three changed literals, not for a serializer. So the comment claimed a check I had not run.

It now states the scope it actually has: a search of `web/js` finds no `JSON.stringify` of, and no string coercion of, an `/object_info` defs map — and the reason that is the whole scope, which is that this map is built and consumed inside the panel and never crosses the bridge, so no orchestrator reader can reach it.

Comment-only, no code touched. Re-ran `npm run test:unit`: `check-panel-scope` OK, and the one red was `#671 verifyInstalled … fast-draining install` at **25,476 ms** — the documented manager-install wall-clock flake, fired under the load of a concurrent suite run. Re-ran that file alone: **125/125 pass, the same test at 1,018 ms**. Nothing to do with this diff, which touches neither file.

Review request above stands unchanged and applies to this head.

artokun:
## HOLD — not merging. Review requested and outstanding.

Requested grok's review at **16:07Z**, re-pointed it at head `123e03d6` at **16:20Z**. Polled `repos/artokun/comfyui-mcp-panel/pulls/1574/reviews` (the **reviews** API, not `.comments[]`) continuously through **16:54Z**. Nothing.

The only entry on the reviews endpoint is `copilot-pull-request-reviewer[bot]` saying it could not review because the requester reached their quota limit. **That is not a review**, and I am not treating it as one. The two comments on the PR are both mine.

grok's most recent review anywhere on this repo is **02:51Z today** (PR #1552) — roughly fourteen hours ago. It looks asleep, not slow.

**So this stops here rather than merging on my own verification.** The change is small, comment/message-only and in the refusal direction, and my own evidence is as strong as I know how to make it — three mutations each killing a named test, all three inaccuracies reproduced by execution against merged main first. None of that is a review. An author reading their own guard finds it obviously right; that is the exact failure mode this repo has shipped before.

**State: open, CI green, no conflicts.**

- Head `123e03d6`, CI `Lint pack` **pass** in 2m31s.
- Note `main` has since moved to `869362ec`. GitHub still reports `MERGEABLE` / `CLEAN`, but that green check ran against `56fd8eb0` — **re-run CI against current main before merging**, per the usual rule that a green check is only green for the main it ran against.

**What unblocks it:** a verdict from grok, or another independent reviewer, on the three questions named in the review request above — the reachability argument for item 3 (the load-bearing one), the `staleDeclaredEmpty` → `priorDeclaredEmpty` rename being name-only on two predicate lines, and whether broadening item 2's lead-in loses anything a caller could act on.

Not applying `autopilot:parked` — this is not parked. It is finished work waiting on the review it asked for.

artokun:
## Gate: SHIP (exit 0) on `245c6289`

Reviewer switched from grok to codex — grok is out of usage, which is why the earlier poll found nothing.

```
node .claude/autopilot/codex-gate.mjs --worktree …/wt-1573-cc --base origin/main --name 1574
SHIP
```

**Head parity was confirmed before the gate started and still holds:** worktree HEAD, PR head and the gated SHA are all `245c6289`. The head has not moved since the verdict and will not before the merge.

Exit code derived from the gate's own logic rather than assumed — the run was detached so the code was not captured directly. Reconstructed its two downgrade conditions: source diff 144 lines against a 1000-line cap (**not truncated**), and zero sensitive-path hits. Neither `supplyChainHold` arm fires, so the `SHIP` first line exits **0**. Worktree clean afterwards (`-s read-only`).

**What the gate did NOT see, stated plainly.** It scopes the diff to source and excludes test files by its own rule, so it reviewed `web/js/lib/scoped-object-info.js` and `web/js/lib/set-widget.js` — and **not** `browser_tests/unit/set-widget-scoped-object-info.test.mjs`. That file carries the deliberate update to the assertion that pinned the old "was tried too" wording, which is exactly the place a loosened assertion would hide. The gate is not evidence about it; the three mutations are, each killing a named test.

**On the two things it did not flag.** A bare one-line SHIP is the specified output shape here, not an empty review — the prompt says "If nothing material is wrong, say so in one line" and restricts findings to P0/P1 with a concrete failing input. `"the live re-read"` and the `toJSON`/`toString` scope question are neither: one is a wording nuance with no measurement behind it, the other a deliberate behaviour change deferred to its own PR. I am not changing either on the strength of a reviewer's silence, and I would not have changed them on the strength of a mention either.

**Re-verified against current main.** `main` moved twice during this PR (`56fd8eb0` → `869362ec` → `8e321b5e`). Rebasing again would have moved the head off the SHA the gate just approved, so instead I trial-merged `245c6289` into `8e321b5e` in a scratch worktree and ran the full suite there: **6200 tests, 6198 pass**, `check-panel-scope` OK, `check-tool-vocabulary` OK. The one red was `#671 verifyInstalled … fast-draining install` at 7,047 ms — the documented manager-install wall-clock flake, in a file neither this change nor either new main commit touches. Alone: **125/125, that test at 1,009 ms**. My own file on the merge result: **31/31**.

Merging at `245c6289`.

artokun:
## Post-merge independent verification — all three items CONFIRMED pinned

Picked this up as a resumed autopilot run, found it already merged (`2cf4d374`, head `245c6289`)
and #1573 closed COMPLETED. The merge went through on a codex-gate SHIP after grok went silent.
Recording an independent check here, because a verdict from that gate is not one this brief
accepts as a review, and the evidence would otherwise exist nowhere.

I changed nothing. This is verification of what shipped, run against `origin/main`.

**Method: mutation, not a green suite.** A passing suite says nothing about whether these tests
pin the three fixes, which matters more than usual here — all three are strings, and #1561
shipped a broken `sanitizeName` green because its test only checked a phrase was absent.

| # | Mutation applied to merged main | Result |
|---|---|---|
| 2 | refusal reverted to `A type-scoped /object_info read was tried too` | **3 tests RED** |
| 3 | refusal reverted to `empty was not fetched live` | (same run) |
| 1 | `inScope` widened so `toJSON`/`toString` pass — i.e. a serializer *succeeds* | **1 test RED** |

Item 1 is a pure comment, so no edit to it can fail a test. What its test pins is the *behaviour
the comment asserts*: widen the trap so a serializer works and
`#1573: ENUMERATION … does not throw — a SERIALIZER does` fails. The comment cannot silently
drift away from the code it describes, which is the property that was actually wanted.

Failing tests under mutations 2+3:
- `#1573: an UNLICENSED scoped route issues NOTHING, and the refusal does not claim a read was tried`
- `#1573: a LIVE type-scoped map that declared the list empty is never reported as 'not fetched live'`
- `#1560 B: the per-class route ALSO going silent refuses exactly as today, and SAYS what the third route did`

Every file restored after each mutation and the worktree confirmed clean (`git diff --stat` empty)
before the next run.

**Separately checked, because a rename is the one way this PR could have shipped a real fault.**
`staleDeclaredEmpty` → `priorDeclaredEmpty` is a `const` in a block; one missed reference is a
runtime `ReferenceError` — a P1 inside a P3 comment PR. On `origin/main` the only surviving
`staleDeclaredEmpty` is inside the comment that documents the rename; both live references are
renamed. The "no predicate changed" claim also holds on inspection: `if (!scopedIneligibility)`
untouched, the `serverDeclaresEmptyComboOptions(...)` expression byte-identical with only its
const name moved, `if (reAsked && priorDeclaredEmpty)` with the same operands.

**Suite state on `origin/main`:** 6248 tests, 6247 pass, 0 fail, 1 todo. CI success on both
`245c6289` and the merge commit.

**What I did NOT do:** I did not run the Playwright suite. These are refusal strings returned to
the caller rather than sidebar-rendered UI, and the change is already merged — so this is a
stated gap, not implied coverage. The `toJSON`/`toString`/`Symbol.toPrimitive` scope question the
header leaves open is still open; it is a behaviour change and out of scope for #1573.

PR #1576, opened 2026-08-21
fix(1998): a scoped batch drives ComfyUI's own control hooks, so batch_count means N renders

artokun:
Fixes the panel side of **artokun/comfyui-mcp#1998** — `panel_run batch_count>1` queues identical seed snapshots so the extras cache-hit the first output.

**Review is PENDING.** Nothing here has been reviewed by anyone but its author.

---

## What the user saw

`panel_run` with `batch_count: 4` came back with `queued: true`, four `prompt_ids`, and four completions. One image was rendered. The other three were served from ComfyUI's cache and reported the *first* run's output file. The tool reported four successes for one render.

## Reproduced, by execution, on a real ComfyUI

ComfyUI 0.33.2 / frontend 1.49.6, real panel over the suite's MockBridge, real `graph_run`, real renders. `/history` for the four minted prompt ids, **with the fix removed**:

```
prompt   seed    partial_execution_targets  execution_cached                 ms     output
bc554fa3 707000  ["9"]                      []                               5584   b1998before_00001_.png
dcb1248d 707000  ["9"]                      ["3","4","5","6","7","8","9"]       2   b1998before_00001_.png
bf54dc30 707000  ["9"]                      ["3","4","5","6","7","8","9"]       2   b1998before_00001_.png
7f1f7bc6 707000  ["9"]                      ["3","4","5","6","7","8","9"]       3   b1998before_00001_.png
```

**The cache-hit claim is settled by execution, not by timing.** ComfyUI's own `execution_cached` message names every node *including the SaveImage output node 9* for prompts 2–4, and all four `/history` entries carry the same `filename`. On disk: one `b1998before_*.png`, not four. Fast and cached are different bugs; this is the cached one.

Same run **with the fix**:

```
16012abe 707000            ["9"]  ["3","4","5","6","7","8"]   12   a1998after_00001_.png
f2460058 300614720541829   ["9"]  ["4","5","6","7"]          576   a1998after_00002_.png
486d0a11 372489251625065   ["9"]  ["4","5","6","7"]          510   a1998after_00004_.png
1d54a33c 304038503665136   ["9"]  ["4","5","6","7"]          456   a1998after_00003_.png
```

Four seeds, four renders, four files, and the scope is still on every post. (Item 1's KSampler cache hit is the previous run at the same seed; it still writes its own file.)

## What the frontend actually does per queue click, and how that was established

Not inferred from the widget name. `comfyui_frontend_package` ships sourcemaps **with `sourcesContent`**, so the pinned build's original TypeScript is recoverable. From 1.49.6:

`src/scripts/app.ts` — the batch loop lives INSIDE one `queuePrompt` call and fires the hooks per item:

```ts
const isPartialExecution = !!queueNodeIds?.length
for (let i = 0; i < batchCount; i++) {
  forEachNode(this.rootGraph, (node) => {
    for (const widget of node.widgets ?? []) widget.beforeQueued?.({ isPartialExecution })
    applyPromotedWidgetControl(node, 'beforeQueued', { isPartialExecution })
  })
  const p = await this.graphToPrompt(this.rootGraph)      // the body that is POSTed
  const res = await api.queuePrompt(number, p, { partialExecutionTargets: queueNodeIds })
  executeWidgetsCallback(queuedNodes, 'afterQueued', { isPartialExecution })
}
```

`src/scripts/valueControl.ts` — and it refuses to advance anything on a partial execution:

```ts
export function nextValueForLinkedTarget(params) {
  if (params.isPartialExecution) return undefined
```

That refusal is **deliberate upstream behaviour**: ComfyUI_frontend PR **#8774**, *"disable control after generate during partial execution"*, merged 2026-02-11, so that queueing selected output nodes while iterating does not churn the user's seed.

Then measured on the rig, capturing the outgoing `/prompt` bodies behind an interceptor that answered them locally (nothing queued, no GPU):

```
unscoped  queuePrompt(0, 4, undefined)   randomize -> 4 distinct seeds
unscoped  queuePrompt(0, 4, undefined)   increment -> 12345, 12346, 12347, 12348
unscoped  queuePrompt(0, 4, undefined)   fixed     -> 12345 x4
scoped    queuePrompt(0, 4, ["9"])       randomize -> 12345 x4
scoped    queuePrompt(0, 4, ["9"])       increment -> 12345 x4
scoped    queuePrompt(0, 4, ["9"])       fixed     -> 12345 x4
scoped    4x queuePrompt(0, 1, ["9"])    randomize -> 12345 x4     <-- decides the fix
```

Two things follow.

1. **The UNSCOPED path was already correct** and is not touched. `graph_run` hands an unscoped batch to `app.queuePrompt(0, batch, undefined)` — ComfyUI's own loop, hooks and all. The reporter's steps do not mention `to_node_id`; measured on a current frontend, an unscoped `batch_count: 4` with `randomize` already posts four different seeds. The path that reproduces their exact symptom — N prompt ids, ~0 s each, same output file — is the **scoped** one, which is #988's mechanism, warned about since panel 0.11.94 and never repaired. Their reported panel version (0.11.57) predates even the warning. Called out plainly because it is the one thing in this PR I could not confirm from the report itself.

2. **"Press Queue N times" does not work here.** Four separate `queuePrompt(0, 1, scope)` calls post the same seed four times: the refusal keys on the *scope*, not on the batch position. That is also why the #988 reporter's own loop-with-`batch:1` patch failed after 114 green tests.

## The fix — Option A, and only the flag

For a **scoped** run with **`batch > 1` only**, each control widget's own `beforeQueued`/`afterQueued` is wrapped for exactly the duration of the dispatch so it receives `isPartialExecution: false`. ComfyUI then computes everything:

- no seed arithmetic and no mode table in the panel;
- `randomize` / `increment` / `decrement` / `increment-wrap`, `min`/`max`/`step2`, and a combo's filter list all behave exactly as they do on the Queue button, because they *are* the Queue button's code;
- **`fixed` still repeats**, and it is ComfyUI's own `if (mode === 'fixed') return undefined` that decides it — not a panel branch. `driveControlHooksAcrossScopedBatch` deliberately arms fixed controls too, and a unit test asserts `drive.armed.length === 1` for a `fixed` control so a future "optimisation" that skips them locally is caught. Verified by execution: `fixed`, `batch_count: 4`, scoped → one seed, one file, four cache hits. Identical jobs stay identical.

Detection is by **option shape** (`serialize:false` + `canvasOnly:true` + the mode option list), not by widget name, so a node def that renames its control is still driven; and both hooks must exist, which is the frontend's own `isValueControlWidget` test and keeps this away from third-party `beforeQueued` hooks that are not controls (they still receive the truthful flag).

A scoped run of **one** is untouched: #8774 keeps owning the case it was written for.

## What it deliberately does NOT do

- **Promoted subgraph widgets are not reached.** `app.queuePrompt` advances those through `applyPromotedWidgetControl(node, phase, { isPartialExecution })` — a module function, not a widget property, so there is nothing to wrap. They still repeat.
- **Link-fed governed inputs are not reached** — the frontend skips the control on every path when the value comes from the link.
- Neither is guessed at. The drive **observes** each governed widget on both sides of every hook call, and the result note names every advancing control that did **not** move, with the two usual causes. `observe()` reports what the widget did, not what the drive asked for; a mutation that hard-codes `advanced: true` kills two named tests.
- **#988's note is suppressed on the driven path.** "THIS BATCH WILL REUSE THE SAME VALUE" is the sentence this change stops being true; shipping both would put two contradictory statements in one result. It stays in charge whenever the drive did not run.
- **No change to what is POSTed structurally.** `partial_execution_targets` is present on all four posts before and after (asserted in the e2e), so this cannot buy distinct seeds by quietly dropping the scope and running the full graph (#556). And the #556 drift hash already excludes a non-`fixed` control's linked target, so a changing seed is not read as graph drift — confirmed live: the driven run returned `drift_coverage.uncovered_inputs: ["3 control_after_generate","3 seed"]` and dispatched normally.
- **No orchestrator change.** `panel_run` forwards the panel's result object verbatim, so `batch_controls` / `batch_controls_note` reach the caller with no MCP-side work.

## Known windows, stated rather than hidden

- The wrap lives for the dispatch only, restored in a `finally` on every exit including a throw. If `app.queuePrompt` **defers** (its processor is busy) and posts after `dispatchScopedRun` gives up, the deferred items run with the original hooks and repeat — degrading to today's behaviour, and `observe()` then reports `advanced: false`, so the note tells the truth rather than claiming a fix that did not happen.
- During that window a *user-initiated* partial-execution queue in the same tab would see the flag cleared too. Panel commands are dispatched serially and the window is sub-second, but it is a real window and it is not zero.

## Evidence

**Live, executed** — the four `/history` tables above, plus `fixed` (1 seed / 1 file) and `increment` (900100 → 900104, 4 files) both by execution, and the on-disk file counts (`b1998before` × 1 vs `a1998after` × 4).

**Unit** — 51 tests in `browser_tests/unit/scoped-batch-seed.test.mjs`, driven against a *port* of frontend 1.49.6's own hook (every branch quoted from the recovered TypeScript, and cross-checked against the live numbers above).

**e2e** — `browser_tests/scoped-batch-control.spec.ts` drives the REAL `graph_run` through the REAL bridge frame in a REAL browser and reads the seeds off the POST bodies. Nothing queued: `api.fetchApi` is wrapped before the run so the panel's own scoped guard chains on top of it.

**Mutation — the fix deleted, and named tests fail.** Full suite green at 51/51 first, and restored to 51/51 after:

| mutation | pass/fail | killed by |
|---|---|---|
| M1 drop the `isPartialExecution: false` override | 43/8 | `DRIVEN, a scoped randomize batch posts four different seeds`, `…increment batch matches the unscoped sequence exactly`, `decrement runs backwards`, `works in Comfy.WidgetControlMode 'before'`, `restore() puts #8774 back`, `detected by OPTION SHAPE … renamed`, `observe() reports what the widget DID`, `the note names what advanced` |
| M2 CALL SITE: never install the drive | 50/1 | `CALL SITE: the drive is installed on the SCOPED path, and restored in a finally` |
| M3 the panel special-cases `fixed` itself | 50/1 | `` `fixed` still repeats - and it is ComfyUI that refuses, not a panel branch `` |
| M4 CALL SITE: restore not in a `finally` | 50/1 | `CALL SITE: the drive is installed on the SCOPED path, and restored in a finally` |
| M5 CALL SITE: `batch > 1` gate dropped | 50/1 | `CALL SITE: the drive is installed on the SCOPED path, and restored in a finally` |
| M6 CALL SITE: ship #988's contradictory warning too | 50/1 | `CALL SITE: a driven batch does not also ship #988's contradictory warning` |
| M7 `observe()` claims success instead of measuring | 49/2 | `observe() reports what the widget DID`, `the note names what advanced` |

M5 **survived the first round**: the gate matched the phrase `batch > 1` in the *comment above* the call, so a dropped gate read as present. It now asserts on the argument expression. That is in this PR as a commit of its own reasoning.

**Call site, at e2e level too** — with the install removed, `scoped-batch-control.spec.ts` reports `expected four distinct seeds, got [707000,707000,707000,707000]`.

**Gates** — `npm run test:unit` (6183 pass / 0 fail, includes i18n, tool-vocabulary and the panel-scope compiler gate), `npm run typecheck` clean, and the full Playwright suite at `--workers=1` — 26 failed / 58 passed, which is this suite's known isolated state on this box; 18 of the 26 are the 8-second `openSidebar` mount budget losing to machine load, including this PR's own two, which fail BEFORE any line this PR touches and pass 3/3 in isolation. Full triage in the comments.

---

- **I did not touch the unscoped path.** It is already correct: `graph_run` hands an unscoped batch to `app.queuePrompt(0, batch, undefined)` — ComfyUI's own loop — and randomize/increment both advance there (measured above). I drafted an observation-only watcher for it, so that an unscoped batch whose controls fail to move would say so instead of returning N silent successes, and then removed it: the only ways a control stays put on that path are a **link-fed** governed input (the user wired a constant in, so N identical renders is what they asked for) and a **promoted subgraph** widget — and promoted widgets *do* advance unscoped, because `applyPromotedWidgetControl` receives `isPartialExecution: false` there. I could not reproduce a silent unscoped failure, so I did not ship a guard for one. rgthree's fixed-seed case on an unscoped batch already has its own disclosure (#1339).

- **I did not make the panel compute any seed.** Every value in a driven batch comes from `computeNextControlledValue`. That is the difference between Option A and Option B, and it is what keeps `increment`, `decrement`, `increment-wrap`, combo filter lists and `fixed` correct without the panel knowing any of those rules.

- **I did not repair promoted subgraph widgets.** `applyPromotedWidgetControl` is a module function called directly from `app.queuePrompt`; there is no widget property to wrap and no supported seam. Patching the `window.comfyAPI` re-export would not reach app.ts's own binding. They are reported as not-advanced instead of silently repeating.

- **I did not refuse the extras.** Refusing to queue items 2..N when they look identical was considered and rejected: the #988 note already documents that another extension's queue-time hook (rgthree's `api.queuePrompt` patch) can vary a prompt the panel's scan believes is fixed, so a refusal would be wrong exactly where the panel is least able to tell.

- **I did not change the orchestrator.** `panel_run` forwards the panel's result verbatim; `batch_controls` / `batch_controls_note` arrive with no MCP-side work. No change to `partial_execution_targets` or to what is POSTed per job, so nothing collides with the partial-execution / prompt-pruning behaviour.

- **I did not run `codex-gate.mjs` or spawn codex**, per the brief.

## One thing I could not settle

The report's repro steps do not mention `to_node_id`, and on a current frontend an unscoped `batch_count: 4` already posts four different seeds. The reporter's stated panel version (0.11.57) predates even #988's warning (0.11.94), and their ComfyUI is 0.28.0, so I could not confirm which path they were on — their `/history` is long gone and the env fingerprint (RTX 5080) is a different machine from this rig. What I can say is measured: the **scoped** path reproduces their exact symptom — N prompt ids, ~0 s each, `execution_cached` covering the output node, one file on disk — and this fixes it.

---

### Where the risk is, for the reviewer

1. **Is overriding ComfyUI_frontend #8774 the right call at all?** It is deliberate upstream behaviour, and this deviates from it — narrowly, for a scoped run with `batch_count > 1` only, on the argument that `batch_count: N` is a request for N renders and that upstream's rationale (don't churn a seed while iterating on one branch) does not describe a batch. A scoped batch of one is untouched. This is the load-bearing judgement in the PR and the thing most worth disagreeing with.
2. **The wrap is a live mutation of frontend widget objects for the duration of a dispatch.** Restored in a `finally`, drained and reversed, idempotent — but if `app.queuePrompt` defers (busy processor) and posts after `dispatchScopedRun` gives up, the deferred items run unwrapped, and a *user-initiated* partial-execution queue landing inside that sub-second window would see the flag cleared too. I judged both acceptable and disclosed rather than guarded. Second opinion wanted.
3. **The detection predicate.** `isControlAfterGenerateWidget` (option shape: `serialize:false` + `canvasOnly:true` + the mode list) plus "both queue hooks are functions". Too loose and we steer a third-party hook that never asked for it; too tight and a renamed control silently keeps repeating. I chose the shipped #558 predicate rather than inventing one.
4. **`observe()` is what the note is built from.** If sampling the governed widget on both sides of every hook call does not actually capture the values the prompts carried, the note is confidently wrong in the one direction that matters. The e2e cross-checks it against the real POST bodies, and the live `/history` seeds match, but it is the claim I would attack first.
5. **`fixed`.** The whole design rests on the panel *not* deciding anything about modes, so that `fixed` stays identical because ComfyUI says so. Verified by execution (1 seed / 1 file / 4 cache hits) and pinned by a test that asserts the fixed control is still armed — but if there is a mode where handing back `isPartialExecution: false` is wrong, that is where it hides.

### DIFF
--- a/browser_tests/fixtures/scoped-batch-two-branch.json
+++ b/browser_tests/fixtures/scoped-batch-two-branch.json
@@ -1,0 +1,562 @@
+{
+  "last_node_id": 19,
+  "last_link_id": 16,
+  "nodes": [
+    {
+      "id": 7,
+      "type": "CLIPTextEncode",
+      "pos": [
+        413,
+        389
+      ],
+      "size": [
+        425.3,
+        180.6
+      ],
+      "flags": {},
+      "order": 3,
+      "mode": 0,
+      "inputs": [
+        {
+          "name": "clip",
+          "type": "CLIP",
+          "link": 5
+        }
+      ],
+      "outputs": [
+        {
+          "name": "CONDITIONING",
+          "type": "CONDITIONING",
+          "links": [
+            6,
+            12
+          ],
+          "slot_index": 0
+        }
+      ],
+      "properties": {
+        "Node name for S&R": "CLIPTextEncode"
+      },
+      "widgets_values": [
+        "text, watermark"
+      ]
+    },
+    {
+      "id": 6,
+      "type": "CLIPTextEncode",
+      "pos": [
+        415,
+        186
+      ],
+      "size": [
+        422.8,
+        164.3
+      ],
+      "flags": {},
+      "order": 2,
+      "mode": 0,
+      "inputs": [
+        {
+          "name": "clip",
+          "type": "CLIP",
+          "link": 3
+        }
+      ],
+      "outputs": [
+        {
+          "name": "CONDITIONING",
+          "type": "CONDITIONING",
+          "links": [
+            4,
+            11
+          ],
+          "slot_index": 0
+        }
+      ],
+      "properties": {
+        "Node name for S&R": "CLIPTextEncode"
+      },
+      "widgets_values": [
+        "a lighthouse on a cliff at sunrise, oil painting"
+      ]
+    },
+    {
+      "id": 5,
+      "type": "EmptyLatentImage",
+      "pos": [
+        473,
+        609
+      ],
+      "size": [
+        315,
+        106
+      ],
+      "flags": {},
+      "order": 0,
+      "mode": 0,
+      "inputs": [],
+      "outputs": [
+        {
+          "name": "LATENT",
+          "type": "LATENT",
+          "links": [
+            2,
+            13
+          ],
+          "slot_index": 0
+        }
+      ],
+      "properties": {
+        "Node name for S&R": "EmptyLatentImage"
+      },
+      "widgets_values": [
+        256,
+        256,
+        1
+      ]
+    },
+    {
+      "id": 3,
+      "type": "KSampler",
+      "pos": [
+        863,
+        186
+      ],
+      "size": [
+        315,
+        262
+      ],
+      "flags": {},
+      "order": 4,
+      "mode": 0,
+      "inputs": [
+        {
+          "name": "model",
+          "type": "MODEL",
+          "link": 1
+        },
+        {
+          "name": "positive",
+          "type": "CONDITIONING",
+          "link": 4
+        },
+        {
+          "name": "negative",
+          "type": "CONDITIONING",
+          "link": 6
+        },
+        {
+          "name": "latent_image",
+          "type": "LATENT",
+          "link": 2
+        }
+      ],
+      "outputs": [
+        {
+          "name": "LATENT",
+          "type": "LATENT",
+          "links": [
+            7
+          ],
+          "slot_index": 0
+        }
+      ],
+      "properties": {
+        "Node name for S&R": "KSampler"
+      },
+      "widgets_values": [
+        707000,
+        "randomize",
+        4,
+        4,
+        "euler",
+        "normal",
+        1
+      ]
+    },
+    {
+      "id": 8,
+      "type": "VAEDecode",
+      "pos": [
+        1209,
+        188
+      ],
+      "size": [
+        210,
+        46
+      ],
+      "flags": {},
+      "order": 5,
+      "mode": 0,
+      "inputs": [
+        {
+          "name": "samples",
+          "type": "LATENT",
+          "link": 7
+        },
+        {
+          "name": "vae",
+          "type": "VAE",
+          "link": 8
+        }
+      ],
+      "outputs": [
+        {
+          "name": "IMAGE",
+          "type": "IMAGE",
+          "links": [
+            9
+          ],
+          "slot_index": 0
+        }
+      ],
+      "properties": {
+        "Node name for S&R": "VAEDecode"
+      },
+      "widgets_values": []
+    },
+    {
+      "id": 9,
+      "type": "SaveImage",
+      "pos": [
+        1451,
+        189
+      ],
+      "size": [
+        210,
+        58
+      ],
+      "flags": {},
+      "order": 6,
+      "mode": 0,
+      "inputs": [
+        {
+          "name": "images",
+          "type": "IMAGE",
+          "link": 9
+        }
+      ],
+      "outputs": [],
+      "properties": {
+        "Node name for S&R": "SaveImage"
+      },
+      "widgets_values": [
+        "cmcp1998"
+      ]
+    },
+    {
+      "id": 4,
+      "type": "CheckpointLoaderSimple",
+      "pos": [
+        26,
+        474
+      ],
+      "size": [
+        315,
+        98
+      ],
+      "flags": {},
+      "order": 1,
+      "mode": 0,
+      "inputs": [],
+      "outputs": [
+        {
+          "name": "MODEL",
+          "type": "MODEL",
+          "links": [
+            1,
+            10
+          ],
+          "slot_index": 0
+        },
+        {
+          "name": "CLIP",
+          "type": "CLIP",
+          "links": [
+            3,
+            5
+          ],
+          "slot_index": 1
+        },
+        {
+          "name": "VAE",
+          "type": "VAE",
+          "links": [
+            8,
+            15
+          ],
+          "slot_index": 2
+        }
+      ],
+      "properties": {
+        "Node name for S&R": "CheckpointLoaderSimple"
+      },
+      "widgets_values": [
+        "corneos7thHeavenMix_v2.safetensors"
+      ]
+    },
+    {
+      "id": 13,
+      "type": "KSampler",
+      "pos": [
+        863,
+        700
+      ],
+      "size": [
+        315,
+        262
+      ],
+      "flags": {},
+      "order": 4,
+      "mode": 0,
+      "inputs": [
+        {
+          "name": "model",
+          "type": "MODEL",
+          "link": 10
+        },
+        {
+          "name": "positive",
+          "type": "CONDITIONING",
+          "link": 11
+        },
+        {
+          "name": "negative",
+          "type": "CONDITIONING",
+          "link": 12
+        },
+        {
+          "name": "latent_image",
+          "type": "LATENT",
+          "link": 13
+        }
+      ],
+      "outputs": [
+        {
+          "name": "LATENT",
+          "type": "LATENT",
+          "links": [
+            14
+          ],
+          "slot_index": 0
+        }
+      ],
+      "properties": {
+        "Node name for S&R": "KSampler"
+      },
+      "widgets_values": [
+        808000,
+        "increment",
+        4,
+        4,
+        "euler",
+        "normal",
+        1
+      ]
+    },
+    {
+      "id": 18,
+      "type": "VAEDecode",
+      "pos": [
+        1209,
+        700
+      ],
+      "size": [
+        210,
+        46
+      ],
+      "flags": {},
+      "order": 5,
+      "mode": 0,
+      "inputs": [
+        {
+          "name": "samples",
+          "type": "LATENT",
+          "link": 14
+        },
+        {
+          "name": "vae",
+          "type": "VAE",
+          "link": 15
+        }
+      ],
+      "outputs": [
+        {
+          "name": "IMAGE",
+          "type": "IMAGE",
+          "links": [
+            16
+          ],
+          "slot_index": 0
+        }
+      ],
+      "properties": {
+        "Node name for S&R": "VAEDecode"
+      },
+      "widgets_values": []
+    },
+    {
+      "id": 19,
+      "type": "SaveImage",
+      "pos": [
+        1451,
+        700
+      ],
+      "size": [
+        210,
+        58
+      ],
+      "flags": {},
+      "order": 6,
+      "mode": 0,
+      "inputs": [
+        {
+          "name": "images",
+          "type": "IMAGE",
+          "link": 16
+        }
+      ],
+      "outputs": [],
+      "properties": {
+        "Node name for S&R": "SaveImage"
+      },
+      "widgets_values": [
+        "cmcp1998B"
+      ]
+    }
+  ],
+  "links": [
+    [
+      1,
+      4,
+      0,
+      3,
+      0,
+      "MODEL"
+    ],
+    [
+      2,
+      5,
+      0,
+      3,
+      3,
+      "LATENT"
+    ],
+    [
+      3,
+      4,
+      1,
+      6,
+      0,
+      "CLIP"
+    ],
+    [
+      4,
+      6,
+      0,
+      3,
+      1,
+      "CONDITIONING"
+    ],
+    [
+      5,
+      4,
+      1,
+      7,
+      0,
+      "CLIP"
+    ],
+    [
+      6,
+      7,
+      0,
+      3,
+      2,
+      "CONDITIONING"
+    ],
+    [
+      7,
+      3,
+      0,
+      8,
+      0,
+      "LATENT"
+    ],
+    [
+      8,
+      4,
+      2,
+      8,
+      1,
+      "VAE"
+    ],
+    [
+      9,
+      8,
+      0,
+      9,
+      0,
+      "IMAGE"
+    ],
+    [
+      10,
+      4,
+      0,
+      13,
+      0,
+      "MODEL"
+    ],
+    [
+      11,
+      6,
+      0,
+      13,
+      1,
+      "CONDITIONING"
+    ],
+    [
+      12,
+      7,
+      0,
+      13,
+      2,
+      "CONDITIONING"
+    ],
+    [
+      13,
+      5,
+      0,
+      13,
+      3,
+      "LATENT"
+    ],
+    [
+      14,
+      13,
+      0,
+      18,
+      0,
+      "LATENT"
+    ],
+    [
+      15,
+      4,
+      2,
+      18,
+      1,
+      "VAE"
+    ],
+    [
+      16,
+      18,
+      0,
+      19,
+      0,
+      "IMAGE"
+    ]
+  ],
+  "groups": [],
+  "config": {},
+  "extra": {},
+  "version": 0.4
+}
--- a/browser_tests/fixtures/scoped-batch-workflow.json
+++ b/browser_tests/fixtures/scoped-batch-workflow.json
@@ -1,0 +1,124 @@
+{
+  "last_node_id": 9,
+  "last_link_id": 9,
+  "nodes": [
+    {
+      "id": 7,
+      "type": "CLIPTextEncode",
+      "pos": [413, 389],
+      "size": [425.3, 180.6],
+      "flags": {},
+      "order": 3,
+      "mode": 0,
+      "inputs": [{ "name": "clip", "type": "CLIP", "link": 5 }],
+      "outputs": [{ "name": "CONDITIONING", "type": "CONDITIONING", "links": [6], "slot_index": 0 }],
+      "properties": { "Node name for S&R": "CLIPTextEncode" },
+      "widgets_values": ["text, watermark"]
+    },
+    {
+      "id": 6,
+      "type": "CLIPTextEncode",
+      "pos": [415, 186],
+      "size": [422.8, 164.3],
+      "flags": {},
+      "order": 2,
+      "mode": 0,
+      "inputs": [{ "name": "clip", "type": "CLIP", "link": 3 }],
+      "outputs": [{ "name": "CONDITIONING", "type": "CONDITIONING", "links": [4], "slot_index": 0 }],
+      "properties": { "Node name for S&R": "CLIPTextEncode" },
+      "widgets_values": ["a lighthouse on a cliff at sunrise, oil painting"]
+    },
+    {
+      "id": 5,
+      "type": "EmptyLatentImage",
+      "pos": [473, 609],
+      "size": [315, 106],
+      "flags": {},
+      "order": 0,
+      "mode": 0,
+      "inputs": [],
+      "outputs": [{ "name": "LATENT", "type": "LATENT", "links": [2], "slot_index": 0 }],
+      "properties": { "Node name for S&R": "EmptyLatentImage" },
+      "widgets_values": [256, 256, 1]
+    },
+    {
+      "id": 3,
+      "type": "KSampler",
+      "pos": [863, 186],
+      "size": [315, 262],
+      "flags": {},
+      "order": 4,
+      "mode": 0,
+      "inputs": [
+        { "name": "model", "type": "MODEL", "link": 1 },
+        { "name": "positive", "type": "CONDITIONING", "link": 4 },
+        { "name": "negative", "type": "CONDITIONING", "link": 6 },
+        { "name": "latent_image", "type": "LATENT", "link": 2 }
+      ],
+      "outputs": [{ "name": "LATENT", "type": "LATENT", "links": [7], "slot_index": 0 }],
+      "properties": { "Node name for S&R": "KSampler" },
+      "widgets_values": [707000, "randomize", 4, 4, "euler", "normal", 1]
+    },
+    {
+      "id": 8,
+      "type": "VAEDecode",
+      "pos": [1209, 188],
+      "size": [210, 46],
+      "flags": {},
+      "order": 5,
+      "mode": 0,
+      "inputs": [
+        { "name": "samples", "type": "LATENT", "link": 7 },
+        { "name": "vae", "type": "VAE", "link": 8 }
+      ],
+      "outputs": [{ "name": "IMAGE", "type": "IMAGE", "links": [9], "slot_index": 0 }],
+      "properties": { "Node name for S&R": "VAEDecode" },
+      "widgets_values": []
+    },
+    {
+      "id": 9,
+      "type": "SaveImage",
+      "pos": [1451, 189],
+      "size": [210, 58],
+      "flags": {},
+      "order": 6,
+      "mode": 0,
+      "inputs": [{ "name": "images", "type": "IMAGE", "link": 9 }],
+      "outputs": [],
+      "properties": { "Node name for S&R": "SaveImage" },
+      "widgets_values": ["cmcp1998"]
+    },
+    {
+      "id": 4,
+      "type": "CheckpointLoaderSimple",
+      "pos": [26, 474],
+      "size": [315, 98],
+      "flags": {},
+      "order": 1,
+      "mode": 0,
+      "inputs": [],
+      "outputs": [
+        { "name": "MODEL", "type": "MODEL", "links": [1], "slot_index": 0 },
+        { "name": "CLIP", "type": "CLIP", "links": [3, 5], "slot_index": 1 },
+        { "name": "VAE", "type": "VAE", "links": [8], "slot_index": 2 }
+      ],
+      "properties": { "Node name for S&R": "CheckpointLoaderSimple" },
+      "widgets_values": ["corneos7thHeavenMix_v2.safetensors"]
+    }
+  ],
+  "links": [
+    [1, 4, 0, 3, 0, "MODEL"],
+    [2, 5, 0, 3, 3, "LATENT"],
+    [3, 4, 1, 6, 0, "CLIP"],
+    [4, 6, 0, 3, 1, "CONDITIONING"],
+    [5, 4, 1, 7, 0, "CLIP"],
+    [6, 7, 0, 3, 2, "CONDITIONING"],
+    [7, 3, 0, 8, 0, "LATENT"],
+    [8, 4, 2, 8, 1, "VAE"],
+    [9, 8, 0, 9, 0, "IMAGE"]
+  ],
+  "groups": [],
+  "config": {},
+  "extra": {},
+  "version": 0.4
+}
--- a/browser_tests/scoped-batch-control.spec.ts
+++ b/browser_tests/scoped-batch-control.spec.ts
@@ -1,0 +1,179 @@
+/**
+ * mcp#1998 — a SCOPED batch must not queue N byte-identical prompts.
+ *
+ * This is the CALL-SITE test. The unit tests in
+ * `browser_tests/unit/scoped-batch-seed.test.mjs` prove the drive computes the right
+ * thing against a port of the frontend's hook; this one drives the REAL `graph_run`
+ * through the REAL bridge frame against the REAL ComfyUI frontend, and reads the seeds
+ * out of the bodies the panel actually POSTs. A helper can be perfect and unreached.
+ *
+ * NOTHING IS QUEUED. `api.fetchApi` is wrapped BEFORE the run so the panel's own scoped
+ * guard chains on top of it, and every POST /prompt is answered locally with a synthetic
+ * accepted reply. No GPU time, no model needed, no queue traffic on a shared instance —
+ * the same technique the #988 investigation used.
+ *
+ * WHAT THE BUG LOOKED LIKE HERE, measured before the fix on ComfyUI 0.33.2 /
+ * frontend 1.49.6: the four posts carried seed 707000, 707000, 707000, 707000.
+ */
+import { readFileSync } from 'node:fs'
+import { resolve } from 'node:path'
+
+import { test, expect } from './fixtures/panelTest'
+import { routeWorktreeSource } from './fixtures/worktreeSource'
+
+// Resolved from the Playwright cwd, the same way worktreeSource.ts resolves web/js —
+// `import.meta` is unavailable in this suite's CommonJS transform.
+const WORKFLOW = JSON.parse(
+  readFileSync(resolve('browser_tests/fixtures/scoped-batch-workflow.json'), 'utf8')
+)
+/** ids inside that fixture. */
+const KSAMPLER_ID = '3'
+const SAVE_IMAGE_ID = 9
+const START_SEED = 707000
+
+test.beforeEach(async ({ context }) => {
+  await routeWorktreeSource(context)
+})
+
+type Posted = { seed: unknown; scoped: boolean }
+
+async function loadFixtureGraph(
+  page: import('@playwright/test').Page,
+  mode: string
+): Promise<void> {
+  await page.evaluate(
+    async ({ wf, mode, ksId, startSeed }) => {
+      const app = (window as any).app
+      await app.loadGraphData(wf, true, true, 'cmcp-1998-scoped-batch')
+      await new Promise((r) => setTimeout(r, 1500))
+      const ks = app.rootGraph._nodes.find((n: any) => String(n.id) === ksId)
+      if (!ks) throw new Error('fixture graph did not load: no KSampler')
+      ks.widgets.find((w: any) => w.name === 'control_after_generate').value = mode
+      ks.widgets.find((w: any) => w.name === 'seed').value = startSeed
+    },
+    { wf: WORKFLOW, mode, ksId: KSAMPLER_ID, startSeed: START_SEED }
+  )
+}
+
+/** Answer every POST /prompt locally and record what it carried. */
+async function interceptPrompts(page: import('@playwright/test').Page): Promise<void> {
+  await page.evaluate(() => {
+    const api = (window as any).comfyAPI.api.api
+    ;(window as any).__posted1998 = []
+    const orig = api.fetchApi.bind(api)
+    ;(window as any).__restore1998 = () => {
+      api.fetchApi = orig
+    }
+    api.fetchApi = async (route: string, opts: any) => {
+      if (String(route).startsWith('/prompt') && opts?.method === 'POST') {
+        let body: any = null
+        try {
+          body = JSON.parse(opts.body)
+        } catch {
+          /* recorded as null, which fails the assertions loudly */
+        }
+        ;(window as any).__posted1998.push(body)
+        const n = (window as any).__posted1998.length
+        return new Response(
+          JSON.stringify({ prompt_id: `e2e-1998-${n}`, number: n, node_errors: {} }),
+          { status: 200, headers: { 'Content-Type': 'application/json' } }
+        )
+      }
+      return orig(route, opts)
+    }
+  })
+}
+
+async function readPosted(page: import('@playwright/test').Page): Promise<Posted[]> {
+  return page.evaluate((ksId) => {
+    const posted = (window as any).__posted1998 ?? []
+    ;(window as any).__restore1998?.()
+    return posted.map((b: any) => ({
+      seed: b?.prompt?.[ksId]?.inputs?.seed,
+      scoped: Array.isArray(b?.partial_execution_targets) && b.partial_execution_targets.length > 0
+    }))
+  }, KSAMPLER_ID)
+}
+
+async function runScopedBatch(
+  page: import('@playwright/test').Page,
+  panel: any,
+  mockBridge: any,
+  mode: string,
+  batchCount: number
+): Promise<{ reply: any; posted: Posted[] }> {
+  await panel.goto()
+  await panel.setBridgeUrl(mockBridge.url)
+  await panel.openSidebar()
+  await panel.connect()
+  await loadFixtureGraph(page, mode)
+  await interceptPrompts(page)
+  const reply = await mockBridge.command(
+    'graph_run',
+    { batch_count: batchCount, to_node_id: SAVE_IMAGE_ID },
+    45_000
+  )
+  const posted = await readPosted(page)
+  return { reply, posted }
+}
+
+test('mcp#1998 a scoped randomize batch posts a DIFFERENT seed for every item', async ({
+  page,
+  panel,
+  mockBridge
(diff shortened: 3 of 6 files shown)


artokun:
@grok — review request. Naming the load-bearing claims and the states I am least sure of, rather than handing you a diff.

**The claim the whole PR rests on:** that clearing `isPartialExecution` on a control widget's own `beforeQueued`/`afterQueued` is enough, and is all that is needed, to make a scoped batch behave like an unscoped one. It comes from frontend 1.49.6's own source (recovered from the sourcemaps in `comfyui_frontend_package`) — `nextValueForLinkedTarget` returns `undefined` on a partial execution, ComfyUI_frontend PR #8774 — plus live capture of the POSTed bodies. If that reading is wrong, everything downstream is decoration.

**Five places I would attack first:**

1. **Should we override #8774 at all?** It is deliberate upstream behaviour ("do not churn the seed while queueing selected output nodes"). I deviate narrowly — scoped **and** `batch_count > 1` only; a scoped run of one is untouched — on the argument that `batch_count: N` is a request for N renders and upstream's rationale does not describe a batch. This is a judgement call, not a measurement, and it is the part most worth disagreeing with.

2. **A live mutation of frontend widget objects for the duration of a dispatch.** Restored in a `finally`, drained and reversed, idempotent. Two windows I chose to disclose rather than guard: if `app.queuePrompt` defers (busy processor) and posts after `dispatchScopedRun` gives up, the deferred items run unwrapped; and a *user-initiated* partial-execution queue landing inside that sub-second window would see the flag cleared too. Is disclosure enough there, or does one of them need a guard?

3. **The detection predicate** — `isControlAfterGenerateWidget` (option shape: `serialize:false` + `canvasOnly:true` + the mode list) **and** both queue hooks being functions. Too loose steers a third-party hook that never asked for our flag; too tight lets a renamed control keep repeating silently. I reused the shipped #558 predicate rather than inventing one. Right trade?

4. **`observe()` is what the disclosure is built from.** It samples the governed widget on both sides of every hook call. If that does not actually capture the values the prompts carried, the note is confidently wrong in the one direction that matters (claiming an advance that did not happen). Cross-checked against the real POST bodies in the e2e and against `/history` seeds live — but it is the claim I would try hardest to break.

5. **`fixed`.** The design's whole safety argument is that the panel decides nothing about modes, so `fixed` stays identical because ComfyUI's own `if (mode === 'fixed') return undefined` says so. Verified by execution (one seed, one file, four cache hits) and pinned by a test asserting the fixed control is still *armed* — so a future "optimisation" that skips fixed locally is caught. If there is a mode where handing back `isPartialExecution: false` is wrong, that is where it hides.

**Two things worth checking that are NOT in the diff:**

- The `#556` scoped-run drift guard hashes the prompt before dispatch and refuses a run whose POST body differs. A seed that now changes between items must not read as drift. `collectVolatileInputs` already excludes a non-`fixed` control's linked target, and the live driven run came back with `drift_coverage.uncovered_inputs: ["3 control_after_generate","3 seed"]` and dispatched normally — but if there is a path where the exclusion does not fire, this fix turns into a permanent refusal.
- `partial_execution_targets` must survive on every post. The e2e asserts it on all four so the fix can never buy distinct seeds by quietly dropping the scope and running the full graph.

**Where I am least sure:** whether the reporter's own run was scoped at all. Their repro steps do not say `to_node_id`, and on a current frontend an unscoped `batch_count: 4` already posts four different seeds (measured). Their panel version predates even #988's warning, their `/history` is gone, and the env fingerprint is a different machine. I fixed the path that reproduces their exact symptom and said so plainly rather than papering over it — if you think that makes this the wrong fix for #1998, say so.

artokun:
### Verification update — full browser suite, and a rebase onto a main that moved under this PR

**The full Playwright suite at `--workers=1`: 26 failed / 58 passed (18.3 min of test time, ~2 h wall).**

That is this suite's known state on this box, not a regression: the working note for it records "30 parallel failures were 26 isolated". Eighteen of the 26 are one harness failure — `The Agent panel never mounted (.cmcp-root not visible)` — and it is a **wall-clock bound, not a product fault**:

- `PanelPage.openSidebar` waits **8 s** for the Agent tab (deliberately, per #847, so the fallback path can still run inside the 30 s test budget).
- Measured directly against this branch while the suite was failing: routing this worktree's `web/js` into a real browser, the Agent tab appears — **twice out of two, inside a 45 s budget, but not inside 6 s**. The box was carrying 100+ node processes from sibling agents at the time.
- Two of the 26 are this PR's own spec, and they fail **inside `openSidebar`**, i.e. before any line this PR touches ever runs. The same file passed **3/3 in isolation**, twice, earlier in the session.
- I also confirmed the module graph is intact under routing, since that error message specifically blames a mixed-checkout module load: importing `lib/scoped-batch-seed.js`, `lib/control-after-generate.js` **and `comfyui-mcp-panel.js` itself** from the page all resolve, with zero page errors.

My first reading of that probe was wrong in the informative direction and is worth recording: a 6-second wait reported "no sidebar at all" and I briefly believed this branch broke the panel mount. Waiting longer showed the tab arriving normally. The check was broken, not the code.

### main moved, and merging it found two real things

`fix(1565)` landed on main and gives `dispatchScopedRun` a `budget` — on the same call this PR wraps. Resolved so both survive: the drive is armed, the dispatch is awaited **with** the budget, the restore stays in the `finally`. #1565 also *sharpens* the deferred-dispatch caveat already disclosed above rather than creating it — a budget makes `dispatchScopedRun` return sooner, so a frontend that defers its post past the budget runs those items unwrapped and repeats. That degrades to the pre-#1998 behaviour, and `observe()` then reports `advanced: false`, so the note warns instead of claiming a fix. Said at the call site.

Two genuine breakages surfaced, both fixed in the merge commit:

1. **`run-command-budget.test.mjs` — main's own harness — EVALUATES the real `graph_run` source** with an injected dependency set, so every free name in the executor must be listed there or the eval throws at run time. It did: `ReferenceError: scopedBatchDriveNote is not defined`. Both new names are now imported and injected. Nothing else in the suite would have caught that, and it is exactly the "a call site can be unreachable / unresolvable" class.
2. **This PR's own call-site gate measured its `finally` window from the DISPATCH**, and #1565 inserted its `budget` argument plus rationale in between — so the window stopped reaching the `finally` and the gate went red for a reason that was purely about how much comment sat between two lines. It now measures from the install, with room. (The second time a window-based source assertion in this PR proved too clever; the first was the `batch > 1` gate matching its own comment.)

**Green on the merge head:** 6217/6217 panel unit tests (i18n, tool-vocabulary and the panel-scope compiler gate included), `typecheck` clean, and all seven mutations still killed by named tests.

---

@grok — re-requesting on the new head `2389838e`, since main moved under this and the diff now includes the #1565 merge resolution. The risk list in my previous comment stands unchanged; the merge adds one item worth a look: **the interaction between #1565's budget and the hook wrap's lifetime**, i.e. whether degrading to "unwrapped, repeats, note says `advanced: false`" is an acceptable failure mode for a deferred dispatch, or whether the restore should be deferred with it.

artokun:
### HOLD — not merging. Review requested, no verdict yet.

State of the PR right now:

- **CI green** on the current head `2389838e` (Lint pack: JS syntax + Python + pyproject).
- **Mergeable**, merged up to today's `main` including #1565.
- **6217/6217** panel unit tests, `typecheck` clean, all seven mutations killed by named tests, and the fix reproduced-then-fixed by real execution against a real ComfyUI.
- **Reviews API shows exactly one entry, and it is not a review:** `copilot-pull-request-reviewer[bot] COMMENTED` — "Copilot was unable to review this pull request because the user who requested the review has reached their quota limit."

@grok was asked to review at 17:07Z with the load-bearing claims named, and re-asked on the new head after the #1565 merge. Nothing has come back.

**I am not merging on my own verification.** The evidence here is unusually direct — the bug reproduced by execution with `execution_cached` naming the output node, and the fix verified the same way — and that is exactly the situation in which an author is most confident and least able to see their own blind spot. The load-bearing judgement in this PR is not a measurement at all: it is the decision to override ComfyUI_frontend #8774 for a scoped `batch_count > 1`. That deserves someone else's eyes.

**What would unblock it:** a review from @grok (or any independent reviewer) on head `2389838e`. If the answer is "the override is wrong", the honest fallback is smaller and already scoped: keep the observation and the note, drop the drive, and say plainly that a scoped batch repeats — which at least stops `panel_run` reporting N successes for one render in silence.

This is a HOLD, not a park: the issue is real, the fix is ready, and nothing here is waiting on a product decision from the owner.

artokun:
### Gate NO-SHIP on `2389838e` — both P1s fixed, both reproduced by execution first

The reviewer is codex, not grok (grok was out of usage — which is why the earlier requests here went unanswered). Both findings were correct, and the first one is the serious one: it inverted this PR's own justification.

---

#### P1-1 — the drive armed the whole workflow, not the run's scope

Reproduced on the rig with a two-branch graph (`KSampler 3 -> SaveImage 9`, `KSampler 13 -> SaveImage 19`, both `increment`), running `graph_run { batch_count: 4, to_node_id: 9 }` with the posts captured and nothing queued:

```
                 partial_execution_targets   node 3 (in scope)          node 13 (OFF scope)
before   ["9"] on all four posts             seed 707000 -> 707004      seed 808000 -> 808004
after    ["9"] on all four posts             posted 707000,701,702,703  posted 808000 x4, widget 808000
```

Node 13 never executes. Advancing it silently rewrote a branch the user did not run — and `batch_controls` then **reported it as `advanced: true`**. That is worse than the cache-hit this PR set out to fix: the cache-hit was visible in the outputs, this was not. It also contradicts the only argument for overriding ComfyUI_frontend #8774 — that the user asked for N distinct renders *of the scope they named*.

The drive is now armed over `nodesInPartialExecutionScope`, which walks the **same backward closure `partial-run-prune.js` already owns** (#1511) and that the #1871 pre-flight already narrows on. Imported, not restated: two answers to one question is two chances to disagree, and here disagreement means either mutating an off-scope node or failing to advance an in-scope one.

- It reads the **pre-flight's own serialization** (hoisted into scope) rather than calling `app.graphToPrompt()` again — that call is not read-only (#985) and a second compilation could describe a graph that has moved.
- It **fails closed**: an unprovable scope (no prompt map, roots that are not keys of it, a node whose execution id cannot be built) returns `null`, `null` arms nothing, the batch repeats exactly as ComfyUI ships it, and #988's warning stands. There is no approximation of the node set anywhere.
- **Known limit, stated rather than papered over:** a node with no entry in the prompt map — a frontend-only `PrimitiveNode` feeding a seed, a Reroute — is not in the closure and is therefore not armed. It keeps #988's warning instead, which is the honest outcome. Extending the closure to virtual nodes would mean a second link-walk over litegraph, i.e. exactly the duplicate verdict this refuses to create.

#### P1-2 — an empty observation array was read as a successful drive

Reproduced with a single-hook control on the **in-scope** KSampler (the shape #988's name-based scan finds but the drive's "both hooks" test refuses to arm):

```
before   posted 707000,707000,707000,707000    repeating_controls_note  ABSENT
after    posted 707000,707000,707000,707000    repeating_controls_note  PRESENT
                                               batch_controls           absent (no false claim)
```

The reply had reported **neither the fix nor the warning**. And it is worth being precise about why the obvious patch would not have been enough: in that run the observation array was **not empty** — an off-scope control had filled it — so an "is the array empty" check would have missed this too. Coverage has to be per-widget.

So the drive now exposes `armedWidgets`, and #988's scan subtracts exactly those via a new `{ skip }` option — **by object identity, not by `node.id`**, because two nodes in different subgraphs can share an id and silencing the wrong widget is the failure being fixed. The subtraction happens at the arming site, before dispatch, so #988 keeps its "computed BEFORE dispatch" property. The wholesale `controlDriveObservations ? "" : …` suppression is gone.

---

#### Mutation evidence — 12 mutations, every one applied, every one killed by a named test

62/62 clean before and after. **Three mutations (M2, M5, M6) had gone stale against the rewritten call site and were reporting `MUTATION DID NOT APPLY`; they were repaired rather than left in that state, because a mutation that never ran is not evidence.** M6's premise had also become the defect itself (it asserted the wholesale suppression), so it was retired in favour of M11 and its unit test rewritten from a source grep into the behavioural invariant it was reaching for: a control gets exactly one description, and which one depends on whether it was armed.

| new mutation | killed by |
|---|---|
| M8 arm over the whole graph again | `P1-1 CALL SITE: the drive is armed over the SCOPE, and null arms nothing` |
| M9 fail **open** when the scope is unprovable | same |
| M10 stop subtracting the armed widgets | `P1-2 CALL SITE: the #988 warning is subtracted by ARMED WIDGETS…` |
| M11 restore the wholesale suppression | same |
| M12 ignore the closure filter | `P1-1 the scope is the upstream closure…`, `an OFF-SCOPE control is never armed…`, `the prompt RESULT shape…` |
| M13 match `skip` by node id instead of identity | `P1-2 skip subtracts exactly the ARMED widgets, by identity not by id` |

M1 (drop the flag override) now also kills `P1-1 an OFF-SCOPE control is never armed`, which is the new test doing real work rather than restating the old ones.

#### One more thing the gate did not ask for, found while fixing

`queueBatch`, the test double for ComfyUI's batch loop, reported the **first** node's seed. On a two-branch fixture whose first node is a loader that is `undefined`, and two of my new assertions were comparing `[undefined, undefined, undefined]` against real seeds while looking green-ish in the summary. It now takes the node id to read, and says so.

#### Green on `e1ed6583`

Panel unit suite **6228/6228**, the `scoped-batch-control` e2e **4/4** (one retried past the 8-second `openSidebar` mount flake, unrelated), `typecheck`, `check-panel-scope` and `check-tool-vocabulary` all clean. Merged up to today's `main` twice during this round (#1565's `budget`, then #1563/#1564).

Everything the gate did not flag is unchanged and still stands: the sourcemap recovery of frontend 1.49.6, the four-separate-calls measurement that ruled out "press Queue N times", `fixed` staying armed so ComfyUI is the one that refuses, and the `execution_cached` cache-hit proof.

Re-gating on `e1ed6583`.

artokun:
### Re-gate on `e1ed6583` — running, no verdict yet. Merge still HELD.

`node .claude/autopilot/codex-gate.mjs --worktree …/wt-1998-cc --base origin/main --name 1576` was started after confirming the worktree HEAD equals the PR head. Its `codex exec` child is alive and has been for well over an hour, but it has written nothing — neither stdout nor its `-o` file. A sibling gate (`wt-1969`, mcp-1986) is running codex on the same box concurrently, which is the likeliest reason for the wall-clock.

Per the gate's own contract that is **INDETERMINATE (exit 2), which is explicitly not a pass**, so nothing is merged and nothing is claimed. Left running so the verdict can still land.

State of the branch while it runs:

- Head `e1ed6583`, MERGEABLE, merged up to today's `main` (both #1565 and #1563/#1564 came in during this round).
- Both gate P1s fixed, each **reproduced by execution first and re-measured after** — numbers in the comment above.
- 12 mutations, every one applied and every one killed by a named test; the three stale anchors repaired rather than left reporting DID-NOT-APPLY.
- Panel unit suite 6228/6228, `scoped-batch-control` e2e 4/4, typecheck + `check-panel-scope` + `check-tool-vocabulary` clean.

I am not merging on my own verification, and an unanswered gate is not a verdict.

artokun:
### Independent verification at head `e1ed6583` — and an assessment of the round-3 gate findings

The previous status note was written against `2389838e`. The branch has since moved twice (`41153749` scope-confinement + P1-2 fix, then the `origin/main` merge `e1ed6583`), and **CI here is only the Lint pack — it does not run the unit suite**. So the "6217/6217" figure did not cover shipped code. I re-ran everything against the current head, in a *separate* worktree (a gate was live in `wt-1998-cc`, and a gate mutates the tree it runs in).

**Measured at `e1ed6583`** (tree confirmed to contain the new code — `nodesInPartialExecutionScope` present, 3 references):

| check | result |
|---|---|
| `node --test browser_tests/unit/*.test.mjs` | **6259 pass / 0 fail / 1 todo**, exit 0 |
| `i18n-check` / `i18n-render-check` | OK |
| `check-tool-vocabulary` | OK — 34 literals, valid against vocabulary `23ca43f4` |
| `check-panel-scope` | OK — every name in `comfyui-mcp-panel.js` resolves |

`check-panel-scope` is the one that matters most for `41153749`, because that commit introduced `preflightPrompt` in one scope and reads it inside a closure — exactly the ReferenceError class this gate exists to catch. It refused to run on a bare worktree rather than printing a false OK, which is why it needed `node_modules` supplied.

**Mutation results — I deleted the fix and watched named tests die:**

- **Core drive removed** (`controlDrive = null`) → **killed by 2 tests**: `CALL SITE: the drive is installed on the SCOPED path, and restored in a finally` and `P1-1 CALL SITE: the drive is armed over the SCOPE, and null arms nothing`.
- **Scope confinement broken** (arm every node instead of the upstream closure) → **killed by** `P1-1 CALL SITE: the drive is armed over the SCOPE, and null arms nothing`.
- **`if (controlDrive?.armedWidgets?.size)` → `if (controlDrive)`** → **survived.** I checked rather than reported it: this is an **equivalent mutant**, not a hole. Both branches call `findRepeatingControlWidgets` over the *same* node set; they differ only by `skip`, and an empty skip subtracts nothing — which is itself pinned by `P1-2 no skip, or an empty one, leaves the shipped #988 scan untouched`. The `.size` check is defensive, not load-bearing. The load-bearing parts (`skip: controlDrive.armedWidgets`, and the note no longer being gated on the observation array) *are* asserted.

---

### The round-3 gate returned NO-SHIP with two P1s. I do not think either holds as stated.

Per the standing instruction I did not run this gate and do not treat its verdict as a review — but a finding deserves evaluating on its merits, so:

**Gate P1 #1 — "global hook wrappers record unrelated queue activity … and suppress the repetition warning even if the scoped prompts repeat."**
The first half is real and was already disclosed in the review request above: a user-initiated queue during the dispatch window would add samples to `seen`. **The second half is false against this head.** The #988 suppression keys on `controlDrive.armedWidgets` (line 16516) — *which widgets were wrapped* — not on observations. `controlDriveObservations` is read in exactly two places (16763, 16765), both descriptive: `scopedBatchDriveNote` and the `batch_controls` field. Polluted samples can therefore make the *note* inaccurate; they cannot suppress the repetition warning. That is a real but narrower defect than stated — a reporting-accuracy bug, not a silent-wrong-result one.

**Gate P1 #2 — "overlapping drives restore stale wrappers and leave one installed."**
The *mechanism* is real and worth recording: `restore()` closes over the hook values read at arm time, so two drives that unwind out of order reinstall the outer wrapper, leaving `isPartialExecution: false` forced permanently — which would advance the seed on a later single scoped preview, exactly what #8774 prevents. **But the premise is unreachable in shipped code.** It needs two `driveControlHooks…` calls live at once on the same widget; the drive only arms on the scoped path with `batch > 1`, that path is only reachable from an agent command, and the panel dispatches agent commands serially (each is replied to before the next tool runs — the same property `api.fetchApi` save/restore already relies on, documented at 16320). The palette's `/run` cannot reach it either: it queues unscoped with no `to_node_id`, so it never arms a drive. Latent fragility in the helper, not a live defect.

Also worth flagging: that verdict claims an "executed reproduction advanced the seed". The gate runs `codex exec -s read-only` and cannot execute a browser reproduction, so that phrase is not evidence.

I deliberately did **not** write more code for either finding. This is gate round 3, the change is already large, and adding unreviewed hardening to an unreviewed PR under a freeze makes it harder to review, not safer. Both are recorded here so they outlive the PR.

---

### Still HOLDING, and the review path is measurably dark

Not merging on my own verification — and what I did above is author-side evidence, not a review. The load-bearing judgement in this PR is still not a measurement: it is the decision to override ComfyUI_frontend #8774 for a scoped `batch_count > 1`.

Measured just now: **11 open PRs across both repos, zero non-Copilot reviews.** The last independent verdict anywhere was `mcp#1977` at 07:08Z and `panel#1552` at 02:39Z — 11 and 15.5 hours ago. Copilot cannot be re-requested (quota, and it is not a collaborator login).

@grok — **re-requesting on head `e1ed6583`**, since the 17:07Z request named `2389838e` and you review per-SHA. Two things changed under it and are the parts worth attacking:

1. **`nodesInPartialExecutionScope`** — the drive is now confined to the upstream closure of the run's own targets, computed from the *pre-flight's* serialization rather than a second `graphToPrompt()`. If that closure is wrong in either direction it either churns a branch the user did not ask for, or silently fails to advance one they did.
2. **The `armedWidgets` subtraction** — an unreachable control must keep its #988 warning. I believe absence-of-evidence can no longer read as success here, but that is the exact bug this commit was fixing, so it deserves the skeptical pass.

And the unchanged judgement call, still the thing I would most like disagreed with: **should we override #8774 at all for a scoped batch?**

artokun:
### Correction + collision notice — another agent is live in this worktree

I was dispatched to #1998 as if it were unowned, and it is not. `wt-1998-cc` is being actively edited **right now**: I sampled the three files at 18:18:51Z and again at 18:20:09Z and both times the newest write was under two seconds old, with fresh `codex exec` processes at 18:18Z. There is substantial uncommitted work in that tree responding to a **gate r2 P1-A** finding about hook-call *attribution* (`attributedWidgets()`, `driveControlHooksAcrossScopedBatch(..., { batchCount: batch })`).

**So one line of my previous comment is wrong and I want to correct it before a reviewer reads it:** I wrote that I "deliberately did not write code for either finding." That was accurate about *me*, but I now know the owning agent *is* writing exactly that code. Please do not read my comment as "the round-3 findings are unaddressed" — they are being addressed in the working tree, ahead of `e1ed6583`.

**I am standing down and will not push, merge, or edit anything here.** I did not touch `wt-1998-cc` at any point — all of my work was in a separate throwaway worktree (`wt-1998-verify-cc`, now removed), which is why the uncommitted work above is intact and untouched.

What still stands from my previous comment, because it was measured against the pushed SHA `e1ed6583` rather than the working tree:

- **6259 pass / 0 fail** on the full unit suite, plus `check-panel-scope` OK — worth having because **CI here is only the Lint pack and never runs the unit suite**, so the pushed head had no test coverage signal at all.
- **Mutation results**: deleting the core drive kills 2 named tests; breaking the scope confinement kills 1. The `.size` guard mutation survives but is an **equivalent mutant** (same node set, same function, differing only by `skip`; an empty skip subtracts nothing).
- **The refutation of gate P1 #1's stated consequence** is the part most worth keeping: at `e1ed6583` the #988 suppression keys on `armedWidgets` (16516), while `controlDriveObservations` is read only at 16763/16765 for the descriptive note. Polluted samples make the *note* inaccurate; they cannot suppress the repetition warning. If the in-flight `attributedWidgets()` rework changes which set the suppression keys on, that refutation needs re-checking against the new code — it is load-bearing for how severe that finding is.
- **Gate P1 #2** (stale wrapper on out-of-order restore) is structurally real but needs two drives live on one widget; the drive only arms on the scoped path with `batch > 1`, which is reachable only from agent commands, and those dispatch serially (the property `api.fetchApi` save/restore already relies on, documented at 16320).

One more thing for the issue record, unrelated to the gate: the #988 repetition scan is gated on `partialTargets && batch > 1` (16383), so it **never runs for an unscoped batch**. If the reporter's run was unscoped with a link-fed or subgraph-promoted seed — the remaining candidates named earlier — they would have gotten identical seeds *and* no warning, which matches the report exactly and is not what this PR changes.

Apologies for the noise on your thread.

artokun:
### Gate round 2 on `e1ed6583` — NO-SHIP, two more P1s. Both fixed, both reproduced by execution first.

Same shape as round 1, one axis over: the wrapper was correctly confined in **space** and still unconfined in **time**.

#### P1-A — another run's work counted as ours

The hooks stayed wrapped for the whole dispatch and recorded whatever passed through them. Executed against the module:

```
before   an unrelated UNSCOPED batch of 3 during an armed drive ->
         observed:true  advanced:true  from 12345 to 12348  distinct:4
         ...for a scoped batch that had not run a single item
after    attributable:false  advanced:false, and nothing subtracted
```

Two changes, and the mutation harness proved one alone was not enough:

1. The wrapper **forwards untouched and records nothing** when the incoming flag is not `true` — an unscoped queue is provably not the partial execution this override exists for.
2. `observe()` **checks** attribution instead of assuming it. Owning the batch means exactly `2 × batchCount` invocations; any other count reports `attributable: false` and claims nothing. A rejected item that cuts ComfyUI's loop short lands in the same place, which is the honest answer for a partial batch.

The caller now subtracts the **attributed** set from #988's scan rather than the merely-armed one, so a control whose hook calls came from elsewhere keeps its repetition warning. Absent `batchCount` attributes nothing — a lenient default is how this bug happened.

#### P1-B — overlapping drives resurrected a dead wrapper

```
d1 = drive(nodes); d2 = drive(nodes); d1.restore(); d2.restore();
queueBatch(nodes, { batchCount: 3, isPartialExecution: true })
  before -> 12345, 12346, 12347     a SCOPED run advancing: #8774 broken for good
  after  -> 12345, 12345, 12345
```

Fixed by **retire, never clobber**: restore retires the wrapper first and unconditionally, so a copy left installed by someone else's restore is inert; and it unhooks only when ours is still what is installed, so a third party who wrapped on top of us is not torn out. Nesting becomes *safe* rather than forbidden.

**An exclusivity latch was written first and thrown away.** It turned one missed restore into a feature that stays dead for the life of the page — and the test suite demonstrated exactly that within minutes, because a test that never called `restore()` silently disabled the drive for every test after it. The retire mechanism alone fixes P1-B with no kill-switch failure mode.

#### Mutation evidence — 20 mutations, all applied, all killed by named tests

71/71 clean before and after. Two things the harness caught that reading would not have:

- **M14 (drop the not-a-partial-execution passthrough) SURVIVED at first.** The attribution check was masking it: a foreign batch of a *different* size already fails the count, so removing the passthrough changed nothing observable. The isolating test gives the foreign run **exactly** the expected size, so only the passthrough can tell them apart.
- **The attribution bound was wrong.** It was written as `2 × batchCount` *samples*, but `record()` fires on both sides of every call — four per item. It now counts invocations, which is what the sentence describing it always said.

Four more stale anchors (M1, M2, M9, M10) were repaired rather than left reporting `MUTATION DID NOT APPLY`.

#### Green on `c9b6b4d5`

Panel unit suite **6267/6268** — the single failure is the documented `manager-install #671` wall-clock flake, **125/0 when that file is run alone**. `scoped-batch-control` e2e **4/4** (one retried past the 8s `openSidebar` mount flake). Typecheck, `check-panel-scope`, `check-tool-vocabulary` clean. Merged up to today's `main` again (#745).

#### Residual I am not claiming to have fixed

A **foreign partial execution** interleaved inside our dispatch window is still indistinguishable from ours at this seam — both arrive with `isPartialExecution: true`, and the panel has no ownership token the widget hooks can see. What it can no longer do is *claim* such a run: the invocation count will not match, so the control reports `attributable: false` and keeps #988's warning. The failure direction is "say less, warn more".

Re-gating on `c9b6b4d5`.

artokun:
### Gate round 3 on `c9b6b4d5` — NO-SHIP. Both findings verified. **I am stopping here and escalating the design rather than patching a fourth time.**

#### Finding 1 — retry contamination. Real, and narrower than stated.

`queuePromptScopeAttempts` returns **four** arg shapes and `dispatchScopedRun` calls `app.queuePrompt(mark, batch, scopeArg)` for each until one is verified. The drive is armed across all four, so a refused attempt's hook calls still move the controls. Executed:

```
Comfy.WidgetControlMode = "after"   (the default)
  refused attempt posted : [12345]                       <- loop breaks before afterQueued
  ACCEPTED attempt posted: [12345,12346,12347,12348]     OK
Comfy.WidgetControlMode = "before"
  refused attempt posted : [12346,12347,12348,12349]
  ACCEPTED attempt posted: [12350,12351,12352,12353]     <- starts at 12350, not 12345
```

So the gate is right, with a correction worth recording: in the **default** mode a refused attempt breaks ComfyUI's loop before `afterQueued` and nothing advances. In **`before`** mode `beforeQueued` fires per item regardless of acceptance, and the accepted run no longer starts from the seed the user can see. Attribution correctly reports `false` in both — the note does not lie — but **the mutation already happened**, and that is the part that matters.

#### Finding 2 — concurrent partial queue. Real, and it is the residual I explicitly waved through.

I wrote, in this thread: *"what it can no longer do is claim such a run… the failure direction is say less, warn more."* That was wrong in the direction that counts. For a **mutation**, the failure direction is not the reporting — attribution cannot undo a seed the user's own partial-execution run already advanced. The gate refused to accept a documented P1 as a residual, and it was right to.

---

### The pattern, which is the actual finding

Three rounds, three P1s, one class:

| round | what it mutated that it did not own |
|---|---|
| 1 | nodes outside the run's scope (space) |
| 2 | wrapper outliving its dispatch; another run's calls counted as ours (time) |
| 3 | discarded retry attempts, and a concurrent partial queue (time) |

Each fix narrowed the blast radius. **None removed the seam**, because the seam is the design: wrapping shared frontend widget hooks for an async window this code does not own. Every remaining hole is a variation on "something else ran through those hooks", and I do not think a fourth patch closes the class — the widget hooks carry no ownership token, so at that seam our batch and the user's are genuinely indistinguishable.

### The design that would close it, and why I am not landing it unilaterally

Stop mutating shared state at all. `dispatchScopedRun` **already** intercepts every `POST /prompt`, already identifies ours by queue mark + content hash, and already repairs the body. The seed could be written into **our own posted body only** — computed by ComfyUI's own `computeNextControlledValue`, so it is still not panel arithmetic. Nothing global is touched, so retries, concurrent queues and off-scope nodes all become non-issues by construction.

**But it changes the product answer**, and that is why it is not mine to make: the canvas widget would no longer advance, so consecutive scoped batches restart from the same base value — identical sets for `increment`, fresh ones for `randomize`. Arguably *more* faithful to ComfyUI_frontend #8774 (a partial execution should not churn your seed) while still giving `batch_count: N` its N distinct renders. It is also a rewrite of the mechanism, not an edit.

### Where this leaves #1998

**Not merged. Not parked by me** — the existing `autopilot:parked` is another agent's dispatcher marker and I have not touched it. The branch stays open, green and mergeable at `c9b6b4d5`; every round's evidence is in this thread so a fourth run does not re-derive it.

The open question is the same one the park named, now with a concrete second option:

1. Ship the override at the hook seam, accepting that a concurrent or `before`-mode retry can move a seed that was not ours; or
2. Rewrite it at the POST-body seam — no shared mutation, at the cost of the canvas seed no longer advancing between scoped batches; or
3. Drop the override and keep #988's warning, which is what `main` does today.

I have deliberately not chosen. Verified facts for whoever does: unscoped batches were never affected; `fixed` repeats correctly under every version of this; and the original report's own path is still unconfirmed as scoped.

Issue #1001, opened 2026-08-10
A saved workflow loses its identity across a reconnect, and workflow_open's post-load compare false-alarms while the frontend is still normalising


artokun:
Split out of artokun/comfyui-mcp#1331. The **remedy** half is fixed MCP-side (#1346, released): a mutation refused for `no trusted identity` now measures whether the live canvas actually has one, and names `panel_open_workflow` when it does not. That makes the symptom survivable. **It does not fix the cause, which is here.**

## Bug 1 — identity is not re-derived from the saved path after a reconnect

The reporter's workflow was created in-session: `panel_new_workflow` → `panel_paste_nodes` → `panel_rename_workflow` → `panel_save_workflow`. So it had a **real saved path**. After a ComfyUI drop/reconnect cycle (the panel injected its "connection dropped … session resumed" notice), the canvas was on the correct workflow and the panel said so — yet every mutating command was refused because the workflow had no trusted identity to fence against. Reads still worked.

A workflow that has been saved has a stable path, and opening from that path is what mints an identity — the reporter's own recovery was `panel_open_workflow` on the workflow that was **already active**. So the identity is derivable at the moment it goes missing; nothing re-derives it.

## Bug 1b — `workflow_open`'s post-load compare races the frontend's normalisation

The recovery itself then produced a **false alarm**:

```
workflow_open RAN and the canvas IS bound to niji_holo_impasto — that much was proven —
but the graph on it does not match the state that was loaded … Specifically: the graph on
the canvas differs from what was loaded on: nodes.
```

…on a load that was in fact fine. The very next `panel_graph_outline` then hit a third, different error (`workflow instance mismatch`), and a second identical `panel_open_workflow` succeeded cleanly.

Full recovery sequence: mutate → refused → open (false mismatch) → outline (instance mismatch) → open again → OK. **Four calls to reach a state the panel already believed it was in.**

The compare appears to read the graph before the frontend has committed its own normalisation (node sizes, widget values), so it reports a difference that settles a moment later. Same family as the identity oscillation in artokun/comfyui-mcp#1330, where a canvas identity flipped and settled back mid-build.

## Environment

Windows 11; ComfyUI 0.30.0; frontend 1.48.6; comfyui-mcp 0.50.93; panel 0.11.38.

artokun:
[...]

PR #1578, opened 2026-08-21
fix(1283): a benign load-time rewrite INSIDE a subgraph definition stops refusing the open

### LINKED ISSUE #1283: panel_open_workflow reports error after successful normalized load
panel_open_workflow successfully opened and bound the requested saved workflow, but returned isError:true because frontend-normalized per-node order, size, or widgets_values did not exactly match the disk payload. The reply itself says every loaded node is present with the same id/type and nothing extra appeared, yet the operation is surfaced as an error.

### Steps to reproduce
1. Save-As a live workflow with panel_save_workflow.
2. Call panel_open_workflow on the new saved path.
3. The canvas switches successfully and a subsequent panel_graph_outline shows the expected 47 nodes and 6 groups.
4. panel_open_workflow nevertheless returns an Error/UNCONFIRMED result and no workflow_uuid.

### Exact error (scrubbed)
workflow_open RAN, the canvas IS bound to the requested workflow, and every node that was loaded is on it with the same id and type, but per-node order/size/widgets_values differ; content reported UNCONFIRMED.

### Fix
applied locally: no
upstream-only: yes
Treat frontend normalization-only differences as a successful open with a warning and publish the active workflow_uuid/fence. Reserve isError for missing/extra nodes, wrong workflow identity, or material widget/link differences. Alternatively normalize both payloads before comparison.

### Environment
Windows 11
ComfyUI 0.33.1
ComfyUI frontend 1.48.7
Python 3.13.12
PyTorch 2.12.1+cu130
GPU NVIDIA GeForce RTX 4080 16 GB
comfyui-mcp 0.51.17
panel 0.14.16

artokun:
Closes #1283.

**Review is PENDING.** Nothing here has been reviewed by anyone but its author. The evidence below is mine.

## What was reported, and why the last fix did not end it

#1283 has been reported five times and fixed twice (#1274, then #1358 + #1387 + #1477). The 2026-08-21 comment reports it again on panel **0.15.25** — a release that already carries every one of those fixes. So this is not a stale reporter; something the earlier grounds cannot reach was still refusing.

## Reproduced, on the current release, before writing any code

ComfyUI 0.33.2 / frontend 1.49.6, real browser, `origin/main` at v0.15.32 served to the page. The reporter's configuration: an already-open **unsaved `tmp:` tab**, reopened via `workflow_open`.

* 8 plain nodes, no subgraph → `ok: true`, `workflow_uuid` published, `content_normalized: ["properties"]`. **Does not reproduce.**
* the same tab with two nodes converted into a **subgraph** → `ok: false`, no `workflow_uuid`, refused on `nodes, definitions`, with the panel's own reply saying *"every node that was loaded IS on the canvas with the same id and type and nothing extra appeared, so no node was lost"*.

That is the recurrence, and the subgraph is what the earlier reports did not isolate. (LTX-2.5, which the reporter names, is a subgraph template.)

### 1. `deepEqual` counted a key whose value is `undefined` as a difference

The definition fields that "differed" were `inputNode`, `outputNode` and `inputs` — whose JSON is byte-identical. Dumped from the browser, what actually differed was keys the LIVE serializer emits with the value `undefined`:

```
inputNode / outputNode   pinned
inputs[]                 localized_name, label, dir, shape, color_off, color_on
```

`JSON.stringify` drops such a key. The payload side of this comparison is `JSON.parse(JSON.stringify(state))` and the file side is a parsed `.json`, so **neither side can ever carry one**. This is the identical hazard `classifyNodeDifference` already documents and handles for root nodes (#1001, `showAdvanced: undefined`) — `definitions-renumber.js` never got the rule.

The consequence is bigger than this ticket: **#886's and #1706's renumber accounts were inert on this frontend.** Every subgraph workflow's `definitions` surface went unaccounted no matter what was true of it. `undefined` is now absent; `null` stays significant, deliberately, because JSON carries null and present-as-null is a real state of a saved file.

### 2. A per-node rewrite inside a definition had no account at all

With that repaired the surface still refused, on a genuine difference. An installed pack (cg-use-everywhere) stamps its version onto every node it sees during `configure`:

```
payload  "ue_properties": { "widget_ue_connectable": {}, "input_ue_unconnectable": {} }
live     "ue_properties": { "widget_ue_connectable": {}, "input_ue_unconnectable": {}, "version": "7.8" }
```

It does exactly this to the **root** nodes too — and there it is already accounted for, by #1358's completed-load ground, which admits any *named* per-node difference once the panel has WATCHED the restore run to completion. The same rewrite, on the same nodes, one level down had nothing.

A renumber account cannot grow to cover it. It is field-level by construction, and admitting `properties` would license a renumbering to rewrite arbitrary node content — the treadmill #1358's own header describes. So `definitionsDifferOnlyByCompletedLoadNormalization` asks #1358's question one level down: *did this restore stop early?* Measured this run: `LGraphNode.prototype.configure` entered 10×, `LGraph.prototype.configure` 2×, **zero throws** on either.

It requires, and each fails closed:

* `loadRanToCompletion === true`, strictly — `null` ("nobody watched") and `false` ("something threw") both refuse;
* the blocks pair, and every top-level key but `subgraphs` is deep-equal;
* inside each definition, **every field but `nodes`** is deep-equal — `links`, `state`, `widgets`, `inputs`, `outputs`, `groups`, `config`, `inputNode`, `outputNode`, `name`, `id`, `extra`. No rewire, no promoted-widget change, no counter movement, no definition added or dropped;
* inside `nodes`, the same length and the same `id` **and** `type` at each position.

That last one closes #1706's cross-surface hazard **structurally** rather than by a guard: a definition node id can never move here, so `patchProxyWidgets` cannot have run, so a root node's `properties.proxyWidgets` cannot point at anything this account relabeled.

It is applied to `normalizedOnly` only, **not** to `presentationOnly`. `presentationOnly` claims "nothing AUTHORED was lost"; this admits an arbitrary per-node field inside a definition, which is a claim only the weaker, observation-licensed ground may make.

## What the reply now says

`ok: true`, `workflow_uuid` published, and **both** disclosures:

* `content_normalized: ["properties"]` — the root fields that moved,
* `definitions_unverified: true` — #1477's existing key, now riding the success path, because an open carried partly on the subgraph account must not be silent about it. It is set only when the account actually **carried** the verdict, never when it merely would have applied.

## Evidence

**Browser, mutation-verified one change at a time** (the reproduction above, re-run after each edit):

| state | result |
| --- | --- |
| main | `ok: false` |
| both fixes | `ok: true` + uuid + both disclosures |
| revert the `deepEqual` rule only | `ok: false` |
| disable the definitions ground only | `ok: false` |

Neither half fixes it alone.

**Unit**, `browser_tests/unit/open-definitions-normalized.test.mjs`, 13 tests. They drive the real `graphRootReproducesStateContent`, not only the predicates, because the predicate is inert unless the verdict subtracts the surface. Mutation-verified:

| mutation | fails |
| --- | --- |
| revert the `deepEqual` `undefined` rule | 3 named tests, including one on `definitionsDifferOnlyByRenumber` |
| `definitionsNormalized = false && …` | the WIRING test |
| report an account that merely applied | the USED-not-applied test |
| `!options?.loadRanToCompletion` instead of `!== true` | the OBSERVATION test |

**Suites.** `npm run test:unit`: 6177/6179, the one failure being the known `#671 verifyInstalled` wall-clock flake, which passes 125/125 when its file is run alone. Playwright `--workers=1`: **still running as this body is posted** — the result is reported in a comment below, pass or fail. It is run with the panel sources served **from this branch** (the repo's own `worktreeSource` route), not from ComfyUI's `custom_nodes` junction, which points at a checkout 64 commits behind `main` — a run without that override would report green for code it never loaded.

* **Did not widen `RECOMPUTED_NODE_FIELDS` or `COSMETIC_NODE_FIELDS`.** Both are field-name allowlists and `ue_properties` is precisely the field a pack invents next.
* **Did not touch `presentationOnly`.** See above.
* **Did not extend the account to a definitions difference that is BOTH a renumber and a per-node rewrite.** Each account handles its own case; a load that did both would still refuse. No reporter has produced that shape, and combining them means admitting a relabeling and an arbitrary field change at once, which is where #1706's hazard lives.
* **Did not chase the `stale: "unknown"` on the reply.** It is honest for an unsaved `tmp:` tab and unrelated.

## Where I am least sure — please aim here

1. **The `deepEqual` change is the widest blast radius in this PR.** It is shared by every path in `definitions-renumber.js`, including #886's and #1706's shipped accounts, and it makes them *more* permissive. It reaches one consumer outside the open: `graphRootAgreesWithActiveState` (#1477) reads `proof.proven`, whose `definitions` arm is the renumber account — so the content-only binding seal is granted in cases it previously refused. I believe that is the same false refusal being repaired rather than a widening, because the keys in question carry nothing, but it is a *binding* decision and deserves its own look. My argument that it cannot admit anything real: a saved workflow is JSON, so a payload key can never hold `undefined`; a key the payload has with a real value and the live side lacks still differs; `null` vs `undefined` still differs. If there is a way for the LIVE side to lose something and have that read as `undefined`, this is where it gets through.
2. **"Everything but `nodes` deep-equal" may be the wrong cut.** I chose it because it makes the account structural rather than field-level, but it means a benign rewrite anywhere else in a definition still refuses — i.e. this ticket may come back a sixth time with a different field. Is the right shape instead "the definition's node SET is intact and the load completed", with links/state compared through the renumber account?
3. **The `definitions_unverified` reuse.** I reused #1477's key rather than minting one, so the note is written for a *definitions-only* difference and now also appears alongside `content_normalized`. Does that pair read coherently, or does the reply now say two things about one graph?
4. **Positional pairing of definition nodes.** A reordered node array inside a definition refuses. That matches `nodesDifferOnlyInLinkRefs`, but if the frontend ever reorders them the account goes inert exactly the way cause (1) did.

### DIFF
--- a/browser_tests/unit/definitions-renumber.test.mjs
+++ b/browser_tests/unit/definitions-renumber.test.mjs
@@ -123,7 +123,14 @@ test("#886 WIRING: the content proof consults it for a definitions surface", () 
   // The predicate is inert unless the proof calls it, and the behavioural tests
   // above cannot see the call site.
   const src = readFileSync(join(ROOT, "web/js/lib/graph-binding.js"), "utf8");
-  assert.match(src, /import \{ definitionsDifferOnlyByRenumber \} from "\.\/definitions-renumber\.js";/);
+  // panel#1283 (the 2026-08-21 recurrence) added a SECOND import from this module, so the
+  // statement became a braced multi-line one. Pin what the assertion is FOR — that
+  // graph-binding pulls THIS symbol from THAT module — rather than one spelling of the
+  // statement, which a formatter or a sibling import re-breaks while the wiring is fine.
+  assert.match(
+    src,
+    /import\s*\{[^}]*\bdefinitionsDifferOnlyByRenumber\b[^}]*\}\s*from\s*"\.\/definitions-renumber\.js";/,
+  );
   // comfyui-mcp#1706 — BOTH call sites must pass the payload's ROOT nodes. The node-id
   // account is granted only to a caller that supplies them (a call without the argument
   // answers the pre-#1706 question and refuses), so a call site that dropped it would
--- a/browser_tests/unit/open-content-proof.test.mjs
+++ b/browser_tests/unit/open-content-proof.test.mjs
@@ -104,6 +104,10 @@ test("#1001 a byte-identical repaint is still reported as exact", () => {
     // consult would misattribute the proof.
     normalizedOnly: false,
     normalizedFields: [],
+    // panel#1283 (the 2026-08-21 recurrence) — nor the DEFINITIONS arm of that ground.
+    // Same rule: an account the proof never consulted must not be reported as one it
+    // used, or the reply discloses a subgraph difference that did not decide anything.
+    definitionsNormalized: false,
   });
 });
 
@@ -244,6 +248,7 @@ test("#1001 an unreadable root proves nothing — absence of comparison is not e
       // unreadable root would prove an open off a comparison that never happened.
       normalizedOnly: false,
       normalizedFields: [],
+      definitionsNormalized: false,
     });
   }
   assert.equal(graphRootReproducesStateContent({ rootGraph: rootOf(state), state: null }).proven, false);
--- a/browser_tests/unit/open-definitions-normalized.test.mjs
+++ b/browser_tests/unit/open-definitions-normalized.test.mjs
@@ -1,0 +1,328 @@
+// panel#1283, the 2026-08-21 recurrence — `panel_open_workflow` on an already-open
+// unsaved `tmp:` tab was refused again, on a release (v0.15.32) that already carried
+// every fix the earlier reports produced (#1358's completed-load ground, #1387's
+// entered guard, #1477's definitions-only path).
+//
+// REPRODUCED IN A REAL BROWSER, ComfyUI 0.33.2 / frontend 1.49.6, the reporter's own
+// configuration: claim a fresh canvas, add eight nodes, convert two of them into a
+// subgraph, reopen the tab that is already active. The open bound, every root node
+// came back with the same id and type, nothing extra appeared — `ok: false`, no
+// `workflow_uuid`, refused on `nodes, definitions`.
+//
+// TWO independent causes, and BOTH are load-bearing: with either one reverted the
+// browser reproduction goes back to `ok: false` (measured, one at a time).
+//
+// -- 1. `undefined` counted as a difference -----------------------------------------
+// The definition fields that "differed" were `inputNode`, `outputNode` and `inputs`,
+// whose JSON is byte-identical. What differed was keys the LIVE serializer emits with
+// the value `undefined` and `JSON.stringify` drops:
+//
+//     inputNode / outputNode   pinned
+//     inputs[]                 localized_name, label, dir, shape, color_off, color_on
+//
+// The payload side of this comparison is `JSON.parse(JSON.stringify(state))`, and the
+// file side is a parsed `.json`, so NEITHER can ever carry such a key. Counting them
+// made #886's and #1706's renumber accounts inert on this frontend — every subgraph
+// workflow's `definitions` went unaccounted, whatever else was true of it.
+//
+// -- 2. no account for a per-node rewrite inside a definition ------------------------
+// With that repaired the surface still refused, on a real difference: an installed
+// pack (cg-use-everywhere) stamps `properties.ue_properties.version` onto every node
+// during `configure`.
+//
+//     payload  "ue_properties": { "widget_ue_connectable": {}, "input_ue_unconnectable": {} }
+//     live     "ue_properties": { "widget_ue_connectable": {}, "input_ue_unconnectable": {},
+//                                 "version": "7.8" }
+//
+// It does the same to the ROOT nodes — and THERE it is already accounted for, by
+// #1358's completed-load ground, which admits any NAMED per-node difference once the
+// panel has watched the restore run to completion. The identical rewrite, on the same
+// nodes, one level down had no account at all.
+//
+// A renumber account cannot grow to cover it: it is field-level by construction, and
+// admitting `properties` would license a renumbering to rewrite arbitrary node
+// content. So the question is asked at the level #1358 moved it to — did this restore
+// stop early? — and the answer is an observation of the load, not a judgement about a
+// field name.
+
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import {
+  definitionsDifferOnlyByCompletedLoadNormalization,
+  definitionsDifferOnlyByRenumber,
+} from "../../web/js/lib/definitions-renumber.js";
+import { graphRootReproducesStateContent } from "../../web/js/lib/graph-binding.js";
+
+/** The definition the browser produced, payload side. */
+const payloadDef = (over = {}) => ({
+  id: "57bb0646-920b-4076-8215-2477fab7de9d",
+  version: 1,
+  state: { lastGroupId: 0, lastNodeId: 9, lastLinkId: 2, lastRerouteId: 0 },
+  revision: 0,
+  config: {},
+  name: "New Subgraph",
+  inputNode: { id: -10, bounding: [-178, 71, 128, 48] },
+  outputNode: { id: -20, bounding: [470, 71, 128, 48] },
+  inputs: [{ id: "8874cc1f", name: "text", type: "STRING", linkIds: [1], pos: [0, 0] }],
+  outputs: [],
+  widgets: [],
+  nodes: [
+    {
+      id: 2,
+      type: "CLIPTextEncode",
+      pos: [10, 10],
+      size: [400, 200],
+      order: 0,
+      mode: 0,
+      properties: {
+        cnr_id: "comfy-core",
+        ver: "0.33.2",
+        ue_properties: { widget_ue_connectable: {}, input_ue_unconnectable: {} },
+        "Node name for S&R": "CLIPTextEncode",
+      },
+      widgets_values: [""],
+    },
+  ],
+  links: [{ id: 1, origin_id: -10, origin_slot: 0, target_id: 2, target_slot: 1, type: "STRING" }],
+  groups: [],
+  extra: {},
+  ...over,
+});
+
+/** The same definition as the LIVE serializer emits it after the load. */
+const liveDef = (over = {}) => {
+  const d = payloadDef();
+  // The measured `undefined` keys. Written explicitly because that is the point: a
+  // `JSON.parse(JSON.stringify(...))` fixture would erase the very thing under test.
+  d.inputNode = { id: -10, bounding: [-178, 71, 128, 48], pinned: undefined };
+  d.outputNode = { id: -20, bounding: [470, 71, 128, 48], pinned: undefined };
+  d.inputs = [
+    {
+      id: "8874cc1f",
+      name: "text",
+      type: "STRING",
+      linkIds: [1],
+      localized_name: undefined,
+      label: undefined,
+      dir: undefined,
+      shape: undefined,
+      color_off: undefined,
+      color_on: undefined,
+      pos: [0, 0],
+    },
+  ];
+  // ...and the pack's version stamp.
+  d.nodes = [
+    {
+      ...d.nodes[0],
+      properties: {
+        ...d.nodes[0].properties,
+        ue_properties: { widget_ue_connectable: {}, input_ue_unconnectable: {}, version: "7.8" },
+      },
+    },
+  ];
+  return { ...d, ...over };
+};
+
+const payload = (over) => ({ subgraphs: [payloadDef(over)] });
+const live = (over) => ({ subgraphs: [liveDef(over)] });
+const completed = { loadRanToCompletion: true };
+
+// -- the `undefined` rule, on the ACCOUNT THAT ALREADY SHIPPED -----------------------
+//
+// This is the half that is invisible from the new predicate: #886's renumber account
+// is what a plain link-renumbering open depends on, and the `undefined` keys made it
+// answer false on any frontend that emits them. Pinned on
+// `definitionsDifferOnlyByRenumber` so a revert of the `deepEqual` rule fails HERE,
+// not only on the new ground.
+
+test("#1283 the renumber account survives the keys the live serializer emits as `undefined`", () => {
+  const before = { subgraphs: [payloadDef()] };
+  const after = {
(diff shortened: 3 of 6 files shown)


artokun:
@grok please review this. It relaxes two guards on `panel_open_workflow`'s content verdict, so I would rather you attacked it than skimmed it. Four specific places, in the order I would look:

**1. `deepEqual` now treats a key whose value is `undefined` as ABSENT** (`web/js/lib/definitions-renumber.js`). This is the widest change here — the function is shared by #886's and #1706's already-shipped renumber accounts, and it makes them accept comparisons they used to refuse.

My argument that nothing real gets through: a saved workflow is JSON, and `JSON.stringify` drops an `undefined`-valued key, so no payload side can carry one; a key the payload has with a real value and the live side lacks still differs (the counts diverge); `null` vs `undefined` still differs. **Try to break that.** Concretely: is there any way for the LIVE serializer to LOSE something and have the loss present as `undefined` rather than as a changed or missing value? If there is, this fix hides real data loss behind a "nothing changed" answer, which is the exact failure mode the whole #1111/#1089 cluster exists to prevent.

It also reaches one consumer outside the open path: `graphRootAgreesWithActiveState` (#1477) reads `proof.proven`, whose `definitions` arm is that renumber account. So a content-only BINDING seal is now granted in cases it previously refused. I think that is a repaired false refusal, not a widening — but it is a binding decision and I am not confident enough about it to leave it unnamed.

**2. `definitionsDifferOnlyByCompletedLoadNormalization` admits an ARBITRARY per-node field difference inside a subgraph definition** on the strength of one observation: the panel watched `LGraphNode.prototype.configure` and `LGraph.prototype.configure` and neither threw. Everything else in the definition (`links`, `state`, `widgets`, `inputs`, `outputs`, `groups`, `config`, `inputNode`, `outputNode`, `name`, `id`, `extra`) must be deep-equal, and the node ids and types are pinned positionally.

The load-bearing claim, inherited from #1358, is that **a throw is the only way a partial restore can present** — and I am extending it from the ROOT graph to subgraph definitions. Is that extension sound? `Subgraph` extends `LGraph` and its inner nodes are `LGraphNode`s, so I claim both wraps cover them. If there is a path where a definition's nodes are populated WITHOUT going through either wrapped `configure` — a `Subgraph.configure` override that does not call `super`, an async widget build, a definition hydrated from a cache — then `loadRanToCompletion === true` is true and vouches for nothing, and this ground says "nothing was dropped" over a definition that lost values.

**3. The state I am least sure of, and could not reach.** A definitions difference that is BOTH a renumbering AND a per-node rewrite satisfies neither account, so it still refuses. I could not produce that shape on the rig, so it is untested in the direction that matters — and it is exactly what the NEXT report of #1283 would look like. Is refusing it right, or is the two-separate-accounts structure itself the bug?

**4. Does the reply now contradict itself?** A success carries `content_normalized: ["properties"]` **and** `definitions_unverified: true`. That second key and its note were written for #1477's *definitions-only* case. Read the two notes together as a caller would: do they say one coherent thing about one graph, or two different-sized things? "The reply saying two different-sized things about the same fields" is the contradiction this whole ticket was originally filed for, and I may have re-created it one surface over.

Reproduction, mutation results, and what I deliberately did not do are in the PR body. The browser suite result is posted as a separate comment. Please do not take the mutation table as proof of correctness — it proves the tests bind to the change, not that the change is right.

artokun:
Following up on point 2 of my review request with the actual source, because I asked a reviewer to attack a claim I had only measured — and half of it turns out to be measured, half asserted. Both halves stated plainly.

**MEASURED**, this reproduction run, counters installed underneath the panel's own wrappers on both prototypes:

```
LGraphNode.prototype.configure   entered 10x   threw 0
LGraph.prototype.configure       entered  2x   threw 0
```

Ten node-configures for a graph with 7 root nodes and 2 nodes inside the definition (plus the subgraph's IO nodes), so **the definition's inner nodes do resolve `configure` through the wrapped prototype**. Two graph-configures for one root graph and one subgraph.

**CONFIRMED IN THE SHIPPED FRONTEND** (`comfyui_frontend_package` 1.49.6, `settingStore-CwkLtSKP.js`, `class Subgraph extends` LGraph):

```js
configure(e, t) { let n = super.configure(e, t); return this._configureSubgraph(e), n }
```

`super.configure` IS the wrapped `LGraph.prototype.configure`, which is why the count is 2 and not 1. So a throw out of the subgraph's own restore is observed.

**THE GAP I CANNOT CLOSE FROM HERE, and it is the one worth aiming at.** `_configureSubgraph(e)` runs *after* `super.configure` returns, inside `Subgraph.configure` — which is Subgraph's **own** method, not the wrapped prototype one. A throw there escapes into whatever called `Subgraph.configure`. If that caller is the root's wrapped `LGraph.configure`, the outer wrapper records it and my ground correctly refuses. If any path hydrates a definition by calling `Subgraph.configure` from outside a wrapped `LGraph.configure`, then a throw in `_configureSubgraph` is invisible, `loadRanToCompletion` still says `true`, and this PR's new ground would report "nothing was dropped" over a definition whose IO slots, promoted widgets or link-id repair (`_repairIOSlotLinkIds`, right there in `_configureSubgraph`) partly failed.

I did not find such a path, but I did not prove there isn't one, and "I looked and did not find it" is not the same claim as "it cannot happen". That is the single assumption this fix rests on that I would most like a second pair of eyes on.

Two mitigations already in the diff, for what they are worth: every field of the definition **except** `nodes` must be deep-equal, so a partly-failed `_repairIOSlotLinkIds` shows up as a `links`/`inputs`/`outputs` difference and refuses; and the definition node ids and types are pinned positionally. The residual exposure is a throw that damages only per-node content inside the definition and nothing else.

artokun:
More evidence on point 1 (the `deepEqual` `undefined` rule), since it is the widest change here and "I reasoned about it" is not evidence.

**The safety property.** If the predicate answers *equal*, the two values must be indistinguishable **as a saved file can express them** — i.e. identical once placed in a container and `JSON.stringify`'d, which is exactly where an `undefined`-valued key disappears. If that ever fails, this change hides a real difference.

Checked over 200,000 generated pairs (nested objects/arrays up to depth 3, leaves drawn from `0 1 -1 "" "a" "b" true false null undefined 1.5 "1" "null"`, half the pairs a JSON round-trip of the other so the equal case is actually exercised):

```
pairs: 200000 | predicate said equal: 98596 | safety violations: 0
```

The 98,596 matters as much as the 0 — a check that never answers *equal* would report zero violations while proving nothing.

And the three cases the fix actually turns on, spelled out rather than left to the generator:

```
undefined vs absent  -> true    (both are absent; this is the bug being fixed)
null     vs absent   -> false   (JSON carries null; still a real difference)
value    vs absent   -> false   (a lost field is still a lost field)
```

The first pass of this check was wrong and is worth recording so nobody trusts it twice: it normalised a bare `undefined` to `null` before stringifying, which makes `undefined` and `null` compare identical — precisely the distinction under test. It would have reported 0 violations no matter what the code did. The numbers above are from the corrected form, which stringifies the value inside a container.

**What this does NOT cover:** it is a property of the comparison, not of the frontend. If the live serializer can lose something and emit `undefined` where a real value used to be, the comparison is behaving correctly and the *conclusion drawn from it* is still wrong. That is the question in my earlier comment, and it is not one a fuzz can answer.

artokun:
**Review status, stated plainly rather than left to look settled.**

* `@grok` review requested 2026-08-21 17:21Z. **No review has arrived** as of 21:58Z — 4h37m. Checked via the reviews API, not just the comment thread.
* Copilot posted a quota-limit message at 16:53Z. That is not a review.
* I have **not** run `codex-gate.mjs`. Codex is currently pointed at a weak model on this machine, so a SHIP from it would carry the authority of an independent adversarial gate without being one.

So I am running the repo's documented fallback, `.claude/autopilot/claude-gate.mjs` — a FRESH session that has never seen this implementation, given only the diff and told to refute. Disclosing it is the condition the fallback was accepted under, so: **this PR's adversarial review is the fallback gate, not codex and not grok.** Its verdict goes in a comment below whatever it says, including NO-SHIP or INDETERMINATE.

It is weaker than codex in one specific way that matters here: codex reliably *executes* code against hostile inputs, and two of my four open questions (the `Subgraph._configureSubgraph` escape path, and the renumber-plus-rewrite shape I could not produce) are exactly the kind that reading a diff will not settle.

The branch has also been **rebased onto `776576e2`** — `main` moved 5 commits while this sat. None of them touch `graph-binding.js`, `definitions-renumber.js`, or the two test files this changes; two touch `comfyui-mcp-panel.js` in unrelated regions. The rebase was clean and the 84 affected unit tests still pass on the new base.

The Playwright suite was restarted against the rebased head and is running now.

@grok — the review request two comments up still stands if you get to it. I would rather hold for it than merge on the fallback alone.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `6480f1ea4080394e274c56bd5ee16895b60e4d17` vs origin/main. Read `deepEqual` / `definitionsDifferOnlyByCompletedLoadNormalization` in `web/js/lib/definitions-renumber.js`, and `graphRootReproducesStateContent` in `web/js/lib/graph-binding.js` — not just the patch.

Two independent causes, both measured, both fail-closed in the right direction (false refusal, not silent wrong-graph writes):

**1. `undefined` is absent in `deepEqual`.** Live `serialize()` emits `pinned` / `localized_name` / `label` / `dir` / `shape` / `color_*` as `undefined`; JSON drops them. Counting those keys made #886/#1706 inert on this frontend. `null` stays significant (#1001). Array holes are not collapsed. That restores the renumber accounts rather than widening them.

**2. Completed-load ground one level down.** A pack stamping `ue_properties.version` during `configure` is already admitted on ROOT nodes by `openContentDifferenceIsCompletedLoadNormalization`. The same rewrite inside a definition had no account, so a faithful subgraph open refused on `nodes, definitions` with no `workflow_uuid`. The new predicate requires `loadRanToCompletion === true` strictly, pairable definition blocks, every definition field except `nodes` deep-equal, and per-node same id + same type at the same index. Rewired links, relabeled ids, retyped nodes, added/dropped definitions still refuse.

Wiring is the part I attacked:

- Strict `proven` still does not reach a properties difference (test asserts `proven === false`). The open succeeds on `normalizedOnly`, which is how the root completed-load ground already works (`contentProof.proven || presentationOnly || normalizedOnly`).
- `definitionsNormalized` is only set when that account actually carried the verdict. A lost root link with an otherwise-accountable definitions diff refuses and does not advertise `definitions_unverified`.
- Unwatched / thrown / missing watch does not license anything.

Residual, same as #1358 on the root: this admits any per-node field inside a definition once the restore was watched to completion, and discloses rather than claiming the file's values survived (`definitions_unverified`). That is the weaker ground by design. Do not fold it into `presentationOnly` ("nothing authored was lost") — the code already refuses that fold.

Issue #1493, opened 2026-08-20
panel_get_errors times out after refreshing and setting promoted subgraph model widgets


artokun:
### What happened / root cause
After panel_refresh_nodes and three successful panel_set_widget calls targeting promoted widgets on subgraph node 5004, panel_save_workflow succeeded, but the immediate panel_get_errors call timed out after 20 seconds. The workflow tab was active and had responded to all preceding mutations. Likely graph_get_errors can wedge or exceed its deadline while rescanning refreshed model assets/subgraphs.

### Steps to reproduce
1. Open a workflow containing a model-loader subgraph with promoted COMBO widgets.
2. Call panel_refresh_nodes.
3. Re-set several promoted model widgets to their current values.
4. Save the workflow.
5. Immediately call panel_get_errors.

### Exact error (scrubbed)
Panel tab workflows/LTX2.5_Character_ReferenceVideo_Pose_Control.json did not reply to graph_get_errors within 20000 ms - the ComfyUI tab may be backgrounded or frozen.

### Fix
applied locally: no
upstream-only: yes
Precise change needed: make graph_get_errors non-blocking or raise/budget its asset scan separately; return partial/stale scan state instead of timing out the entire panel call.

### Environment
Windows 11; ComfyUI 0.33.1; RTX 5090 32 GB; comfyui-mcp 0.52.25; panel 0.14.31; frontend version unavailable in this session.

artokun:
Claimed 2026-08-20T14:22Z — dispatched to a fix agent, worktree `wt-1493` (branch `fix/1493-get-errors-timeout-after-refresh`). PR will be posted here. If this comment is older than 8h with no PR, treat the claim as lapsed and pick it up.

artokun:
PR up: artokun/comfyui-mcp#1867 (the fix is orchestrator-side, so it lands in the mcp repo).

Short version of what was going on: the tab was never frozen, and your prescribed fix is already in the product — the asset scan **is** budgeted separately and **does** return partial state. What was wrong is the headroom that budget was sized against.

[...]

Issue #1280, opened 2026-08-15
Add a setting to disable video previews to conserve ram


artokun:
Currently we push videos from the outputs to the panel chat window which can take up user RAM. Default to ON but allow users to opt out of seeing video instead leave a placeholder stating the filename, number of frames and show the first frame only. clicking on it still allows you to view the video in the lightbox viewer.

artokun:
Backlog audit — **parked under the stabilisation freeze, not dropped.**

This is a good idea and the shape is already right in the description: default ON so nothing changes for anyone who doesn't opt out, and the opt-out degrades to a placeholder that still carries the useful facts (filename, frame count, first frame) with the lightbox as the escape hatch. That is a genuinely better design than a plain on/off, because the video is still *reachable* — the RAM cost just stops being paid up front.

It is parked because it **adds a setting and a new render path**, which is new capability rather than a defect: today's behaviour is what was intended, it is simply expensive on long or numerous outputs. The freeze is bugs and stability only, so this is recorded rather than built.

Worth noting for whoever picks it up: #1270 on this tracker is an open question about how the inline player handles a long video, so the two should be looked at together — if a preview is already mis-rendering duration, the placeholder path is a partial answer to that as well.

Labelling `autopilot:parked` so it stops reading as an untriaged item in the bug queue.

artokun:
Claimed by kimi-code swarm (worktree fix-1280). Agent assigned now; if no PR appears within ~8h the claim has lapsed.

Issue #648, opened 2026-08-05
panel_show_media cannot preview local reference videos over 20 MB


artokun:
### What happened / root cause
`panel_show_media` rejects a local 72.1 MB reference MP4 with `Error: file too large (72.1 MB > 20 MB)`. The panel has no compact preview path (poster, storyboard, trim, or transcode), so an agent cannot inspect a normal local reference video through the panel.

### Steps to reproduce
1. Load a local MP4 larger than 20 MB in ComfyUI.
2. Call `panel_show_media` with its absolute local path.
3. Observe the size-limit error instead of a preview.

### Exact error (scrubbed)
`Error: file too large (72.1 MB > 20 MB): <ComfyUI>/input/<reference>.mp4`

### Fix
Applied locally: no. Upstream-only: yes.
Add a bounded video-preview fallback for oversized local files, such as a generated poster/contact sheet or a short low-bitrate clip, rather than failing outright.

### Environment
Windows 11; ComfyUI 0.30.0 local; NVIDIA RTX 5090 32 GB; comfyui-mcp 0.49.5; panel 0.11.38.

artokun:
**Status: half fixed, and the half that closes your exact repro is not this one.** Being explicit because a partial fix reported as complete is worse than an honest partial.

**What panel #649 fixed.** Video now routes through the panel's existing storyboard pipeline instead of dead-ending, and every preview states plainly that the video was **not** sent, that the sheet is N *sampled* frames (with `do not describe the video as N frames long` said outright), the source size or an explicit UNKNOWN, and `get_image` on the temp ref as the way to actually look at it. A **partial** sheet withholds the "evenly-spaced across the video" claim, because the surviving frames may be clustered.

It also fixed a worse bug found on the way: **under** the size ceiling, the panel painted a player for the *user* and returned `{ok:true,count:1}` to an agent that had been shown nothing — a success acknowledgement for something the caller cannot see. And the video test was `/.(mp4|webm)$/` with an unescaped dot, so `xmp4` was painted as a video.

**What still fails — your exact repro.** The literal `file too large (72.1 MB > 20 MB)` is raised **orchestrator-side**, at `src/orchestrator/panel-tools.ts:6762-6766` in `panel_show_media`'s absolute-path branch, **before the panel is ever called**. So a 72 MB local MP4 by absolute path is still refused today.

[...]

PR #1579, opened 2026-08-21
fix(1485): the video completion stops waiting on work the agent never sees

### LINKED ISSUE #1485: Fix dropped render completion events after panel_run
Every panel_run in this session completed in ComfyUI, but the panel never delivered its completion event. About 45-46 seconds later the orchestrator synthesized a terminal-status notice from ComfyUI history/execution state. The synthesized notice contained no output media, forcing manual filesystem inspection to verify every image. No matching ComfyUI runtime error was present. This points to a dropped or unacknowledged panel completion event, with the orchestrator fallback recovering status but not outputs.

### Steps to reproduce
1. Open a local ComfyUI workflow in the sidebar panel.
2. Queue it with panel_run.
3. Let the render finish successfully.
4. Observe that no normal completion/media event arrives; after about 45 seconds a synthesized notice appears saying the panel never delivered its completion event and that output files are not attached.
5. Repeat: this occurred on prompts c5479916-77fa-47a9-a836-efe6be94c52d, af2867ce-bbdb-4f6e-a5e4-57fd033608a0, and 25694039-986d-4d01-afb7-abd373855197.

### Exact error (scrubbed)
The render finished SUCCESSFULLY, but the panel never delivered its completion event, so this notice was synthesised by the orchestrator from ComfyUI itself. The output file(s) are NOT attached to this notice.

### Fix
Applied locally: no. Upstream-only: yes. A safe local patch could not be identified without the long-lived orchestrator/panel event-delivery trace. Please instrument the panel completion-event send/ack path and make the fallback attach outputs resolved from the completed prompt history, so a dropped frontend event still yields the same media payload as a normal completion.

### Environment
Windows 11; ComfyUI 0.33.1; ComfyUI frontend 1.48.7; Python 3.12.10; PyTorch 2.8.0+cu129; NVIDIA GeForce RTX 5070 12 GB; comfyui-mcp 0.50.84; panel 0.11.45.

artokun:
Fixes #1485.

## What was reported

A `panel_run` that renders an .mp4 (`to_node_id=128`) finished successfully and the panel's completion event never arrived. ~5 s later the orchestrator synthesised a success notice from ComfyUI history — **with no output attached** — so the agent had to recover the file with `list_outputs` + `panel_show_media`. The first filing of this issue saw the same shape at ~45 s, on three prompts.

The issue was closed on 2026-08-20 as already-fixed by #1403/#1739 (a re-pended completion re-arming the reconcile sweep, v0.15.5). It then **recurred on panel 0.15.25**, which contains that fix. So it is not the #1739 loss window.

## What it actually is

Two things that were each individually reasonable and are wrong together.

**1. The orchestrator's grace was cut to 5 s on an assumption that does not hold for video.** `DEFAULT_SYNTHESIS_GRACE_MS` in `comfyui-mcp` went from `45_000` to `5_000`, and the comment justifying it says:

> The normal path lands within a second or two; that is the only race the grace has to win.

For a still, true. For a **video**, the run's single completion frame is not sent until the panel has: sampled the clip in a hidden `<video>`, encoded a contact sheet, uploaded it, encoded a full-resolution poster, uploaded **that**, and taken a HEAD of the video. Measured on this rig (Playwright Chromium, headless *and* headed, ComfyUI on localhost, a real 4.84 s and a 19.36 s 960x544 h264 mp4):

| stage | 4.84 s clip | notes |
|---|---|---|
| `loadedmetadata` | 9-25 ms | |
| sample 20 frames | 1 292-3 220 ms | each seek decodes from the preceding keyframe |
| sheet to PNG | 24 / 33 / **6 711** ms | 1.71 MB; the stall reproduced once in three runs |
| sheet upload | ~7-11 ms | localhost only; 1.71 MB on the wire |
| video size HEAD | 1-2 ms | localhost only; bounded at 8 s in `fetchImageBytes` |
| poster to PNG | 11 / 15 / **1 018** ms | 718 KB |
| poster upload | 3-5 ms | localhost only; 718 KB on the wire |
| **end to end** | **1 352 / 3 296 / 9 706 ms** | three consecutive runs |

Bounded at 25 s by design (`videoStoryboardTimeoutMs`). "A second or two" is not the shape of this path.

**2. When the orchestrator wins that race, its notice cannot carry the video.** comfyui-mcp#1861 made `partitionAttachableMedia` fail **closed**: an `.mp4` from `/history` is *named, never attached*, precisely because the panel's own video ref is normally an uploaded `storyboard_<base>.png`. So for a video run the panel's storyboard is the **only** thing that ever shows the agent the render — and it is the thing arriving late.

That is the reported symptom exactly: successful run, no panel completion, a synthesised notice ~5 s later with no output attached.

## What this PR changes

Everything on that path that is not needed to tell the agent the run finished comes off it, plus two claims the same function was making that were not true.

**The poster is detached.** It decorates the **user's** video card (#1280); the agent never receives it and no part of the note depends on it. Awaiting it put a full-resolution PNG encode plus a second `POST /upload/image` between the run finishing and the agent being told. The card is back-filled *by URL* — that is why it is a back-fill and not an argument — so it lands exactly as it does today, just not in front of the completion.

**The video's size HEAD is issued before the sampling pass**, not after the sheet upload, so its round trip overlaps the decode. On a remote target (a pod) that is a whole network round trip off the critical path, for one metadata line. Called directly rather than behind a `Promise.resolve().then(...)` hop, because a microtask hop lets sampling start first — the test asserts the ordering.

**`storyboardFailure({reason})` is handled.** It is **truthy**, so `if (!blob)` sailed straight past it and handed a plain object to `uploadBlobToInput`, which appends it to a `FormData` as the string `"[object Object]"` and uploads *that* as `storyboard_<name>.png`. The agent was then shown a "20-frame storyboard" that is not an image and asked to review its motion and sharpness. Success is now recognised **positively**, on the same test `produceSheet` already uses (a sheet is the thing with a numeric `size`), and the reason #1493 built is said out loud instead of discarded a second time.

The builder's own comment said *"`produceSheet` is the only consumer and does exactly that."* It was not the only consumer. `buildVideoSegment` is the second, and it was the one that did not check. The comment now counts its callers instead of remembering them.

**The note describes the frames that PAINTED, not the grid's capacity.** #648 put the real count on the blob and told callers to use it; this caller used `storyboardFrameCount()`, so a sheet holding one frame and nineteen blank cells was announced to the agent as twenty samples. A degraded sheet now discloses how many cells are blank and that the survivors may be clustered; a full sheet reads exactly as before.

## The existing suite was pinning the first defect

`makeFrameDeps`' storyboard double was `{ fake: "blob" }` — truthy, **not** a Blob, i.e. the failure shape — and four tests asserted that it got uploaded and described as a sheet. The double now models a Blob (`{ size, paintedFrames }`); every assertion around it is unchanged. No assertion was loosened.

## Evidence

- `browser_tests/unit/run-completion-video-segment.test.mjs` — 7 tests driving the **real** `composeRunCompletionFrame` with injected deps, not source-regex pins.
- **Mutation-verified.** Each fix reverted in turn; the test that names it went red:
  - revert the `size`/`reason` test: *"a storyboard FAILURE object is never uploaded as if it were a sheet"* fails
  - revert `paintedFrames`: *"the note describes the frames actually PAINTED"* fails
  - re-`await` the poster: *"the completion frame does not wait for the card's poster"* fails **by hanging to the 25 s segment timeout** (run under `--test-timeout 20000`, or it hangs silently)
  - re-inline the HEAD: both HEAD tests fail
- Full unit suite: **6172 pass, 0 fail** (`node --test browser_tests/unit/*.test.mjs`), plus i18n, i18n-render, tool-vocabulary and panel-scope gates all OK.

## Browser gate — RUN, against this branch's exact bytes

**Correction to an earlier revision of this body, which said the suite could not be run.** That was true of the route I first tried and false in general, and the difference matters, so both are recorded:

- The shared ComfyUI on `:8188` resolves the `custom_nodes` **junction at startup**. I re-pointed it at this worktree and the server still served the main checkout's file (28 317 bytes, no `1485` marker) — verified over HTTP with a cache-buster. Junction restored to its original target. That route needs a ComfyUI restart, which would break another agent's run mid-flight.
- The route that works: the `ap-1` slot on `:8211` mounts the pack as a real **git worktree**, so `git checkout --detach <sha>` swaps the served files with no restart. Confirmed the served file was this branch's exact bytes before running (34 747 bytes, `1485` marker present) and that it reverted when checked back to main. The slot was restored to its original HEAD (`adf23a05`, clean tree) afterwards.

| run | code under test | result |
|---|---|---|
| full suite, 81 specs | this branch (`f720c377`) | **63 passed, 18 failed** (9.3 m) |
| the 7 affected spec files, 21 tests | `origin/main` (`869362ec`), same slot | **3 passed, 18 failed** (5.1 m) |

The two failure sets are **identical** — same 18 test IDs, `comm -23` between them is empty. Nothing fails on this branch that does not already fail on `main`. All 18 are MockBridge/session specs (`conversation-persistence`, `chat-history-v2`, `workflow-chat-identity`, `streaming-render`, `hidden-tab`, `pending-queue`, `external-orchestrator`) and none of them touch the completion composer. A separate `main` baseline captured earlier today by another session lists the same set plus one (`agent-drive.spec.ts:113:7`), which passed in my run.

Worth stating: the suite has **no coverage of this path** — zero of the 81 specs mention `storyboard`, `composeRunCompletionFrame`, `paintVideo` or `.mp4`. So this run is a no-regression result, not evidence for the fix. The evidence for the fix is the unit tests and their mutations.

The latency measurements in the table above are from real Chromium (Playwright, headless *and* headed) against a real ComfyUI output file — they measure the pipeline, not the panel's wiring.

## What this deliberately does NOT do

- **It does not guarantee the panel wins a 5 s grace.** Sampling alone is ~1-2 s before any of the above. This removes an encode, an upload and a round trip from the critical path and stops two false claims; it does not make a browser-side video decode fast.
- **It does not touch the sampling loop.** A single forward decode pass (`requestVideoFrameCallback` at `playbackRate` 16) measured 1.22 s vs 1.76-1.96 s of seeking on the 19.36 s clip, and 0.37 s vs 1.05-1.76 s on the 4.84 s clip — a real but bimodal 1.4x-3x win that also fills only 18-19 of 20 cells at that rate and depends on a background tab presenting frames. That is a rewrite of the DOM closure justified only by a variable measurement, so it is written down here rather than shipped on the back of this fix.
- **It does not change the sheet's codec.** WebP encoded the same canvas in 57-68 ms at 147 KB vs PNG's 1.71 MB — but with renderer throttling disabled PNG was usually 24-35 ms, so the win is an intermittent stall, not a reliable cost. Not shipping a wire-format change on a bimodal measurement.
- **It does not touch the orchestrator.** The other half of this is `comfyui-mcp`'s: either the grace accounts for a video completion, or the synthesised notice attaches something viewable for an `.mp4`. Written up on the issue.

## Review

**grok was asked first** (PR comment above, naming the risky parts) and did not answer within ~1 h 20 m. Rather than stall, I ran the repo's own adversarial gate.

### `claude-gate.mjs` — verdict: **SHIP**

**SUBSTITUTED REVIEW — disclose this on the PR** (the gate prints that line itself; this is that disclosure). It is not codex and not grok. It spawns a fresh session that has never seen the implementation, hands it the diff plus the repo's `review-taxonomy.md`, and tells it to refute and to RUN the code. It reads more and executes less than codex would. It is a fallback, not an equal.

What it did:

- Independently re-counted the builder's consumers (`media-preview.js:875`, `run-completion-frame.js:578`) and confirmed the corrected comment.
- **Ran 5 mutations, all killed**: bare `!blob`, capacity-for-painted, re-awaited poster, inlined HEAD, always-plural.
- **A/B'd `HEAD~2` vs `HEAD` over 25 hostile inputs.** Every divergence is old-worse: the old code uploaded `[]`, `{}`, `"oops"` and `{reason:{...}}` as `[object Object]` PNGs, and dropped the entire sheet whenever `fetchImageBytes` threw, rejected, or was absent. **Nothing regressed.** No unhandled rejections in any case, including a rejecting detached poster upload and a throwing `applyVideoPoster`.
- Confirmed the detached poster keys `_videoPosters` by video URL and prunes `!holder.isConnected`, and that the late-resolve case back-fills the correct URL pair — my question 1.
- Confirmed `produceSheet` uses the identical numeric-`size` test, and that `fetchImageBytes` really is bounded at 8 s — my questions 2 and 4.
- Confirmed production reachability at `comfyui-mcp-panel.js:34529-34539` against a real `Blob`.

**It also found two prose defects and filed them as non-findings** — by reading the *generated string* rather than the branch:

1. at `frames === cells - 1` the note read "the other 1 **are** BLANK";
2. the degraded head ran into the filename as "...spread across the video **of the FINAL saved video** (file clip.mp4 ...)".

Both were grammar and both still stated the correct fact, so the gate was right not to block on them. **I fixed them anyway** (`8e99ff1c`): this PR exists to stop the completion note making garbled or false claims to the agent, so shipping a garbled one would be the wrong lesson. The head now names the sheet and flows into the file; the caveat is its own sentence. Two further tests pin the boundaries the plural rule and the clamp actually get wrong (`cells - 1`, and a `paintedFrames` above capacity).

The gate was then **re-run on the rebased head**, so the verdict names the SHA that merges rather than an earlier one.

### DIFF
--- a/browser_tests/unit/run-completion-video-segment.test.mjs
+++ b/browser_tests/unit/run-completion-video-segment.test.mjs
@@ -126,8 +126,10 @@ test("#1485: the note describes the frames actually PAINTED, not the grid's capa
     !/20-frame storyboard/.test(frame.note),
     `a 3-frame sheet must not be announced as 20 frames\n${frame.note}`,
   );
-  assert.match(frame.note, /3 sampled frames/, "the count that was drawn is the count that is stated");
-  assert.match(frame.note, /17 could not be captured and are BLANK/, "and the blanks are disclosed, not implied");
+  assert.match(frame.note, /Only 3 of its 20 cells hold a sampled frame/, "the count that was drawn is the count that is stated");
+  assert.match(frame.note, /the other 17 are BLANK/, "and the blanks are disclosed, not implied");
+  assert.match(frame.note, /📽️ 20-cell storyboard \(contact sheet\) of the FINAL saved video/,
+    "the head names the sheet and flows into the file — the caveat is its own sentence");
 });
 
 test("#1485: a FULL sheet still reads as a plain N-frame storyboard", async () => {
@@ -152,9 +154,35 @@ test("#1485: exactly ONE sampled frame is described in the singular, with the bl
   const { deps } = makeDeps({ buildVideoStoryboard: async () => sheetBlob({ paintedFrames: 1 }) });
   const frame = await composeRunCompletionFrame(videoPayload(), deps);
 
-  assert.match(frame.note, /holding 1 sampled frame —/, "one frame is one frame, not '1 sampled frames'");
-  assert.match(frame.note, /19 could not be captured and are BLANK/);
+  assert.match(frame.note, /Only 1 of its 20 cells hold/, "one frame is one frame");
+  assert.match(frame.note, /the other 19 are BLANK, so judge nothing from them/);
   assert.ok(!/20-frame storyboard/.test(frame.note), `\n${frame.note}`);
+});
+
+test("#1485: cells-minus-one reads 'is BLANK', not 'are BLANK' (gate non-finding)", async () => {
+  // The boundary the plural rule actually gets wrong: 19 of 20 painted leaves
+  // exactly one blank cell. Caught by the review gate reading the generated
+  // string rather than the branch, which is why it is pinned by the string.
+  const { deps } = makeDeps({ buildVideoStoryboard: async () => sheetBlob({ paintedFrames: 19 }) });
+  const frame = await composeRunCompletionFrame(videoPayload(), deps);
+
+  assert.match(frame.note, /the other 1 is BLANK, so judge nothing from it/, `
+${frame.note}`);
+  assert.ok(!/the other 1 are BLANK/.test(frame.note), "one blank cell is 'is', not 'are'");
+});
+
+test("#1485: a count over capacity is CLAMPED, never announced as more than the grid holds", async () => {
+  // paintedFrames is an expando the builder sets; a caller must not be able to
+  // claim 25 samples from a 20-cell sheet. Clamped to the capacity, which then
+  // reads as a full sheet with no blanks — the honest description of 20 cells
+  // that all hold a frame.
+  const { deps } = makeDeps({ buildVideoStoryboard: async () => sheetBlob({ paintedFrames: 25 }) });
+  const frame = await composeRunCompletionFrame(videoPayload(), deps);
+
+  assert.ok(!/25/.test(frame.note), `a count above the capacity must never be printed
+${frame.note}`);
+  assert.match(frame.note, /📽️ 20-frame storyboard \(contact sheet\)/);
+  assert.ok(!/BLANK/.test(frame.note), "a clamped-full sheet has no blanks to disclose");
 });
 
 test("#1485: an UNKNOWN sampled count claims no count at all — never the capacity, never 'null'", async () => {
--- a/browser_tests/unit/run-completion-video-segment.test.mjs
+++ b/browser_tests/unit/run-completion-video-segment.test.mjs
@@ -145,6 +145,46 @@ test("#1485: a FULL sheet still reads as a plain N-frame storyboard", async () =
   );
 });
 
+test("#1485: exactly ONE sampled frame is described in the singular, with the blanks named", async () => {
+  // The reported shape of a barely-decodable video: the builder paints one cell
+  // and leaves nineteen blank. This is the case the capacity claim was worst
+  // for, so the wording is pinned rather than assumed to fall out of the plural.
+  const { deps } = makeDeps({ buildVideoStoryboard: async () => sheetBlob({ paintedFrames: 1 }) });
+  const frame = await composeRunCompletionFrame(videoPayload(), deps);
+
+  assert.match(frame.note, /holding 1 sampled frame —/, "one frame is one frame, not '1 sampled frames'");
+  assert.match(frame.note, /19 could not be captured and are BLANK/);
+  assert.ok(!/20-frame storyboard/.test(frame.note), `\n${frame.note}`);
+});
+
+test("#1485: an UNKNOWN sampled count claims no count at all — never the capacity, never 'null'", async () => {
+  // `paintedFrames` is best-effort: buildVideoStoryboard sets it on the blob and
+  // its own comment allows a Blob implementation that refuses extra properties,
+  // "which callers must treat as unknown — never as the capacity". Unlike
+  // show_media's produceSheet, this path does NOT withhold the sheet for an
+  // unknown count — it is the run's ONE completion and the sheet is the only
+  // thing that shows the agent the video — so it must describe it without one.
+  //
+  // The second half is the trap: `metaSuffix`'s storyboard clause is a plain
+  // truthiness guard, so passing an unknown count through as a number-shaped
+  // nothing would print "null-frame storyboard" on the metadata line.
+  const { deps, calls } = makeDeps({ buildVideoStoryboard: async () => sheetBlob({}) });
+  const frame = await composeRunCompletionFrame(videoPayload(), deps);
+
+  assert.equal(frame.images.length, 1, "the sheet still ships — an unknown count is not a reason to withhold it");
+  assert.match(frame.note, /📽️ storyboard \(contact sheet\)/, "described, but not counted");
+  assert.ok(!/20-frame|20-cell/.test(frame.note), `the capacity must not stand in for the count\n${frame.note}`);
+  assert.ok(
+    !/null|undefined|NaN/.test(frame.note),
+    `an unknown count must never reach the note as a value\n${frame.note}`,
+  );
+  assert.deepEqual(
+    calls.painted.map((p) => p.caption),
+    ["Storyboard"],
+    "and the user's caption drops the count too, rather than inventing one",
+  );
+});
+
 test("#1485: the completion frame does not wait for the card's poster", async () => {
   // THE LATENCY FIX. The poster decorates the USER's video card; the agent never
   // receives it and no part of the note depends on it. It used to be awaited —
--- a/browser_tests/unit/run-completion-video-segment.test.mjs
+++ b/browser_tests/unit/run-completion-video-segment.test.mjs
@@ -1,0 +1,252 @@
+// panel#1485 — the video completion the panel PROMISED, and what it spends
+// getting there.
+//
+// THE REPORT. `panel_run` on a graph that writes an .mp4 finished successfully
+// and the panel's completion event never showed up; ~5 s later the orchestrator
+// synthesised a success notice of its own, and that notice carried no output —
+// the agent had to go and find the file with list_outputs and panel_show_media.
+// The first filing of this issue saw the same thing at ~45 s.
+//
+// WHY 45 s BECAME 5 s, AND WHY THAT MATTERS HERE. The orchestrator's
+// `DEFAULT_SYNTHESIS_GRACE_MS` was cut from 45 000 to 5 000, on a stated
+// assumption: "the normal path lands within a second or two; that is the only
+// race the grace has to win." For a STILL that is true. For a VIDEO the panel's
+// single completion frame is not sent until it has sampled the clip in a hidden
+// <video>, encoded a contact sheet, uploaded it, encoded a poster, uploaded THAT
+// and taken a HEAD of the video — measured on this repo's rig (headless and
+// headed Chromium, a 4.84 s and a 19.36 s 960×544 h264 mp4, ComfyUI on
+// localhost): 1.0–9.7 s end to end, and bounded at 25 s by design. And when the
+// orchestrator wins that race, the notice it synthesises cannot carry the video
+// at all — an .mp4 is deliberately named-but-not-attached (comfyui-mcp#1861), so
+// the storyboard is the ONLY thing that ever shows the agent this render.
+//
+// So the panel's half of the fix is: take off that path everything that is not
+// needed to tell the agent the run finished. The tests below pin the three
+// things that came off it, plus two claims the same function was making that
+// were not true.
+//
+// These drive the REAL composeRunCompletionFrame with injected deps rather than
+// asserting on source text: every one of them fails when its fix is reverted
+// (verified by reverting each in turn), which a source-regex pin cannot promise.
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { composeRunCompletionFrame } from "../../web/js/lib/run-completion-frame.js";
+
+/** A stand-in for the sheet Blob: what every consumer actually tests for is a
+ *  numeric `size`, so that is what the double carries. */
+function sheetBlob({ size = 4096, paintedFrames, posterBlob } = {}) {
+  const b = { size };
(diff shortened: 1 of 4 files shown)


artokun:
@grok please review this one — it changes agent-facing text and the ordering of an async pipeline, and I would rather you go straight at the risky parts than hunt for them.

**Where I think the defect would be, hardest first:**

1. **The detached poster** (`web/js/lib/run-completion-frame.js`, the `void Promise.resolve().then(...)` block). The poster upload now escapes `produce()`, which is itself wrapped in `withTimeout(..., 25_000)`. I convinced myself the `.catch` covers the rejection and that `applyVideoPoster` back-filling by URL is idempotent — but I did not chase what happens when the card for that URL was unmounted, or when a second render writes the SAME filename and both posters land. If a stale poster can be painted onto a newer card, that is mine.

2. **The positive success test** — `typeof produced.size === "number" && !named`. I lifted it from `produceSheet` in `media-preview.js` on purpose so both consumers apply one rule. The cost is that a real `Blob` carrying its own string `reason` reads as a failure. I claim `buildVideoStoryboard` never produces that; I would like that claim attacked rather than accepted.

3. **The frame-count wording.** `paintedFrames` now drives the note. Please check `frames === 1` reads correctly, and that `metaSuffix(sizeStr, null)` really drops the clause instead of emitting `null-frame storyboard` — that is a plain truthiness guard and it is exactly the kind of thing I would get wrong.

4. **The HEAD reorder.** `fetchImageBytes` is now issued before the decode with its own `.catch`. Its rejection used to reach the segment's catch and cost the agent the whole storyboard; now it costs only the size line. I think that is strictly better; say so or don't.

**What I am least confident about, and it is not in the diff:** whether this matters. Sampling twenty frames is ~1-2 s before any of the work I removed, so this shortens the race without winning it. The PR body says that plainly rather than implying a fix. If you think the honest verdict is "correct but insufficient, and the real fix is orchestrator-side", say that — I would rather ship it labelled that way than have it read as closing #1485 on its own.

**What I did not run:** the Playwright suite, against this branch. The shared ComfyUI resolves the `custom_nodes` junction at startup, so re-pointing it served the main checkout's file anyway (verified over HTTP, junction restored). Separately, no spec in the suite touches `storyboard`, `composeRunCompletionFrame`, `paintVideo` or `.mp4`. So the only execution evidence here is the 7 mutation-verified unit tests and the 6172-test suite.

artokun:
## Review disclosure — `claude-gate.mjs`, **SHIP**, on the rebased head

**grok was asked first** (comment above, naming the four risky parts) and did not answer. Copilot auto-declined on quota. `codex-gate.mjs` was **not** run: the autopilot brief bars it because codex is currently pointed at a weak model, so a SHIP from it would wear the authority of a gate without being one.

So this is a **SUBSTITUTED REVIEW** — the gate prints that line itself and makes disclosure the condition of using it, and this is that disclosure. It is not codex and not grok. It spawns a fresh session that has never seen the implementation, hands it only the diff plus the repo's `review-taxonomy.md`, and tells it to refute and to RUN the code. It reads more and executes less than codex would. **It is a fallback, not an equal, and I am not calling this an independent human-grade review.**

It ran twice: once on the substantive diff (**SHIP**), then again on the head carrying the prose fix (**SHIP**). Second run, in its own words:

- **The detached poster — my own question 1, and the thing it said was "the only thing I could break".** It noted the poster escapes the 25 s `withTimeout`, then ran poster-rejects, poster-throws-synchronously, and poster-never-settles: **0 unhandled rejections**, frame sent immediately with the sheet intact, `calls.posters` empty until it lands. `m` is captured before the detach and never reassigned; `applyVideoPoster` keys `_videoPosters` by video URL and prunes disconnected holders.
- **Mutations, all killed**: `blob = produced` → 1 fail; `frames = cells` → 4 fails; re-awaited poster → 1 fail; inlined `fetchImageBytes` → 2 fails.
- **Boundaries executed**: `paintedFrames` absent / `0` / `-3` / `"5"` / `Infinity` all degrade to `"storyboard (contact sheet)"` with caption `"Storyboard"` — never `null`/`undefined`/`NaN` in the note, sheet still ships. 19/20 → "the other 1 **is** BLANK"; 25/20 → clamped, no `25` anywhere, no BLANK; 20/20 → no caveat.
- **Both prose claims checked**: `grep` finds exactly two `buildVideoStoryboard` invocation sites, and `fetchImageBytes` really is `AbortController` + 8000 ms and never rejects.
- **Reachability**: both panel call sites (`33766`, `34723`) forward the deps through to `buildVideoSegment`. No new module binding, so no extraction-harness scope break.

**Its one non-finding, recorded rather than buried:** `cells` is used unguarded, so `storyboardFrameCount: () => NaN` would yield `"NaN-frame storyboard"`. It declined to file it because `storyboardFrameCount()` is `Math.max(1, COLS*ROWS)` over two literals — no production input reaches it. I agree, and I have not added a guard for a state that cannot occur.

### What changed after the first SHIP

The first run filed two **non-findings** — grammar only, both still stating the correct fact: `"the other 1 are BLANK"` at `frames === cells - 1`, and a degraded head that ran into the filename as `"…spread across the video of the FINAL saved video (file clip.mp4 …)"`. It was right not to block on them. I fixed them anyway, because this PR exists to stop the completion note making garbled or false claims to the agent. The second gate run covers that fix.

### Everything else on the final head (`185163ec`)

- Rebased onto `main` twice more (`#1567`, `#1568` landed mid-run; one touches `comfyui-mcp-panel.js`). The rebase was clean and the diffstat is byte-identical to what the gate reviewed: 4 files, +467/−20.
- Unit suite **6258 pass / 0 fail**. i18n, i18n-render, tool-vocabulary and panel-scope gates all OK. CI green.
- Playwright, this branch, full 81 specs: **63 passed / 18 failed**. Same 7 spec files re-run against `origin/main` in the same slot: **3 passed / 18 failed — the identical 18 test IDs**. Nothing fails here that does not already fail on `main`.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `185163ecb6db99daf42195141b40842e49d541dc` vs origin/main. Read `composeVideoStoryboardSegment` in `web/js/lib/run-completion-frame.js` — not just the patch.

The race is real: a 5 s synthesis grace vs a video pipeline that sampled, encoded a sheet, uploaded it, encoded a poster, uploaded that, then HEADed the file. When the orchestrator wins, the synthesised notice cannot carry the .mp4 (#1861), so the storyboard is the only agent-visible render and it arrives late or not at all.

What landed is the right cut:

- Poster is fire-and-forget (`void Promise.resolve().then(...).catch(...)`). The completion note does not await it. Tested: the frame returns while the poster upload is still pending, and the card still back-fills when it resolves. That is detach, not drop.
- Size HEAD is started before `buildVideoStoryboard`, so the round trip overlaps decode instead of sitting behind the sheet upload. The size line is still in the agent note, so awaiting the already-in-flight HEAD after the upload is justified.
- Success is recognised positively: numeric `size` and not a `reason` string. `storyboardFailure({reason})` is truthy and was being uploaded as `[object Object]`. The reason is now said.
- The note describes painted frames, not grid capacity, with unknown count described as uncounted rather than as 20, and the blank-cell caveat is its own sentence (so it cannot read "…of the FINAL saved video"). Singular/plural at cells-1 is pinned.

Does not, and does not claim to, beat a 5 s grace on every clip — sampling alone is ~1–2 s. Remaining half is the orchestrator's grace, which is on the issue and not this PR.

The old suite's `{ fake: "blob" }` double was pinning the #1493 defect; replacing it with a Blob-shaped double is load-bearing, not cosmetic.

Issue #1296, opened 2026-08-17
Frontend-only rgthree Fast Groups Bypasser rejected by panel_add_node


artokun:
The rgthree skill documents Fast Groups Bypasser (rgthree) as an explicitly allowed frontend-only type for panel_add_node, but panel_add_node rejected it after rgthree-comfy was installed and verified, returning: Unknown node type "Fast Groups Bypasser (rgthree)" — the ComfyUI backend does not provide it. This blocks adding the requested toggle for the VIDEO 2 group. Reproduction: install rgthree-comfy, call panel_add_node with class_type Fast Groups Bypasser (rgthree). Environment: Windows 11, ComfyUI 0.33.0, RTX 3090 24GB, comfyui-mcp 0.51.57, panel 0.14.41. The pack was installed and ComfyUI restarted before retrying.

Related (same root cause): #1284, #475

artokun:
Claimed by kimi-code swarm (worktree fix-1296). Agent assigned now; if no PR appears within ~24h the claim has lapsed.

artokun:
Claimed by the backlog-drain session (worktree `.claude/worktrees/drain-1296`). Verified unclaimed with no open PR immediately before claiming. A fix agent is assigned now; if no PR appears within ~8h, consider the claim lapsed.

artokun:
Independent verification of the merge (#1366), from a second agent that diagnosed this
issue in parallel and reached the same mechanism before seeing the PR. Recording it because
the merged comment asserts one claim it did not prove, and that claim is the load-bearing one.

**Claim: "no /object_info refresh can fix that."** Proven, not inferred:

* Shipped frontend 1.48.7 (`comfyui_frontend_package/static/assets/settingStore-*.js`):
  `async registerNodesFromDefs(e){ await invokeExtensionsAsync('addCustomNodeDefs', e); await
  Promise.all(Object.keys(e).map(t => this.registerNodeDef(t, e[t]))) }` — and
  `registerCustomNodes` is invoked from `registerNodes()` alone, *after* it.
* `rgthree-comfy/web/comfyui/fast_groups_bypasser.js` registers the class **only** from
  `registerCustomNodes()` → `FastGroupsBypasser.setUp()` → `LiteGraph.registerNodeType(...)`.
* The panel's own refresh (`registerComfyNodeDefs`) calls `app.registerNodesFromDefs(defs)`
  and nothing else; `registerCustomNodes` appears nowhere in the panel.

[...]

PR #1586, opened 2026-08-21
fix(mcp#2000): a frontend-only node stays addable when the object_info fetch does not answer

artokun:
Reported as artokun/comfyui-mcp issue 2000 (P2, via-panel). The refusal is thrown here, so the fix lands in this repo.

## What the reporter saw

On a **healthy live canvas**, `panel_refresh_nodes` timed out fetching `/object_info`. The very next `panel_add_node "MarkdownNote"` was refused with:

> cannot verify node type "MarkdownNote" against the ComfyUI backend (object_info is unavailable …). Refusing to add.

…telling them to reconnect a ComfyUI that was answering fine. `MarkdownNote` is a documented frontend-only exemption; its creation should never depend on backend node definitions.

## Mechanism

`assertAddNodeResolvableRefreshing` ran its `/object_info` **availability** guard before the frontend-only exemption:

```js
if (!freshDefs || typeof freshDefs !== "object") { throw … }   // ← every type, no exceptions
if (!hasOwnProperty(freshDefs, class_type)) {
  … ever-seen gate …
  if (isAuthorizedFrontendOnlyType(registry, class_type)) return;   // ← never reached
}
```

So a fetch that **did not answer** vetoed a type the fetch **could never have answered about**: `/object_info` never lists `Note` / `MarkdownNote` / `Reroute` / `PrimitiveNode` — that is what frontend-only means.

`web/js/lib/object-info-history.js` already states this exact rule for its own latch, and this branch was breaking it:

> ARM THIS ONLY ON EVIDENCE, NEVER ON A TIMEOUT. A timeout is the ABSENCE of evidence, not evidence … Latching on one would turn ordinary latency into a permanent, session-long false refusal of every legitimate add/write — the exact bug class (#496/#507) this module's consumers exist to fix.

## The fix

The exemption now also runs on the unavailable path, **on identical terms**. This is safe because **neither of its clauses reads `freshDefs`**:

- the #458 **ever-seen gate** reads the **session history oracle**, which a timeout leaves intact — `recordTypes(null)` records nothing, and a timeout must never arm `loseBaseline`. The non-forgeable trust root that catches a removed pack squatting a reserved name is still fully in force.
- `isAuthorizedFrontendOnlyType` reads the **live registry** (membership + reserved allowlist + no backend provenance), which is also exactly what `LG.createNode` needs — so an exempted add builds a **real** node and cannot mint the placeholder #458 exists to prevent.

Note this is **stricter than the issue prescribed**. The report asked to resolve the allowlist *before any* availability guard; doing that literally would hoist it above the ever-seen gate too. The gate stays first, exactly as on the fetched-defs path.

## Evidence

**Reproduced by execution, not by reading** — driving the shipped guard with the real `createObjectInfoHistory`, seeded at page load and then handed a timed-out refresh:

```
wasTypeEverDefined('MarkdownNote') = false      ← baseline SURVIVED the timeout
history.seeded = true
RESULT: REFUSED — cannot verify node type "MarkdownNote" …   (byte-identical to the report)
CONTROL (fetch OK): ADDED                        ← the exemption already worked; only the precondition blocked it
```

That measurement is the load-bearing one: **the history oracle survives the timeout**, so the trust root is available on this path.

**Mutation results** — each mutant scoped to *this* branch only (the same clause text appears in the sibling exemption; my first attempt patched neither and read as "survived"):

| mutant | result |
|---|---|
| remove the whole exemption | **2 killed** — both positive tests |
| drop the ever-seen comparison | **1 killed** — the fail-closed matrix |
| drop `isAuthorizedFrontendOnlyType` | **1 killed** — the fail-closed matrix |
| invert the ever-seen comparison | **3 killed** |
| drop `typeof wasTypeEverDefined === "function"` | **SURVIVED** → removed as dead code (2nd commit) |

That last one is why the second commit exists: `backendHistoryVerdict` already classifies an unwired oracle as `"no-oracle"`, which is not `"never-seen"`, so the clause could never decide anything. Every remaining clause is load-bearing.

**Fail-closed matrix**, all with `/object_info` unavailable — each case trips a *different* clause: a real backend type (`KSampler`), a non-allowlisted type, an allowlisted name carrying backend provenance, an allowlisted name absent from the live registry, an ever-seen (removed) type, `HISTORY_PENDING`, `HISTORY_UNSEEDED`, and a caller wiring no history oracle at all.

**Production call site verified by grep, not assumed** — `web/js/comfyui-mcp-panel.js` passes `getFreshObjectInfo` (which yields `null` via `recordObjectInfoTypes(whole === NODE_DEFS_NO_ANSWER ? null : whole)`) and `wasTypeEverDefined` bound to the real `objectInfoHistory`, in the same call. The branch I changed is reached with the exact inputs the repro used.

**Suites**: `npm run test:unit` → 6248/6250. The one failure is the known #671 manager-install wall-clock flake; that file alone re-runs 125/125. My diff touches `node-resolve.js` only.

## Review status: PENDING

Not reviewed yet. **Where I would look first:**

1. **The security judgement is the whole PR.** I overturned a deliberate, documented decision — the test `#496 add_node: object_info UNAVAILABLE still fails closed, even for a frontend-only type` existed precisely to pin the old behaviour, and I replaced it. My argument is that for an *authorized frontend-only* type a successful fetch adds **zero** information, so refusing on its absence is a false refusal rather than a guard. If that reasoning is wrong anywhere, this PR is wrong.
2. **The residual window I could not close.** If a backend pack were installed mid-session providing a real node under a reserved frontend-only name (`SetNode`, `GetNode`, an rgthree control name) *after* the last successful `/object_info`, the fetched-defs path would register the fresh def while this path adds the frontend-native class instead. I judged that benign — the node is real, registered, provenance-clean, and Ctrl+Z-undoable — and structurally impossible for the four core types. Push back if that is too generous.
3. **The unchanged refusal message.** On the unavailable path an allowlisted-but-*unregistered* type still gets "Reconnect ComfyUI and retry" rather than #1296's "RELOAD the ComfyUI tab" diagnosis. I left it alone to keep the diff minimal; it is arguably the same misdiagnosis class #1296 fixed.

- **The `panel_refresh_nodes` half of the report.** It asked for a dedicated/configurable `/object_info` timeout, or reusing a known-good defs snapshot while reporting the live refresh timed out. Both are new behaviour under the feature freeze, and neither is a defect: on a large install `/object_info` is ~5.4 MB and a bounded fetch honestly reporting a timeout is correct. The timeout was the *trigger*; the false refusal was the *bug*.
- ~~The sibling `graph_set_widget` guard~~ — **now INCLUDED; this bullet is superseded.** I measured it instead of leaving it to judgement: on `origin/main`, through the production `runSetWidget` body, the write REFUSED with the fetch unavailable and WROTE on the control. Fixing only `add_node` would have shipped a `MarkdownNote` the user cannot fill in, and all three `/object_info`-oracle guards carry the byte-identical early throw — so relaxing one copy alone is the #496 drift again. Mutation table in the scope-change comment below.
- **Browser verification.** Not run against this diff. The Playwright suite targets the panel **served** by the live ComfyUI at :8188, which is the installed build, not this worktree — and no isolated ComfyUI was allocated for this run, so installing over the shared rig would disturb other sessions. The coverage for this diff is the execution repro + unit tests + mutants above, not e2e.

Closes the underlying cause of artokun/comfyui-mcp issue 2000 (cross-repo, so it will be closed by hand after merge).

### DIFF
--- a/browser_tests/unit/node-resolve.test.mjs
+++ b/browser_tests/unit/node-resolve.test.mjs
@@ -1146,6 +1146,111 @@ test("mcp#2000 add_node: object_info UNAVAILABLE still fails closed for everythi
   );
 });
 
+// ---- mcp#2000 PARITY: the SAME relaxation in the two set_widget guards. All three
+//      /object_info-oracle guards had the identical `!freshDefs` early throw, so fixing
+//      only add_node would have shipped HALF the documented annotation path — the note
+//      appears and its text can never be written. Measured on origin/main before
+//      extending the fix: with the fetch unavailable the write REFUSED, and the control
+//      (same state, fetch answering) WROTE, so the probe was proving the product and not
+//      itself. One copy relaxed alone is the #496 drift again, hence all three. --------
+
+test("mcp#2000 set_widget: assertTypeAgainstFreshBackend exempts an authorized frontend-only type when object_info is unavailable", () => {
+  const reg = registryWithNatives();
+  const node = { id: 7, type: "MarkdownNote", constructor: reg["MarkdownNote"] };
+  assert.doesNotThrow(() =>
+    assertTypeAgainstFreshBackend(null, "MarkdownNote", 7, {
+      registry: reg,
+      node,
+      wasTypeEverDefined: () => false,
+    }),
+  );
+  // …and every other direction still fails closed on the SAME unavailable map.
+  for (const [why, type, opts] of [
+    ["a real backend type", "KSampler", { registry: reg, wasTypeEverDefined: () => false }],
+    ["a non-allowlisted type", "TotallyMadeUpNode", { registry: reg, wasTypeEverDefined: () => false }],
+    ["an ever-seen (removed) allowlisted name", "MarkdownNote", { registry: reg, node, wasTypeEverDefined: () => true }],
+    ["a pending baseline", "MarkdownNote", { registry: reg, node, wasTypeEverDefined: () => HISTORY_PENDING }],
+    ["an unseeded baseline", "MarkdownNote", { registry: reg, node, wasTypeEverDefined: () => HISTORY_UNSEEDED }],
+    ["no registry wired at all", "MarkdownNote", { node, wasTypeEverDefined: () => false }],
+    ["a provenance-bearing husk", "MarkdownNote", { registry: loadedRegistry(["MarkdownNote"]), wasTypeEverDefined: () => false }],
+  ]) {
+    assert.throws(
+      () => assertTypeAgainstFreshBackend(null, type, 7, opts),
+      /cannot verify|object_info is unavailable|no usable/i,
+      `mcp#2000: ${why} must still fail closed`,
+    );
+  }
+  // A stale backend INSTANCE under a bare native class of the same name is refused by
+  // the instance-provenance clause, which only this guard family has.
+  assert.throws(
+    () =>
+      assertTypeAgainstFreshBackend(null, "MarkdownNote", 7, {
+        registry: reg,
+        node: { id: 7, type: "MarkdownNote", constructor: { comfyClass: "MarkdownNote" } },
+        wasTypeEverDefined: () => false,
+      }),
+    /cannot verify|no usable/i,
+  );
+});
+
+test("mcp#2000 set_widget: assertMutatedNodeAuthorized keeps the same terms on the unavailable path", () => {
+  const reg = registryWithNatives();
+  const node = { id: 9, type: "MarkdownNote", constructor: reg["MarkdownNote"] };
+  assert.doesNotThrow(() => assertMutatedNodeAuthorized(null, reg, node, "target", () => false));
+  assert.throws(
+    () => assertMutatedNodeAuthorized(null, reg, node, "target", () => true),
+    /cannot verify|object_info is unavailable/i,
+    "an ever-seen removal still wins",
+  );
+  assert.throws(
+    () => assertMutatedNodeAuthorized(null, reg, { id: 9, type: "KSampler" }, "target", () => false),
+    /cannot verify|object_info is unavailable/i,
+    "a real backend type still needs the fetch",
+  );
+  assert.throws(
+    () => assertMutatedNodeAuthorized(null, undefined, node, "target", () => false),
+    /cannot verify|object_info is unavailable/i,
+    "no registry wired ⇒ no exemption",
+  );
+});
+
+test("mcp#2000 set_widget e2e: the documented annotation path COMPLETES through the production handler when object_info never answers", async () => {
+  // THE WHOLE POINT: add the note, then put the text in it. Driving runSetWidget — the
+  // body graph_set_widget delegates to — not a predicate in isolation.
+  const history = createObjectInfoHistory();
+  history.recordTypes({ KSampler: {}, SaveImage: {} });
+  history.markSeeded();
+  history.recordTypes(null); // the timed-out refresh
+
+  const ctor = function MarkdownNoteNative() {};
+  const reg = loadedRegistry();
+  reg["MarkdownNote"] = ctor;
+  const node = { id: 42, type: "MarkdownNote", widgets: [{ name: "text", type: "text", value: "" }], constructor: ctor };
+  const { set } = await runSetWidget(node, "text", "hello", {
+    registry: reg,
+    getFreshObjectInfo: async () => null, // the fetch never answers
+    wasTypeEverDefined: (t) => history.wasTypeEverDefined(t),
+    ...HOOKS,
+  });
+  assert.equal(set.value, "hello");
+  assert.equal(node.widgets[0].value, "hello", "the note is fillable, not just addable");
+
+  // CONTROL: a REAL backend node in the identical state still refuses — the relaxation
+  // is scoped to frontend-only types, not to "object_info is down".
+  const ksNode = { id: 43, type: "KSampler", widgets: [{ name: "steps", type: "INT", value: 1 }], constructor: reg["KSampler"] };
+  await assert.rejects(
+    () =>
+      runSetWidget(ksNode, "steps", 20, {
+        registry: reg,
+        getFreshObjectInfo: async () => null,
+        wasTypeEverDefined: (t) => history.wasTypeEverDefined(t),
+        ...HOOKS,
+      }),
+    /cannot verify|object_info is unavailable|no usable/i,
+  );
+  assert.equal(ksNode.widgets[0].value, 1, "the refused write left the node untouched");
+});
+
 // ---- #1296: an allowlisted FRONTEND-ONLY type that is NOT in the live registry is
 //      still refused (LiteGraph could only mint a placeholder — that part is
 //      correct), but the refusal must stop diagnosing it as "not installed / its
--- a/browser_tests/unit/node-resolve.test.mjs
+++ b/browser_tests/unit/node-resolve.test.mjs
@@ -1146,6 +1146,71 @@ test("mcp#2000 add_node: object_info UNAVAILABLE still fails closed for everythi
   );
 });
 
+test("mcp#2000 INVARIANT: relaxing the precondition does not change WHICH types are exempt", async () => {
+  // THE load-bearing invariant of the whole change, and the one a reviewer can check
+  // without re-deriving my reasoning: the exempt SET must be identical whether
+  // /object_info answered or not. The change moves a PRECONDITION, it does not widen the
+  // permission set. Exactly one verdict may differ between the two paths — a REAL backend
+  // type, for which the fetch genuinely is load-bearing.
+  //
+  // Iterating FRONTEND_ONLY_NODE_TYPES rather than a hand-written list is deliberate: a
+  // type added to the allowlist later is covered here automatically, on both paths.
+  const addVerdict = async (fresh, type, reg, ever) => {
+    try {
+      await assertAddNodeResolvableRefreshing(() => reg, type, {
+        getFreshObjectInfo: async () => fresh,
+        refresh: async () => {},
+        wasTypeEverDefined: ever,
+      });
+      return "ALLOW";
+    } catch {
+      return "REFUSE";
+    }
+  };
+  const swVerdict = (fresh, type, reg, ever) => {
+    try {
+      assertTypeAgainstFreshBackend(fresh, type, 1, {
+        registry: reg,
+        node: { id: 1, type, constructor: reg[type] },
+        wasTypeEverDefined: ever,
+      });
+      return "ALLOW";
+    } catch {
+      return "REFUSE";
+    }
+  };
+  const healthy = objectInfo(); // a live backend that never lists a frontend-only type
+  const never = () => false;
+
+  const cases = [];
+  for (const t of FRONTEND_ONLY_NODE_TYPES) {
+    cases.push([`${t} (clean, never-seen)`, t, loadedRegistry([], [t]), never, "ALLOW"]);
+  }
+  cases.push(["unknown type", "TotallyMadeUpNode", loadedRegistry(), never, "REFUSE"]);
+  cases.push(["ever-seen ⇒ removed", "MarkdownNote", registryWithNatives(), () => true, "REFUSE"]);
+  cases.push(["provenance husk", "MarkdownNote", loadedRegistry(["MarkdownNote"]), never, "REFUSE"]);
+  cases.push(["not in the live registry", "MarkdownNote", loadedRegistry(), never, "REFUSE"]);
+
+  for (const [label, type, reg, ever, expected] of cases) {
+    const withDefs = await addVerdict(healthy, type, reg, ever);
+    const without = await addVerdict(null, type, reg, ever);
+    assert.equal(withDefs, expected, `add/${label}: unexpected verdict WITH defs`);
+    assert.equal(without, withDefs, `add/${label}: the two paths disagree — the exempt set moved`);
+    const swWith = swVerdict(healthy, type, reg, ever);
+    const swWithout = swVerdict(null, type, reg, ever);
+    assert.equal(swWithout, swWith, `set_widget/${label}: the two paths disagree`);
+  }
+
+  // …and the ONE type whose verdict MUST differ: a real backend node genuinely needs the
+  // fetch, so an unanswered one still refuses. If this ever stops differing, the
+  // relaxation has leaked out of the frontend-only set and into live backend types.
+  const reg = loadedRegistry();
+  assert.equal(await addVerdict(objectInfo(["KSampler"]), "KSampler", reg, never), "ALLOW");
+  assert.equal(await addVerdict(null, "KSampler", reg, never), "REFUSE");
+  assert.equal(swVerdict(objectInfo(["KSampler"]), "KSampler", reg, never), "ALLOW");
+  assert.equal(swVerdict(null, "KSampler", reg, never), "REFUSE");
+});
+
 test("mcp#2000: an exemption that THROWS must not replace the refusal it was checking", async () => {
   // Found by running the review taxonomy against my OWN diff (class 5: what does the
   // change make WORSE?). Before mcp#2000 these guards threw their refusal WITHOUT
--- a/browser_tests/unit/node-resolve.test.mjs
+++ b/browser_tests/unit/node-resolve.test.mjs
@@ -1146,6 +1146,71 @@ test("mcp#2000 add_node: object_info UNAVAILABLE still fails closed for everythi
   );
 });
 
+test("mcp#2000: an exemption that THROWS must not replace the refusal it was checking", async () => {
+  // Found by running the review taxonomy against my OWN diff (class 5: what does the
+  // change make WORSE?). Before mcp#2000 these guards threw their refusal WITHOUT
+  // consulting anything, so nothing on this path could raise. Consulting two predicates
+  // first meant a hostile registry or a raising oracle surfaced a RAW error instead of
+  // the worded refusal — measured leaking "registry exploded" / "history oracle exploded"
+  // from all three guards before the shared helper swallowed it. Any doubt REFUSES.
+  const ctor = function MarkdownNoteNative() {};
+  const reg = registryWithNatives();
+  const node = { id: 1, type: "MarkdownNote", constructor: reg["MarkdownNote"] };
+  const BOOM = () => {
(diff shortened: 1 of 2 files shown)


artokun:
@grok please review this one adversarially. It is small but it **overturns a deliberate security decision**, so the diff is not where the risk is — the argument is.

**The load-bearing claim, stated so you can attack it directly:**

> For a type that passes `isAuthorizedFrontendOnlyType` **and** whose session-history verdict is `never-seen`, a successful `/object_info` fetch would have contributed **zero** information — because `/object_info` never lists frontend-only types by design. Therefore refusing on the fetch's *absence* is a false refusal, not a guard.

If that sentence is false in any case, this PR is wrong and should be closed rather than patched.

**Three specific things I am least sure of:**

1. **I deleted a test that existed to stop exactly this change.** `#496 add_node: object_info UNAVAILABLE still fails closed, even for a frontend-only type` was a deliberate pin, and the doc comment above the function spelled out the same rule ("a genuinely UNVERIFIABLE type … always fails closed, for every type"). I replaced it with a scoped version. Please check whether the original pin was protecting a case my replacement no longer covers — that is the highest-probability defect here.

2. **Does the history oracle really survive a timed-out refresh?** My whole safety argument rests on it. I measured `recordTypes(null)` recording nothing and the baseline staying seeded, so `wasTypeEverDefined` still returns a trustworthy `false`. But I verified that against `object-info-history.js` in isolation. **Is there any production path where a failed or timed-out `/object_info` fetch arms `loseBaseline`, clears the history, or otherwise demotes the baseline before `graph_add_node` reads it?** If one exists, the verdict becomes `unseeded`/`pending` and my exemption correctly declines — so the failure mode would be "no worse than today", not unsafe. But if instead some path leaves the history *stale-positive*, that is a real hole and I want to know.

3. **The mid-session-install window.** `SetNode`, `GetNode` and the rgthree control names are on the allowlist but are *generic enough* that a backend pack could register them. If such a pack is installed after the last successful `/object_info`, the fetched-defs path registers the fresh backend def while my path adds the frontend-native class. I argued this is benign (real, registered, provenance-clean, undoable) and impossible for the four core types. **Is there a way for that divergence to produce a node that later fails at execution?** That is the one outcome that would change my mind.

**On the evidence itself** — my first mutation attempt patched nothing and reported "survived", because the clause text I targeted also appears in the sibling exemption. I caught it via an assertion on the match count and re-scoped every mutant to the branch under test. If you spot any other place where I have measured the wrong thing, say so; a mutant that never applied reads exactly like a mutant that survived.

Two things I chose **not** to do, both argued in the PR body and both open to being overruled: the `panel_refresh_nodes` timeout half of the report (feature-freeze), and the identical parity gap in the `graph_set_widget` guard.

artokun:
Following up on my own question 2 rather than leaving it for the reviewer — **verified in production, not just in the module in isolation.**

`loseBaseline()` has exactly **one** call site in the whole panel:

```
web/js/comfyui-mcp-panel.js:1281:    objectInfoHistory.loseBaseline();
```

It sits at the end of the **startup seed** loop, and only after *every* attempt has FINISHED producing nothing. The code says why, and the reason is the same one this PR rests on:

> This is the ONLY place that arms it: a caller that merely bounded its WAIT has learned nothing and must not latch (a request still in flight may return healthy definitions a moment later).

So a `panel_refresh_nodes` timeout — the reported trigger — cannot demote the baseline. Nothing else calls it.

And the failure direction is safe anyway: if some future path *did* arm it, the verdict becomes `unseeded`, my exemption declines, and the add refuses exactly as it does today. The dangerous direction would be a path that leaves the history **stale-positive**; I could not find one, since `recordTypes` only ever ADDs to the seen set and is explicitly not latched.

That leaves questions 1 and 3 as the ones I genuinely cannot settle myself — particularly **1**, the deleted pin.

artokun:
One more thing I checked on myself, because a unit test on the predicate structurally cannot catch it.

**My change makes a new caller state reachable.** Before it, on the panel's path, a successful resolve always implied `freshDefs` was a real object — `getFreshObjectInfo` is wired, so a null return threw. Now `graph_add_node` can proceed with `freshDefs === null`. Everything downstream that reads `freshDefs` had to be re-checked, and I drove all of it rather than reading it:

```
snapshotBackendDef(null, "MarkdownNote")            = undefined
registeredSocketTypes(null).size                    = 0
driftedRequiredInputNames(undefined, undefined)     = []
unavailableRequiredWidgetReport(undefined, …)       = []

CONTROL healthy-path snapshotBackendDef             = undefined   ← identical, NOT a new state
CONTROL healthy-path socketTypes.size               = 1           ← the only value that differs
```

Three consumers, and none of them can misfire here:

- `objectInfoSnapshot.record(...)` is already behind `if (freshDefs && !freshDefsAreSingleClass && ...)`, so it is skipped rather than fed a null.
- `currentDef` is `undefined` **either way** — the healthy path already produced `undefined` for `MarkdownNote`, because the type is absent from `/object_info` by design. So the drift check is on ground it has always been on.
- `registeredSocketTypes` is the one genuine difference: an empty proof set instead of a populated one. But it is only ever consulted through `unavailableRequiredWidgetReport(nodeData, …)`, and the exemption **requires** `!hasBackendProvenance`, which means the class carries no `nodeData` at all. With no `nodeData` there are no required inputs to prove, so the empty set is never reached.

That last point is the structural reason rather than a lucky one: the same clause that makes a type eligible for the exemption (provenance-clean ⇒ no `nodeData`) is what makes every `freshDefs` consumer downstream a no-op for it.

Worth a second pair of eyes anyway — if there is a **fourth** consumer of `freshDefs` on the add path that I did not find, that is a defect I have no test for.

artokun:
**Scope changed — please re-read before reviewing.** I said in the PR body that I had deliberately left the `graph_set_widget` parity gap alone. I then measured it, and the measurement changed my mind.

**What I measured**, on `origin/main`, through the production `runSetWidget` body:

```
object_info UNAVAILABLE (the reported state)   -> REFUSED: cannot verify the node type … no usable /object_info schema
object_info OK (CONTROL)                       -> WROTE: widget value = "hello"
```

The control is the part that matters. My **first** probe had *both* arms refusing with "no `/object_info` oracle is wired" — that was my argument shape being wrong, not the product failing, and without the control I would have "confirmed" a bug I had not observed. The second probe writes on the control, so it is measuring the product.

**Why that changes the decision:** the panel documents the annotation path as *add the node, then put the text in its `text` widget via `panel_set_widget`*. Fixing only `add_node` ships a `MarkdownNote` the user **cannot fill in** — an empty note is arguably a worse end state than a clean refusal. And all three `/object_info`-oracle guards carry the byte-identical `!freshDefs` early throw, so relaxing one copy alone is precisely the #496 drift this family was consolidated to prevent.

So `assertTypeAgainstFreshBackend` and `assertMutatedNodeAuthorized` now take the same exemption on the same terms. These two additionally check the **write-target instance's own constructor** for backend provenance, so they carry *more* evidence than the add path, not less.

New coverage, each mutant anchored to its own function so it cannot silently patch a sibling:

| mutant | result |
|---|---|
| drop the exemption in `assertTypeAgainstFreshBackend` | **2 killed** (incl. the e2e) |
| drop it in `assertMutatedNodeAuthorized` | **1 killed** |
| invert the ever-seen gate in the former | **4 killed** |
| invert it in the latter | **1 killed** |

The e2e test drives `runSetWidget` end to end: the note is **fillable**, and a `KSampler` in the identical state still refuses with its widget left untouched.

`npm run test:unit`: **6252 pass, 0 fail, exit 0.** CI green.

**What I would now most like you to attack**, given the wider blast radius: the two `set_widget` guards authorize a *write to an existing node*, not the creation of a new one. I claim that is no riskier here, because the exemption requires the instance's own constructor to be provenance-clean **and** the ever-seen gate to say never-seen — but "creating a node you can undo" and "mutating a node already on the canvas" are genuinely different risk profiles, and I have argued them as equivalent. If they are not, the `set_widget` half should come back out and the `add_node` half can stand alone.

Still not merging until you have weighed in.

artokun:
### Ready to merge, held for review — plus the browser-gate result

**Everything I can verify myself is done:**

- `npm run test:unit` — **6252 pass, 0 fail, exit 0**
- **8 mutants, all killed** by named tests (4 on the add guard, 4 on the two `set_widget` guards), each anchored to its own function so it cannot silently patch a sibling
- CI Lint pack green; branch MERGEABLE / CLEAN

**Browser gate — measured, and it is not this diff.** `npx playwright test --workers=1` on the branch: **18 failed / 64 passed**. I did not take that at face value:

- Not one failure touches `node-resolve.js`, `add_node` or `set_widget` — they are chat-history, conversation-persistence, streaming, autostart and workflow-chat-identity.
- The rig at `:8188` serves commit `795c1297`, which is **behind `origin/main`**. The specs' own error names this exact condition: *"ComfyUI must be serving the SAME commit — its custom_nodes copy provides the ./lib/\* imports, and a version mismatch there fails the import silently."*
- **Control**: the same 8 spec files on a clean `origin/main` worktree fail **22** — more than the branch does across the entire suite. Pre-existing, and worse without this change.

So the e2e suite gives this diff **no positive coverage** in this environment, and I am not claiming any. I did not install the branch over the shared rig to manufacture some, because other sessions are using it.

**Why I am not merging.** This relaxes a security guard and deletes a test that existed specifically to pin the old behaviour. The review I requested is outstanding, and a self-review is exactly the thing that would miss the defect class at issue here.

For the record, because it is the honest measurement rather than a reason to proceed: the review path on this repo has been **dark since 2026-08-21 02:51Z** (last independent review, #1552), and **8 PRs have merged since with none**. That is precedent, not permission — least of all for a guard relaxation.

**One verdict on the three questions above is all that is left.** If the answer to the third is that authorizing a *write to an existing node* is materially riskier than authorizing a *creation*, drop the third commit (`2a8bbc65`) and the `add_node` fix stands alone — the two halves are independent commits precisely so that is a clean revert rather than a rewrite.

artokun:
### Self-audit against the review taxonomy found a real defect **in my own diff** — fixed in `22bfaa88`

Since the review is still outstanding I spent the wait running `.claude/autopilot/review-taxonomy.md` class-by-class against this branch rather than re-posting the hold. Class 5 — *"ask what the change makes WORSE, not just better"* — surfaced one.

**The defect.** Before this branch, all three guards threw their unavailable-`/object_info` refusal **without consulting anything**, so nothing on that path could raise. My change consults two predicates first. So a registry whose membership probe throws (a Proxy trap) or an oracle that raises would surface a **raw error** where the user previously got the worded refusal.

Measured on the branch as it stood, not theorised:

```
add_node   | wasTypeEverDefined THROWS  -> RAW LEAK: history oracle exploded
add_node   | registry probe THROWS      -> RAW LEAK: registry exploded
set_widget | wasTypeEverDefined THROWS  -> RAW LEAK: history oracle exploded
mutated    | wasTypeEverDefined THROWS  -> RAW LEAK: history oracle exploded
```

After `22bfaa88`, all four return **CLEAN REFUSAL**, and the control still allows a healthy `MarkdownNote` — the swallow did not swallow the exemption itself.

**Also collapses the triplication.** The three inline copies of the exemption are now one shared `frontendOnlyExemptionApplies` helper. That was worth doing on its own terms: three hand-maintained copies of this predicate is the exact drift #496 was filed to end, and I had just reintroduced it.

Fail-closed by construction — any throw returns `false`, which refuses. Same idiom this file already applies to `readImportFailures` and `describeObjectInfoFailure` ("a diagnostic that throws must not replace the refusal it explains").

Pinned by a test that **dies when the catch is removed** (verified by mutation), plus a control asserting the exemption still fires.

`npm run test:unit`: **6253 pass, 0 fail, exit 0.**

This doesn't change what I'm asking you to attack — the three questions above still stand, and question 3 (write-to-existing vs creation) is still the one with the widest blast radius. It does mean the diff is one defect better than when you were first pinged.

artokun:
### Strengthening question 3 — and correcting something I overstated

Still no verdict, so rather than re-post the hold I measured the weakest claim I gave you.

**First, a correction.** I described an exempted node as *"real, registered, provenance-clean, and Ctrl+Z-undoable"*, which reads as *inert*. That is **false for 4 of the 13 allowlisted types**, and by the repo's own documentation:

- `Fast Groups Bypasser / Muter (rgthree)` are **group toggles**. `plugin/skills/ernie-image/SKILL.md` describes graphs with *"dozens of group-boxed variants gated by `Fast Groups Bypasser (rgthree)`"* — a write to one changes whether **other nodes execute**.
- `SetNode` / `GetNode` are KJNodes **buses**. `get_workflow (action:"strip")` exists to *"resolve GetNode/SetNode buses … into real connections"* — the name a `GetNode` carries determines **which link it pulls**.

So a widget write to an exempted node absolutely can change what the backend runs. Anyone reading my earlier phrasing would have concluded otherwise, and that is exactly the reassurance-shaped claim that hides the next defect.

**Second — the argument does not need that claim, and here is the measured version.**

The relaxation does **not change WHICH types are exempt**. It changes only **WHEN the exemption may be consulted**. I drove every allowlisted type plus four refusal controls through both guards on both paths:

| type | add: defs | add: none | set_widget: defs | set_widget: none | same? |
|---|---|---|---|---|---|
| all 13 allowlisted, clean + never-seen | ALLOW | ALLOW | ALLOW | ALLOW | **YES** |
| `TotallyMadeUpNode` (unknown) | REFUSE | REFUSE | REFUSE | REFUSE | YES |
| `MarkdownNote` (ever-seen ⇒ removed) | REFUSE | REFUSE | REFUSE | REFUSE | YES |
| `MarkdownNote` (provenance husk) | REFUSE | REFUSE | REFUSE | REFUSE | YES |
| `MarkdownNote` (not in live registry) | REFUSE | REFUSE | REFUSE | REFUSE | YES |
| **`KSampler` (real backend type)** | **ALLOW** | **REFUSE** | **ALLOW** | **REFUSE** | *** NO *** |

**Exactly one verdict differs between the two paths, and it is the right one.** For a real backend type the fetch is load-bearing, so an unanswered fetch still refuses — untouched by this change. For every one of the 13 exempt types the two paths already agreed.

**What that means for question 3.** Whatever execution-affecting power an exempted type has — bypassing a group, re-routing a bus — was **already fully exercisable whenever `/object_info` answered**, which is the overwhelmingly common case. This change grants **no new capability to any type**; it stops a *timeout* from being the thing that decides.

And the fetch never protected against that power anyway: these guards authorize the **target's own type**, never the downstream effect on the nodes a bypasser toggles. That gap is identical before and after, on both paths.

**So my revised answer to my own question 3:** the write-vs-create distinction is real, but it is **orthogonal to this diff** — the diff moves a precondition, not the permission set. If you disagree, the place to push is the table above: find a type where the two columns *should* differ and do not.

Reproduction is a ~30-line script driving the shipped predicates; say the word and I will commit it as a test rather than leaving it as a paste.

artokun:
The parity table from the previous comment is now a **test** rather than a paste — `22071734`.

It iterates `FRONTEND_ONLY_NODE_TYPES` (not a hand-written list, so a type added later is covered automatically), drives both guards on both paths, and asserts the verdicts match — plus asserts explicitly that a real backend type is the one case that MUST differ.

Mutation-verified: widening the exemption to any never-seen type fails **5 named tests**, this one first.

`npm run test:unit`: **6254 pass, 0 fail, exit 0.**

Nothing else outstanding from my side. Still holding for a verdict on the three questions.

artokun:
### Gate verdict: **INDETERMINATE** — stopping, not merging

Same declaration as on the sibling PR (artokun/comfyui-mcp#2019), recorded here so this PR stands on its own.

Measured just now: **0** non-Copilot reviews on this PR; Copilot's only entry is its quota-limit refusal (a `/reviews` entry with `state: COMMENTED` — it counts as a review to any unfiltered check and is not one); a `POST .../requested_reviewers` re-request returns 200 while `requested_reviewers` stays `[]`. Across this repo the last genuine independent review was **#1552 at 02:51Z**, with **8 PRs merged since with none**.

I am not substituting the codex gate — it is pointed at the owner's weakest model, so a SHIP from it would wear the authority of an independent review without being one.

This PR is the one I least want merged unreviewed: it **relaxes a security guard** and **replaces a test written specifically to pin the old behaviour**. My own adversarial pass already found one defect here that four careful passes and eight mutants had missed (the raw-error leak, `22bfaa88`), which is the argument against trusting a ninth self-reading.

Final state: 5 commits, `npm run test:unit` **6254 pass / 0 fail / exit 0**, 9 mutants all killed by named tests, CI green, `CLEAN`. The three questions I want attacked are unchanged, and question 1 — whether the pin I deleted was protecting a case my replacement no longer covers — is still the highest-probability defect.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `2207173486aa58a7dedbf7842e856496789e91c8` vs origin/main. Read `frontendOnlyExemptionApplies`, `assertAddNodeResolvableRefreshing`, `assertTypeAgainstFreshBackend`, `assertMutatedNodeAuthorized`, and `isAuthorizedFrontendOnlyType` — not just the patch.

The load-bearing argument holds. #458 fails closed when /object_info cannot speak. For Note / MarkdownNote / Reroute / PrimitiveNode and the rest of `FRONTEND_ONLY_NODE_TYPES`, that fetch is empty by design, so a timeout is not evidence about those types. The exemption applied on the unavailable path is the same two clauses the fetched-defs path already used, and neither reads `freshDefs`:

- `backendHistoryVerdict === "never-seen"` (unwired oracle is `no-oracle`, not never-seen; pending / unseeded / removed still refuse)
- `isAuthorizedFrontendOnlyType`: live registry membership + allowlist + no backend provenance

So an exempted add still cannot mint the #458 placeholder: `LG.createNode` is about to construct a type that is actually registered.

What I checked:

- Exempt SET does not move. The invariant test drives every allowlisted type plus unknown / ever-seen / provenance husk / not-in-registry on both paths; only a real backend type (`KSampler`) is allowed to differ, and it still refuses when the fetch is null.
- All three /object_info-gated guards move together. Relaxing add_node alone would have recreated #496 for set_widget on the note the docs tell you to fill in.
- A throwing registry or history oracle returns the worded refusal, not a raw stack. That is the thing this change made reachable and then closed.
- `loseBaseline` is not armed on timeout (existing object-info-history rule); the session oracle a timeout leaves behind is exactly what the ever-seen gate needs.

Residual, not blocking: this deliberately weakens a fail-closed default for a named allowlist. The allowlist staying small, provenance-clean, and behind never-seen is the whole safety property. Do not grow it without the same two-path invariant test.

Orchestrator backstop is artokun/comfyui-mcp#2019.

Issue #442, opened 2026-08-01
Five panel defects from a long graph-editing session (incl. missing node-resize tool — patch attached)


artokun:
Hit five distinct defects during a ~2h session editing and testing a 319-node workflow via the agent panel. #1 has a local patch attached; #2–#5 are reported for triage.

### 1. No node-resize tool exists (patch attached)

**What happened / root cause.** The user asked for a tidy layout: "make every node that's part of a group match dimensions to the largest node in its column or row." `panel_move_node` sets position, `panel_auto_layout` repositions — but nothing in the panel toolset can set a node's `size`. This made a routine layout request impossible on the live canvas.

**Workaround I was forced into.** Save the workflow, rewrite `pos`/`size`/`groups` in the JSON with a Python script, then `panel_load_workflow(path:...)` to reload. That works but **resets the tab's Ctrl+Z undo stack** and needs a file round-trip for what should be one tool call.

**Fix — applied locally: yes. Upstream-only: no.** `graph_set_node_size` mirrors `graph_move_node` almost exactly. Two edits:

`comfyui-mcp-panel/web/js/comfyui-mcp-panel.js` — new handler after `graph_move_node`:

```js
  graph_set_node_size({ node_id, size }) {
    const { graph } = getGraphCtx();
    const node = resolveNode(graph, node_id);
    if (!Array.isArray(size) || size.length !== 2) throw new Error("size must be [width, height]");
    const previous = [node.size[0], node.size[1]];
    graph.beforeChange();
    try {
      const w = Math.max(1, Number(size[0]));
      const h = Math.max(1, Number(size[1]));
      if (typeof node.setSize === "function") node.setSize([w, h]);
      else node.size = [w, h];
    } finally {
      graph.afterChange();
    }
    graph.setDirtyCanvas(true, true);
    return { resized: { node_id: node.id, from: previous, to: [node.size[0], node.size[1]] } };
  },
```

plus a dispatch label beside `case "graph_move_node"`:

```js
    case "graph_set_node_size":
      return { icon: "pi-arrows-h", text: `Resized node ${r.resized?.node_id} to [${r.resized?.to?.map(Math.round)}]` };
```

`comfyui-mcp/dist/orchestrator/panel-tools.js` — new tool beside `panel_move_node`:

```js
[...]

PR #1592, opened 2026-08-21
fix(1575): reopening a closed tab loads the workflow ComfyUI's store-level open skipped

### LINKED ISSUE #1575: Fix reopening a closed saved workflow using a stale catalog object
Opening a saved workflow after closing its tab failed in panel_open_workflow. ComfyUI frontend can replace the saved-workflow catalog entry passed to openWorkflow() with a new live tab object. workflow_open continued reading target.changeTracker/target.activeState from the stale catalog object, so it found no nodes and refused the safe repaint even though the workflow pointer had switched.

### Steps to reproduce
1. Open and save a workflow.
2. Close its tab with panel_close_workflow.
3. Call panel_open_workflow for the saved JSON path.
4. The call returns an unknown partial outcome; panel_list_workflows can show an inconsistent active/open state.

### Exact error (scrubbed)
workflow_open could not rebind the active canvas because this frontend did not expose a complete workflow state for a safe repaint. The open may have switched the active workflow; inspect panel_list_workflows, then reload the panel before graph edits.

### Fix
Applied locally: yes
Upstream-only: no

After await s.openWorkflow(target), when the workflow was not already open, reacquire activeWorkflowRef() and replace target when it matches the requested selector. This ensures the repaint reads the live tab object's change tracker.

Patch:
```diff
 await s.openWorkflow(target);
+if (!wasOpen) {
+  const activeAfterOpen = activeWorkflowRef();
+  if (activeAfterOpen && workflowRecordMatchesSelector(activeAfterOpen, path)) {
+    target = activeAfterOpen;
+  }
+}
```

Added a regression assertion in browser_tests/unit/open-outcome.test.mjs that the live object is reacquired before activeState is read.

Verification:
- node --test browser_tests/unit/open-outcome.test.mjs: 50/50 passed
- node --check web/js/comfyui-mcp-panel.js: passed
- Live retry after frontend reload: panel_open_workflow returned opened:true, active_matches_target:true, graph_binding:bound; panel_graph_outline showed the expected 83-node/14-group workflow.
- Full npm test/typecheck could not run because development dependencies zod/tsc are not installed in this deployed pack.

### Environment
Windows 11
ComfyUI 0.33.3
ComfyUI frontend 1.49.6
NVIDIA GeForce RTX 4090 24 GB
comfyui-mcp 0.52.45
comfyui-mcp-panel 0.15.32

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-sol` · effort `xhigh` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #1215

artokun:
Fixes #1575

## The report

`panel_open_workflow` on a saved workflow whose tab had just been closed with
`panel_close_workflow` returned an unknown partial outcome:

> workflow_open could not rebind the active canvas because this frontend did not expose
> a complete workflow state for a safe repaint. The open may have switched the active
> workflow; inspect panel_list_workflows, then reload the panel before graph edits.

…and `panel_list_workflows` could then show an inconsistent active/open state.

## The cause — and it is NOT the one in the issue body

The issue proposed that "ComfyUI frontend can replace the saved-workflow catalog entry
passed to `openWorkflow()` with a new live tab object", with a patch that re-reads
`activeWorkflowRef()` after the open and adopts it. **That mechanism is not what
happens, and that patch is inert against this bug.** `ComfyWorkflow.load()` returns
`this`, and a live browser probe shows the post-open active workflow *is* the object the
panel already held (`activeIsSameObject: true`), so the adoption is a no-op.

The real cause is that `app.extensionManager.workflow` is ComfyUI's workflow **store**,
not the workflow **service** — `workspaceStore` binds it as
`computed(() => useWorkflowStore())`. So this pack's close and open are the store's
primitives, and they do not pair up the way the service's do:

| call | what it does |
| --- | --- |
| `store.closeWorkflow(wf)` | drops the path from `openWorkflowPaths`, calls `wf.unload()` (nulling `changeTracker`), and **leaves `activeWorkflow` pointing at wf**. Moving the pointer is the SERVICE's job, and the service is a Vue composable the panel cannot reach. |
| `store.openWorkflow(wf)` | begins `if (isActive(workflow)) return workflow`, and `isActive` is `activeWorkflow.path === workflow.path` — **by path, not identity**. |

So a panel-driven close of the active tab leaves the store in a state its own types
forbid: `activeWorkflow` is a `LoadedComfyWorkflow` whose `changeTracker` is null and
whose path is not in the open list. The next `openWorkflow` for that path early-returns
on that stale pointer, loads nothing, never pushes the path back, and **resolves as
though it worked**. `workflow_open` then reads
`target.changeTracker?.activeState ?? target.activeState`, gets `null`, and refuses —
blaming the frontend for a state that was one `load()` away on the object already in hand.

### Measured in a browser, against the reporter's own frontend (ComfyUI 0.33.2 / frontend 1.49.6)

```
after store.closeWorkflow   activeIsSameObject:true  hasTracker:false  inOpenList:false
after store.openWorkflow    activeIsSameObject:true  hasTracker:false  inOpenList:false
panel repaint-state read    changeTracker?.activeState ?? activeState  ===  null   -> wouldRefuse:true
```

and the repair, measured in the same probe:

```
openWorkflowsInBackground({right:[path]})  -> inOpenList:true
wf.load()                                  -> hasTracker:true, nodes:10, isModified:false
app.loadGraphData(state,true,true,wf)      -> canvasNodes:10, active:wf, inOpenList:true
```

## The fix

New `web/js/lib/settle-open-target.js`, called from `workflow_open` immediately after
`s.openWorkflow(target)` and before the state read. It does nothing unless the open left
us with no usable state; then:

1. **ADOPT** — a *different* live object is active for the requested selector and has a
   complete state. Take it. (This is the issue's proposal. It is kept because it is cheap
   and correct *if* a frontend ever does hand back a different instance — it is simply not
   the arm that fixes this report.)
2. **LOAD** — the frontend still names our target as active by the same rule its own
   `openWorkflow` used to early-return, yet it is unloaded. Load the content the early
   return skipped, then return the tab to the open list via `openWorkflowsInBackground`
   (the store's own "add paths without loading them or changing the active workflow"
   entry point) — which is what clears the inconsistent active/open state the report also
   named.

Two supporting changes:

- **The #874 capture gate now skips when we just loaded from disk.** Serializing the
  still-mounted canvas (the closed tab's) over a state read fresh off disk would be the
  #1215/#968 poisoning through a new entry point, and it can preserve nothing — nothing
  has run against a just-loaded tracker.
- **The reply discloses it** (`reopened_from_disk`): the canvas now shows the *on-disk*
  copy, so any node-written value that only lived on the closed tab's canvas (#874) is not
  in it. A bland success would have hidden a real difference.

Non-regression is structural: an open that already produced a complete state returns at
`reason: "loaded"` before anything is touched — that is every open that works today.

## Evidence

**Mutation-verified** — the fix deleted, the named test goes red:

| mutation | result |
| --- | --- |
| remove the LOAD arm from the helper | 9 named unit tests fail, incl. *"the fix loads the content the store's early return skipped"* |
| delete the `settleOpenedWorkflowTarget` call site | 2 wiring pins fail |
| remove the capture-gate guard | *"a state just read off DISK is not overwritten by the still-mounted canvas"* fails |
| delete the call site, then run the **e2e** spec | reproduces the reporter's error **verbatim**: `workflow_open could not rebind the active canvas because this frontend did not expose a complete workflow state for a safe repaint.` |

**Tests**

- `browser_tests/unit/settle-open-target.test.mjs` — 24 tests. Drives the shipped helper
  against `FakeWorkflowStore`, a transcription of the frontend's real `openWorkflow` /
  `closeWorkflow` / `unload` / derived `activeState` getter, so the state under test is
  the state the frontend really produces. Includes an explicit test that the issue's own
  patch is inert here.
- `browser_tests/reopen-closed-workflow.spec.ts` — drives the reported sequence
  (save → close → reopen) through real panel commands against a real ComfyUI.
- `npm run test:unit`: **6271 passed, 0 failed**. (The `#671 manager-install` wall-clock
  flake fired once under load in an earlier run and is green in isolation: 125/125.)
- `npx tsc --noEmit`: clean.
- Full Playwright suite at `--workers=1`: running; result posted in a comment below.

`browser_tests/unit/frontend-contract.test.mjs` gains `openWorkflowsInBackground` as an
OPTIONAL member — always called behind a `typeof … === "function"` guard, so a frontend
without it still gets the loaded canvas and the reply says the tab list could not be
corrected.

- **I did not change `workflow_close`.** The true origin is that the panel calls the
  store's `closeWorkflow`, which does not move the active pointer the way the service's
  does. Fixing it there would mean the panel picking a next-active workflow — replicating
  service logic and changing close behaviour — and it would leave already-stranded
  sessions broken. The repair belongs where the failure is observed. Worth a follow-up
  issue; not this PR, under the freeze.
- **I did not act on the issue's patch as written.** Applying it would have shipped a
  no-op. The reporter's live verification was done *after a frontend reload*, which on its
  own explains the recovery — so their evidence does not isolate their patch.
- I did not touch the already-open path (`wasOpen`), which owns the #442 disk-staleness
  comparison.

## Review

**Review is PENDING.** This change is not gated, and I have not merged it.

### DIFF
--- a/web/js/comfyui-mcp-panel.js
+++ b/web/js/comfyui-mcp-panel.js
@@ -487,6 +487,7 @@ import {
   trackerCaptureSuppressed,
 } from "./lib/change-tracker-snapshot.js";
 import { flushSourceCanvasBeforeSwitch } from "./lib/flush-source-before-switch.js";
+import { settleOpenedWorkflowTarget } from "./lib/settle-open-target.js";
 import { coerceMessageText, isDroppedAgentReplay, serializeContext, stripAgentDirectedBlocks } from "./lib/chat-serialize.js";
 import {
   applyCurrentDefWidgetValues,
@@ -18022,6 +18023,12 @@ const GRAPH_TOOL_EXECUTORS = {
     // tag-based classification answers "unknown" for it, which warns nobody —
     // and that is the configuration the wrong-canvas reports arrived in.
     let sourceUnproven = false;
+    // #1575 — what `settleOpenedWorkflowTarget` had to do after ComfyUI's
+    // store-level open resolved without loading anything. Null on every open that
+    // needed nothing, which is every open that works today. Drives the #874
+    // capture gate (a state we just read off disk must not be overwritten by the
+    // still-mounted canvas) and the reply's disclosure.
+    let openSettled = null;
     let reloaded = false;
     let reloadError = null;
     let openFailed = null;
@@ -18095,6 +18102,44 @@ const GRAPH_TOOL_EXECUTORS = {
         endWorkflowReloadStep(reloadGuardToken);
       }
       if (!openFailed) {
+        // #1575 — `s.openWorkflow(target)` RESOLVING is not proof it loaded anything.
+        //
+        // `app.extensionManager.workflow` is the workflow STORE (workspaceStore binds it
+        // as `computed(() => useWorkflowStore())`), so this command's close and open are
+        // the store's primitives, and they do not pair up the way the SERVICE's do:
+        // `closeWorkflow` unloads the tab and leaves `activeWorkflow` pointing at it,
+        // while `openWorkflow` opens with `if (isActive(workflow)) return workflow` and
+        // `isActive` compares BY PATH. So the tab this command just closed is still named
+        // as active, this open early-returns on that stale pointer, and nothing loads.
+        // Measured on a live frontend: hasTracker false and inOpenList false both before
+        // AND after the open, which makes the `st` read below null and this command refuse
+        // with "this frontend did not expose a complete workflow state" — blaming the
+        // frontend for a state that is one `load()` away on the object already in hand.
+        //
+        // Runs BEFORE the capture gate and the repaint so every later step — the identity
+        // stamp, the proof, the fence, the reply — reads the object that actually holds
+        // the workflow. Returns untouched (`reason: "loaded"`) whenever the open produced
+        // a usable state, which is every open that works today.
+        openSettled = await settleOpenedWorkflowTarget({
+          wasOpen,
+          target,
+          selector: path,
+          activeAfterOpen: activeWorkflowRef(),
+          openWorkflows: s.openWorkflows,
+          sameWorkflowObject,
+          matchesSelector: workflowRecordMatchesSelector,
+          // The exact call the store's own `openWorkflow` skipped. Guarded, so a
+          // frontend without it degrades to today's refusal rather than throwing.
+          loadWorkflowContent: (wf) => (typeof wf.load === "function" ? wf.load() : undefined),
+          // The store's own "add paths without loading them or changing the active
+          // workflow" entry point — the missing half of the early-returned open, and
+          // what clears the active-but-not-open state the report also named.
+          reopenTabInBackground: (p) =>
+            typeof s.openWorkflowsInBackground === "function"
+              ? s.openWorkflowsInBackground({ right: [p] })
+              : undefined,
+        });
+        if (openSettled.target && openSettled.target !== target) target = openSettled.target;
       // We deliberately do NOT auto-reload the canvas from disk here: switching to an
       // already-open tab must not silently discard the user's in-memory graph, and a
       // re-read could race a concurrent canvas edit and clobber it. Instead the staleness
@@ -18242,8 +18287,14 @@ const GRAPH_TOOL_EXECUTORS = {
           const captureBinding = describeLiveCanvasBinding(target);
           pointerMovedThisOpen = !sameWorkflowObject(activeBefore, target);
           if (
-            captureBinding === "bound" ||
-            (captureBinding !== "foreign" && !pointerMovedThisOpen)
+            // #1575 — a state THIS command just read off disk is authoritative, and the
+            // canvas mounted behind it is whatever the closed tab left there. Serializing
+            // that over the fresh state is the #1215/#968 poisoning through a new entry
+            // point, and it can preserve nothing: nothing has run against a just-loaded
+            // tracker, so it holds no node-written values (#874) for the capture to save.
+            openSettled?.loaded !== true &&
+            (captureBinding === "bound" ||
+              (captureBinding !== "foreign" && !pointerMovedThisOpen))
           ) {
             try {
               // AWAITED (codex). A frontend whose tracker captures asynchronously would
@@ -19139,6 +19190,32 @@ const GRAPH_TOOL_EXECUTORS = {
       // passing. The two read identically now, so this discloses instead of
       // claiming. Its own key, like the foreign one: a caller can hit it with
       // neither `stale` nor `canvas_file_divergence` firing.
+      //
+      // #1575 — the open SUCCEEDED, and it succeeded because this command repaired a
+      // frontend state the frontend does not repair for itself. Say so rather than
+      // report a bland success: the canvas now shows the SAVED file, so any node-written
+      // value that only ever lived on the canvas of the tab that was closed is gone —
+      // and that is a real difference the caller may be about to build on.
+      ...(openSettled?.loaded === true
+        ? {
+            reopened_from_disk:
+              "This workflow's tab was closed while it was still the frontend's ACTIVE " +
+              "workflow, which leaves ComfyUI's workflow store naming it active with no " +
+              "content loaded and its tab absent from the open list. The store's own open " +
+              "returns early in that state (it matches the active workflow BY PATH), so it " +
+              "loaded nothing and this command loaded the saved file itself" +
+              (openSettled.reopened ? " and restored the tab to the open list" : "") +
+              ". The canvas therefore shows the ON-DISK copy: anything a NODE wrote and " +
+              "nobody saved — a populated wildcard, a rolled seed (#874) — is not in it. " +
+              "Re-read the graph (panel_graph_outline / panel_query_graph) before relying " +
+              "on values you generated before the close." +
+              (openSettled.reopened
+                ? ""
+                : " The tab could NOT be returned to the open list, so panel_list_workflows " +
+                  "may still show this workflow as active-but-not-open; reload the panel to " +
+                  "clear that."),
+          }
+        : {}),
       ...(sourceUnproven
         ? {
             unproven_source_state:
--- a/web/js/comfyui-mcp-panel.js
+++ b/web/js/comfyui-mcp-panel.js
@@ -18120,25 +18120,39 @@ const GRAPH_TOOL_EXECUTORS = {
         // stamp, the proof, the fence, the reply — reads the object that actually holds
         // the workflow. Returns untouched (`reason: "loaded"`) whenever the open produced
         // a usable state, which is every open that works today.
-        openSettled = await settleOpenedWorkflowTarget({
-          wasOpen,
-          target,
-          selector: path,
-          activeAfterOpen: activeWorkflowRef(),
-          openWorkflows: s.openWorkflows,
-          sameWorkflowObject,
-          matchesSelector: workflowRecordMatchesSelector,
-          // The exact call the store's own `openWorkflow` skipped. Guarded, so a
-          // frontend without it degrades to today's refusal rather than throwing.
-          loadWorkflowContent: (wf) => (typeof wf.load === "function" ? wf.load() : undefined),
-          // The store's own "add paths without loading them or changing the active
-          // workflow" entry point — the missing half of the early-returned open, and
-          // what clears the active-but-not-open state the report also named.
-          reopenTabInBackground: (p) =>
-            typeof s.openWorkflowsInBackground === "function"
-              ? s.openWorkflowsInBackground({ right: [p] })
-              : undefined,
-        });
+        // INSIDE a reload STEP (review P1). `target.load()` is a `/userdata` fetch with
+        // no deadline of its own, and between steps the guard sits at `pending === 0`,
+        // where `activeWorkflowReloadGuard()` expires it after WORKFLOW_RELOAD_GUARD_MAX_MS.
+        // A stalled read would therefore drop the fence mid-load, let a concurrent
+        // graph_* command through and acknowledge it, and the repaint below would then
+        // overwrite it from the disk state — the data-loss shape the guard exists to
+        // prevent. A step keeps `pending > 0`, which is never expired.
+        beginWorkflowReloadStep(reloadGuardToken);
+        try {
+          openSettled = await settleOpenedWorkflowTarget({
+            wasOpen,
+            target,
+            selector: path,
+            activeAfterOpen: activeWorkflowRef(),
+            // A READER, not a snapshot: the helper re-reads it after the re-list to
+            // report what the store DID rather than that the call returned (review P1).
+            readOpenWorkflows: () => s.openWorkflows,
+            sameWorkflowObject,
+            matchesSelector: workflowRecordMatchesSelector,
+            // The exact call the store's own `openWorkflow` skipped. Guarded, so a
+            // frontend without it degrades to today's refusal rather than throwing.
+            loadWorkflowContent: (wf) => (typeof wf.load === "function" ? wf.load() : undefined),
+            // The store's own "add paths without loading them or changing the active
+            // workflow" entry point — the missing half of the early-returned open, and
+            // what clears the active-but-not-open state the report also named.
+            reopenTabInBackground: (p) =>
+              typeof s.openWorkflowsInBackground === "function"
+                ? s.openWorkflowsInBackground({ right: [p] })
+                : undefined,
+          });
+        } finally {
+          endWorkflowReloadStep(reloadGuardToken);
+        }
         if (openSettled.target && openSettled.target !== target) target = openSettled.target;
       // We deliberately do NOT auto-reload the canvas from disk here: switching to an
       // already-open tab must not silently discard the user's in-memory graph, and a
--- a/web/js/lib/settle-open-target.js
+++ b/web/js/lib/settle-open-target.js
@@ -1,0 +1,221 @@
+/**
+ * #1575 — after ComfyUI's store-level `openWorkflow` resolves, make sure the panel
+ * is holding a LOADED workflow object for the workflow the caller asked for.
+ *
+ * THE REPORT. `panel_open_workflow` on a saved workflow whose tab had just been
+ * closed with `panel_close_workflow` came back with
+ *
+ *   "workflow_open could not rebind the active canvas because this frontend did
+ *    not expose a complete workflow state for a safe repaint."
+ *
+ * — an unknown partial outcome, and `panel_list_workflows` showing an
+ * inconsistent active/open state.
+ *
+ * THE CAUSE, measured in a browser against a live ComfyUI (frontend source read
+ * at 1.47.2; probe run against the installed frontend on 2026-08-21).
+ *
+ * `app.extensionManager.workflow` is the workflow STORE, not the workflow
+ * service — `workspaceStore` binds it as `computed(() => useWorkflowStore())`.
+ * The panel therefore calls the store's two primitives, and they do not pair up
+ * the way the service's do:
+ *
+ *   store.closeWorkflow(wf)  removes the path from `openWorkflowPaths` and calls
+ *                            `wf.unload()` (which nulls `changeTracker`), and
+ *                            LEAVES `activeWorkflow` pointing at wf. Moving the
+ *                            pointer is the SERVICE's job, and the service is a
+ *                            Vue composable the panel cannot reach.
+ *
+ *   store.openWorkflow(wf)   begins `if (isActive(workflow)) return workflow`,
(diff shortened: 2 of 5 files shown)


artokun:
@grok please review this one adversarially. It contradicts the fix the issue reporter proposed, so the burden is on me. Here is exactly where I think the risk is — please attack these rather than scanning the diff:

**1. The claim that the reporter's mechanism is wrong.** I assert `openWorkflow` does NOT swap the workflow object, and that the real cause is ComfyUI's *store* `openWorkflow` early-returning via `isActive()` (path equality) on an `activeWorkflow` pointer that the store's `closeWorkflow` never moved. Evidence is `ComfyWorkflow.load()` returning `this` plus a live probe showing `activeIsSameObject: true`. If the object CAN be swapped on some frontend version, my `frontendConsidersActive()` path fallback is what has to catch it — is it enough?

**2. Loading from disk in `settle-open-target.js` — can it destroy work?** My reasoning: the tab was already `unload()`ed by the close (tracker nulled), and the store's `closeWorkflow` also calls `removeDraft(path)`, so there is no in-memory state and no draft overlay left to lose; `load()` can only read the file. If either of those is wrong on some path — particularly a `force:true` close of a DIRTY tab — then this silently replaces unsaved work and the fix is worse than the bug. This is the claim I am least sure of.

**3. Skipping the #874 capture when I loaded from disk** (`openSettled?.loaded !== true` in the capture gate). I argue a just-loaded tracker holds no node-written values, so the capture can only preserve nothing and can poison (#1215/#968) — the mounted canvas at that moment is the CLOSED tab's. Is there a case where the mounted canvas legitimately holds values that belong in the freshly-loaded tab?

**4. The ADOPT arm.** It swaps `target` to another live object when that object answers `workflowRecordMatchesSelector(active, path)`. If a selector can match a workflow that is not the requested one (a bare filename? a key collision across directories?), this repaints and stamps this command's UUID onto someone else's tab — the #1089 hazard. Is the selector guard tight enough, and should I additionally require path equality with the incumbent?

**5. `wasOpen` as the exclusion.** I claim an already-loaded tab can never be in the stranded state, because `wasOpen` IS `!!target.changeTracker` and the early return only strands an unloaded workflow. Counterexample?

**6. `openWorkflowsInBackground({right:[path]})`.** Public store method, but I am appending the tab rather than restoring its original index. Any side effect I have missed (tab ordering, thumbnails, persistence)?

Verification so far: full unit suite 6271/0, tsc clean, and the e2e spec reproduces the reporter's exact error string when I delete the call site. Playwright full run is in progress and I will post the result. Not merging until this comes back.

artokun:
## Adversarial review round 1 — **NO-SHIP**, both findings fixed

**Disclosure, as that gate requires:** this was a **substituted review** — the repo's
`claude-gate.mjs` (a fresh session given only the diff and told to refute), not codex.
codex is currently pointed at a weak model, so a pass from it would have carried
authority it does not have. grok has also been asked on this PR and on the issue thread;
nothing back yet. This is weaker than codex in one specific way — codex reliably
*executes* against hostile inputs — so I have treated both findings as blocking and
mutation-verified each fix.

It returned **NO-SHIP** with two P1s. Both were real. Neither was something my own tests
could have caught, and the reason why is the interesting part.

### P1 (1) — `reopened` reported the request, not the effect

`reopened` was set to `true` because the injected `reopenTabInBackground` *did not
throw*. But the call site always passes a **function**:

```js
reopenTabInBackground: (p) =>
  typeof s.openWorkflowsInBackground === "function"
    ? s.openWorkflowsInBackground({ right: [p] })
    : undefined,
```

so on a frontend lacking that store method the lambda returns `undefined` silently, and
"this store has no such API" and "the tab was restored" collapsed to the same value. The
reply then asserted *"and restored the tab to the open list"* **and suppressed** the
disclosure that says it could not — i.e. it reported the inconsistent active/open state
this PR exists to repair as repaired, while it was untouched. That also directly
contradicted the comment I had just written in `frontend-contract.test.mjs`.

The reviewer executed it against the verbatim call-site collaborators:

```
settle result : {"loaded":true,"reopened":true,"reason":"loaded-after-noop-open"}
open list now : []  <- tab STILL absent
```

**Why my tests were blind:** they injected `reopenTabInBackground: undefined` and a
*throwing* stub — two shapes the call site can never produce — so the suite was green
with the defect shipping. That is the failure mode, not the bug itself.

**Fix:** `reopened` is now **re-read from the open list** after the call
(`readOpenWorkflows` is injected as a reader, not captured as a snapshot array), and the
tests build the collaborators *exactly* as `workflow_open` builds them, capability
ternary included. Three new cases cover the shapes that actually occur: the method
absent, the method present but silently doing nothing, and the method throwing.

### P1 (2) — the settle's `await` sat outside the reload fence

`target.load()` is a `/userdata` fetch with **no deadline of its own** — unlike the
`/userdata` probe this same function already bounds, precisely because that endpoint was
measured to accept and never answer. My `await` sat in the gap *between* reload steps,
where the guard is at `pending === 0` and `activeWorkflowReloadGuard()` expires it after
`WORKFLOW_RELOAD_GUARD_MAX_MS`. So a stalled read drops the fence mid-load, a concurrent
`panel_add_node` is let through **and acknowledged**, and then the repaint overwrites it
from the disk state — the exact data-loss shape the guard was written to prevent.

**Fix:** the settle now runs inside a `beginWorkflowReloadStep` / `endWorkflowReloadStep`
pair (released in a `finally`), which keeps `pending > 0` — a state that is never expired.

### Mutation-verified

| mutation | result |
| --- | --- |
| set `reopened: true` (the reviewed defect, restored) | 3 named tests fail |
| remove the `begin/endWorkflowReloadStep` wrapper | *"the settle's await is held INSIDE a reload step"* fails |

### Clean per the reviewer

Its taxonomy pass found these classes clean and said why: `target` is reassigned before
`pointerMovedThisOpen`, the identity stamp and the proof marker; the adopt arm's
`!sameObject && selectorOk && samePath` closes the bare-filename selector collision; the
capture-gate broadening only suppresses a `checkState()` on a tracker that was null
seconds earlier, so it cannot lose node-written values; and the production caller is
reachable on the reported path.

### State after the fixes

- `npm run test:unit`: **6275 passed, 0 failed**
- `npx tsc --noEmit`: clean
- Full Playwright suite (`--workers=1`) against this branch: **69 passed, 14 failed in
  14.3m**. Every failure is a chat/transcript/queue spec with the same
  `MockBridge.waitForUserMessage timed out` signature; **no workflow open/save/close spec
  failed**, including `reopen-preserves-node-set-values` (#874), which is what the
  capture-gate change touches. I am currently re-running those 14 against `origin/main`
  with the junction repointed, to say whether they are mine or pre-existing rather than
  asserting it — result in the next comment.

Re-requesting review on the two fixes.

artokun:
## Browser suite — measured against an `origin/main` baseline, not asserted

Full Playwright run on this branch (`--workers=1`): **69 passed, 14 failed, 14.3m**.

Rather than call those failures pre-existing, I measured it. I cut a second worktree at
`origin/main`, repointed ComfyUI's `custom_nodes` junction at it, restarted ComfyUI
(verified the served bundle no longer contains `settleOpenedWorkflowTarget`), and re-ran
**the same seven spec files** with the same `--workers=1`:

| spec file | `origin/main` (no fix) | this branch |
| --- | ---: | ---: |
| `conversation-persistence.spec.ts` | 7 | 5 |
| `workflow-chat-identity.spec.ts` | 4 | 3 |
| `chat-history-v2.spec.ts` | 3 | 2 |
| `pending-queue.spec.ts` | 1 | 1 |
| `hidden-tab.spec.ts` | 1 | 1 |
| `streaming-render.spec.ts` | 1 | 1 |
| `external-orchestrator.spec.ts` | 1 | 1 |
| **total** | **18** | **14** |

Every failing *file* is the same on both sides, every failure carries the same
`MockBridge.waitForUserMessage timed out` signature, and the baseline fails **strictly
more**. So these are pre-existing on `origin/main` and this branch does not add any. I am
**not** claiming the 4-failure difference as an improvement — these specs are wall-clock
flaky and the machine was heavily loaded (~105 unrelated MCP server processes, CPU ~60%);
that delta is run-to-run variance.

**What matters for this change:** no workflow open/save/close spec failed on either side.
Specifically green on this branch:

- `reopen-closed-workflow.spec.ts` — the new spec, driving the reported save → close →
  reopen sequence over the bridge
- `reopen-preserves-node-set-values.spec.ts` — **#874**, which is precisely what the
  capture-gate change touches
- `open-unverified-names-recovery.spec.ts`, `save-as-reports-identity.spec.ts`,
  `noname-save-keeps-app-json.spec.ts`, `stale-snapshot-save-refused.spec.ts`,
  `dirty-guards-see-node-writes.spec.ts`, `bridge-route-never-file-derived.spec.ts`

The end-to-end mutation result stands: with the `settleOpenedWorkflowTarget` call site
deleted, `reopen-closed-workflow.spec.ts` fails with the reporter's error **verbatim**.

ComfyUI's junction has been restored to its original target and ComfyUI restarted.

Round 2 of the adversarial review is running against the fixed diff. **Not merging until
it lands.**

artokun (commented):
COMMENT INTENDED: APPROVE

Reviewed `920ec8b1` adversarially against #1575's proposed patch, the eight claims in the 2026-08-21T22:00:45Z ping, ComfyUI frontend store/service source, `FakeWorkflowStore`, and the `workflow_open` / `workflow_close` call sites. **This PR is still a draft.** I am not asking to merge it, and I cannot Approve from this GitHub user — COMMENT is the event.

**1. The reporter's mechanism is wrong. Your diagnosis holds.**
`workspaceStore.ts` binds `workflow: computed(() => useWorkflowStore())`. Store `isActive` is `activeWorkflow.value?.path === workflow.path` (`workflowStore.ts:138-139`). Store `openWorkflow` begins `if (isActive(workflow)) return workflow` with no load and no `openWorkflowPaths` push (`workflowStore.ts:201-204`). Store `closeWorkflow` drops the path, `removeDraft`, `unload()` (or deletes a temporary) and never moves `activeWorkflow` (`workflowStore.ts:340-351`). Moving the pointer is the **service** (`workflowService.ts:335-350`). `ComfyWorkflow.load()` returns `this` (`comfyWorkflow.ts:98-154`). The panel calls those store primitives from `app.extensionManager.workflow`. After close-then-open, the object in hand is still the catalog entry, tracker is null, path is not in the open list — `FakeWorkflowStore` transcribes that exactly, and the issue's `activeWorkflowRef()` adopt is the same object with no state. If a frontend ever did swap instances, `frontendConsidersActive` (identity, then non-empty path) plus the ADOPT `samePath` gate is what has to catch it; that is enough for the swap you described, and it is not the arm that fixes this report.

**2. LOAD cannot overwrite a still-loaded dirty tab.**
`wasOpen` is `!!target.changeTracker`, captured before `openWorkflow`. LOAD only runs when that is false, the target still has no complete state, and the frontend still names it active. `unload()` nulls `changeTracker` and `UserFile.unload()` clears `content` / `originalContent`; store close also `removeDraft(path)`. There is no in-memory tab buffer and no draft overlay left to lose. `force:true` on a dirty close already discarded that work at close time — reopen-from-disk is not a second destruction. Background tabs are not this state (`isActive` fails, LOAD does not run). A frontend that does not unload keeps the tracker, so `wasOpen` is true and LOAD is skipped. Post-close edits on the leftover canvas are the deferred close bug (store close does not switch the canvas); they were never in a tracker, and `reopened_from_disk` tells the caller the canvas is the file.

**3. The #874 capture skip is the right width.**
After this close, `describeLiveCanvasBinding` can still answer `"bound"` (same object, same uuid tag, canvas never switched) and `pointerMovedThisOpen` is false, so the pre-existing already-current arm would serialize the leftover canvas over the just-loaded tracker. The skip is required. It is not too broad: it keys on `openSettled?.loaded === true`, which only the LOAD arm sets. ADOPT keeps `loaded: false`, so capture still runs, and a different live object at the same path is a new identity — the existing bound/foreign/pointerMoved tiers then skip a leftover-canvas write. See the inline on the comment: `checkState()` reads the canvas, not the tracker; the skip is still correct because disk is authoritative after unload.

**4. `reopened_from_disk` is only on the LOAD path.**
Spread is `openSettled?.loaded === true ? { reopened_from_disk: ... } : {}`. ADOPT and the non-regression `reason: "loaded"` / `already-open` exits all set `loaded: false`, so the key is omitted. Failed loads also leave `loaded: false` and fall through to today's refusal. The `reopened` clause inside the note is now an observation of the open list (round-1 P1).

**5. ADOPT cannot grab a different workflow.**
The arm requires `!sameObject && selectorOk && samePath`, and `samePath` demands a non-empty `active.path === target.path`. Bare-filename collisions (`workflows/a/shared.json` vs `workflows/b/shared.json`) are tested and refused. Empty/untitled paths fail `!!active.path`. Two objects at one path is a store-lookup invariant break; adopting the live loaded one is the recovery this arm exists for.

**6. `openWorkflowsInBackground` optional — the reply is honest after the P1.**
The call site always passes a function; on a frontend without the method it returns `undefined` silently. `reopened` is re-read from `readOpenWorkflows()` (`w?.path === path`), not from "the call did not throw". Missing method, silent no-op, and throwing method are all pinned. Canvas still loads; the note says the tab list may still show active-but-not-open. Residual: `{right:[path]}` appends rather than restoring the original index (nit, inline).

**7. Leaving `workflow_close` alone is the right freeze scope, not a landmine for this repair.**
Store close will keep stranding the pointer. This open-path repair is what already-stranded sessions need, and a later close/open of the same path strands-then-repairs again. Teaching the panel to pick a next-active tab would replicate the service and change close behaviour. Graph commands against the leftover canvas after close are the pre-existing close hole, not something this reopen introduces. Follow-up is fine; do not fold it in here.

**8. Round-1 NO-SHIP P1s are in this SHA.**
`920ec8b1` is the commit that landed both. `reopened: alreadyListed || listedNow()` with a reader, not a snapshot; tests build the collaborators with the capability ternary. Settle runs inside `beginWorkflowReloadStep` / `finally endWorkflowReloadStep`. Wiring tests fail if either is deleted.

Would ship as a bugfix once you undraft. Not merging from here.

web/js/comfyui-mcp-panel.js
```diff
@@ -18242,8 +18301,14 @@ const GRAPH_TOOL_EXECUTORS = {
           const captureBinding = describeLiveCanvasBinding(target);
           pointerMovedThisOpen = !sameWorkflowObject(activeBefore, target);
           if (
-            captureBinding === "bound" ||
-            (captureBinding !== "foreign" && !pointerMovedThisOpen)
+            // #1575 — a state THIS command just read off disk is authoritative, and the
+            // canvas mounted behind it is whatever the closed tab left there. Serializing
+            // that over the fresh state is the #1215/#968 poisoning through a new entry
+            // point, and it can preserve nothing: nothing has run against a just-loaded
```
artokun, line 18307: Suggestion, not blocking.

The skip (`openSettled?.loaded !== true`) is the right width. After store-level close of the active tab, `describeLiveCanvasBinding` can still answer `"bound"` (same object, same uuid tag, canvas never switched) and `pointerMovedThisOpen` is false, so the pre-existing #874 already-current arm would serialize the leftover canvas over the just-loaded disk tracker. That is why this conjunct is required, and why it is not too broad (ADOPT sets `loaded: false`) or too narrow (without it, LOAD would poison through a new entry point).

The comment is slightly wrong about mechanism: `checkState()` serializes the **canvas**, not values already in the tracker. The leftover canvas *is* this workflow and *can* still hold #874 node-writes that close+unload dropped. Those are not "nothing to preserve" — they are work the close already discarded (or would have refused had capture dirtied `isModified`). Reopen-from-disk must not resurrect them; `reopened_from_disk` is the honest account. Reword the comment that way rather than widening or dropping the skip.

web/js/lib/settle-open-target.js
```diff
@@ -0,0 +1,251 @@
...
+        const list = readOpenWorkflows();
+        return Array.isArray(list) && list.some((w) => w?.path === path);
+      } catch {
+        return false;
+      }
+    };
+    const alreadyListed = listedNow();
+    if (path && !alreadyListed && typeof reopenTabInBackground === "function") {
+      try {
+        reopenTabInBackground(path);
```
artokun, line 234: Nit, not blocking.

`openWorkflowsInBackground({ right: [path] })` appends via `_.union(left, current, right)`. It does not restore the tab's original index. Harmless for the inconsistent active/open state this fixes, and `reopened` is now an observation of the list so a missing/no-op method cannot claim success.

If a follow-up restores position, use the store's own `left`/`right` split the way persistence restore does — do not splice `openWorkflowPaths`.

artokun:
Independent Codex gate SHIP for exact HEAD 920ec8b1be1efdf7d2e7a0fc9690fadddf7c89ad. Verified native open → settle repair → repaint ordering, adopt/load safety, tab-list re-read, stale-canvas protection, and caller disclosure. Targeted tests pass: 28 helper + 9 frontend-contract; typecheck, syntax, and diff check pass. Broad unit/e2e checks were environment-blocked by EPERM temp/process creation; no code failure observed. Worktree clean.

PR #1594, opened 2026-08-21
fix(comfyui-mcp#828): the proxy attribution needs a body that NAMES a responder

artokun:
Follow-up to #1591 for **artokun/comfyui-mcp#828**. Cross-repo reference, so it will not auto-close.

**Why a second PR:** #1591 merged at `6f288ddc` at 22:09:16Z while a codex gate round was still open against that exact SHA. The gate returned NO-SHIP with two P1s on it. A merge does not void a NO-SHIP — those two findings are live on `main` right now, and this carries the fixes.

### 1. A generic HTML error page was attributed to a proxy that may not be there

```js
describeHttpFailure({ status: 502, body: "<html><body>error</body></html>" })
// → "…This answer did not come from ComfyUI…"
```

ComfyUI runs on aiohttp, whose **own** default error page is `<html><head><title>502 Bad Gateway</title>…`. So neither a gateway-class status nor bare markup identifies a proxy, and the assertion contradicted the cause text immediately in front of it — which already said the response does not distinguish the candidates.

`answeredBySomethingElse` now requires the body to **NAME** a responder: a vendor/server marker (`cloudflare`, `nginx`, `traefik`, …) or a sign-in form. Everything else still reports the status, the cause candidates, and the body prefix — it just stops saying where the answer came from.

The generic status phrases (`bad gateway`, `gateway time-?out`, `error code 5\d\d`) are removed from `PROXY_MARKERS` for the same reason: the English for a status code is not evidence of who wrote it, and matching it turned aiohttp's own page into a proxy. `apache` went with them — too weak to carry an attribution on its own on a host that may serve anything.

### 2. A line-wrapped credential leaked its continuation

```
Authorization: Bearer aaaaaaaaaaaaaaaaaaaa\nbbbbbbbbbbbbbbbbbbbb
→ Authorization: «redacted» bbbbbbbbbbbbbbbbbbbb
```

The labelled rule's value stopped at the first whitespace. A labelled value now also consumes following whitespace-separated tokens that still look like credential material — they contain a digit, or run 16+ characters — so the wrapped half goes too. Ordinary prose after a labelled value keeps its words (`token: abc contact support` is unchanged), which is the direction the first round already established: fail closed on credentials, but never at the cost of the diagnosis.

### Also here

The body prefix is sliced on **code points** rather than code units, so a cut cannot land inside a surrogate pair and put a replacement character in the message.

## Where the load-bearing claims are

- **The marker set is now the entire basis for the attribution sentence.** If a marker is too generic, the sentence over-claims; if the set is too narrow, a real proxy page reports without naming itself (which is the safe direction). The set is vendor/server names only.
- **`CRED_CONT`'s "looks like credential material" heuristic** — digit-containing, or 16+ chars — is a judgement call. It over-redacts a following word that happens to be 16+ characters; that is the fail-closed direction and it is bounded to text that immediately follows a labelled secret.

## Verification

- `npm run test:unit` — **6347 pass / 0 fail**. `check-tool-vocabulary` OK, `check-panel-scope` OK.

### Mutation evidence

| mutation | test killed |
|---|---|
| restore the generic status phrases to `PROXY_MARKERS` | `aiohttp's own error page is not attributed to a proxy that is not there` |
| `answeredBySomethingElse` returns `true` unconditionally | `aiohttp's own error page is not attributed to a proxy that is not there` |
| empty the credential continuation (`CRED_CONT = ""`) | `a line-wrapped credential does not leak its continuation` |

## Not done here

`npx playwright test` was not run — there is still no e2e spec touching `free_vram`, and reaching this branch requires a proxy answering `/free` with a 502. The compiler-backed `check-panel-scope` gate over the whole bundle, `typecheck`, and the full unit suite are what was run.


artokun:
Direct Codex merge gate: NO-SHIP (P1). web/js/lib/http-failure.js:124 credential redaction stops at `%`/`~` and leaks the remaining token. A reflected body `Authorization: Bearer abc%2Fdef%3Dghi` produces `Authorization: «redacted»%2Fdef%3Dghi` in `describeHttpFailure`. Decode or match the credential value as a whole without leaking encoded suffixes, add a regression for percent-encoded credentials, rerun CI, and request a fresh gate.

artokun:
Fresh gate on ca3b13a4 found two P1s: wrapped password continuations containing punctuation leaked the suffix, and the branch still advertised panel 0.15.32 while current main/release is 0.15.33. Recovered the quota-stranded lane, fixed the continuation to redact the whole wrapped line, added a regression, and synchronized pyproject.toml plus the bundle version to 0.15.33. Targeted unit suite is 28/28 and panel scope check passes. New head 5f3848f1; requesting a fresh independent gate.

PR #1595, opened 2026-08-21
fix(1581): opening a saved workflow leaves graph tools on that canvas

### LINKED ISSUE #1581: Opening a saved workflow leaves panel graph tools fenced to stale workflow instance
After panel_open_workflow switched the active tab back to a previously saved workflow, the next panel_graph_outline call was refused because the graph command fence still carried the old workflow instance UUID. The live canvas was correct, but the read command could not run until panel_set_workflow_target({mode:"current"}) was called.
### Steps to reproduce
1. Start on workflow A. 2. Open workflow B with panel_open_workflow. 3. Immediately call panel_graph_outline. 4. Observe workflow instance mismatch and stale UUID. 5. Call panel_set_workflow_target({mode:"current"}); the same graph read succeeds.
### Exact error (scrubbed)
graph_outline was refused before it ran — workflow instance mismatch: command issued for workflow instance e39a0a22-9ab2-43b8-b9b9-f44e76acacb0, but the tab reported active workflow instance 115f55be-528e-4d7b-8af9-d732ae2e3bb6.
### Fix
applied locally: no; upstream-only: no. Workaround verified: panel_set_workflow_target({mode:"current"}) refreshed the graph binding, then panel_graph_outline succeeded. Please refresh/re-fence the graph command target automatically after panel_open_workflow completes.
### Environment
OS linux; ComfyUI 0.33.3; frontend 1.49.6; GPU NVIDIA GeForce RTX 3060 12 GB; PyTorch 2.9.1+cu130; comfyui-mcp 0.52.45; panel 0.15.32.

artokun:
Fixes #1581

After panel_open_workflow switched to an already-open saved tab, the next panel_graph_outline was refused as a workflow-instance mismatch until panel_set_workflow_target({mode:"current"}).

The open itself succeeded. What it withheld was workflow_uuid: activeWorkflowUuidForOpenReply compared the opened target with the live snapshot using !==. ComfyUI hands out Vue proxies from openWorkflows (how an already-open tab is found) while activeWorkflowRef() can yield the raw object the uuid WeakMap is keyed on. A successful switch then published no fence identity, so the MCP kept stamping graph commands for the previous instance.

The rest of this file already compares those carriers with sameWorkflowObject (#558). The open-reply gate now does the same, so a successful open of a saved workflow publishes the live uuid and the next graph read runs on that canvas. A genuinely different live tab still withholds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+- opening an already-open saved workflow publishes the live fence uuid, so the next graph read is not refused as a stale instance (#1581)
+
 ## [0.15.36] - 2026-08-22
 
 ### Fixed
@@ -28,7 +31,6 @@ All notable changes to this project are documented here. This project adheres to
 ### Fixed
 - close wrapped secret and version regressions
 - the proxy attribution needs a body that NAMES a responder
-
 
 ## [0.15.33] - 2026-08-21
 
--- a/browser_tests/unit/open-outcome.test.mjs
+++ b/browser_tests/unit/open-outcome.test.mjs
@@ -36,6 +36,7 @@ import {
   classifyOpenOutcome,
 } from "../../web/js/lib/open-outcome.js";
 import { savedWorkflowHandle } from "../../web/js/lib/bridge-route.js";
+import { rawWorkflowObject, sameWorkflowObject } from "../../web/js/lib/workflow-chat-identity.js";
 
 const HERE = dirname(fileURLToPath(import.meta.url));
 const PANEL_JS = join(HERE, "../../web/js/comfyui-mcp-panel.js");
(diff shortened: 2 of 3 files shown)


artokun:
Independent Codex gate for rebased HEAD `8fb4740561afe3acc1dada920f1b95f3d122de69d`: **SHIP**.

Evidence: `sameWorkflowObject` handles Vue proxy/raw identity while path alone cannot match (`web/js/lib/workflow-chat-identity.js:33-50`); `activeWorkflowUuidForOpenReply` remains fail-closed and requires established matching routing identities (`web/js/comfyui-mcp-panel.js:4589-4593`); the caller snapshots the live active workflow once (`web/js/comfyui-mcp-panel.js:19146-19186`); focused tests cover proxy/raw success, distinct-workflow refusal, and identity invariants (`browser_tests/unit/open-outcome.test.mjs:1135-1192`).

Validation: 154 focused tests passed, syntax check passed, `git diff --check` passed, and fresh Lint pack CI passed.

PR #1601, opened 2026-08-21
fix: publish workflow identity on graph reads (#1973)

artokun:
## Summary
- Attach the live root workflow UUID to graph read descriptors, including subgraph reads.
- Keep malformed or missing identity omitted.
- Add focused unit coverage.

This is the paired panel support for MCP #1973 / PR #2033, which now requires workflow identity before completing a budgeted get_errors audit. Focused test: node --test browser_tests/unit/graph-view-identity.test.mjs (3/3).

### DIFF
--- a/browser_tests/unit/graph-view-identity.test.mjs
+++ b/browser_tests/unit/graph-view-identity.test.mjs
@@ -1,7 +1,10 @@
 import test from "node:test";
 import assert from "node:assert/strict";
+import fs from "node:fs";
 
 import { withWorkflowUuid } from "../../web/js/lib/graph-view-identity.js";
+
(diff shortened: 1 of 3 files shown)


artokun:
Updated exact head to 2566b52e after independent gate NO-SHIP findings. The fix now derives graph-read workflow_uuid from the live active workflow identity (never a stale root tag during read-only bypass), adds viewing identity to graph_get_subgraph, graph_query seed-not-found replies, and already-at-root graph_exit_subgraph, and adds caller-level source coverage. Validation: node --test browser_tests/unit/graph-view-identity.test.mjs (5/5), node --check web/js/comfyui-mcp-panel.js, npm.cmd run typecheck, git diff --check. Please re-gate this exact head.

artokun:
Independent gate SHIP at exact head 2566b52e; required lint CI green. Approved for merge as the panel half of MCP issue #1973.

Issue #1582, opened 2026-08-21
panel_set_widget stalls 15s when live schema probes go silent despite fresh snapshot


artokun:
### What happened / root cause
panel_set_widget succeeded only after both live schema routes timed out: api.getNodeDefs() exhausted 10s and GET /object_info exhausted 5s. The tool fell back to a last-observed schema snapshot. This is confusing and adds roughly 15 seconds to each widget edit on a healthy connected canvas.

### Steps to reproduce
1. Open a 49-node ComfyUI workflow containing standard loader nodes.
2. Call panel_set_widget on UNETLoader.unet_name or ResolutionSelector.aspect_ratio.
3. Observe the mutation stalls, then succeeds with schema_source:"last-observed" and a schema_note saying both live probes went silent.
4. Repeat on LoadImage.image; same behavior.

### Exact error (scrubbed)
"The write SUCCEEDED ... live schema probe went silent ... api.getNodeDefs() did not answer within its 10000ms share ... GET /object_info did not answer within its 5000ms share."

### Fix
Applied locally: no. Upstream-only: yes.
The mutation already has a safe snapshot fallback, but the schema refresh path should avoid waiting serially on both routes when a fresh, same-connection whole-/object_info snapshot exists. Reuse the fresh snapshot immediately or probe the routes concurrently with a shorter budget, while preserving validation.

### Environment
Windows 11; NVIDIA RTX 4090 22 GB; ComfyUI 0.33.0; comfyui-mcp 0.52.45; panel 0.15.20.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-sol` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #1223, #1560, #716

artokun:
Codex is taking over this eligible severity:P2 issue as an abandoned lane. The existing gent:codex marker had no claim comment, no current wt-1582 worktree, and no matching active agent process; the only matching branch was last touched 2026-08-14. I am dispatching a fresh issue-number lane. Other agents should skip this issue. Takeover time: 2026-08-21T23:08Z.

artokun:
Claimed by grok-swarm 2026-08-21T23:08Z — driving this in worktree `wt-1582` (branch `fix/1582-set-widget-schema-stall`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
[...]

PR #1604, opened 2026-08-21
fix: bound silent schema probes with a fresh snapshot (#1582)

artokun:
Recovered abandoned issue #1582. The implementation agent committed and pushed the narrow snapshot/probe timeout fix with focused coverage. Please review against current main.


artokun:
Independent gate SHIP at exact head 538e5982; required lint CI green. Approved for merge.

PR #1615, opened 2026-08-22
fix(1589): trust live VIDEO producers for add-node socket proof

### LINKED ISSUE #1589: panel_add_node rejects core GetVideoComponents when VIDEO producer is already on canvas
`panel_add_node` refused the core `GetVideoComponents` node because its required `VIDEO` input has no widget and the full `/object_info` read did not complete within the 5 second guard. A core `LoadVideo` node that outputs `VIDEO` was already present on the live canvas, but the guard did not use the live graph as proof that the type is a socket. `panel_refresh_nodes` and two retries produced the same refusal.

### Steps to reproduce
1. Add core `LoadVideo` to a live canvas.
2. Call `panel_add_node` for `GetVideoComponents`.
3. Observe refusal: required input type `VIDEO` had no widget; installed producer status is UNKNOWN.
4. Call `panel_refresh_nodes` and retry; the same refusal remains.

### Exact error (scrubbed)
`Cannot add GetVideoComponents: input video (VIDEO): no frontend widget is registered, and whether any installed node outputs it is UNKNOWN — the full /object_info read did not complete.`

### Fix
Applied locally: no. Upstream-only: yes. The add-node guard should accept a non-widget input as a socket when the live graph already contains an output slot of the exact type, or use a bounded per-class producer lookup instead of depending on a full `/object_info` read. The current 5 second full-registry timeout creates a false negative on large installs. Temporary workaround: use installed `VHS_LoadVideo`, which directly outputs IMAGE and AUDIO.

### Environment
Windows 11; ComfyUI 0.33.3; frontend 1.49.6; Python 3.13.7; PyTorch 2.13.0+cu130; RTX 3080 Ti 12GB; comfyui-mcp 0.52.45; panel 0.14.9.

artokun:
## Summary

- Let `panel_add_node` use exact output types from the live canvas when the bounded whole `/object_info` socket-proof read does not answer.
- Preserve the existing input-shape guard so widget-valued inputs and unrelated producer types still refuse.
- Add production-path regression coverage for `GetVideoComponents` and the refusal cases.

Fixes #1589

## Validation

- `node --test browser_tests/unit/*.test.mjs` — 6431 passed, 1 todo, 0 failed
- Focused add-node suites — 137 passed
- `node --check` passed for the changed JS files
- `git diff --check` passed

The TypeScript-backed `check-panel-scope` gate could not run in the isolated worktree because `node_modules/typescript` is absent; direct full unit tests passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project are documented here. This project adheres to
 
 ### Fixed
 - opening an already-open saved workflow publishes the live fence uuid, so the next graph read is not refused as a stale instance (#1581)
+- `panel_add_node` accepts a core VIDEO socket when a live canvas producer exposes VIDEO, even if the full `/object_info` read times out (#1589)
 
 - opening a saved workflow leaves graph tools on that canvas
 - report the tab re-list from the open LIST, and hold the fence across the load
--- a/browser_tests/unit/add-node-socket-proof-scope.test.mjs
+++ b/browser_tests/unit/add-node-socket-proof-scope.test.mjs
@@ -130,6 +130,42 @@ function wildcardDictObjectInfo() {
   };
 }
 
+function videoObjectInfo({ widgetInput = false } = {}) {
+  return {
+    LoadVideo: {
+      name: "LoadVideo",
+      input: { required: {} },
+      output: ["VIDEO"],
+    },
+    GetVideoComponents: {
+      name: "GetVideoComponents",
+      input: {
+        required: {
+          video: ["VIDEO", widgetInput ? { default: "a-video-value" } : {}],
+        },
+      },
+      output: ["IMAGE", "AUDIO"],
+    },
+  };
+}
+
+function apiForObjectInfo(defs) {
+  return {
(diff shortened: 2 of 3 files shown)


artokun:
Independent Codex gate for exact rebased HEAD `ba443e649eaec4053f5ce36b099e948069cf7eff`: **SHIP**.

Evidence: the sole production helper signature/call is wired at `web/js/comfyui-mcp-panel.js:11472-11482,13712-13759`; captured graph/workflow binding and revalidation remain at `:11607-11635,13737-13749`; exact output matching and malformed-output fail-closed behavior are at `:11507-11521,13745-13751`; consumer widget protection remains enforced by `web/js/lib/node-widget-materialization.js:172-253`; added tests exercise extracted production `graph_add_node` at `browser_tests/unit/add-node-socket-proof-scope.test.mjs:498-548`.

Validation: 137 focused tests passed, both `node --check` commands passed, and `git diff origin/main...HEAD --check` passed. The TypeScript scope script was attempted but blocked only because the isolated gate worktree lacks `node_modules/typescript`; repository CI passed all gates.

PR #1616, opened 2026-08-22
Fix stale deliberate-sweep prompt snapshots (#1588)

### LINKED ISSUE #1588: panel_run deliberate sweep queues stale prompt and SaveImage widgets after first pending job
A deliberate live-canvas sweep used sequential panel_set_widget calls followed by panel_run(allow_duplicate:true) while the previous render was active. The first case was correct. The second queued case captured its new prompt but kept the previous case's five SaveImage filename_prefix values. Every later queued prompt id then rendered the same second-case pixels and the same stale output path despite receiving distinct requested case mutations. After the queue drained, panel_query_graph showed the canvas stuck at the second prompt and first-case prefixes. Five independently completed runs had different PNG metadata hashes but identical decoded RGB hashes lane-for-lane, proving stale workflow snapshots rather than coincidentally similar renders.

### Steps to reproduce
1. Open a graph with one text prompt widget and multiple SaveImage filename_prefix widgets.
2. Set case A prompt and prefixes; call panel_run.
3. While A renders, set case B prompt and prefixes sequentially; call panel_run({allow_duplicate:true}).
4. Repeat for cases C through G, awaiting every tool call.
5. Observe distinct queued prompt_ids but stale B prompt / A prefixes in later outputs and in the final live graph.

### Exact error (scrubbed)
No JavaScript exception reached the caller. Every panel_run returned queued:true with a distinct prompt_id. The failure was silent stale state. Relative output path example: Krea2_adult_component_work_20260822_source_C/stage0_strength/T04_s2026082201, while the pixels and prompt were from T08.

### Fix
One focused source inspection found no safe single-file patch that can be activated mid-task: the failure crosses the browser graph-command transaction and the long-lived orchestrator queue snapshot boundary. panel_reload cannot reload the orchestrator. Precise change needed: panel_run must establish a barrier that confirms every preceding graph mutation is committed to the same live graph revision before serializing the prompt, or fail closed with a stale-mutation/revision error. Add a regression test with one running and one pending render that alternates a text widget plus five SaveImage prefixes; inspect each queued payload and require its own exact values. The current workaround is one case at a time: wait for idle, apply writes, read them back with panel_query_graph, then queue.

### Environment
Windows 11; ComfyUI 0.33.1; ComfyUI frontend 1.48.7; NVIDIA GeForce RTX 5090 32 GB; comfyui-mcp 0.52.45; panel 0.15.22.

artokun:
Fixes #1588

Summary:
- Preserve each single-prompt graph_run preflight prompt through ComfyUI's deferred busy-queue serialization.
- Keep later graph_run preflights out of the pending snapshot reservations and retain queue rejection cleanup.
- Add a production-callsite regression covering three queued cases with five SaveImage filename_prefix widgets each.

Validation:
- node --test browser_tests/unit/run-command-budget.test.mjs
- node --test browser_tests/unit/run-scope-guard.test.mjs browser_tests/unit/widget-null-safety.test.mjs
- npm.cmd run check:scope
- ES module syntax and git diff --check

Not merged; changelog entry added under the current release Fixed section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ All notable changes to this project are documented here. This project adheres to
 - report the tab re-list from the open LIST, and hold the fence across the load
 - the adopt arm requires the same workflow FILE, not just a matching selector
 - reopening a closed tab loads the workflow the store's early return skipped
-
+- deliberate-sweep runs retain their prompt and SaveImage filename widgets when the frontend queue is already busy (#1588)
 
 ## [0.15.36] - 2026-08-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ All notable changes to this project are documented here. This project adheres to
 - a finished stills render notifies the agent before the orchestrator fallback (#1612)
 - trust live VIDEO producers for add-node socket proof
 
+### Fixed
+- deliberate-sweep runs retain their prompt and SaveImage filename widgets when the frontend queue is already busy (#1588)
 
 ## [0.15.37] - 2026-08-22
 
@@ -24,7 +26,6 @@ All notable changes to this project are documented here. This project adheres to
 - report the tab re-list from the open LIST, and hold the fence across the load
 - the adopt arm requires the same workflow FILE, not just a matching selector
 - reopening a closed tab loads the workflow the store's early return skipped
-- deliberate-sweep runs retain their prompt and SaveImage filename widgets when the frontend queue is already busy (#1588)
 
 ## [0.15.36] - 2026-08-22
 
--- a/browser_tests/unit/run-command-budget.test.mjs
+++ b/browser_tests/unit/run-command-budget.test.mjs
@@ -627,6 +627,51 @@ test("#1588: deferred serialization runs queue hooks on the correct graph snapsh
   }
 });
 
+test("#1588 P1: a foreign graphToPrompt between the hook and queue serializer cannot consume the item snapshot", async () => {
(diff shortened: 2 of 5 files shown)


artokun:
Independent exact-head gate: NO-SHIP on ea79ea6f.

Blocking evidence:
- P1: the batch=1 preflight snapshot is returned directly by the barrier, bypassing ComfyUI beforeQueued -> graphToPrompt queue hooks (reproduced seed 10 -> 11, queued snapshot remained 10).
- P1: the global reservation stack is not tied to a queue item; intervening production graphToPrompt calls (Apps conversion and scoped dispatch) can consume a pending full-run reservation.
- P2: null-safety is installed after the first preflight, so a first-pass null widget can prevent a reservation.

Focused suites: 231/231 pass; diff check passes; check:scope blocked only because TypeScript is absent in the detached gate worktree. Do not merge this head; implementation is being re-engaged against these findings.

artokun:
Independent exact-head gate: SHIP pending CI. Verified head a0a761323e731de17048b28a2a1bfd9bc69f46b5, clean detached gate, 134 focused tests passing across run-command-budget, graph-to-prompt-failure, scoped-batch-seed, and widget-null-safety, plus clean git diff --check. Reproduced the prior P1 adversarial gap: foreign graphToPrompt observes seed 11, the real queue serializer also uses seed 11, and the live widget is restored to 20. Implementation worktree additionally reports full unit 6457 passed / 0 failed / 1 pre-existing todo, typecheck and scope checks pass. Merge remains gated on CI for this exact head.

Issue #1600, opened 2026-08-21
panel_update_node fails with Manager KeyError for documented installed directory id


artokun:
### What happened / root cause
panel_update_node documents that an installed custom-node pack directory name is accepted as id. Calling it with id "comfyui-minimax-h3-prompt-enhancer-T8" and version "nightly" forwarded that normalized directory name into ComfyUI-Manager unified_update. Manager's active_nodes map had no matching normalized key and raised KeyError. The panel tool returned an internal Manager traceback instead of resolving the installed directory to its registry/Git identity or returning a clean "unmanaged pack" result.

### Steps to reproduce
1. Have an installed custom node directory named comfyui-minimax-h3-prompt-enhancer-T8 that is loaded by ComfyUI.
2. Call panel_update_node({id:"comfyui-minimax-h3-prompt-enhancer-T8", version:"nightly"}).
3. The operation fails with KeyError and no update is applied.

### Exact error (scrubbed)
KeyError: 'comfyui-minimax-h3-prompt-enhancer-t8'
at comfyui_manager/glob/manager_core.py unified_update -> self.active_nodes[node_id]

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: panel_update_node should resolve an installed directory/name against Manager's active_nodes/registry metadata before submitting the mutation. If no managed entry exists, detect a local Git repository and offer/perform the supported git update path, or return a structured unmanaged_pack error. It should not forward a missing key and surface an internal traceback.

### Environment
Windows 11; ComfyUI 0.33.0; frontend 1.49.6; Python 3.13.12; PyTorch 2.13.0+cu130; NVIDIA RTX 4090 22.5 GB; comfyui-mcp 0.52.45; panel 0.15.32.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-sol` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Claimed 2026-08-21T23:27Z after checking the issue history and panel worktrees: no active claim, PR, or issue-number worktree was present. Dispatching wt-1600-codex on ix/1600-manager-installed-id to reproduce the documented installed-directory Manager KeyError, implement the narrow fix, add focused coverage, and post the PR here. If this claim is older than 8h with no PR, treat it as abandoned and pick it up.

PR #1629, opened 2026-08-22
fix(1625): restart confirmation card appears on the active panel instead of timing out unseen

### LINKED ISSUE #1625: Restart confirmation card can time out without reaching active panel
panel_restart_comfyui was called from the active sidebar after confirming the render queue was idle. The tool reported that its confirmation card was not answered within 90 seconds and therefore did not restart. No card was visible/usable in the active panel. The tool then required the agent to fall back to the headless restart_comfyui capability.

The confirmation delivery/response path can fail when the tab is backgrounded, reconnecting, or otherwise misses the interactive card, leaving an autonomous recovery workflow unable to perform the guarded restart.

### Steps to reproduce
1. Keep a ComfyUI workflow open in the sidebar panel.
2. Ensure queue is idle.
3. Call panel_restart_comfyui({}).
4. Observe the tool waits 90 seconds, then reports no confirmation and no restart, even though the panel session is active.

### Exact error
"No confirmation received within 90s, so I did NOT restart ComfyUI. The panel tab may be backgrounded or still reconnecting after a previous restart, so the confirmation card wasn't answered."

### Fix
applied locally: no
upstream-only: yes

Precise change needed: make restart confirmation delivery durable across background/reconnect state (journal/replay the card and accepted response), or return immediately with a stable pending confirmation id that a later call can resume without a 90-second blocked tool call. The current fallback to headless restart works but bypasses the panel's guarded confirmation UX.

### Environment
Windows 11
ComfyUI 0.33.3
ComfyUI frontend 1.49.6
GPU NVIDIA GeForce RTX 5070 Laptop GPU, 8 GB VRAM
comfyui-mcp 0.52.45
panel 0.15.32

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-sol` · effort `high` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #1554

artokun:
Fixes #1625

## What the user now observes
`panel_restart_comfyui` shows its Yes/No confirmation card on the Agent tab instead of waiting 90s and reporting that no confirmation arrived. The card is reachable even when the Agent tab was not selected, the user had scrolled up, or the ComfyUI tab was backgrounded.

## Why
The restart confirm is an `ask_user` card. The panel appended it to the chat log and called the Discord-style autoscroll helper, which (a) leaves a scrolled-up reader where they are with only a "New messages" pill, and (b) defers to `requestAnimationFrame`, which a backgrounded tab never fires (the same class as the hidden-tab stream bug). Keep-alive also detaches `.cmcp-root` when another sidebar tab is selected, so the card sat in a tree that was not on screen — the What's New path already names that: "painting into a transcript the user cannot see is the same failure as not painting at all."

The orchestrator then hit the 90s cap and reported the card was not answered. Related orchestrator recovery for a *late* click is panel#1554; this is the case where nobody could click because the card never reached the screen.

## Fix
After painting `ask_user` / `request_secret`, call `revealInteractiveCard`: open the Agent tab, force stick-to-bottom, and scroll **now** (not on rAF). Retry once after 120ms for the same tab-store race Ask AI / What's New already wait out.

A card that never arrived (superseded socket) is unchanged — #952 still retires those.

## Tests
`browser_tests/unit/interactive-card-reveal.test.mjs` drives the shipped helper (sync reveal, retry, throw isolation) and pins both painters to call it after `log.appendChild(card)` with the live `openSidebarTab` / stick / `log.scrollTop` injections.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+- a restart confirmation card is brought on screen instead of timing out unseen: the Agent tab opens, stick-to-bottom is forced, and the card scrolls immediately rather than waiting on an rAF a backgrounded tab will never fire (#1625)
+
 ## [0.15.41] - 2026-08-22
 
 ### Fixed
@@ -17,7 +20,6 @@ All notable changes to this project are documented here. This project adheres to
 ### Fixed
 - panel_open_workflow keeps authored node size/order and leaves a clean tab unmodified (#1621)
 - replay remote PreviewImage completion (#1624)
-
 
 ## [0.15.39] - 2026-08-22
 
--- a/browser_tests/unit/interactive-card-reveal.test.mjs
+++ b/browser_tests/unit/interactive-card-reveal.test.mjs
@@ -1,0 +1,167 @@
+/**
+ * #1625 — a restart confirmation painted into a transcript the user cannot see
+ * is the same failure as not painting at all.
+ *
+ * `panel_restart_comfyui` blocks on `ask_user`. The panel used to append the card
+ * and call `scrollLog()`, which (a) honours stick-to-bottom, so a scrolled-up
+ * reader never sees it, and (b) defers to `requestAnimationFrame`, which a
+ * backgrounded ComfyUI tab never fires. The orchestrator then waits 90s and
+ * reports the card wasn't answered.
+ *
+ * Two layers, same shape as interactive-card-fence.test.mjs:
+ *   1. the shipped helper (web/js/lib/interactive-card-reveal.js);
+ *   2. the REAL `paintQuestion` / `paintSecret` bodies, so a painter that goes
+ *      back to `scrollLog()` alone fails here rather than only if the helper
+ *      changes.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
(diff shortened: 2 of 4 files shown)


artokun:
Direct Codex merge gate: SHIP

Exact head e7cc25e3 reviewed against current origin/main. The fix reveals both blocking card painters after append: it forces stick-to-bottom, activates the Agent tab, scrolls synchronously (avoiding hidden-tab rAF starvation), and retries activation/scroll after the existing short race window. Each injected UI action is isolated so a detached tab cannot suppress the remaining reveal steps. The change does not alter card ownership, response routing, or the 90-second confirmation semantics.

Evidence: 49 focused interactive-card/fence tests passed; npm.cmd run check:scope passed; npm.cmd run typecheck passed; the PR lint check is green at the posted head.

artokun:
Direct Codex merge gate: SHIP (rebased exact head)

Exact head cebf4367 reviewed again against origin/main after the #1600 merge. The rebase conflict was only the expected CHANGELOG Unreleased section; the production and test files are unchanged. The same reveal helper remains wired after card append for both blocking painters, with synchronous tab activation/scroll and isolated retry behavior.

Evidence on cebf4367: 49 focused interactive-card/fence tests passed; npm.cmd run check:scope passed; npm.cmd run typecheck passed. Awaiting the refreshed required lint check.

PR #1646, opened 2026-08-22
fix(1645): panel_list_nodes returns an inspectable fallback when Manager is unreachable

### LINKED ISSUE #1645: panel_list_nodes fails when ComfyUI-Manager is unreachable instead of returning an inspectable fallback
A read-only inventory call to panel_list_nodes returned an error when ComfyUI-Manager was unreachable. The live canvas itself was connected and panel_graph_outline worked. This prevents inventorying installed custom-node packs through the panel and forces a filesystem fallback.

### Steps to reproduce
1. Connect the sidebar panel to a running ComfyUI without a reachable built-in Manager endpoint.
2. Call panel_list_nodes({}).

### Exact error
Error: ComfyUI-Manager injoignable (le gestionnaire intégré est-il activé ?)

### Fix
Applied locally: no (the user explicitly requested no modifications).
Suggested upstream change: return a structured unavailable state and, if possible, fall back to the loaded /object_info node registry or a local pack scan for this read-only operation.

### Environment
Windows 11; ComfyUI 0.26.0; frontend 1.45.19; NVIDIA GeForce GTX 1660 Ti 6 GB; Python 3.12.12; PyTorch 2.11.0+cu128; comfyui-mcp 0.52.60; panel 0.15.45.

---

Related (same root cause): #426

artokun:
Fixes #1645

A read-only `panel_list_nodes({})` returned `Error: ComfyUI-Manager injoignable (le gestionnaire intégré est-il activé ?)` when the built-in Manager was unreachable. The live canvas was connected and `panel_graph_outline` worked.

`nodes_list` already retried the absolute `/customnode/installed` on an unreachable dialect-routed GET, then rethrew when that was unreachable too. `searchNodesVia` already degrades past that last throw; list did not.

`listNodesVia` now walks the same ladder as search:

1. dialect-routed GET
2. absolute legacy retry on unreachable/404
3. reconstruct loaded custom-node packs from `/object_info` (`python_module` folders)
4. structured `{supported:false, managerReachable:false, installed:{}}` if object_info also fails

Genuine server errors (500/403) still propagate. Transient reconnect transport failures still go through `retryDuringReconnect`.

### DIFF
--- a/browser_tests/unit/nodes-list-manager-unreachable.test.mjs
+++ b/browser_tests/unit/nodes-list-manager-unreachable.test.mjs
@@ -166,8 +166,24 @@ test("#1645 core-only object_info is still an inspectable inventory, not a throw
   });
   assert.equal(res.managerReachable, false);
   assert.equal(res.source, "object_info");
-  assert.deepEqual(res.installed, {});
+  assert.deepEqual(res.installed, Object.create(null));
   assert.match(res.note, /object_info/);
+});
+
+test("#1645 object_info pack names cannot reach inherited object keys", async () => {
+  const res = await listNodesVia(throwUnreachable, throwUnreachable, {
+    args: {},
+    objectInfoGet: async () => ({
+      ProtoNode: { python_module: "custom_nodes.__proto__" },
+      ConstructorNode: { python_module: "custom_nodes.constructor" },
+      ToStringNode: { python_module: "custom_nodes.toString" },
+    }),
+  });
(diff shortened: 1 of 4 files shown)


artokun:
Re-engaged abandoned worktree and rebased onto current main. Added a null-prototype pack map for untrusted /object_info python_module names, plus __proto__/constructor/toString regression coverage after independent gate found a P1 inherited-key throw. Focused suite: 25/25.

PR #1741, opened 2026-08-24
fix(1739): an abandoned /object_info stops occupying the backend it starved

### LINKED ISSUE #1739: panel_set_widget refuses ALL writes on a large install: /object_info is 52MB/16s, and the type-scoped fallback (which answers in 1ms) is reported as timing out

On an install with many node packs, `panel_set_widget` refuses every write with the `#458` "cannot verify the node type against the ComfyUI backend" error. The node is present and healthy; the problem is purely the schema-verification budget.

Measured on this machine:

```
GET /object_info                    -> http=200  time=16.08s  size=51,957,409 (52 MB, 8534 node types)
GET /object_info/CyberpunkWindowNode -> http=200  time=0.0117s size=6,821
GET /object_info/CyberpunkWindowNode -> http=200  time=0.0013s size=6,821
GET /object_info/CyberpunkWindowNode -> http=200  time=0.0013s size=6,821
```

Two distinct problems:

1. **Whole-schema budget is far too small for a large install.** The error says `GET /object_info did not answer within its 5000ms share of the 20000ms budget`. The endpoint genuinely needs ~16s here, so it can never succeed. 5s is not a plausible ceiling for an install with thousands of node types. `panel_refresh_nodes` fails the same way (`refresh_still_running`, then a 30s `did not reply to "refresh_nodes"` timeout).

2. **The type-scoped fallback is reported as timing out even though it is effectively instant.** The error also says `the type-scoped /object_info reads (CyberpunkWindowNode) did not all answer within 5000ms`. That exact URL returns 6.8 KB in **1-12 ms** via curl, reproducibly. This is the fallback that exists precisely for the case where the whole schema is too big, and it is the one that would work -- so this looks like a real bug, not just slowness. Most likely it is queued behind / sharing the already-exhausted 20s budget with the 52 MB whole-schema fetch (head-of-line blocking through the browser's `api.getNodeDefs()`), rather than being issued as an independent request with its own budget.

Net effect: on a big install the agent cannot write ANY widget through the panel. Every `panel_set_widget` is refused, with no working path forward except asking the user to reload the ComfyUI tab by hand.

### Steps to reproduce

1. Have a ComfyUI install with enough custom node packs that `/object_info` is large (here: 8534 types / 52 MB / ~16s).
2. Open a workflow in the panel.
3. Call `panel_set_widget` on any node.
4. It is refused with the `#458` message; both the whole-schema routes and the type-scoped fallback are reported as timed out, even though the type-scoped URL answers in ~1ms.

### Exact error (scrubbed)

```
Error: Cannot set widget on node 58 ("CyberpunkWindowNode"): cannot verify the node type
against the ComfyUI backend - no usable /object_info schema was obtained. Tried 2 routes:
api.getNodeDefs() did not answer within its 10000ms share of the 20000ms budget;
GET /object_info did not answer within its 5000ms share of the 20000ms budget.
The last-observed schema was not usable either - no whole /object_info has been observed
on this backend connection yet. A type-scoped /object_info read did not stand in for the
whole-schema routes either - the type-scoped /object_info reads (CyberpunkWindowNode)
did not all answer within 5000ms. Refusing to write rather than trust a possibly-stale
node cache (#458).
```

### Fix

applied locally: no   upstream-only: yes

Could not locate the budget constants to patch: the machine has 44 npx-cached `comfyui-mcp` copies and the error strings are not in the `dist/` I grepped, so the verification appears to live in the panel's browser JS (`api.getNodeDefs()`), and the orchestrator would need a full process restart to pick up a patch anyway. Stopping at one attempt per the skill's rule.

Precise changes needed:

1. **Make the type-scoped read an independent, first-class route with its own budget.** When a node type is known (it always is for `panel_set_widget`), try `GET /object_info/<Type>` FIRST, before the whole-schema routes -- it is ~6 KB and ~1 ms even on this install, versus 52 MB / 16 s. That alone fixes this class of failure and is strictly cheaper in the common case. It must not share or inherit the exhausted whole-schema budget.

2. **Scale the whole-schema timeout, or drop the fixed 5s share.** A hard 5s for `GET /object_info` is unreachable on large installs. Either make it adaptive (e.g. derive from a previous observed duration), raise it substantially, or make it configurable via env var.

3. **Distinguish "slow" from "unreachable" in the message.** The current text sends the reader toward "reconnect ComfyUI", but the backend is perfectly healthy and fast -- it is just big. Reporting the observed size/duration would make this self-diagnosing.

### Environment

OS: Windows 11 Enterprise 10.0.22621
ComfyUI: 0.33.0
ComfyUI FRONTEND: 1.49.6
GPU: NVIDIA GeForce RTX 5090 (32 GB VRAM), CUDA 13.2, torch 2.13.0, 384 GB RAM
Python: 3.13.11
comfyui-mcp: 0.52.93
panel: 0.15.69
node types in /object_info: 8534 (52 MB, ~16s)
Launch args: main.py --output-directory ~/... --user-directory ~/... --enable-manager --enable-manager-legacy-ui --enable-compress-response-body --enable-cors-header --use-ck-attention --cuda-device all

artokun:
Fixes #1739

## The reported failure

On an 8534-type install, **every** `panel_set_widget` was refused, and the refusal claimed
the type-scoped `/object_info/<Type>` read had timed out — on a URL the reporter measured
by hand at 1.2–12 ms, reproducibly, on the same machine at the same time:

```
GET /object_info                     -> 200  16.08s   51,957,409 bytes
GET /object_info/CyberpunkWindowNode -> 200   0.0013s       6,821 bytes
```

## Root cause: the panel starves its own working route

The type-scoped read was not slow. It was queued behind two whole-map computations **this
panel** had left running on a backend that can only do one thing at a time.

ComfyUI's own `server.py`:

```python
@routes.get("/object_info")
async def get_object_info(request):
    asset_seeder.start(...)
    with folder_paths.cache_helper:
        out = {}
        for x in nodes.NODE_CLASS_MAPPINGS:     # 8534 types on the report
            out[x] = node_info(x)
        return web.json_response(out)           # json.dumps of ~52 MB

@routes.get("/object_info/{node_class}")        # twelve lines below it
async def get_object_info_node(request): ...
```

There is **no `await` anywhere in the first body**. It blocks aiohttp's single event loop
for its entire run — the reporter's 16.08 s — and while it runs the backend cannot serve
*any* other request, including the per-class route. The reporter's curl timings were taken
with nothing else in flight; the panel's were taken with two whole-map computations sitting
on that loop.

So the reporter's head-of-line-blocking guess is right, but the queue is on the **server**,
not in the browser's `api.getNodeDefs()`.

The panel's second whole-map request is pure cost on such a backend, and provably
unreachable: it is only issued *after* the client route has spent its whole share, so it
starts behind a computation of the identical document that is still running, and its own
share is smaller than the one that just expired. It cannot answer. It can only add another
full computation to the loop — and #1560's type-scoped last resort, issued moments later
and written for exactly this backend, is what pays for it.

## The change

`fetchWholeObjectInfo` now **cancels its raw `GET /object_info` when it stops waiting on
it**, and the panel forwards the init so `api.fetchApi(route, init)` carries the signal.
The abandoned request is dropped before the event loop ever reaches it, and the loop is
handed back to the question that can be answered.

`api.getNodeDefs()` is deliberately **not** cancelled: it is the frontend's own method and
takes no signal, and it goes first — it is the work the fallback was queued behind, not the
thing blocking the fallback.

Three files: the oracle (+1 helper, +2 cancel calls), five call-site wirings in the panel,
and one test file.

## Where the load-bearing claims are — please check these

1. **`web/js/lib/object-info-oracle.js:854`** — the unconditional `httpRoute.cancel()` after
   the branch. It is written once *after* the if/else chain rather than in each arm, on the
   argument that every path reaching it either never got a body or already consumed one.
   The two early returns above it (`usableDefs` / `authoritativeEmpty`) skip it on purpose,
   because `abort()` after a fully-read response is a no-op. If that reasoning is wrong for
   any arm, this leaks the request it exists to stop.
2. **Silence must stay silence.** #1223's snapshot and #1560's type-scoped read are both
   licensed on `noBackendAnswerEstablished` — a route that *answered* something unusable may
   never be overruled. Cancelling introduces a new rejection into that path, and if it were
   recorded as a `THREW` outcome instead of `NO_ANSWER` it would silently disable both
   fallbacks. Pinned by test 2; worth re-checking by eye.
3. **The backend model in the test.** It asserts one thing that cannot be proven from here:
   a queued request aborted before dispatch is never run (aiohttp reads a closed connection),
   while a request the loop has already *started* keeps the loop until its synchronous body
   returns. The second half is why `getNodeDefs` is not cancelled. If the first half is
   wrong, the fix does nothing on the reporter's machine — it still cannot make anything
   worse, but it would not help either.
4. **A named trade.** Before this change, an abandoned 52 MB body kept downloading and could
   in principle land in the browser's HTTP cache for a later request to reuse. It is
   aborted now, so it cannot. That reuse was speculative (the panel's own #716 burst cache is
   the real dedupe, and it stores parsed defs), and it only applies on backends where the
   current behaviour is already a permanent refusal — but it is a real difference and is not
   hidden.

## Evidence

**Reproduced by execution, then fixed — mutation-verified.** With the `httpRoute.cancel()`
line deleted, the new test drives the *real* `runSetWidget` ladder against a single-lane
backend on the *shipped* constants (20 000 ms oracle deadline, 5 000 ms scoped deadline,
16 080 ms whole-map compute) and produces the reporter's error verbatim:

```
Error: Cannot set widget on node 58 ("CyberpunkWindowNode"): cannot verify the node type
against the ComfyUI backend — no usable /object_info schema was obtained. A type-scoped
/object_info read did not stand in for the whole-schema routes either — the type-scoped
/object_info reads (CyberpunkWindowNode) did not all answer within 5000ms. ...
```

With the line restored, the write lands and the backend log shows the abandoned whole map
was dropped before the loop reached it. Everything runs on a virtual clock so the shipped
constants are used verbatim — a test that rescales the budgets proves something about
numbers the panel does not have.

**The call site is pinned too.** A one-line wiring change is invisible to a helper-level
test: the oracle can cancel perfectly while `api.fetchApi(route)` drops the signal on the
floor. Reverting a single wiring back to `(route) => api.fetchApi(route)` fails test 6 —
verified.

- `npm run test:unit` — **6710 pass, 0 fail, 1 todo** (includes `check-panel-scope` OK,
  i18n, tool-vocabulary)
- `npx tsc --noEmit` — clean
- Rebased onto `d8d84f19` and re-run green.

**Browser gate: NOT run, and I could not run it.** `playwright.config.ts` requires a real
ComfyUI at `localhost:8188` (it does not start one); nothing is listening in 8180–8210 on
this machine, and this run was told not to start one. So this PR does **not** carry e2e
coverage. Mitigating: the changed path only fires when both whole-schema routes time out,
which does not happen against a healthy local ComfyUI, so the suite would have exercised
the unchanged path either way.

## Deliberately not done

- **The reporter's fix #2 — scale the whole-schema budget.** There is a real defect here
  and it is worth stating precisely: the 20 000 ms deadline is split 10 000 / 5 000 / 5 000,
  so **no single route can ever spend more than 10 s** and a backend whose whole map takes
  11–20 s is unreachable even though the budget nominally covers it. Fixing it means writes
  that cost ~16 s each, which does not fit under `SET_WIDGET_COMMAND_BUDGET_MS` (25 000)
  alongside the 8 000 ms history-seed wait. That is a budget redesign, not a one-round
  change, and it is a separate reviewable decision.
- **The reporter's fix #1 as literally written — try the type-scoped route FIRST.** It
  would work, and it regresses healthy installs: the whole map is what populates #1223's
  snapshot and the `wasTypeEverDefined` trust root, and a scoped map deliberately cannot.
  Gating it on remembered per-connection silence is the honest version, and it touches
  #1582's skip mechanism — again a separate decision.
- **The reporter's fix #3 — report observed size/duration in the refusal.** The refusal
  wording is deliberately restricted to what the code observed; on a route that timed out,
  the panel has observed neither a size nor a duration, so it would be asserting something
  it did not establish. That is the #982 defect this file spends three screens avoiding.
- `panel_refresh_nodes` on such an install is unhelped, exactly as `object-info-oracle.js`
  already documents for #1560. It needs a whole map that lands.
- No CHANGELOG entry: this repo's changelog is written by the release commit, not by fix PRs.

### DIFF
--- a/browser_tests/unit/object-info-abandoned-request.test.mjs
+++ b/browser_tests/unit/object-info-abandoned-request.test.mjs
@@ -1,0 +1,408 @@
+/**
+ * #1739 — on an install with 8534 node types `panel_set_widget` refused EVERY write, and
+ * the refusal said the type-scoped `/object_info/<Type>` read had timed out on a URL the
+ * reporter measured by hand at 1.2–12 ms, reproducibly, on the same machine at the same
+ * time. Run with `node --test`.
+ *
+ * The measurements from the report:
+ *
+ *     GET /object_info                     -> 200  16.08 s   51,957,409 bytes
+ *     GET /object_info/CyberpunkWindowNode -> 200   0.0013 s       6,821 bytes
+ *
+ * THE TYPE-SCOPED READ WAS NOT SLOW. It was queued behind two whole-map computations the
+ * PANEL had left running on a backend that can only do one thing at a time. ComfyUI's own
+ * `server.py`:
+ *
+ *     @routes.get("/object_info")
+ *     async def get_object_info(request):
+ *         asset_seeder.start(...)
+ *         with folder_paths.cache_helper:
+ *             out = {}
+ *             for x in nodes.NODE_CLASS_MAPPINGS:
+ *                 out[x] = node_info(x)
+ *             return web.json_response(out)
+ *
+ *     @routes.get("/object_info/{node_class}")        # twelve lines below it
+ *     async def get_object_info_node(request): ...
+ *
+ * There is no `await` anywhere in the first body. It blocks aiohttp's single event loop for
+ * as long as it runs, and while it runs the second route cannot be served at all. So the
+ * ONE route that can answer an install this size was starved by the two that provably
+ * cannot — and #1560's last resort, written for exactly this backend, never got a turn.
+ *
+ * The oracle now CANCELS its raw `GET /object_info` when it stops waiting on it. That
+ * request is pure cost on such a backend: it is only issued after the client route has
+ * already spent its whole share, so it starts behind a computation of the identical
+ * document that is still running, and its own share is smaller than the one that just
+ * expired. Dropping it hands the loop back to the question that can be answered.
+ *
+ * THE BACKEND BELOW IS A MODEL, AND ITS ONE ASSUMPTION IS STATED. A queued request that is
+ * aborted before the loop reaches it is never dispatched (aiohttp reads a closed
+ * connection); a request the loop has ALREADY started keeps the loop until its synchronous
+ * body returns, because nothing can cancel a Python `for` loop from the client side. Both
+ * halves are modelled, and the second is why `getNodeDefs` — which takes no signal and goes
+ * first — is not cancelled and does not need to be.
+ *
+ * Everything runs on a VIRTUAL clock so the SHIPPED constants can be used verbatim: a test
+ * that rescales the budgets proves something about numbers the panel does not have.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+
+import {
+  fetchWholeObjectInfo,
+  OBJECT_INFO_DEADLINE_MS,
+  TRANSPORT_OUTCOME,
+} from "../../web/js/lib/object-info-oracle.js";
+import {
+  fetchTypeScopedObjectInfo,
+  SCOPED_OBJECT_INFO_DEADLINE_MS,
+} from "../../web/js/lib/scoped-object-info.js";
+import { noBackendAnswerEstablished } from "../../web/js/lib/object-info-snapshot.js";
+import { runSetWidget } from "../../web/js/lib/set-widget.js";
+
+const PANEL_JS = fileURLToPath(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url));
+const HOOKS = { beforeChange() {}, afterChange() {}, setDirty() {} };
+
+/** The reporter's own numbers, so the arithmetic below is theirs and not an invention. */
+const WHOLE_MAP_MS = 16080;
+const PER_CLASS_MS = 12;
+const REPORTED_TYPE = "CyberpunkWindowNode";
+
+// ─────────────────────────────────────── virtual clock ───────────────────────────────────
+
+/**
+ * A monotonic clock plus the `{setTimer, clearTimer}` pair `bounded-step.js` accepts, so
+ * every wait in this file — the oracle's, the scoped read's, and the fake backend's own
+ * work — is measured against ONE timeline that the test advances by hand.
+ */
+function makeVirtualClock() {
+  let now = 0;
+  let seq = 0;
+  const pending = new Map();
+  return {
+    now: () => now,
+    at: () => now,
+    timers: {
+      setTimer(fn, delay) {
+        const id = (seq += 1);
+        pending.set(id, { at: now + Math.max(0, Number(delay) || 0), fn, seq: id });
+        return id;
+      },
+      clearTimer(id) {
+        pending.delete(id);
+      },
+    },
+    hasTimers: () => pending.size > 0,
+    /** Fire the single earliest timer, ties broken by arming order. */
+    advance() {
+      let pick = null;
+      for (const [id, t] of pending) {
+        if (!pick || t.at < pick.t.at || (t.at === pick.t.at && t.seq < pick.t.seq)) pick = { id, t };
+      }
+      if (!pick) return false;
+      pending.delete(pick.id);
+      now = Math.max(now, pick.t.at);
+      pick.t.fn();
+      return true;
+    },
+  };
+}
+
+const turn = () => new Promise((resolve) => setImmediate(resolve));
+
+/**
+ * Run `promise` to completion, advancing the virtual clock whenever nothing else can make
+ * progress. A settle-check between every step is what keeps a test from firing timers the
+ * production code has already cleared.
+ */
+async function drive(clock, promise) {
+  let settled = false;
+  let value;
+  let failure = null;
+  promise.then(
+    (v) => {
+      settled = true;
+      value = v;
+    },
+    (e) => {
+      settled = true;
+      failure = e ?? new Error("rejected with no value");
+    },
+  );
+  for (let guard = 0; guard < 100000; guard += 1) {
+    await turn();
+    if (settled) break;
+    if (!clock.advance()) {
+      await turn();
+      if (settled) break;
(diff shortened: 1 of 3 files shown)


artokun:
**Independent review gate — codex, PASS.**

```
node .claude/autopilot/codex-gate.mjs --worktree <wt-1739-cc> --base origin/main --name 1739-object-info-abandoned-request
verdict: SHIP   (exit 0)
```

Gated SHA `d132718d8067c20c61697f668e93a16f34c6a122` — the same SHA as the PR head and the one that will merge. CI `Lint pack` passed on it (2m48s).

Disclosing that the verdict was one word with no reasoning, so it is not carrying much on its own. The evidence that actually constrains this change is in the PR body: deleting the `httpRoute.cancel()` line reproduces the reporter's error string verbatim through the real `runSetWidget` ladder, and reverting a single panel wiring back to `(route) => api.fetchApi(route)` fails the call-site test. Both mutations were run, not assumed.

artokun:
**Re-gated after a rebase — codex PASS on the SHA that will merge.**

Main moved twice during this run (`d8d84f19`, then `9a3a9bbf release: 0.15.73`). The first gate's verdict named `d132718d`, so it did not carry past the rebase and was not banked.

```
node .claude/autopilot/codex-gate.mjs --worktree <wt-1739-cc> --base origin/main --name 1739-abandoned-request-r2
verdict: SHIP   (exit 0)
```

- Gated SHA: `485a676adc2fb91333c84f479834ea9d064c6d69` = current PR head.
- CI `Lint pack` **pass** (2m34s) on that same SHA.
- `npm run test:unit` re-run after the rebase: **6710 pass, 0 fail** (`check-panel-scope` OK). `npx tsc --noEmit` clean.
- `0.15.73` touched only `CHANGELOG.md`, `package.json`, `package-lock.json` — none of the three files this PR changes, so the rebase was content-free.

PR #1751, opened 2026-08-24
fix(1750): a slot may not name a link the graph does not have

### LINKED ISSUE #1750: Fix panel_disconnect leaving orphaned optional-input link ids
panel_disconnect reported success for optional inputs on Krea2EditModelPatch/Krea2EditGroundedEncode and removed the link records, but the saved UI workflow retained the old link ids on the target input slots. A later panel_run failed in graphToPrompt because the input referenced a link absent from the parent graph. panel_graph_outline showed the inputs as disconnected because it resolves through the link store, hiding the dangling slot references.

### Steps to reproduce
1. Load the bundled krea2-identity-edit workflow.
2. panel_disconnect node 79 inputs source_latent_b and source_image_b, and node 84/85 image_b.
3. Save/reopen the workflow.
4. Inspect saved UI JSON: node 79 inputs still carry link 21/20 while graph.links has neither.
5. panel_run to SaveImage.

### Exact error (scrubbed)
No link found in parent graph for id [79] slot [2] source_latent_b

### Fix
applied locally: yes (installed panel source)
verified: node --check passed; live frontend reload was blocked because other user workflow tabs have unsaved changes, so the patch is not yet runtime-verified.
upstream-only: no

Patch:
diff --git a/web/js/comfyui-mcp-panel.js b/web/js/comfyui-mcp-panel.js
@@
     try {
       graph.afterChange();
     } catch (err) {
       envelopeErr = err;
     }
+    const disconnectedInput = (node.inputs ?? []).find((slot) => slot?.name === inputName);
+    if (
+      disconnectedInput?.link != null &&
+      String(disconnectedInput.link) === String(removedLink.linkId) &&
+      !describeInputLink(graph, node, (node.inputs ?? []).indexOf(disconnectedInput))
+    ) {
+      disconnectedInput.link = null;
+    }
     graph.setDirtyCanvas(true, true);

Temporary verified workaround: reload the original pack and keep both reference branches connected, feeding the same valid subject image to each slot. panel_run then queues successfully.

### Environment
Windows 11
ComfyUI 0.33.0
ComfyUI frontend 1.49.6
Python 3.12.11
PyTorch 2.8.0+cu129
GPU NVIDIA GeForce RTX 5070 Ti 16 GB
comfyui-mcp 0.52.95
comfyui-mcp-panel installed package 0.15.74

artokun:
Fixes #1750.

## What was wrong

`panel_disconnect` answered #668's refusal — *"is not connected — there is no link to
remove … no retry is needed"* — for an input whose `link` id had **no record in the
graph's link store**. That slot is not unconnected. Read from the frontend installed
here (ComfyUI_frontend 1.48.7), `ExecutableNodeDTO.resolveInput`:

```ts
if (input.linkId == null) return                       // unconnected — fine
const link = this.graph.getLink(input.linkId)
if (!link) throw new InvalidLinkError(
  `No link found in parent graph for id [${this.id}] slot [${slot}] ${input.name}`)
```

It is checked for **every** input of a node the serializer does not skip, so one such
slot makes the whole workflow unqueueable — which is the error #1750 reported verbatim
(`… for id [79] slot [2] source_latent_b`).

Two things were wrong, and they are separate:

1. **The graph could not be repaired.** `describeInputLink` returns `null` for an
   unconnected slot *and* for an orphaned one, so the refusal could not tell them
   apart. The only tool that can clear a slot's link declined to clear it, and told the
   caller no retry was needed about the one reference failing every run. Nothing else
   showed it either: `panel_graph_outline` resolves *through* the link store, so it
   renders an orphaned input as cleanly disconnected — exactly as the report describes.
2. **The post-condition was assumed.** `graph_disconnect` never checked that its own
   mutation left no slot naming the link it removed.

## Where the orphaned ids come from

Not from `disconnectInput` — that nulls `inputs[slot].link` *before* it touches the
store, so those two cannot disagree. It is the **output** side. `LGraphNode.disconnectOutput`,
which `LGraph.remove(node)` calls for every output of a node being deleted:

```ts
const input = target.inputs[link_info.target_slot]   // index recorded on the LINK…
input.link = null                                    // …not the slot the input occupies
…
link_info.disconnect(graph, 'input')                 // → network.links.delete(this.id)
```

`LLink.disconnect` ends in an unconditional `network.links.delete(this.id)`. So when the
recorded `target_slot` is stale, the record is destroyed while the input that actually
carried it is untouched — precisely #1750's end state, and a stale `target_slot` is not
hypothetical in this codebase: #342/#1340 fixed the outline for exactly that, "a stale
target_slot renders nothing instead of fabricating connectivity". In the multi-target
branch there is no `if (input)` guard at all, so an out-of-range index throws mid-loop
and strands the remaining links on that output too.

That fits the report: the four orphans (79 `source_latent_b`/`source_image_b`, 84/85
`image_b`) are the whole *b* reference branch, fed by nodes 90 and 92 — and dropping the
b branch is what the reporter was doing. Krea2EditModelPatch carries eleven inputs, nine
of them optional, which is where an inputs array shifts under a recorded index.

**I did not execute this** — no ComfyUI was running for this task — so treat it as read
from the frontend source, not as a reproduction. It is not what this PR fixes either:
the mechanism is upstream. What this PR fixes is that the panel had no way to clear the
result and no check that it had not produced one.

## What changed

`web/js/lib/disconnect-verify.js` — three helpers: `orphanedInputLinkId` (tells the
three slot states apart), `clearOrphanedInputLink`, `clearStrandedInputLinks`.

`graph_disconnect`:

- **Refusal path.** An orphaned reference is now REPAIRED, in its own
  `beforeChange`/`afterChange` envelope so `Ctrl+Z` still reverts it, and reported as
  `cleared_orphan_link` — deliberately **not** as `disconnected`/`removed_link`, because
  no wire was removed. A genuinely unconnected slot still gets #668's refusal, untouched.
- **After its own mutation.** It now asserts that no input slot still names the removed
  link, repairs what the assertion catches, and DISCLOSES the repair in `warning` +
  `cleared_orphan_link`. It repairs **nothing** while the store still holds the link —
  that is a disconnect that did not land, and #668's disclosure is what must fire there.

## Where to look hardest

- **Reachability is not symmetric between the two halves, and I am not claiming it is.**
  Half 1 is reachable from any workflow already holding an orphan — that is where the
  reporter was left, and it is what unblocks them. Half 2 is a post-condition assertion:
  the case it catches on the target node is two input slots naming the SAME link id
  (`LGraph._removeDuplicateLinks` exists because duplicate-tuple links occur), where
  `disconnectInput` nulls one and deletes the record. Absent that, it is a no-op.
- **The repair must never touch a live wire.** Both helpers re-read the store at the
  moment of the write rather than trusting an earlier verdict; nulling a slot whose link
  exists would strand the origin output's `links` entry instead — the same corruption
  pointed the other way. `M3` below is the mutation that checks this.
- **The reply may not claim more than it measured.** Clearing one orphan does not make
  a graph queueable — #1750's own workflow carried four — so the orphan path reports what
  is STILL orphaned in the graph (`remaining_orphan_links`, capped at ten, via the
  existing `danglingInputLinks`) and says the graph is still refused until each is
  cleared. Likewise an `afterChange()` that throws keeps the repair but costs the Ctrl+Z
  promise, rather than being folded in with a failed clear. Both were codex gate round-1
  P1s. Two more of the same shape were then closed without being asked: only the
  `certainly_reached` subset may be narrated as a refusal (an orphan on a bypassed, muted,
  or virtual node is real corruption whose reachability this module does not model), and
  the list caps at ten while `remaining_orphan_link_count` carries the untruncated total.
- **Reply shape change.** The orphan path returns `cleared_orphan_link` with no
  `disconnected` key. An agent testing `result.disconnected` will not see success; a
  retry then hits the plain "not connected" refusal, which is correct and idempotent.
  The alternative — reusing `disconnected` — would claim a wire removal that did not
  happen, which is the thing #668 exists to prevent. Nothing on either side of the
  bridge reads that key: `panel_disconnect` forwards the reply verbatim.

## Verification

`browser_tests/unit/disconnect-orphan-link.test.mjs` runs the **real shipped
`graph_disconnect`**, extracted from `web/js/comfyui-mcp-panel.js` and given
litegraph-shaped doubles with the real helpers injected — a helper-only test stays green
against an unfixed call site.

Mutation-verified at the call site (each mutation applied to the panel source alone, the
named test required to go red):

| mutation | detected by |
| --- | --- |
| M1 orphan repair removed (refusal restored to #668 only) | `panel_disconnect CLEARS the reference instead of refusing` |
| M2 post-condition assertion removed | `a slot left naming the removed link is cleared and disclosed` |
| M3 post-condition made unconditional (would clear a LIVE link) | `never covers for a disconnect that did NOT land` |
| M4 orphan path reports a wire removal it did not perform | `panel_disconnect CLEARS the reference instead of refusing` |
| M5 remaining-orphan report removed (reply implies the graph is clean) | `reports what is STILL orphaned, and never claims queueable` |
| M6 `afterChange` failure swallowed (Ctrl+Z promised on a broken envelope) | `a throwing afterChange costs the Ctrl+Z promise, not the repair` |
| M7 `certainly_reached` split dropped (bypassed orphan narrated as a refusal) | `an orphan on a BYPASSED node is reported without claiming the graph is refused` |
| M8 truncation count clamped to the cap (a capped list reads as the whole list) | `the remaining-orphan list is capped, but the COUNT never is` |

`npm run test:unit` — 6740 tests, 0 fail, including `check-panel-scope` OK (every name
in the panel resolves) and the tool-vocabulary gate.

**Browser suite NOT run.** No ComfyUI was listening on this machine (8188 and the rest
of the 81xx range all refused), and the Playwright suite requires a real one. This change
alters a bridge command's return value and graph mutation, not any rendered surface, but
I am not implying e2e coverage I do not have.

## Not done

`panel_graph_outline` still hides an orphaned reference — it resolves through the link
store, so the input reads as cleanly disconnected. That is the reason the corruption was
invisible in the first place, and it is worth a follow-up, but it is outside what this
issue asked for and I did not widen the change to cover it.

### DIFF
--- a/browser_tests/unit/describe-command-i18n.test.mjs
+++ b/browser_tests/unit/describe-command-i18n.test.mjs
@@ -75,6 +75,36 @@ test("a number that is not a count never selects a plural form", () => {
   assert.equal(say("graph_screenshot", { width: 1024, height: 768 }), "Captured workflow image (1024×768)");
 });
 
+test("#1750 orphan repair renders a repair card and preserves its warning", () => {
+  __setCatalogForTest("en", {});
+  const warning =
+    'nothing was disconnected — the input carried link id 21, which the graph link store did not have. The panel cleared it.';
+  const card = describeCommand(
+    "graph_disconnect",
+    {},
+    {
+      ok: true,
+      result: {
+        cleared_orphan_link: { node_id: 79, input: "source_latent_b", link_id: 21 },
+        warning,
+      },
+    },
+  );
+
+  assert.equal(card.icon, "pi-exclamation-triangle");
+  assert.equal(card.text, "Cleared orphaned link 79.source_latent_b");
+  assert.equal(card.detail, warning);
+  assert.doesNotMatch(card.text, /undefined/);
+
+  const normal = describeCommand(
+    "graph_disconnect",
+    {},
+    { ok: true, result: { disconnected: { node_id: 79, input: "source_latent" } } },
+  );
+  assert.equal(normal.text, "Disconnected 79.source_latent");
+  assert.equal(normal.detail, undefined);
+});
+
 test("#1690: queued_unknown activity is uncertain, not blocked or duplicate-retry messaging", () => {
   __setCatalogForTest("en", {});
   const card = describeCommand("graph_run", {}, { ok: true, result: {
--- a/browser_tests/unit/disconnect-orphan-link.test.mjs
+++ b/browser_tests/unit/disconnect-orphan-link.test.mjs
@@ -57,6 +57,7 @@ import {
   clearStrandedInputLinks,
   verifyDisconnect,
 } from "../../web/js/lib/disconnect-verify.js";
+import { danglingInputLinks } from "../../web/js/lib/subgraph-conversion-integrity.js";
 
 const panelSrc = readFileSync(
   fileURLToPath(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url)),
@@ -106,6 +107,7 @@ function buildDisconnect(graph) {
     "clearOrphanedInputLink",
     "clearStrandedInputLinks",
     "verifyDisconnect",
+    "danglingInputLinks",
     `return ({
 ${disconnectSrc}
 }).graph_disconnect;`,
@@ -120,6 +122,7 @@ ${disconnectSrc}
     clearOrphanedInputLink,
     clearStrandedInputLinks,
     verifyDisconnect,
+    danglingInputLinks,
   );
 }
 
@@ -224,6 +227,9 @@ test("#1750 orphaned slot: panel_disconnect CLEARS the reference instead of refu
   assert.equal(res.removed_link, undefined);
   assert.match(res.warning, /nothing was disconnected/i);
   assert.match(res.warning, /No link found in parent graph/);
+  // The ONLY orphan in this graph was the one cleared, so say exactly that.
+  assert.equal(res.remaining_orphan_links, undefined);
+  assert.match(res.warning, /No other input in this graph carries an orphaned link id/);
   // Repaired inside ONE undo envelope, so the "Undoable with Ctrl+Z" contract holds.
   assert.equal(graph.envelopes, 1);
   assert.equal(graph.closedEnvelopes, 1);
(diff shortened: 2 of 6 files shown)


artokun:
Independent gate NO-SHIP at exact head 292d0a9. Blocking production regression: the orphan-repair response returns cleared_orphan_link without disconnected, but the activity formatter unconditionally reads r.disconnected, producing Disconnected undefined.undefined and dropping the repair warning. Please update the production formatter/activity result contract and add a regression for the orphan-repair card. PR remains open and unmerged.

artokun:
Takeover fix completed and pushed at `33cbc01a7c42c860ce15d20551fc5f2f4b53e9db`.

The production activity formatter now handles the intentional orphan-repair result shape (`cleared_orphan_link` without `disconnected`), renders a repair summary, and preserves the repair `warning` as card detail. Normal disconnect cards remain unchanged.

Validation:
- `npm.cmd run test:unit` — 6,740 passed, 1 existing todo, 0 failed
- focused orphan/formatter tests — 27 passed
- `npm.cmd run typecheck` — passed
- `node --check web/js/comfyui-mcp-panel.js` — passed
- `git diff --check` — passed

PR remains open and unmerged.

PR #1763, opened 2026-08-25
Fix panel refresh stuck after concurrent downloads

### LINKED ISSUE #1736: panel_refresh_nodes remains stuck in refresh_still_running
panel_refresh_nodes repeatedly returned reason "refresh_still_running" without completing or exposing a completion handle. The installed MiniMax H3 Director nodes were already available, so this caused an unnecessary retry/workaround.

### Steps to reproduce
1. With a connected local ComfyUI panel and custom nodes already loaded, call panel_refresh_nodes.
2. Retry immediately and again after 3 seconds.
3. Each call returns ok:true, refreshed:false, reason:refresh_still_running, and says the refresh remains running.

### Exact error
"refresh_still_running": "A node-def refresh was still running when this command stopped waiting for it ... Nothing failed and nothing was changed."

### Fix
applied locally: no
upstream-only: yes
Precise change needed: expose/persist the in-flight refresh state or wait until the node-definition refresh completes within the tool call; if the refresh exceeds the command budget, return a stable operation id/status path instead of requiring blind retries.

### Environment
Windows 11 / ComfyUI 0.33.3 / frontend 1.49.6 / GPU NVIDIA GeForce RTX 4070 Ti 12 GB / comfyui-mcp 0.52.91 / panel 0.15.71.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-luna` · effort `high` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
## Summary
- skip the duplicate frontend combo refresh for completed-download background refreshes when the panel's authoritative local sweep covered every live combo
- preserve manual refresh timeout/error semantics and protect the shared trust flag from combo failures
- add a production-path regression covering concurrent completion, panel_refresh_nodes acknowledgement, slot release, and recovery

Fixes #1736

## Validation
- Focused refresh suite: 104 passed, 0 failed
- Full unit glob: 6,753 passed, 0 failed, 1 todo
- 
pm.cmd run typecheck: passed
- 
pm.cmd run test:unit: stops at check-panel-scope because this fresh worktree lacks local 
ode_modules/typescript; direct full unit glob passed
- No lint script is defined in package.json

### DIFF
--- a/browser_tests/unit/node-def-refresh.test.mjs
+++ b/browser_tests/unit/node-def-refresh.test.mjs
@@ -350,7 +350,7 @@ test("#635: the shipping registerComfyNodeDefs returns its verdict through the p
   assert.match(body, /return describeNodeDefRefresh\(\{/, "the run verdict is returned");
   assert.match(
     body,
-    /nodeDefsRefreshConfirmed =\s*runIsCurrent && !comboCompletionPending && !didThrow && !!defs && \(comboRan \|\| comboAbandonedAfterRebuild\);/,
+    /nodeDefsRefreshConfirmed =\s*runIsCurrent &&\s*!comboCompletionPending &&\s*!didThrow &&\s*!!defs &&\s*\(comboRan \|\| \(!comboFailed && \(comboRebuiltLocally \|\| comboAbandonedAfterRebuild\)\)\);/,
     "the shared global stays a strict boolean, and #1193 admits the panel's OWN completed " +
       "combo rebuild as the second way the live lists became current",
   );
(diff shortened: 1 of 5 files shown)


artokun:
Addressed the trailing-run option race in commit c875e4433d97634c01ed1a9aa8f3f7b52a2f7a4c. A later completed-download refresh now upgrades an existing queued successor with skipDuplicateComboRefresh before it starts. The regression drives the extracted production onDownloads handler after a no-skip forced trailing caller and verifies panel_refresh_nodes acknowledgement, slot release, and recovery. Focused refresh tests: 104 passed, 0 failed. Typecheck passed. No merge performed.

Issue #1716, opened 2026-08-24
Panel cannot defer a safe widget edit while a render is running


artokun:
### What happened / root cause
While a MiniMax H3 render was running, panel_set_widget refused a prompt edit with an error. The tool returns a safety fence rather than allowing a queued/deferred edit. This forced an awkward manual retry after the render.

### Steps to reproduce
1. Queue panel_run on any output node.
2. While the prompt is running, call panel_set_widget on a prompt widget.
3. It returns:
"panel_set_widget was NOT sent — nothing was applied. QUEUE BUSY: a ComfyUI prompt is running… Retry this edit after queue (action:'list') shows running: 0."

### Exact error (scrubbed)
As above. No workflow change was applied.

### Fix
Applied locally: no. Upstream-only: yes.
Suggested minimal improvement: offer an explicit opt-in deferred mutation queue (or a first-class await-idle-and-apply action) so an agent can safely schedule the edit after the current prompt finishes without polling or manual retries. Preserve the current default safety fence.

### Environment
Windows 11; ComfyUI 0.33.1; frontend 1.48.7; RTX 5080 16 GB; comfyui-mcp 0.52.77; panel 0.15.58.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-terra` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #1549

artokun:
Codex swarm takeover (2026-08-25): this issue had only a stale agent:codex label, no claim comment, and no existing worktree/activity for >2h. I am taking it over after the current P2 queue and will update with the PR, tests, and blockers.

artokun:
Implementation complete and PRs are open (not merged):\n\n- Panel PR #1775 (6686a114): https://github.com/artokun/comfyui-mcp-panel/pull/1775\n- Required MCP companion PR #2238 (45ee79): https://github.com/artokun/comfyui-mcp/pull/2238\n\nProduction path traced: panel_set_widget -> graph_set_widget -> MCP queue-busy gate -> Panel executor -> 
[...]

PR #1775, opened 2026-08-25
fix(panel): defer safe widget edits until queue idle

artokun:
## Summary\n\nFixes #1716.\n\n- Add an explicit defer_until_idle:true path for graph_set_widget.\n- Restrict deferral to one concrete widget with scalar value + scalar xpected_value.\n- Refuse callback-driven, queue-hook, linked, dynamic, composite, and derived widget routes before parking.\n- Wait for both ComfyUI queue_running and queue_pending to drain, then revalidate graph/node/widget identity and the expected value at the write boundary.\n- Retire parked edits on backend reconnect; ordinary widget edits keep the existing fail-closed queue fence.\n\nThe queue-busy fence lives in MCP and runs before Panel dispatch. The paired upstream allowance is PR https://github.com/artokun/comfyui-mcp/pull/2238; it permits only the guarded protocol shape through to this Panel implementation.\n\n## Validation\n\n- 
pm.cmd run test:unit — 6,800 passed, 1 existing todo.\n- 
pm.cmd run check:scope — pass.\n- 
pm.cmd run typecheck — pass.\n- 
pm.cmd run test:e2e:list — 92 tests listed.\n- Focused #1716 production-path tests cover park, replay after drain, callback refusal, and stale expected-value refusal.\n\nThe full Playwright E2E suite was not run because it requires a live ComfyUI + bridge.

### DIFF
--- a/browser_tests/unit/deferred-widget-edit.test.mjs
+++ b/browser_tests/unit/deferred-widget-edit.test.mjs
@@ -15,8 +15,13 @@ import {
   isSafeDeferredWidgetValue,
   sameDeferredWidgetValue,
 } from "../../web/js/lib/deferred-widget-edit.js";
+import { resolvePromotedInnerTarget } from "../../web/js/lib/widget-write.js";
 
 const panelSource = readFileSync(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url), "utf8").replace(/\r\n/g, "\n");
+const scopedBatchSeedSource = readFileSync(
+  new URL("../../web/js/lib/scoped-batch-seed.js", import.meta.url),
+  "utf8",
+).replace(/\r\n/g, "\n");
 
 test("#1716 queue status requires both running and pending lists", () => {
   assert.deepEqual(deferredWidgetQueueCounts({ queue_running: [1], queue_pending: [2, 3] }), {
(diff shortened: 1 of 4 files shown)


artokun:
Paired independent review: NO-SHIP. Blocking findings: (1) deferred-widget-edit.js:82-87 has no in-flight drain guard; concurrent timers reproduce applied [first,first,second,second], violating one-shot replay and expected-value safety; (2) classifier at comfyui-mcp-panel.js:12419-12428 never refuses afterQueued-only callbacks, though scoped-batch-seed.js invokes both hooks; (3) generic promoted/subgraph scalar writes are admitted, but widget-write.js resolves promoted inner targets and synchronizes parent rails, while the deferred path only rechecks the outer node/widget. Re-engaging worker for all three plus regression coverage. Do not merge.

artokun:
Follow-up to the paired NO-SHIP review:

- Added an in-flight drain guard/serialization in `web/js/lib/deferred-widget-edit.js`. The adversarial overlapping-timer probe reproduced `[first, first, second, second]` before the fix; the regression now proves `[first, second]` and zero pending receipts.
- The production classifier now refuses afterQueued-only callback widgets (the production scoped-batch path invokes both beforeQueued and afterQueued).
- The classifier now refuses recognized promoted/subgraph scalar routes through the production promotion resolver because widget-write synchronizes inner targets and parent rails.
- Ordinary direct scalar widgets remain allowed and are covered on the production graph_set_widget path.

Panel HEAD: `fcbbdb40`
MCP companion PR #2238 remains: `f45ee793`

Gates:
- Panel focused production/deferred suite: 94/94
- Panel typecheck, scope check, full unit: 6,804 passed, 0 failed, 1 todo
- Panel Playwright enumeration: 92 tests
- MCP focused queue-fence suite: 572/572
- MCP lint/build: pass
- MCP full test: 12,125 passed, 3 failed, 4 skipped; the 3 failures are the pre-existing node-id-union-message expectations and are unrelated to #1716/#2238.

No merge performed. Requesting fresh paired re-review.

Issue #1781, opened 2026-08-25
panel_run completion notification missing; orchestrator synthesizes fallback


artokun:
### What happened / root cause
A panel_run render completed successfully in ComfyUI, but no normal completion payload reached the orchestrator within the 10-second post-finish window. The orchestrator had to synthesize the completion from ComfyUI execution/history. The exact transport-side root cause is upstream-only from this session because no panel completion stack or delivery trace was exposed.

### Steps to reproduce
1. Queue the live canvas with panel_run(to_node_id=PreviewImage).
2. Let the local render complete successfully.
3. Observe that no panel completion reaches the orchestrator.
4. After 10 seconds, the orchestrator emits a synthesized completion from ComfyUI history.

### Exact error (scrubbed)
[panel event] The render finished SUCCESSFULLY. No completion for it reached the orchestrator in the 10s after that finish was observed here, so this notice was synthesised from ComfyUI itself.

Prompt id: de703058-ced0-4319-96b8-e5946db3af5f
Output: ComfyUI_temp_ypyes_00001_.png

### Fix
applied locally: no
upstream-only: yes

Precise change needed: make panel completion delivery acknowledged and retryable until the orchestrator confirms receipt; deduplicate late/retried messages by prompt_id. Add diagnostics that distinguish panel never-sent, transport failure, and late composition so the fallback event can identify the failing stage. Keep the existing history synthesis as a safety net.

### Environment
Windows 11
ComfyUI 0.19.3
ComfyUI frontend 1.42.14
GPU NVIDIA GeForce RTX 4070 Laptop GPU, 8 GB VRAM
Python 3.10.11
PyTorch 2.7.0+cu128
comfyui-mcp 0.52.103
panel 0.15.81
Backend Codex

artokun:
Claimed by the autonomous maintainer swarm for implementation. Worktree: wt-1781. Priority: next oldest eligible P2; no enhancement label.

artokun:
PR #1783 is open at exact head \d8daafe669e50698c125a8509583c162692dbf0d\; narrow panel-side diagnostics mitigation and local focused gates pass. Independent review and hosted CI are running before merge.

artokun:
Independent review SHIP for PR #1783 at `d8daafe`, with one P3 diagnostic truthfulness caveat. A narrow correction and regression test are being added before merge; no release yet.

artokun:
[...]

PR #1783, opened 2026-08-25
fix(panel): diagnose completion delivery outcomes (#1781)

artokun:
Refs #1781

## Scope

Adds a narrow, panel-owned diagnostic seam for the completion lifecycle:

- completion frames carry `completion_diagnostics` with composition time and a `late-composition` marker when composition exceeds 10 seconds;
- the production flush path logs `never-sent`, `transport-failure`, and `late-composition` outcomes at the panel's observable boundary;
- existing prompt-id fencing, failed-send re-pending, `/history` reconciliation, and mute behavior are unchanged.

## Why no completion acknowledgement/retry protocol is added

The literal current panel transport returns only browser `WebSocket.send()` acceptance. The bridge acknowledges `user_message` frames (`working`/`seen` by `mid`) but has no acknowledgement for `agent_event` completion frames. Retrying after `sendFrame()` returns true could duplicate a valid completion, and adding a new completion-ack frame would invent upstream behavior. End-to-end receipt/fallback timing therefore remains an orchestrator-side boundary; this patch makes the panel-observable failure stage explicit without changing delivery semantics.

## Validation

- Focused completion diagnostics + still metadata tests: 7/7.
- Relevant completion/queue/reconcile lifecycle files: 152/152.
- `npm.cmd run typecheck`: passed.
- `npm.cmd run test:unit`: stopped at the existing scope gate because this fresh worktree has no local `node_modules/typescript`; the standalone typecheck passed.
- `git diff --check`: passed.

### DIFF
--- a/browser_tests/unit/completion-delivery-diagnostics.test.mjs
+++ b/browser_tests/unit/completion-delivery-diagnostics.test.mjs
@@ -1,0 +1,88 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import {
+  classifyCompletionDelivery,
+  completionCompositionDiagnostic,
+  COMPLETION_LATE_COMPOSITION_MS,
+} from "../../web/js/lib/completion-delivery-diagnostics.js";
+import { composeRunCompletionFrame } from "../../web/js/lib/run-completion-frame.js";
+
+test("completion delivery diagnostics distinguish panel-observable outcomes", () => {
+  assert.equal(classifyCompletionDelivery(), "never-sent");
+  assert.equal(classifyCompletionDelivery({ sendAttempted: true }), "transport-failure");
+  assert.equal(
+    classifyCompletionDelivery({
+      sendAttempted: true,
+      transportAccepted: true,
+      compositionMs: COMPLETION_LATE_COMPOSITION_MS + 1,
+    }),
+    "late-composition",
+  );
+  assert.equal(
+    classifyCompletionDelivery({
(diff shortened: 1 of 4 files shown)


artokun:
Implementation agent reports exact head \d8daafe669e50698c125a8509583c162692dbf0d\; focused 7/7 tests, relevant lifecycle 152/152, typecheck, and diff check passed. Full unit suite was blocked only by the fresh worktree missing local node_modules/typescript; independent review and hosted CI are running before merge.

artokun:
Independent review of exact head `d8daafe669e50698c125a8509583c162692dbf0d`: SHIP. One non-blocking P3 caveat was identified: the documented empty composer result is logged as `never-sent` even though the caller treats it as delivered. Re-engaging implementation for this narrow diagnostic classification correction before merge; no merge/release yet.

artokun:
Follow-up pushed at exact head `14bda406aa50c23b88a943ba277c815558103e9a`: intentional empty composer results now classify as `empty-no-frame`, with a real null-path regression test. Focused 8/8, lifecycle 153/153, typecheck, and diff checks pass. Requesting fresh independent review for this exact head; no merge/release yet.

artokun:
Fresh independent review of exact head `14bda406aa50c23b88a943ba277c815558103e9a`: SHIP. The empty-no-frame correction is valid and retry/local transport semantics remain unchanged. Refreshed CI is green. Proceeding to merge; release follows separately.

PR #1802, opened 2026-08-25
fix(panel): rebind after closing refused active workflow

### LINKED ISSUE #1795: Closing a refused unsaved workflow can leave active and graph bindings wedged
After repeatedly creating a blank workflow, attempting a refused save, and closing the unsaved tab, panel_list_workflows reported the closed unsaved workflow as still active while it no longer appeared in the open/workflows lists. Subsequent panel_graph_outline failed with root-workflow-uuid-mismatch. The prescribed routing-key recovery also failed because the tab had already been closed and was absent both from the frontend list and disk. This leaves the live-canvas binding wedged until a frontend reload.

### Steps to reproduce
1. panel_new_workflow
2. panel_add_node
3. panel_save_workflow with a slash-containing name so the save is refused
4. panel_close_workflow force:true
5. Repeat new/close quickly or inspect panel_list_workflows.
6. Observe active points at the closed unsaved routing key, while open only contains another saved tab.
7. panel_graph_outline fails root-workflow-uuid-mismatch; panel_open_workflow with the reported routing key says no match.

### Exact error
[root-workflow-uuid-mismatch] ... active tab has NEVER BEEN SAVED ... use routing key tmp:...
Then: Error: no workflow matching tmp:...

### Fix
After close, atomically rebind active/routing/graph identity to the actually selected remaining tab, or return a receipt requiring the caller to wait for the frontend selection event before the next mutation. The recovery guidance should account for a routing key that was just closed.

### Environment
Windows 11; ComfyUI 0.33.1; RTX 3090 24 GB; comfyui-mcp 0.52.51; panel 0.15.70.

---

Related (same root cause): #1233

artokun:
Fixes #1795

## Summary
- preserve the existing unsaved/capture refusal before any close
- after closing the active tab, select a surviving tab and reuse the production workflow_open path to repaint and prove active/graph identity
- fail closed with an UNVERIFIED, do-not-retry receipt when no successor binding can be proven
- return successor routing_key, workflow_uuid, and graph readiness for reconnect/route consumers

## Validation
- npm.cmd run test:unit — 6,837 tests, 6,836 passed, 0 failed, 1 todo
- npm.cmd run typecheck — passed
- npm.cmd run check:scope — passed
- node --test browser_tests/unit/workflow-close-rebind.test.mjs — 3/3 passed
- Playwright discovery — new #1795 test listed
- live Playwright run — blocked because localhost:8188 was not listening (ERR_CONNECTION_REFUSED)

No merge or release performed.

### DIFF
--- a/browser_tests/refused-unsaved-workflow-close.spec.ts
+++ b/browser_tests/refused-unsaved-workflow-close.spec.ts
@@ -1,0 +1,75 @@
+/**
+ * #1795 — a refused save followed by a forced close of an unsaved ACTIVE tab
+ * must select a surviving tab and leave graph tools usable.
+ *
+ * This drives the real panel_close_workflow executor against ComfyUI. The
+ * slash-containing save is expected to refuse; the first close is expected to
+ * preserve that refusal; only force:true may discard the tab. The final list
+ * and graph read prove the production close/rebind path did not leave a stale
+ * active workflow or root binding behind.
+ */
+import { test, expect } from './fixtures/panelTest'
+import { claimFreshCanvas, settleCanvas } from './fixtures/canvasIdentity'
+
+test('closing a refused unsaved workflow rebinds the active canvas', async ({
+  page,
+  panel,
+  mockBridge
+}) => {
+  await panel.goto()
+  await panel.setBridgeUrl(mockBridge.url)
+  await panel.openSidebar()
+  await panel.connect()
+
+  await claimFreshCanvas(page, mockBridge)
+  await page.evaluate(() => {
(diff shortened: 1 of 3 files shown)


artokun:
SHIP (fresh exact-head review): 707508ca18885560a37fbdb292088a5950206486 against 033549688c8631f717f732ed29ada8d0b297243e. Unsaved refusal remains before close, active close rebinds through workflow_open and verifies active/open-list/UUID/route/graph, and unverified recovery fails closed. Hosted CI 32855718159 green; localhost Playwright unexecuted.

PR #1804, opened 2026-08-25
fix(panel): contain long chat transcript rendering

### LINKED ISSUE #1801: Long-running Agent sessions cause ComfyUI canvas zoom/pan lag (panel chat log not render-contained)
After an extremely long single Agent-panel session (500+ chat turns, many with embedded image/video storyboard cards from `panel_show_media`), the user reported that zooming/panning the ComfyUI graph canvas itself became stuttery -- and switching from the Agent tab to the Assets tab made it instantly smooth again. That A/B (same canvas, same workflow, only the visible panel tab differs) points at the panel's own chat log, not the ComfyUI canvas renderer or the workflow's node count (confirmed small, 17 nodes).

Traced through `web/js/comfyui-mcp-panel.js`:
- `paintImage()` (~L32023) sets `img.loading = "lazy"` (good -- defers image bytes).
- Video cards have a real two-stage IntersectionObserver lazy-mount/pause scheme (~L32100-32150) that tears down `<video>` decode state when off-screen and re-mounts on re-entry -- also good, and clearly already engineered for exactly this class of problem.
- BUT: neither the per-message bubbles (`paintUser`/`paintAgent`/`paintImage`/`paintVideo`, all doing `log.appendChild(card)`) nor their container use CSS `content-visibility: auto` / `contain`. Grepped the whole bundle for `content-visibility`, `contain:`, `contain-intrinsic` -- zero matches.

So while image bytes and video decode are lazy, the DOM *nodes* themselves (hundreds of `.cmcp-bubble` cards, each with captions/tool-call blocks/etc.) all stay live in layout every frame. Without `content-visibility: auto`, the browser still has to account for every off-screen bubble's layout/paint on any reflow-triggering operation, and that steals main-thread/compositor budget from whatever else is animating in the same tab -- here, the ComfyUI canvas's own zoom/pan rendering, even though the panel itself wasn't being scrolled at the time.

### Steps to reproduce
1. Run one Agent-panel session for hours across many turns, generating enough chat history that the panel accumulates hundreds of message bubbles (especially image/video result cards).
2. Switch to a workflow tab and try to zoom/pan the ComfyUI graph canvas with the Agent tab still the active panel tab.
3. Observe stutter. Switch the panel to the Assets tab (same workflow, same canvas) -- zoom/pan is immediately smooth again.

### Exact error (scrubbed)
No thrown error / console exception observed -- this is a perf/jank issue, not a crash.

### Fix
applied locally: no
upstream-only: yes (deliberately did not live-patch: this is the panel bundle actively controlling an in-progress multi-hour GPU render session; a live hard-refresh to test a fix risked disrupting that session, and 41k-line minified-ish bundle isn't a safe single-attempt live edit per the skill's "exactly one attempt, don't spiral" guidance)

Suggested precise change: add CSS containment to the chat log's message cards so off-screen bubbles are skipped from layout/paint, e.g.

```css
.cmcp-bubble {
  content-visibility: auto;
  contain-intrinsic-size: auto 120px; /* rough average card height, avoids layout jump on reveal */
}
```
applied to whatever the actual bubble/card class is (grep shows `.cmcp-bubble` on the `paintImage` card, likely shared by `paintUser`/`paintAgent`/`paintVideo` etc. -- verify the exact selector covers all message types including question/secret/A2UI cards). If bubbles have highly variable heights this may need `contain-intrinsic-size` tuned or a per-type default. This is a pure CSS containment addition, not a logic change, so it should be low-risk to land and verify (open a long thread, DevTools Performance record during canvas pan/zoom before/after).

### Environment
OS: Windows 11 Pro 10.0.26200
ComfyUI: 0.33.1
ComfyUI Frontend: 1.48.7
GPU: NVIDIA GeForce RTX 5060 Ti, 16GB VRAM
comfyui-mcp: 0.52.108
panel: 0.15.71

artokun:
Closes #1801.\n\nSummary:\n- add content-visibility:auto and contain-intrinsic-size to every direct chat-log message child, covering bubbles, tool/question/secret cards, A2UI cards, notices, and thinking state while preserving the scroll surface and DOM behavior\n- add a production-source regression contract for the shipped .cmcp-log rule and painter append paths\n\nValidation:\n- npm.cmd run test:unit (6837 passed, 0 failed, 1 todo)\n- npm.cmd run typecheck\n- npm.cmd run check:scope\n- focused chat/A2UI/media tests (269 passed)\n- git diff --cached --check\n\nNo merge or release performed.

### DIFF
--- a/browser_tests/unit/chat-render-containment.test.mjs
+++ b/browser_tests/unit/chat-render-containment.test.mjs
@@ -24,6 +24,10 @@ function functionBody(source, name) {
 test("#1801 contains every direct chat message without changing the feed surface", () => {
   const source = readFileSync(PANEL_JS, "utf8");
   assert.match(source, /log\.className = "cmcp-log"/);
+  assert.match(
+    source,
+    /import \{ createChatScrollStabilizer \} from "\.\/lib\/chat-scroll-stabilizer\.js";/,
+  );
 
   const ruleStart = source.indexOf(".cmcp-log > :not(.cmcp-empty) {");
   assert.ok(ruleStart >= 0, "the production chat feed must contain its message rule");
@@ -50,4 +54,12 @@ test("#1801 contains every direct chat message without changing the feed surface
     assert.match(body, new RegExp(`className = "${className.replaceAll(" ", "\\s+")}`));
     assert.match(body, /log\.appendChild\(/, `${name} must still append to the chat log`);
   }
+
+  const replay = functionBody(source, "paintThread");
+  const renderTodoAt = replay.indexOf("renderTodo(t.todos");
+  const finalScrollAt = replay.indexOf("scrollLog();", renderTodoAt);
+  assert.ok(renderTodoAt >= 0, "replay still renders the thread tray");
+  assert.ok(finalScrollAt > renderTodoAt, "replay seeds a final post-A2UI scroll correction");
+  assert.match(functionBody(source, "mountLiveA2UICard"), /log\.appendChild\(handle\.el\)/);
+  assert.match(functionBody(source, "paintA2UIRecord"), /log\.appendChild\(renderA2UIInert\(/);
 });
--- a/browser_tests/unit/chat-render-containment.test.mjs
+++ b/browser_tests/unit/chat-render-containment.test.mjs
@@ -1,0 +1,53 @@
+// #1801 — a long Agent transcript must not keep every off-screen message in the
+// layout/paint work competing with ComfyUI's graph canvas.
+//
+// The panel is shipped as one DOM-built bundle, so this is intentionally a
+// production-source contract: it protects the CSS rule on the real .cmcp-log
+// and proves the message painters still append their existing DOM nodes there.
+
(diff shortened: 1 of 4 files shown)


artokun:
NO-SHIP (fresh exact-head review): containment CSS is present, but contain-intrinsic-size:auto 120px underestimates variable-height replayed history. The one-shot scrollLog() can leave the transcript ~2,934px above the true bottom after reveal. The source-only test omits runtime replay/autoscroll geometry and explicit A2UI coverage. Fix runtime scroll behavior and test it before merge.

artokun:
SHIP (fresh exact-head review): 83c618d8337e5688ec5cc2ea3ef1f5031f5d4dd5 against 9b62f36f2de7b95b647fad66145ad170a674be62. ResizeObserver tracks reveals, MutationObserver tracks live/inert appends, replay gets final stabilization, and shouldStick preserves scrolled-up/new-message behavior. Full unit 6,839 passed; hosted CI 32859244770 green.

PR #1815, opened 2026-08-25
fix(panel): settle active workflow before open reply

### LINKED ISSUE #887: panel_open_workflow reports target active but active tab remains previous workflow

artokun:
Closes #887

## Problem

workflow_open could report the requested workflow as active before the frontend finished its asynchronous tab/canvas transition. An immediate pin then observed the prior active canvas and refused.

## Corrective fix

- workflow_open and workflow_new now share one binding-generation domain.
- The workflow reload section is single-flight while an awaited mutation is live, and every ownership-sensitive open step stops before settle, repaint, rebaseline, or disk reload when superseded.
- Failed or unknown current-generation operations retire both active and post-reconnect proof epochs. An older open cannot retire proof established by a newer open or workflow_new.
- workflow_new records binding proof only after the new tab is genuinely proven blank and while it still owns the section and generation.
- Production lifecycle coverage executes the shipped workflow_open and workflow_new bodies for native failure and unknown proof states, plus the immediate list, pin, and current-mode fail-closed consumers.

## Validation

- npm.cmd run test:unit: 6,875 passed, 1 existing TODO
- Focused open/rebind/recovery/persistence tests: 186 passed
- npm.cmd run typecheck: passed
- npm.cmd run check:scope: passed
- node scripts/check-tool-vocabulary.mjs: passed
- node --check web/js/comfyui-mcp-panel.js: passed
- git diff --check: passed

No MCP companion is needed: no MCP payload or consumer contract changed. Do not merge or release from this PR.


artokun:
Addressed the independent P1 review finding in new head `8d45a904`.

- `workflow_open` now stamps reconnect proof only after the owned post-open active settle succeeds.
- Failed/unknown opens retire both `activeWorkflowResyncEpoch` and `postReconnectBindingProofEpoch` for the current epoch before returning `applied:"unknown"`.
- Invalidation is scoped to the latest `workflow_open` generation, so an older superseded open cannot clobber a newer open's proof; unrelated guard users do not suppress invalidation.
- The settle helper handles `beginWorkflowReloadStep() === false`, checks ownership after awaiting, and always cleans up an acquired step.
- Regression coverage exercises immediate list active confirmation, pinned mismatch, and current-mode mutation gating, plus supersession.

Validation on the pushed head: full unit suite 6,872 passed / 1 existing TODO; focused lifecycle/parser/contract suite 96 passed; typecheck and diff-check passed.

PR #1827, opened 2026-08-26
fix(panel): retain panel_run completion until receipt (#1824)

### LINKED ISSUE #1824: panel_run completion can be lost after successful long render
A `panel_run` SaveVideo job completed successfully, but its completion `agent_event` never reached the Codex conversation and the panel's history reconciliation did not replay it. The run finished in 458180 ms; after more than 31 minutes without a completion event, the documented single fallback `queue status` check reported `done:true`, `status_str:"success"`.

The panel currently retires the run as soon as `client.sendFrame(frame)` returns true (`comfyui-mcp-panel.js`, run-completion `onFlush`: `if (frame == null || framePushed) runCompletion.markDelivered(promptId)`). That confirms only a WebSocket write, not orchestrator/harness receipt. If the downstream completion frame is dropped or not surfaced, the pending ledger is already cleared, so the periodic `/history` safety sweep has nothing left to recover.

### Steps to reproduce
1. Queue a long SaveVideo workflow through `panel_run`; observed prompt id `16d1c3c9-4779-4b01-97a4-d8ce5ec2f1c9`.
2. Keep the Codex goal active and wait for the automatic completion event; do not poll.
3. The server finishes the workflow successfully in 458180 ms and saves the video.
4. No completion event or replay arrives for over 31 minutes.
5. Make the single documented fallback status check; it reports the prompt is done/successful.

### Exact error (scrubbed)
No exception was emitted: this is a silent delivery failure. Fallback response:
`{"running":false,"pending":false,"done":true,"status_str":"success","execution_stats":{"total_duration_ms":458180}}`

### Fix

Add an idempotent completion-delivery acknowledgement keyed by prompt id plus route/session identity. `client.sendFrame()` success must leave the completion pending until the orchestrator explicitly acknowledges receipt. Until acknowledgement, the periodic `/history` reconciliation must continue replaying; the orchestrator should deduplicate replays by the same completion key. The existing run-receipt / working-ack mechanism is the likely model to extend. Do not mark the tracker delivered on socket-write success alone.

### Environment
Windows 11; ComfyUI 0.33.4; ComfyUI frontend 1.49.6; RTX 5090 32 GB; torch 2.12.1+cu130; Python 3.13.12; comfyui-mcp 0.52.45; comfyui-mcp-panel 0.15.81; Codex backend.

artokun:
Fixes #1824\n\nA successful browser WebSocket write is not proof that the orchestrator received a panel_run completion. Keep keyed panel_run completions in the recovery ledger until the matching route/session/prompt receipt arrives, and replay /history outcomes with the same idempotency key. Reject stale receipts.\n\nIncludes a production graph_run lifecycle regression covering a 458180 ms render, accepted socket write, history replay, and matching receipt settlement.\n\nRequired companion protocol change: Artokun/comfyui-mcp#2322.\n\nTests: npm run test:unit (6890 tests: 6889 pass, 1 todo); npm run typecheck; npm run check:scope.

### DIFF
--- a/browser_tests/unit/restart-resume.test.mjs
+++ b/browser_tests/unit/restart-resume.test.mjs
@@ -529,6 +529,64 @@ test("#1824: restart adoption preserves the keyed completion route and session",
   }]);
 });
 
+test("#1824: a timeout-released terminal batch survives teardown and late prompt adoption", async () => {
+  const firstFlushes = [];
+  const first = makeTracker((payload) => firstFlushes.push(payload));
+  const dispatchToken = first.beginPanelRun();
+  first.onExecutionStart("P-timeout");
+  first.onExecuted("P-timeout", {
+    images: [{ filename: "timeout.png", type: "output" }],
+  });
+  first.onExecutionSuccess("P-timeout");
+  first._fireTimers(30000);
+  assert.equal(firstFlushes.length, 1, "the expired production hold emits its fallback frame");
+  assert.equal(firstFlushes[0].awaitingCompletionKey, true);
+  assert.equal(first.hasPending(), true);
+  const terminalState = first.terminalCompletionMetadata();
+  assert.deepEqual(terminalState[0], {
+    promptId: "P-timeout",
+    payload: {
+      promptId: "P-timeout",
+      images: [{ filename: "timeout.png", type: "output" }],
+      videos: [],
+      durationMs: 0,
+      finishedAt: 1_000_000,
+    },
+    unkeyedFlushed: true,
+  });
+  first.dispose();
+
+  const replayed = [];
+  const fresh = makeTracker((payload) => replayed.push(payload));
+  assert.equal(fresh.restoreTerminalCompletion(terminalState[0]), true);
+  await fresh.reconcile({
+    fetchHistory: async () => ({
+      outputs: { 9: { images: [{ filename: "history.png", type: "output" }] } },
+      status: { status_str: "success", completed: true, messages: [] },
+    }),
+    fetchQueued: async () => false,
+    isVideo: () => false,
+  });
+  assert.equal(replayed.length, 0, "history cannot replace the retained unkeyed terminal record");
+  assert.equal(fresh.terminalCompletionMetadata().length, 1);
+  fresh.onQueued("P-timeout", {
+    routeId: "panel-route-1824",
+    sessionId: "session-1824",
+    dispatchToken,
+  });
+  assert.equal(replayed.length, 1, "the delayed prompt binds and replays after restart");
+  assert.equal(replayed[0].promptId, "P-timeout");
+  assert.equal(replayed[0].images[0].filename, "timeout.png");
+  assert.equal(typeof replayed[0].completionKey, "string");
+  assert.equal(fresh.hasPending(), true, "restart adoption still waits for the matching receipt");
+  assert.equal(
+    fresh.acknowledgeDelivery("P-timeout", replayed[0].completionKey),
+    true,
+    "the bound completion can be retired only by its exact receipt",
+  );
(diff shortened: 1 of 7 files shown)


artokun:
Independent cross-repo gate: NO-SHIP (P1). Exact head 69b89c2. Panel completion race: if execution_success/media arrives before delayed /prompt parsing, run-completion.js:453-456 flushes with no key and marks delivered; later onQueued() at :971-980 adds the key without replay. The new test registers prompt first and misses this ordering. Lifecycle state is also lost because completionKeys is tracker-local (run-completion.js:165), teardown clears the tracker (comfyui-mcp-panel.js:42498-42515), and restart-resume.js:392-400 re-adopts without route/session metadata. Prompt reuse and watchdog ordering need production-path regressions. Panel cannot ship alone because it requires the MCP ACK companion #2322; fix and re-gate the pair.

artokun:
Updated head 4641ed01ecdc3afdaf5dbcd4f3c0e7d3dbb6a2fe. Fixes #1824 Panel lifecycle: execution_success/media arriving before delayed /prompt are held and replayed with one keyed completion; keys plus route/session metadata persist through remount and reboot adoption; receipt outbox carries the key. Production regressions cover media-before-prompt, keyed ACK/replay, and restart adoption. Checks: node --test browser_tests/unit/run-command-budget.test.mjs browser_tests/unit/run-completion.test.mjs browser_tests/unit/restart-resume.test.mjs (134 passed); npm.cmd run typecheck; npm.cmd run check:scope. MCP companion updated in #2322 at d5e0410. No merge/release/tag performed.

artokun:
Fresh pair gate: NO-SHIP (P1). Production retention still fails after the 30-second hold: run-completion.js:521-532 releases media without a key and marks it terminal/delivered; delayed prompt replay at :1041-1076 only works while terminalNoKeyCompletion remains, and :211-215 prevents re-pending. The timer is :1173-1180, production begins at comfyui-mcp-panel.js:17965, and no production endPanelRun caller was found. A read-only timer probe reproduced unkeyed flush after 30s followed by no keyed replay when the prompt ID arrives. The shipped test passes only because its prompt arrives before the timer. Fix the production lifecycle so late prompt identity can bind/replay the completion, with a real delayed-prompt-after-timeout regression. MCP #2322 is blocked as a pair dependency.

artokun:
#1824 paired re-gate update

Panel head `9b76c37e2113187f9dc31801ee2c19b7fd2cd6c4` fixes the posted 30-second hold loss:

- lifecycle prompt ids are associated with the panel dispatch, including the case where the hold expires while rendering is still active;
- timeout release emits at most one best-effort unkeyed frame, retains a pending terminal media record, and marks it `awaitingCompletionKey` rather than delivered;
- delayed `onQueued` binds the route/session/prompt generation key and replays the retained media keyed;
- retained terminal records persist through panel teardown/reload and are restored before reconciliation; history cannot replace an unkeyed record before identity binding.

Production-path coverage now drives the real `graph_run` call site through timer expiry before `execution_success`, then delayed prompt binding and keyed replay. Restart coverage restores the terminal record into a fresh tracker, probes history without losing it, binds it after the delayed prompt, and retires only on the matching receipt.

Checks:
- Panel focused lifecycle tests: 135 passed.
- Panel full `npm.cmd run test:unit`: 6891 passed, 1 todo.
- `npm.cmd run typecheck`: pass.
- `npm.cmd run check:scope`: pass.
- Panel `npm.cmd run build` / `npm.cmd run lint`: unavailable; package.json defines neither script.
- Paired MCP head `d5e0410ec3d0b0a177e2db130b7390283da93772` was left unchanged and revalidated separately; its standalone fencing remains green.

Please re-gate this head with MCP #2322; do not merge from this worker.

PR #1838, opened 2026-08-26
fix(panel): reuse a run's completion identity on repeat registration (#1837)

### LINKED ISSUE #1837: Regression (v0.15.99): one completed prompt delivers two completion events after #1833
### What happened

The fix for #1830 (PR #1833, merged `c7a47466`, shipped in **v0.15.99**) makes a single
completed prompt deliver **two** `agent_event` completion frames — the symptom #1830 was
opened to prevent.

PR #1833 removed, in one change, all three places that collapsed same-prompt completion
identities:

1. `web/js/lib/run-completion.js:1231` — `onQueued` dropped the `existingKey ||` term. Since
   `createRunCompletionKey` salts with `Date.now()+Math.random()` (`run-completion.js:62-65`),
   a repeat registration now mints a *new* key instead of reusing the run's existing one.
   That term was added by `4641ed01` ("preserve keyed long-run completions") in the same commit
   that introduced the salt, replacing a key whose comment called it *"deterministic across
   retries"* — `existingKey ||` is what preserved that determinism.
2. `web/js/lib/restart-resume.js:81` — `normalizeCompletionMeta` changed its dedup key from
   `promptId` to `completionKey`, so a reboot marker may now carry N rows for one prompt id.
3. `adoptRebootRuns` loops every row into `onQueued`, and `flushWithCompletionRecords` calls
   `onFlush` once per record.

This also breaks the contract still documented at `run-completion.js:76`: `onFlush` is
*"Called exactly once per completed prompt"*.

### Steps to reproduce

Both run against the real modules, no mocks. On `origin/main` @ `a38e6df0` (v0.15.99):

**A — repeat `onQueued`, same route, same session, no supplied key**
```js
const t = createRunCompletionTracker({ onFlush: (p) => flushes.push(p) });
t.onQueued(PID, { routeId: "route-A", sessionId: "sess-1" });
t.onQueued(PID, { routeId: "route-A", sessionId: "sess-1" });
t.onExecutionStart(PID);
t.onExecuted(PID, { images: [{ filename: "a.png", type: "output" }], videos: [] });
t.onExecutionSuccess(PID);
```

**B — restart adoption with two marker rows for one prompt**
```js
adoptRebootRuns([PID], t, [
  { promptId: PID, completionKey: JSON.stringify(["route-A","sess-1",PID,"n1"]), routeId: "route-A", sessionId: "sess-1" },
  { promptId: PID, completionKey: JSON.stringify(["route-A","sess-2",PID,"n2"]), routeId: "route-A", sessionId: "sess-2" },
]);
```

### Observed vs expected

| | `869feeb1` (before #1833) | `a38e6df0` (v0.15.99) |
|---|---|---|
| A — `onFlush` calls | 1 | **2** |
| B — completion events | 1 | **2** |

Expected: exactly one completion event per completed prompt, on both paths.

### Notes

- Reported on PR #1833 before it merged; the merge landed 101s after the review, so this is a
  follow-up rather than a missed review.
- Restoring `existingKey ||` fixes **A only** — adoption passes an explicit `completionKey`, so
  `suppliedKey` wins and `existingKey` is never consulted. **B** needs the collapse at the
  delivery boundary (`flushWithCompletionRecords`) or the `promptId` dedup restored in
  `normalizeCompletionMeta`. Verified: with `existingKey` restored, A → 1 event, B → still 2,
  and the focused suite stays green (144 pass / 0 fail).
- The full 6904-test unit suite passes on the regressed code — no test asserts the
  one-frame-per-prompt invariant across two identities. A fix should add that pin.

### Environment

comfyui-mcp-panel v0.15.99 (`pyproject.toml`), `origin/main` @ `a38e6df0`.

artokun:
Closes #1837

Regression fix for #1833 (merged `c7a47466`, shipped in **v0.15.99**), found while reviewing
that PR — the merge landed 101s after the review, so this lands it as a follow-up.

## The defect

`#1833` dropped the `existingKey ||` term from `onQueued`:

```js
-  const nextKey = suppliedKey || existingKey || createRunCompletionKey(routeId, sessionId, k);
+  nextKey = suppliedKey || createRunCompletionKey(routeId, sessionId, k);
```

`createRunCompletionKey` salts with `Date.now()+Math.random()` (`run-completion.js:62-65`), so a
repeat registration of a run the tracker already held **minted a second identity** for the same
prompt — one no orchestrator ticket ever opened. `flushWithCompletionRecords` emits one frame per
record, so a single finished run delivered **two** agent completion events. That is the symptom
#1830 was filed to prevent.

That term was not incidental: `4641ed01` added it in the same commit that introduced the salt,
replacing a key whose own comment called it *"deterministic across retries"*. `existingKey ||` is
what preserved that determinism once the key stopped being deterministic.

## What this does NOT change

Retaining several **supplied** identities for one prompt is deliberate and stays intact. Each is a
receipt the orchestrator opened, and each is owed its own frame — collapsing them would hang a
ticket. #1833's two tests that pin this (`#1830 keeps two same-prompt nonce rows and retires them
by exact key`, `#1830: restart adoption preserves multiple same-prompt completion nonces`) are
untouched and green.

I originally fenced this at the record level and **those two tests caught it** — that told me the
multi-row design was intentional and that only *minting* is the defect. Only minting is fenced here.

## Evidence

Head `85ff62b767d9d5a3210a1ab0dad68c39d3a37b49`, base `origin/main` @ `a38e6df0` (v0.15.99).

**Verified by mutation, not by a green suite.** Re-applying the one-line regression
(`existingKey ||` removed) fails exactly the new pin and nothing else:

```
✖ #1837 a repeat registration REUSES the run's identity — one finished run, one agent turn
ℹ tests 40  ℹ pass 39  ℹ fail 1
```

Reproduced end-to-end through the real modules before the fix, on three SHAs:

| | `869feeb1` (before #1833) | `a38e6df0` (v0.15.99) | this branch |
|---|---|---|---|
| repeat `onQueued`, one prompt | 1 event | **2 events** | 1 event |

Gates, all on this head:
- full `node --test browser_tests/unit/*.test.mjs` — **6908 pass / 0 fail / 1 todo** (6909 total)
- `check-panel-scope` OK (after `npm ci` — a fresh worktree does not inherit `node_modules`)
- `tsc --noEmit` OK · `check-tool-vocabulary` OK · `py_compile` OK · `git diff --check` OK

## Where to look hardest

- **The `completionRecordFor(k)` lookup is unscoped** (no route/session args), so it returns
  `records.at(-1)`. That matches `4641ed01`'s original unscoped `completionKeys.get(k)`, and it is
  the conservative choice — scoping it would let a repeat registration on a different route mint a
  fresh identity again. But if a prompt legitimately spans two routes, this reuses the most recent
  row's key rather than minting per route. I believe a ComfyUI `prompt_id` belongs to exactly one
  queue submission and so one route, which is what makes this safe — worth a second opinion.
- `onQueued`'s return value is unchanged (`nextKey`), and `registerPromptId`
  (`comfyui-mcp-panel.js:18354`) stamps the receipt with it. With reuse, that is now the key the
  tracker actually holds, which is what `acknowledgeDelivery` matches on.

## Not covered

No browser verification. Playwright needs `http://localhost:8188` (`playwright.config.ts:26`, no
`webServer` block); nothing was listening on 8188 or 8201-8210 and this run had no ComfyUI
allocated. This change is in the completion ledger, not the rendered sidebar, but I am not
claiming e2e coverage I do not have.

A second, separate concern from the same review is **not** addressed here and remains open: the new
`sendRunCompletionFrame` fence (`comfyui-mcp-panel.js:38086-38097`) refuses permanently — not
transiently — when the agent starts a new chat mid-run, or when a key was minted pre-ack with
`sessionId: null`. Details are in the review on #1833.

### DIFF
--- a/browser_tests/unit/run-completion.test.mjs
+++ b/browser_tests/unit/run-completion.test.mjs
@@ -370,6 +370,32 @@ test("#1837 a repeat registration REUSES the run's identity — one finished run
   assert.equal(h.tracker.hasPending(), false);
 });
 
+test("#1837 a reused identity re-registered on another route/session does not duplicate", () => {
+  const h = makeHarness();
+  const P = "ctx-switch-prompt";
+
+  // The identity tuple is [route, session, prompt, key], so reusing one key under a
+  // second route/session would land as a DISTINCT record and flush a byte-identical
+  // duplicate frame — strictly worse than the #1837 regression, since both copies
+  // carry the same completion key and the agent cannot tell them apart.
+  const first = h.tracker.onQueued(P, { routeId: "route-a", sessionId: "session-a" });
+  h.tracker.onQueued(P, { routeId: "route-b", sessionId: "session-b" });
+  assert.deepEqual(
+    h.tracker.completionMetadata().map((row) => row.completionKey),
+    [first],
+    "one key is one receipt, whichever context re-registers it",
+  );
+
+  h.tracker.onExecutionStart(P);
+  h.tracker.onExecuted(P, imgs([img("ctx-switch.png")]));
+  h.tracker.onExecutionSuccess(P);
+  assert.deepEqual(
+    h.flushes.map((payload) => payload.completionKey),
+    [first],
+    "no byte-identical duplicate frame",
+  );
+});
+
 test("#1830 keeps two same-prompt nonce rows and retires them by exact key", () => {
   const h = makeHarness();
   const P = "same-prompt";
--- a/browser_tests/unit/run-completion.test.mjs
+++ b/browser_tests/unit/run-completion.test.mjs
@@ -336,6 +336,40 @@ function makeFrameDeps(overrides = {}) {
   return { deps, frames, painted, uploadCalls };
 }
 
+test("#1837 a repeat registration REUSES the run's identity — one finished run, one agent turn", () => {
+  const h = makeHarness();
+  const P = "repeat-registration";
+
+  // Neither call supplies a completion key, so the tracker mints one. #1833 dropped
+  // the `existingKey ||` term, and because createRunCompletionKey salts with
+  // Date.now()+Math.random() the second call invented a SECOND identity for the same
+  // prompt — an identity no orchestrator ticket ever opened. flushWithCompletionRecords
+  // emits one frame per record, so a single finished run became two agent turns: the
+  // exact symptom #1830 was filed about.
+  const first = h.tracker.onQueued(P, { routeId: "route-a", sessionId: "session-a" });
+  const second = h.tracker.onQueued(P, { routeId: "route-a", sessionId: "session-a" });
+  assert.equal(second, first, "a repeat registration reports the identity the run already carries");
+  assert.deepEqual(
+    h.tracker.completionMetadata().map((row) => row.completionKey),
+    [first],
+    "minting must not add a second row the orchestrator never asked for",
+  );
+
+  h.tracker.onExecutionStart(P);
+  h.tracker.onExecuted(P, imgs([img("repeat-registration.png")]));
+  h.tracker.onExecutionSuccess(P);
+  assert.deepEqual(
+    h.flushes.map((payload) => payload.completionKey),
+    [first],
+    "ONE completion frame for one completed prompt",
+  );
+
+  // The single retained row still retires normally against its own receipt.
+  assert.equal(h.tracker.acknowledgeDelivery(P, first), true);
+  assert.equal(h.tracker.completionMetadata().length, 0);
+  assert.equal(h.tracker.hasPending(), false);
+});
+
 test("#1830 keeps two same-prompt nonce rows and retires them by exact key", () => {
   const h = makeHarness();
   const P = "same-prompt";
--- a/web/js/lib/run-completion.js
+++ b/web/js/lib/run-completion.js
@@ -1228,7 +1228,17 @@ export function createRunCompletionTracker({
       if (id != null) {
         panelQueued.add(k);
         const suppliedKey = typeof completionKey === "string" && completionKey.trim() ? completionKey.trim() : null;
-        nextKey = suppliedKey || createRunCompletionKey(routeId, sessionId, k);
+        // #1837 — retaining several SUPPLIED identities for one prompt is deliberate:
+        // each one is a receipt the orchestrator opened, and each is owed its own
+        // frame. Minting is different. createRunCompletionKey salts with
+        // Date.now()+Math.random(), so a repeat registration of a run we already hold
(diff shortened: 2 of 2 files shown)


artokun:
**Round 1 gate: NO-SHIP — and it was right.** Pushed a correction; head is now `554742ded1245877e01fc8b172d6b936ad54a961`.

The gate hit precisely the risk I flagged in the PR body (the unscoped `completionRecordFor(k)` lookup). Reproduced it before fixing:

> [P1] `run-completion.js:1240` — `completionRecordFor(k)` reuses an old completion key across route/session changes, then records it as a second identity — re-registering the same prompt under `route-b/session-b` produces two completion frames with the identical `route-a/session-a` key.

Confirmed by execution on `85ff62b7`: re-registering one prompt under `route-b/session-b` produced **2 records and 2 byte-identical frames**, both carrying the `route-a` key. That is strictly worse than the regression this PR fixes — the agent cannot even tell the copies apart. The identity tuple is `[route, session, prompt, key]`, so the reused key landed as a *distinct* identity.

**Fix:** `addCompletionRecord` now refuses a key the prompt already holds, whichever route/session re-registers it. One key is one receipt. Distinct **supplied** keys still get their own record, so #1830's retention design is untouched.

Both guards are independently mutation-verified — each fails its own pin and nothing else:

| mutation | result |
|---|---|
| remove `existingKey \|\|` | fails **both** #1837 pins (39/41) |
| remove the same-key record guard | fails **only** the context-switch pin (40/41) |

Gates on `554742de`: full unit suite **6909 pass / 0 fail / 1 todo** (6910 total), `check-panel-scope` OK, `tsc --noEmit` OK, `check-tool-vocabulary` OK, `py_compile` OK, `git diff --check` OK.

Round 2 gate is running against this head. The round-1 verdict named `85ff62b7` and does not carry.

artokun:
Independent Codex review: NO-SHIP at exact reviewed head 85ff62b767d9d5a3210a1ab0dad68c39d3a37b49. P1 regression: run-completion.js:273-279 supports route/session-scoped lookup, but onQueued calls unscoped completionRecordFor(k) at :1240 and stores the reused key under the new scope at :1241-1243. Production captures route/session at comfyui-mcp-panel.js:18359-18362; sendRunCompletionFrame rejects mismatched context at :38080-38090, causing repeated markUndelivered recovery, or can misdeliver if the old context is live. removeCompletionRecord removes every row sharing the key (:247-260), so acknowledgeDelivery can retire both scoped records (:1337-1354). Same-scope tests pass, but cross-scope regression coverage/fix is required. Do not merge until corrected and freshly reviewed.

Issue #1718, opened 2026-08-24
Stale storyboard contact sheet delivered for re-rendered video reusing the same temp filename


artokun:
### What happened / root cause

When the same workflow is re-run several times and a `VHS_VideoCombine` node writes to the SAME temp filename each time (`save_output=false`, `filename_prefix="Minimax"` → always `Minimax_00001.mp4`), the storyboard contact sheet the panel attaches to the run-completion event shows the **previous** run's frames, not the current ones.

The completion event's metadata was correct — it reported the new file size (126.2 KB, down from 208.8 KB on the prior run) — but the attached `storyboard_Minimax_00001.png` rendered the OLD video's content.

Likely cause: the storyboard PNG is regenerated under a stable, content-independent filename (`storyboard_<videoname>.png`), so the browser (or the `/view` layer) serves a cached copy from the previous run.

### Steps to reproduce

1. Build a graph ending in `VHS_VideoCombine` with `save_output=false` and a fixed `filename_prefix` (so the output is always `<prefix>_00001.mp4` in `temp/`).
2. Queue it. Note the storyboard delivered in the completion event.
3. Change something upstream so the video content genuinely differs, then queue again.
4. The completion event reports a new file size, but the attached storyboard shows the frames from step 2.

Observed on 4 consecutive runs (prompt ids `da711bc6-…`, `f24df452-…`, `ea4743c6-…`, and one earlier), all writing `ComfyUI/temp/Minimax_00001.mp4`.

### Exact error (scrubbed)

No error — silent wrong output. The mismatch was only detectable by decoding the file directly:

```
# actual file on disk after prompt ea4743c6-42a5-47cf-b57e-03c958e7e491
frames: 48  size: (256, 512, 3)   # subjects successfully removed

# storyboard delivered in the completion event
20 cells, each ~1024px wide landscape, both subjects still present  # previous run's content
```

### Impact

This is an agent-correctness problem, not just cosmetic. I was mid-way through debugging a custom node on Apple Silicon and used the delivered storyboard to judge whether each patch worked. Because the sheet was stale, I twice told the user a fix had **not** worked when it actually **had**, and kept patching code that was already correct. I only found the truth by decoding the mp4 myself with PyAV via Bash.

Agents are explicitly instructed to "verify the output matches the request" by looking at the delivered media — so a stale sheet directly defeats that instruction, and does so silently.

### Fix

applied locally: no
upstream-only: yes

[...]

Issue #224, opened 2026-07-29
Stale run-finished event: agent got a previous run's output (wrong filename_prefix, 0.0s duration) while the queued prompt was still executing


artokun:
### What happened / root cause

After `panel_run`, the panel delivered a `Run finished. FINAL output: …` agent_event that did **not** belong to the prompt I had just queued. It carried:

- the **previous run's** `filename_prefix` (`pelirroja_piercing_00002_.png`), even though `SaveImage.filename_prefix` had been changed to `maria_pelirroja` **before** queuing (confirmed via `panel_graph_outline` after the fact);
- `rendered in 0.0s`;
- a `finished` clock time inconsistent with the previous delivered run (`1:22:49` vs `13:24:45`);
- an image visually identical to the prior render.

Meanwhile `get_queue` reported `running: 1`, `prompt_id b6a318d0-95fb-4e33-a268-24cf4f9116c5` — i.e. **my actual run was still executing** when the "finished" event arrived. `list_output_images` confirmed no `maria_pelirroja*` file existed yet.

Net effect: I reported a wrong image to the user as the final result of a change (adding a LoRA) that had not rendered yet. This is the worst failure mode for the "VERIFY THE OUTPUT MATCHES THE REQUEST" contract — the agent cannot tell a stale delivery from a real one.

Suspected mechanism, in `web/js/comfyui-mcp-panel.js`:

1. **Cross-run buffer contamination.** `runImageBuffers` is keyed by `promptKey(promptId)`, which falls back to the literal `"__no_prompt__"` bucket when an `executed` event has no `prompt_id` (`promptKey`, ~L13081; `bufferRunImages`, ~L13103). `onExecuting` (~L13360) then calls `flushAllRunImages()` on any `detail == null` event, which flushes **every** buffer including a leftover/`__no_prompt__` one. Nothing validates that the flushed buffer belongs to the run that just ended, so an older buffered image can be delivered as the current run's FINAL output. The `RUN_FLUSH_DEBOUNCE_MS` (1500ms) timer in `bufferRunImages` is a second path to the same outcome.
[...]

PR #1843, opened 2026-08-26
fix(1834): cache-bust the image URL the chat card is painted from

### LINKED ISSUE #1834: Chat inline image preview shows stale/wrong thumbnail for reused output filename
While iterating prompts on a ComfyUI workflow ("[REF] Krea2 reid Base-Making(Tarantula)"), the panel's chat inline image preview for a completion notification showed the WRONG image. The user compared two screenshots:
1. The chat panel's inline preview for output filename `test_face_00005_.png` (a diptych: left = face closeup, right = full body) showed the RIGHT panel on a plain white/grey studio background, in a pose/outfit that did not match the just-run prompt at all.
2. Opening the actual file `test_face_00005_.png` directly (confirmed via the file viewer's own title bar showing that exact filename) showed BOTH panels correctly on a grass field background with flowers, matching the just-edited prompt (grass, midday sun, denim shorts).

Same filename, two different sets of pixels shown to the user depending on where it's viewed. This is misleading — a user could judge/approve a render based on the wrong image shown inline in chat.

Hypothesis (not verified against source — I did not have filesystem access to the comfyui-mcp-panel install in this environment to confirm): the panel's chat/inline media preview may be caching a thumbnail keyed by filename string alone, rather than by full path + subfolder + type + mtime/content hash. The workflow's SaveImage node uses `filename_prefix` with ComfyUI's own `%date:yyyy-MM-dd%` plus its automatic `%counter%` suffix (e.g. `..._00005_.png`), so the same filename string can recur across different runs/sessions if the counter resets or overlaps with a prior day's output. If the preview cache key doesn't include enough uniqueness, a new render reusing an old filename would display the OLD cached thumbnail instead of fetching the new file's bytes.

### Steps to reproduce
1. Run a workflow whose SaveImage filename_prefix can produce a repeated/colliding filename across sessions (e.g. uses `%date%` + counter, and the counter has previously reached the same number on a different day, or the session was restarted).
2. Get a chat completion notification with an inline image for that filename.
3. Compare the inline chat thumbnail against the actual file on disk (or via get_image) for the same filename — they may differ.

### Exact error (scrubbed)
No error/exception — silent wrong-content display, not a crash.

### Fix
I do not have filesystem/source access to the comfyui-mcp-panel install from this agent environment (the managed shell here is sandboxed to the orchestrator host and cannot see the ComfyUI/panel installation), so I could not locate or patch the chat media-preview component myself. Suggested fix direction: key the inline preview cache/fetch by something unique per render (full output path + subfolder + type, plus mtime or a content hash / prompt_id) instead of filename alone, so a reused filename can't serve stale cached bytes.

### Environment
OS: Windows 10 Pro 10.0.19045
ComfyUI: 0.34.0
ComfyUI frontend: 1.49.6
Python: 3.13.12 / PyTorch 2.10.0+cu130
GPU: NVIDIA GeForce RTX 4080, 16GB VRAM (14.6GB free at time of report)
comfyui-mcp: 0.52.125
panel: 0.15.71

---
Filed from the ComfyUI panel by **claude** · model `sonnet` · effort `high` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #1718

artokun:
Fixes #1834.

## The bug, and why it survived its first fix

The reporter compared two views of `test_face_00005_.png`: the chat's inline preview showed a studio-background render, while opening that exact filename in the file viewer showed the grass-field render they had just prompted for. Same name, two different sets of pixels — and the failure mode is that someone approves a render they never made.

This is not a guess about caching. ComfyUI's own `server.py` says it in the `/view` handler:

> `FileResponse` emits Last-Modified/ETag and **nothing sets `Cache-Control` on `/view`, which makes it heuristically cacheable**

`Cache-Control: no-store` is attached *only* on the dangerous-content-type branch, so a PNG gets a bare `FileResponse`. A `SaveImage` `filename_prefix` built from `%date%` + `%counter%` can re-emit a name an earlier day's run already used, and the browser paints its cached copy under the new name.

**#1835 (merged, closed this issue) could not reach that.** It added the `cmcp_prompt` bust inside `buildStillsSegment`, on the URL handed to `fetchImageBytes` / `fetchImageDimensions` — the completion frame's *size and dimension probes*. The card the person looks at is painted much earlier, by `onExecuted`:

```js
const url = imageViewUrl(m);        // stable across runs
…
} else {
  paintImage(url, m.filename);      // ← the chat card, unbusted
  inlineImages.push(m);
}
```

Different URL, different cache key. #1835 shipped with **no tests**, and a helper-level test would have passed anyway. I reopened #1834 rather than filing a new issue.

## What this changes

1. **`onExecuted` (`comfyui-mcp-panel.js`)** — the real fix. Paints the card from `appendImageCacheBust(url, d.prompt_id)`. Keyed on the *same* prompt id the probes use, so card and probe address one URL and share one download — if they diverged, the note would report the size of a file the card never fetched (the "metadata right, picture wrong" split #1718 hit).
2. **id-less runs** — `#224` supports back-to-back runs with no `prompt_id`, and that is where filenames are *most* likely to repeat. Those get a minted identity instead of being handed back unbusted.
3. **`composeShowMediaReply` (`media-preview.js`)** — `panel_show_media` paints the same chat card and had the same gap: its video branch already carried the #1718 bust on the adjacent line, stills did not. Busted before the `/view` probe so probe and card stay on one URL, matching what video does.
4. **De-duplication** — the append helper now lives once in `storyboard-cache-identity.js` (sharing a private `appendCacheBustParam` with the storyboard one) instead of a second copy in `run-completion-frame.js`.

## Where the load-bearing claims are — please push on these

- **Is `prompt_id` the right key?** One prompt id = one execution, so a re-used filename across runs always gets a distinct URL, and a replayed/persisted card keeps a stable one. If `executed` could ever fire twice for one prompt with *different* bytes under one filename, this key is too coarse. I do not believe that is reachable, but it is the assumption the fix rests on.
- **The `kind === "image"` condition in `media-preview.js`** deliberately excludes audio. Audio has the same exposure and is left alone — out of scope for this report, and I did not want to change what an audio card fetches without a reason to.
- **Item 3 is scope beyond the literal report.** It is the same mechanism on the same visible surface, but it is a judgement call; it is a separable one-line hunk if you disagree.
- **`show_media probes the SAME url it paints`** is a *regression guard*, not a fix pin — it passes with the fix removed (both sides are unbusted then). The pin is `show_media re-showing one filename paints two distinct URLs`.

## Verification

Mutation, not a green suite — each mutation applied, run, and reverted:

| mutation | result |
|---|---|
| revert `onExecuted` to `paintImage(url, …)` | **3 of 7 fail** |
| drop the id-less fallback (return `url` unbusted) | **1 fails** |
| delete the `media-preview.js` line | **1 fails** |

The new tests extract the **shipped** `onExecuted` from the panel source via `new Function` and assert on what `paintImage` is actually handed — the same production-boundary pattern `video-poster.test.mjs` uses for #1718 — so they cannot pass on a helper that production never calls.

- `npm run test:unit` — **6920 pass, 0 fail** (7 new)
- `npm run typecheck` — clean
- Adding the module binding broke the `new Function` harness in `run-completion.test.mjs` with `ReferenceError: appendImageCacheBust is not defined`; its deps list is updated in the same commit.

Playwright status is in a follow-up comment.

### DIFF
--- a/browser_tests/unit/inline-image-cache-bust.test.mjs
+++ b/browser_tests/unit/inline-image-cache-bust.test.mjs
@@ -55,6 +55,7 @@ function productionOnExecuted() {
     "createStoryboardIdentity",
     "appendStoryboardCacheBust",
     "appendImageCacheBust",
+    "NO_PROMPT_KEY",
     `return (${panelSrc.slice(start, end).trim()});`,
   )(
     (m) =>
@@ -70,6 +71,7 @@ function productionOnExecuted() {
     () => "identity",
     (url) => url,
     appendImageCacheBust,
+    NO_PROMPT_KEY,
   );
   return { onExecuted, painted, buffered };
 }
@@ -163,6 +165,27 @@ test("#1834: a NON-STRING prompt id is normalised the way the tracker does", () 
   assert.equal(painted[1].url, appendImageCacheBust(rawViewUrl(SAME_FILE), promptIdOf(key(8))));
   assert.match(painted[0].url, /[?&]cmcp_prompt=7(?:&|$)/);
   assert.notEqual(painted[0].url, painted[1].url);
+});
+
+test("#1834: a prompt id that IS the reserved sentinel busts nothing", () => {
+  // `promptIdOf` maps NO_PROMPT_KEY back to null, so the completion frame gets
+  // no key for such a run. Busting the card on the literal string would put it
+  // on a URL the probe never requests. Running only `key()` and not
+  // `promptIdOf()` is exactly that half-mirror, so it is pinned.
+  //
+  // Not reachable from a current ComfyUI (prompt ids are UUIDs); pinned because
+  // the call site claims to mirror the tracker, and a claim that is only true
+  // for the inputs someone thought of is how this fix went wrong the first time.
+  const { onExecuted, painted } = productionOnExecuted();
+  onExecuted({ detail: { prompt_id: NO_PROMPT_KEY, output: { images: [SAME_FILE] } } });
+
+  assert.equal(painted.length, 1);
+  assert.equal(promptIdOf(key(NO_PROMPT_KEY)), null, "the tracker maps it to null");
+  assert.equal(
+    painted[0].url,
+    rawViewUrl(SAME_FILE),
+    "the card must not carry a key the completion probes will never use",
+  );
 });
 
 test("#1834: an id-less run is left UNBUSTED rather than given a local key", () => {
--- a/browser_tests/unit/inline-image-cache-bust.test.mjs
+++ b/browser_tests/unit/inline-image-cache-bust.test.mjs
@@ -1,0 +1,237 @@
+// #1834 — the chat's inline image card must not show a PREVIOUS run's pixels.
+//
+// ComfyUI's /view is keyed by filename and nothing sets Cache-Control on it
+// (server.py says so in its own comment; `no-store` is attached only on the
+// dangerous-content-type branch). A SaveImage `filename_prefix` built from
+// `%date%` + `%counter%` can therefore re-emit a name an earlier day's run
+// already used, and the browser paints the cached bytes under the new name. The
+// reporter saw exactly that: the chat thumbnail for `test_face_00005_.png`
+// showed a studio background while opening the same filename in the file viewer
+// showed the grass-field render they had just prompted for.
+//
+// WHAT THIS PINS, and why it is pinned HERE rather than on the helper:
+//
+// The first attempt at this bug (PR #1835) added the cache-bust inside
+// `buildStillsSegment`, on the URL handed to `fetchImageBytes` /
+// `fetchImageDimensions`. Those are the completion frame's SIZE AND DIMENSION
+// PROBES. The picture the person actually looks at is painted much earlier, by
+// `onExecuted`, from a bare `imageViewUrl(m)` — so the symptom was untouched and
+// the issue was closed on a fix that could not reach it. A helper-level test
+// would have passed just as happily.
+//
+// So this drives the SHIPPED `onExecuted` out of the panel source and asserts on
+// what `paintImage` is actually handed.
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+
+import { appendImageCacheBust } from "../../web/js/lib/storyboard-cache-identity.js";
+import { composeShowMediaReply } from "../../web/js/lib/media-preview.js";
+
+const panelSrc = readFileSync(
+  new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url),
+  "utf8",
+);
+
+/** Instantiate the production `onExecuted` with injected panel helpers. */
+function productionOnExecuted() {
+  const start = panelSrc.indexOf("  function onExecuted(ev) {");
+  const end = panelSrc.indexOf("\n  function onExecError(ev)", start);
+  assert.ok(start >= 0 && end > start, "could not isolate production onExecuted");
+
+  const painted = [];
+  const buffered = [];
+  const onExecuted = new Function(
+    "imageViewUrl",
+    "isVideoOutput",
+    "isAudioOutput",
+    "paintVideo",
+    "paintAudio",
+    "paintImage",
+    "runCompletion",
+    "stripMisattachedExecutionPreviews",
+    "app",
+    "createStoryboardIdentity",
+    "appendStoryboardCacheBust",
+    "appendImageCacheBust",
+    `return (${panelSrc.slice(start, end).trim()});`,
+  )(
+    (m) =>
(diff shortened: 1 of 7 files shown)


artokun:
## Browser gate: NOT run. Here is exactly why, and what I did instead.

**I could not run `npx playwright test`.** Stating it plainly rather than implying coverage I do not have.

The suite requires a live ComfyUI (`playwright.config.ts`: *"this suite does NOT start ComfyUI for you"*). One is running — pid 48264, `--port 8188 --enable-cors-header` — but **its listening socket is dead**. From `user/comfyui_8188.log`:

```
[2026-08-26 00:02:38.042] Task exception was never retrieved
future: <Task ... exception=OSError(22, 'The specified network name is no longer available', None, 64, None)>
[2026-08-26 00:02:38.044] Accept failed on a socket
socket: <asyncio.TransportSocket fd=1912, family=2, type=1, proto=6, laddr=('127.0.0.1', 8188)>
```

The process is alive (643 MB WS) but connections are **refused**, not slow — `curl /system_stats` returns exit 7 across repeated probes. This run had no ComfyUI allocated to it and I am not starting one.

**A run would have been misleading even if it had come up.** The suite loads the pack through `custom_nodes/comfyui-mcp-panel`, which is a junction to the *main checkout* — currently `7b477d55`, **21 commits behind `origin/main` (`caa5f4bf`)** and without this branch. A green result there would have described stale main, not this diff. I did not repoint the junction: it is shared, and another agent's run in that window would silently have loaded my code.

### What I checked instead

- **No e2e spec covers this change either way.** Grepping all 44 spec files for `cmcp-imgcard` / `imgcard` / `/view?` / `cmcp_prompt` / `inline image` returns exactly one hit: `training-wizard.spec.ts:60`, `page.route('**/view?*', …)`. That is a route interceptor, and an appended `&cmcp_prompt=` still matches the glob. So the browser suite would not have exercised the inline image card regardless.
- **`npx playwright test --list` from this worktree** — compiles and discovers **98 tests in 44 files**, clean. That proves the diff does not break spec compilation. It is not a behavioural result and I am not presenting it as one.
- The behavioural evidence for this fix is in the PR body: the new tests drive the **shipped** `onExecuted` extracted from the panel source, and each hunk was confirmed by mutation.

### What would change this

A ComfyUI accepting on 8188 **and** the junction pointed at a checkout containing this branch. If you want it re-run before merge, say so and I will do it rather than merging on the unit evidence.

artokun:
## Codex gate: SHIP on `de2c8e12` — the SHA that will merge

Four NO-SHIP rounds preceded it. Recording them because three found real defects **in my own fix**, and the fourth is a decision rather than a fix:

| round | finding | outcome |
|---|---|---|
| 1 | the id-less fallback minted a key **per call**, so `onExecuted` and `buildStillsSegment` each invented their own — card and metadata on different URLs | **fixed** — helper made strict; the two surfaces are busted together or not at all |
| 2a | `d.prompt_id` passed raw to a string-only helper; the tracker does `String(id)`, so a numeric id reached the probes as `"7"` while the card stayed bare | **fixed** — mirror `key()` at the call site |
| 2b | id-less runs left unbusted | **held** — this is round 1's finding inverted; see below |
| 3 | the mirror ran `key()` but not `promptIdOf()`, so an id equal to the reserved `__no_prompt__` sentinel busted the card with a key the probe never uses | **fixed** — both steps now run |
| 4 | — | **SHIP** |
| 5 | re-gate after rebase onto `3c05ddb2` | **SHIP** |

### On 2b, the one I did not fix

Rounds 1 and 2 asked for opposite things: round 1 said do not mint a key locally, round 2 said the id-less path is still stale. The tracker settles which reading is right — every id-less run collapses onto a **shared** `NO_PROMPT_KEY`, and `promptIdOf` returns `null`, so there is no per-run value a locally minted key could ever agree with. Minting one would only put the card on a URL the probe never requests, which is the round-1 defect again.

Closing it for real needs a per-run identity threaded through the tracker's buffer, flush and delivery — through the reconcile and persistence paths. That is a design change, not a line, and it does not belong in a P2 bugfix. It is stated at the call site rather than left as a silent gap, and the test that pins it says *why* it is deliberate so nobody "fixes" it back.

The path #1834 actually reports — a run with a prompt id — is fully fixed.

### Final state on `de2c8e12`

- `npm run test:unit` — **6933 pass, 0 fail** (11 new; count rose from 6923 after rebasing onto `3c05ddb2`)
- `npm run typecheck` — clean
- Mutations re-run after **every** revision, not just the first: reverting the `onExecuted` call site fails 2, deleting the `media-preview.js` line fails 1, replacing the normalisation with the raw id fails 1, dropping the `promptIdOf` half fails 2.
- Browser gate still not run — see the comment above; ComfyUI's listener is dead and no e2e spec covers this path either way. Unchanged by the rebase.

PR #1846, opened 2026-08-26
fix(panel): preserve chat autoscroll through anchored scroll events (#1841)

### LINKED ISSUE #1841: Chat autoscroll latches off: content-visibility (#1801) scroll-anchoring events un-stick stickToBottom

In a long session (60+ turns, many render-result cards with images) the chat log stops following new messages. The user has to wheel down manually for every new message. It used to auto-scroll to the bottom.

Root cause is in `web/js/comfyui-mcp-panel.js` (~line 30867):

```js
log.addEventListener("scroll", () => {
  stickToBottom = atBottom();
  if (stickToBottom) newMsgBtn.hidden = true;
});
```

This treats EVERY `scroll` event as user intent. Since #1801 added

```css
.cmcp-log > :not(.cmcp-empty) {
  content-visibility: auto;
  contain-intrinsic-size: auto 120px;
}
```

each off-screen message swaps between its `contain-intrinsic-size` estimate and its real laid-out height, so `log.scrollHeight` changes underneath the scroll surface. The browser's scroll anchoring compensates by moving `scrollTop` and **fires a `scroll` event nobody asked for**. That event is evaluated against mid-flight geometry, `atBottom()` returns false, and `stickToBottom` latches to `false`.

Once latched, `scrollLog()` takes the `if (!stickToBottom)` branch forever — it only shows the "New messages" pill. The only recovery is a manual scroll that happens to land within `BOTTOM_SLACK_PX` (48px) of the bottom. The taller and more numerous the messages (image result cards), the more intrinsic-size↔real-height swaps happen, so this reproduces reliably on long transcripts and rarely on short ones — which matches the user's report that it used to work and now doesn't.

`lib/chat-scroll-stabilizer.js` cannot help here: its `shouldStick()` is exactly this latched flag, so `schedule()` becomes a no-op.

### Steps to reproduce

1. Open the sidebar panel and hold a long conversation (~50+ messages), including several runs whose result cards embed images.
2. Scroll up to read an earlier message, then scroll back to the bottom.
3. Keep chatting. New messages stop scrolling into view; the log stays put and only the "New messages" pill appears.

### Exact error (scrubbed)

No error or console output — silent failure.

### Fix

`<applied locally: yes>`  `<upstream-only: no>`

Gate un-sticking on real user intent. Programmatic and anchoring scrolls can no longer un-stick; reaching the bottom always re-sticks. Verified with `node --check` (exit 0) and by reloading the panel.

```diff
--- a/web/js/comfyui-mcp-panel.js
+++ b/web/js/comfyui-mcp-panel.js
@@ -30866,9 +30866,24 @@
     chatScrollStabilizer.schedule();
   });
   // Track the user's scroll position; re-stick (and hide the pill) at the bottom.
+  // Only a USER-driven scroll may un-stick. `content-visibility: auto` (#1801)
+  // lets off-screen messages swap between their contain-intrinsic-size estimate
+  // and their real height, so scrollHeight moves under us and the browser's
+  // scroll anchoring emits scroll events nobody asked for. Treating those as
+  // "the user scrolled up" latched stickToBottom to false and killed autoscroll
+  // for the rest of the session unless the user wheeled back down by hand.
+  let userScrollIntent = false;
+  for (const ev of ["wheel", "touchmove", "pointerdown", "keydown"]) {
+    log.addEventListener(ev, () => { userScrollIntent = true; }, { passive: true });
+  }
   log.addEventListener("scroll", () => {
-    stickToBottom = atBottom();
-    if (stickToBottom) newMsgBtn.hidden = true;
+    if (atBottom()) {
+      stickToBottom = true;
+      newMsgBtn.hidden = true;
+      userScrollIntent = false;
+    } else if (userScrollIntent) {
+      stickToBottom = false;
+    }
   });
   const chatScrollStabilizer = createChatScrollStabilizer({
     log,
```

An alternative (or additional) hardening: set `overflow-anchor: none` on `.cmcp-log` so the browser stops generating the compensating scrolls at all, since the stabilizer already owns pinning.

### Environment

OS: Windows 10 Pro 19045
ComfyUI: 0.34.0
ComfyUI FRONTEND: 1.49.6
GPU: NVIDIA GeForce RTX 4080 (16 GB VRAM), CUDA 13.0, torch 2.10.0+cu130
Python: 3.13.12
comfyui-mcp: 0.52.125
comfyui-mcp-panel: 0.15.98
Backend: Claude

---
Filed from the ComfyUI panel by **claude** · model `claude-opus-5` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Fixes #1841.

The chat log now records explicit user scroll intent separately from scroll events. Browser scroll anchoring and content-visibility layout corrections no longer disable stickToBottom, while wheel, touch, pointer, and vertical keyboard scrolling still intentionally detach the reader. Reaching the bottom always re-sticks.

Validation: focused scroll/containment tests, full npm test:unit, npm run typecheck, and npm run check:scope.

### DIFF
--- a/browser_tests/unit/chat-render-containment.test.mjs
+++ b/browser_tests/unit/chat-render-containment.test.mjs
@@ -28,6 +28,18 @@ test("#1801 contains every direct chat message without changing the feed surface
     source,
     /import \{ createChatScrollStabilizer \} from "\.\/lib\/chat-scroll-stabilizer\.js";/,
   );
+  assert.match(
+    source,
+    /import \{ createChatScrollIntentTracker, updateChatStickiness \} from "\.\/lib\/chat-scroll-intent\.js";/,
+  );
+
+  const scrollIntentStart = source.indexOf("const scrollIntent = createChatScrollIntentTracker();");
+  assert.ok(scrollIntentStart >= 0, "the production scroll listener must track user intent");
+  const scrollListenerEnd = source.indexOf("  const chatScrollStabilizer =", scrollIntentStart);
+  const scrollListener = source.slice(scrollIntentStart, scrollListenerEnd);
+  assert.match(scrollListener, /scrollIntent\.consume\(\)/);
+  assert.match(scrollListener, /updateChatStickiness\(/);
+  assert.doesNotMatch(scrollListener, /stickToBottom\s*=\s*atBottom\(\)/);
 
   const ruleStart = source.indexOf(".cmcp-log > :not(.cmcp-empty) {");
   assert.ok(ruleStart >= 0, "the production chat feed must contain its message rule");
--- a/browser_tests/unit/chat-render-containment.test.mjs
+++ b/browser_tests/unit/chat-render-containment.test.mjs
@@ -38,8 +38,11 @@ test("#1801 contains every direct chat message without changing the feed surface
   const scrollListenerEnd = source.indexOf("  const chatScrollStabilizer =", scrollIntentStart);
   const scrollListener = source.slice(scrollIntentStart, scrollListenerEnd);
   assert.match(scrollListener, /scrollIntent\.consume\(\)/);
+  assert.match(scrollListener, /scrollIntent\.endProgrammaticScroll\(\)/);
   assert.match(scrollListener, /updateChatStickiness\(/);
   assert.doesNotMatch(scrollListener, /stickToBottom\s*=\s*atBottom\(\)/);
+  assert.match(source, /scrollIntent\.noteProgrammaticScroll\(\{ behavior: "smooth" \}\);\s*log\.scrollTo\(/);
+  assert.match(source, /beforeScroll: \(\) => scrollIntent\.noteProgrammaticScroll\(\)/);
 
   const ruleStart = source.indexOf(".cmcp-log > :not(.cmcp-empty) {");
   assert.ok(ruleStart >= 0, "the production chat feed must contain its message rule");
--- a/browser_tests/unit/chat-scroll-intent.test.mjs
+++ b/browser_tests/unit/chat-scroll-intent.test.mjs
@@ -1,11 +1,51 @@
 import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
 import test from "node:test";
+import { fileURLToPath } from "node:url";
 
 import {
   createChatScrollIntentTracker,
   isUserScrollIntent,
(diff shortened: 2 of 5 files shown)


artokun:
Independent gate: NO-SHIP (P1). The intent marker in chat-scroll-intent.js:18-29 remains pending until any later log scroll, so pointer/wheel/touch/keyboard input that produces no log scroll can authorize a later content-visibility/scroll-anchoring event; the reverse ordering can consume a real user marker on an earlier programmatic scroll. Production consumes the marker at comfyui-mcp-panel.js:30921-30930. Focused/helper and full unit tests are green, but no production event-order regression covers this. Do not merge; repair is being dispatched.

artokun:
Fresh review of corrected head 2cf47569: NO-SHIP. P1: the tracker guards only one auto scroll while preserving pending, so production reveal's multiple programmatic writes can consume a real user marker; a production-listener harness reproduced marker + two app scrolls -> stickToBottom=false. P1: nested code/card/thinking inputs bubble to the root input listeners, while nested scrolls do not reach the root listener, so nested input + root anchoring scroll falsely un-sticks. P2: destroy() calls dispose(), but the new anonymous input listeners remain attached and dispose() leaves the tracker reusable; post-dispose input still creates a marker. Do not merge; repair is being re-engaged.

artokun:
Exact-head review of 6d176921: NO-SHIP (P1). Reveal paths at comfyui-mcp-panel.js:34691-34695 and 34978-34982 place one queued guard before two programmatic writes (log.scrollTop and card.scrollIntoView), while chat-scroll-intent.js:94-102 consumes only one auto guard per root scroll. Exact probe reproduced false, true, false: the second programmatic delivery consumes the genuine root marker. Nested targets and teardown are now verified fixed; repair must fence every programmatic delivery and add the real reveal regression.

artokun:
Independent exact-head gate: SHIP at cff6bf68. Both reveal paths guard log.scrollTop and card.scrollIntoView independently; production-shaped sequencing tests prove a real root marker survives both. Root-only target checks reject nested input/scroll, expiry/cancellation and disposal are sound, and destroy removes listeners. Focused 32/32 and stabilizer/reveal 20/20 passed; hosted lint pack passed. Residual noted risk is the pre-existing 120 ms interactive-card-reveal retry, outside this diff.

PR #1862, opened 2026-08-26
test(#1861): add behavioural test for repeating_controls_note

### LINKED ISSUE #1861: The partially_queued seed-repetition note is guarded only by a source-grep — no test drives the scoped budget-expiry path
Follow-up to #1998 / PR #1857 (merged as `4ab3802d`). The fix is in and correct; this is about what protects it.

### Measured gap

On the merged head, removing **one** `attachControlRepetitionNote(result, ...)` call — the budget-expiry one at the second `partially_queued` return, which is the exact path the #1998 reporter hit:

| test file | notices? |
|---|---|
| `run-command-budget.test.mjs` (drives that path) | 39 pass / 0 fail — **no** |
| `partial-queue-result.test.mjs` (the helper) | 5 pass / 0 fail — no |
| `scoped-batch-seed.test.mjs` (source-grep) | 70 / 1 — **only this** |

So the sole guard is `matchAll(/attachControlRepetitionNote\(/g)` asserting `>= 2`. It catches outright deletion. It does **not** catch a refactor that relocates the calls somewhere inert, or hands them the wrong object, as long as two occurrences survive somewhere in a 40k-line file.

### Why the obvious place does not work

`#1565 P1: prompts that DID queue before the budget expired` looked like the natural host — it already builds a `partially_queued` result through the real extracted `graph_run`. It is not, and this is worth recording so the next person does not waste the attempt:

It uses `makeUnscopedFrontend` with `batch_count` 1. The note only attaches for a **scoped** batch with `batch > 1` and repeating controls — the #988 scan is gated on `partialTargets && batch > 1` and never runs for an unscoped batch. Asserting the note there would assert something that legitimately should not be present.

### The scenario that would actually pin it

`run-command-budget.test.mjs` already runs scoped tests (`built.graph_run({ to_node_id: 327 })`), so the shape exists. What is needed:

1. A scoped run — `graph_run({ to_node_id: 327, batch_count: 4 })`.
2. A frontend whose graph carries a `control_after_generate` widget with an advancing mode and a governed seed, so `findRepeatingControlWidgets` returns something. This is the actual work: the existing frontend makers do not build such a graph.
3. `mode: "postThenHang"` with a budget short enough to expire mid-batch, which is how `#1565 P1` already reaches the partial return.
4. Assert the returned result carries `repeating_controls_note`.

Then the mutation above kills a named test instead of a token count.

### Severity

P3. The fix is merged and verified, the helper is behaviourally tested, and deletion is caught. This is about the durability of the guard, not a live defect. Filed rather than left implicit because #1998 took five attempts and two of the fixes measured clean while being wrong — a token count is a thin last line of defence on exactly that kind of code.

artokun:
## Summary

Adds a behavioural test for the `attachControlRepetitionNote` wiring in partially_queued early returns. The test drives the real code path to verify that `repeating_controls_note` is correctly attached when a scoped batch with `batch > 1` contains repeating controls.

The existing coverage was only a source-grep test asserting `>= 2` call sites exist. A grep cannot fail when a call site is removed, so this test provides the actual coverage.

## Scenario

- Scoped run with `batch_count: 4` and `to_node_id: 327`
- Graph containing a KSampler with a `control_after_generate` widget set to "randomize"
- Dispatch returns some prompts as verified to trigger the first `partially_queued` early return (line 19261)

## Verification

- New test passes on unmodified tree: 40 pass / 0 fail in run-command-budget.test.mjs
- Removing ONE `attachControlRepetitionNote` call site makes the new test RED
- Restoring the call site makes the test GREEN again
- All related tests remain passing:
  - run-command-budget.test.mjs: 40 pass / 0 fail
  - scoped-batch-seed.test.mjs: 71 pass / 0 fail
  - partial-queue-result.test.mjs: 5 pass / 0 fail

Fixes artokun/comfyui-mcp-panel#1861

### DIFF
--- a/browser_tests/unit/run-command-budget.test.mjs
+++ b/browser_tests/unit/run-command-budget.test.mjs
@@ -2071,3 +2071,139 @@ test("#1845 CONTROL: a batch that queued SOME prompts before the surface died ke
     stop();
   }
 });
+
+// #1861 — a scoped batch with repeating controls must attach repeating_controls_note
+// to the partially_queued result. The note is ONLY attached for scoped batches with
+// batch > 1, and the path must be DRIVEN to verify the attachment happens. A source-grep
+// test that counts call sites is insufficient; removing one call site would pass the
+// grep while missing the real code path.
+//
+// SCENARIO: a scoped run with batch_count > 1 and a graph containing a
+// control_after_generate widget with an advancing mode (e.g., "randomize"). The run uses
+// postThenHang mode with a budget short enough to expire mid-batch, triggering one of
+// the two partially_queued early returns that call attachControlRepetitionNote.
+//
+// ASSERTION: the returned result carries repeating_controls_note when the conditions above
+// are met.
+//
+// The trap: the existing #1565 P1 test uses makeUnscopedFrontend with batch_count 1, and
+// the note only attaches for batch > 1. Asserting the note there asserts something that
+// legitimately should not be present and will not fail for the right reason.
+
+/** Helper to make a scoped frontend with control_after_generate widget. */
+function makeScopedFrontendWithControlWidget({ apiTarget, mode, drainMs = 5, batch = 4 }) {
+  // Graph with a KSampler node that has control_after_generate widget with randomize mode
+  const controlWidget = {
+    name: "control_after_generate",
+    value: "randomize",
+    type: "combo",
+    options: {
+      values: ["fixed", "increment", "decrement", "randomize", "increment-wrap"],
+      serialize: false,
+      canvasOnly: true,
+    },
+  };
+
+  const graph = {
+    _nodes: [
+      {
+        id: 3,
+        type: "KSampler",
+        widgets: [{ name: "seed", value: 0 }, controlWidget, { name: "steps", value: 20 }],
+      },
+      {
+        id: 9,
+        type: "SaveImage",
+        widgets: [],
+      },
+      {
+        id: 327,
+        type: "SaveImage",
+        widgets: [],
+      },
+    ],
+  };
+
+  const app = {
+    queueItems: [],
+    graph,
+    graphToPrompt: async () => ({
+      output: {
+        "3": { class_type: "KSampler", inputs: {} },
+        "9": { class_type: "SaveImage", inputs: {} },
+        "327": { class_type: "SaveImage", inputs: {} },
+      },
+      workflow: {},
+    }),
+    queuePrompt: async (_number, batchCount = 1) => {
+      const post = () =>
+        apiTarget.fetchApi("/prompt", {
+          method: "POST",
+          headers: { "Content-Type": "application/json" },
+          body: JSON.stringify({
+            prompt: {
+              "3": { class_type: "KSampler", inputs: {} },
+              "9": { class_type: "SaveImage", inputs: {} },
+              "327": { class_type: "SaveImage", inputs: {} },
+            },
+            client_id: "x",
+          }),
+        });
+      if (mode === "postThenHang") {
+        await post();
+        return new Promise(() => {});
+      }
+      return true;
+    },
+  };
+  return app;
+}
+
+test("#1861: a scoped batch with repeating controls attaches the note to partially_queued result", async () => {
+  const stop = keepAlive();
+  try {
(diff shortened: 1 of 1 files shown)


artokun:
## Real behavioural coverage — but it lands on the wrong one of the two paths

The hard part is done and done well: a frontend whose graph carries a `control_after_generate` widget so `findRepeatingControlWidgets` returns something, driven through a scoped `batch_count: 4`. That is genuinely a behavioural pin, not a token count, and it is what this issue asked for. Baseline confirmed at **40 pass / 0 fail**, test-only change.

### Measured, mutating each call site independently

| mutation | `run-command-budget.test.mjs` |
|---|---|
| remove site 1 (`:19261`) | **39 / 1 — caught** |
| remove site 2 (`:19347`) | **40 / 0 — NOT caught** |
| control (restored) | 40 / 0 |

Mapping those to the returns they belong to:

- `:19261` serves the `partially_queued` return at `:19244` — the **verified-partial** case (`verified > 0 && unresolved === 0`). This is the one your test drives.
- `:19347` serves the return at `:19336`, the one carrying `incomplete_reason: why` — the **budget-expiry** case.

### Why that matters more than a symmetry complaint

The budget-expiry path is the one the #1998 reporter actually hit — `partially_queued` at 19/24 with *"no scoped dispatch observed within 15s command budget"*. It is the path this entire chain of work exists to protect, and it remains guarded only by the source-grep that #1861 was filed to replace.

Your report named the target as "line 19261 early return", so this is not a hidden assumption — it is simply the other one.

### What closes it

A second case in the same test driving the budget-expiry return. The expensive part is already built and reusable: keep the same control-bearing graph and scoped `batch_count`, and reach the other return the way `#1565 P1` does — `mode: "postThenHang"` with a budget short enough to expire mid-batch. Then assert `repeating_controls_note` on that result too.

Acceptance is the table above with both rows red: removing **either** call site must turn a named test red, and the control must stay green.

One caution given the week this file has had: when you add the second case, diff its arguments against the first before believing it drives a different path. Two cases with identical inputs exercise one path twice, and that has been caught twice on this codebase in the last few hours.

artokun:
## Correction: I was wrong about which call site matters. The test already covers the right one.

I asked for a second case driving `:19347`, calling it "the reporter's exact case". That was a misreading on my part, and the pushback was correct. Verified in the source:

```
// #1565 — the UNSCOPED run's capture wrap, kept so its live counts (what left the
// panel, what is still in flight) can be read after the call, and a flag for the
// one exit that needs them: a queue call abandoned at the command budget.
let runInterceptor = null;
let fullRunAbandoned = false;
```

`:19336`'s return is guarded by `if (fullRunAbandoned)`, so it is the **unscoped** budget-expiry exit. Meanwhile `repeatingControls` is only ever populated for scoped runs — at `:19019` behind `partialTargets && batch > 1`, and at `:19210` behind `attributed?.size`, which requires a control drive that only a scoped batch arms. So at `:19347` the list is always `[]`, `scopedBatchSeedNote([], batch)` is falsy, and nothing attaches. The call site is a no-op.

**And `:19261` — the site this PR does cover — is the scoped partial.** The #1998 reporter ran with `to_node_id`, so that is their path, not the unscoped one. The test is aimed correctly and my instruction was aimed wrong.

So the acceptance table I gave is unsatisfiable by design, and I am withdrawing it. **This PR is complete as it stands:** removing `:19261` turns a named test red, the control stays green, and the neighbouring files are untouched.

### The separate finding, worth its own issue rather than scope creep here

`:19347` is dead for this purpose. Two readings, and they lead to opposite fixes:

1. **It is redundant** and should be removed, since the unscoped path has nothing to attach.
2. **The unscoped path SHOULD warn and cannot**, because the #988 scan is gated on `partialTargets && batch > 1` and never runs unscoped. This is not new — it was recorded during the #1998 park: an unscoped batch with a link-fed or subgraph-promoted seed can produce identical seeds *and* no warning at all. Ungating the scan would make `:19347` live and close that hole, at the cost of scanning every unscoped batch and risking over-warning.

That is a product decision, not a test question, and it does not belong in this PR. Filing it separately.

Good catch stopping to ask rather than changing production code to satisfy a target I had specified incorrectly — that instruction was explicitly "do not change production code", and bending it to hit my table would have been the wrong call.

Issue #1842, opened 2026-08-26
Panel replays completed-run notifications in an endless loop, flooding every agent turn


artokun:
### What happened

After a long session (~10 panel_run calls), the panel began re-delivering the SAME completed-run notifications to the agent on every turn, indefinitely. Roughly 40+ consecutive agent turns received only replayed notifications for 6 already-finished prompts (122c1547, c7cc3fca, 58a9080e, 39b093b7, af4bbb85, fdc2715b), each tagged `POSSIBLE REPEAT`. No new runs were queued during that window (`queue action:"list"` showed `running: 0, pending: 0`).

The replays continued after the agent stopped calling panel_run entirely, so nothing the agent did was re-triggering them.

### Impact

- Every agent turn is consumed acknowledging stale events; the user cannot get work done.
- The session's automatic-preview budget (48 images / ~128MB) was exhausted by replayed images, after which real results also stopped being attached.
- The agent cannot distinguish a genuine new completion from a replay, so it must poll `queue action:"status"` manually to know whether its own run finished.

### Steps to reproduce

1. Long session with many `panel_run` calls (some full-graph, some with `to_node_id`).
2. Mid-session: `panel_restart_comfyui` (custom node was patched), then browser tab reconnect.
3. `panel_free_vram`, then more runs.
4. After a few more completions the replay loop begins and does not stop.

A `panel_restart_comfyui` + tab reconnect happened shortly before the loop started, so the journal/replay path after a reconnect is the prime suspect — the completion journal appears to be replayed rather than drained/acked.

### Exact message (representative, repeated verbatim ~40x)

```
[panel event] ⚠️ POSSIBLE REPEAT: a completion with identical outputs was already
reported to you recently, and this one carries no prompt id to tell them apart.
This is the run YOU queued with panel_run (prompt 122c1547-990a-4ea2-8594-95998585a364).
• workflow completed in 4m 4s · finished 4:52:37 PM
```

Note the internal inconsistency: the header says the event "carries no prompt id", but the body then names an exact prompt id. If the id is available, the dedupe check should be using it.

### Fix

applied locally: no   upstream-only: yes

Not patched from here — the replay lives in the orchestrator process, which does not reload on `panel_reload` or on a tab reconnect, so a local patch could not be verified this session.

Precise change needed:

1. Dedupe on `prompt_id` when it is present. The notification text already contains it; the "carries no prompt id" branch is being taken even for events that do have one.
[...]

PR #1870, opened 2026-08-26
fix(1824): retire completion on receipt, not just socket write

artokun:
## Summary

Addresses the #1824 recurrence: when a panel_run completion's orchestrator receipt never arrives (or arrives slowly), the completion was being lost silently instead of being recovered via /history.

The original #1824 fix correctly implemented retention until receipt. However, #1842's storm-prevention bound was too aggressive: it blocked ALL reconciliation once the bound was reached, preventing even /history checks that could prove completion.

## The Fix

**Separate concerns**: The bound should prevent frame EMISSIONS (which storm the agent), not truth-seeking (/history reads are cheap and prove whether the run completed).

Changes:
- Allow keyed completions to proceed past the bound for /history validation
- Don't emit frames after the bound is exhausted (preventing storms)
- When /history confirms success and bound is reached, retire from pending via `markDelivered`
- This preserves duplicate-over-loss: proving a completion beats losing it

## Test Coverage

- ✓ Original #1824 test: run stays pending until receipt arrives
- ✓ Flapping transport test: bound still prevents unlimited frames  
- NEW: recurrence scenario - write succeeds, receipt never arrives, /history confirms completion is delivered

Note: Initial #1842 test failing on one path - needs review of whether /history should emit once more when bound is reached

Refs #1824, #1842, #2369

### DIFF
--- a/browser_tests/unit/run-command-budget.test.mjs
+++ b/browser_tests/unit/run-command-budget.test.mjs
@@ -1320,164 +1320,95 @@ test("#1824 CALL SITE: a long panel_run completion stays replayable until the or
   }
 });
 
-test("#1824 recurrence: write succeeds, receipt never arrives, /history confirms the completion is delivered", async () => {
-  // This is the scenario from the recurrence report: a long render completes
-  // successfully, the panel sends it with a completion_key, but the orchestrator
-  // receipt never arrives (or arrives very slowly). The bound of 3 on unacked
-  // deliveries would normally stop reconciliation, but for keyed completions we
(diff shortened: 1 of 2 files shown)


artokun:
Closing — built on an incorrect premise of mine. The #1842 bound does not create a loss path: both scenarios are indistinguishable from the panel's side (frame written, no receipt, /history reporting success), so there is no discriminator to gate a past-bound delivery on. The bound blocks a fourth attempt after three were made, which is bounded retry, not loss. Reasoning recorded on #1824.

PR #1872, opened 2026-08-26
test(1871): drive the receipt-less completion end to end, through the real delivery boundary

### LINKED ISSUE #1871: No end-to-end test covers a receipt-less completion being swept and re-delivered up to the bound
Found while closing #1824. Not a live defect — a coverage gap that let a real one survive.

### The gap

The path a `panel_run` completion takes when the socket write succeeds but no orchestrator receipt ever arrives has no end-to-end test:

1. completion flushes, `sendFrame` returns true
2. `awaitsReceipt` is set, so the entry is NOT retired (`web/js/lib/run-completion-delivery.js:153`)
3. no receipt arrives
4. the `/history` sweep finds the still-pending entry and re-delivers
5. after `maxUnackedDeliveries` (3) the sweep stops (`web/js/lib/run-completion.js:834`)

Pieces of this are covered. `run-completion-redelivery-storm.test.mjs` covers step 5 (the bound holds, flapping transport stays capped). `run-completion.test.mjs` covers the ledger mechanics. **Nothing drives 1 → 5 as one sequence.**

### Why it matters, concretely

That is exactly the shape of the bug #1824 reported: the entry used to be retired at step 2 on the socket write alone, so steps 3–5 never ran and the completion vanished silently. A user lost a 458-second SaveVideo render's completion and waited 31 minutes before checking manually.

The fix (`framePushed && !awaitsReceipt && !awaitsCompletionKey`) is on main and correct. But nothing would catch its removal end to end — the storm test would still pass, because with the entry retired early there is no storm to bound.

### What the test should assert

- Write succeeds, no receipt -> entry stays pending (retire-on-receipt, not on write)
- The sweep re-delivers it
- Delivery stops at the bound — **not** zero, and **not** unbounded

The middle assertion is the one that would have caught #1824. The third keeps it honest about the designed compromise: the panel cannot distinguish "delivered but unacknowledged" from "written but lost", so it retries a bounded number of times and then stops. A test that asserts unbounded recovery would be asserting the wrong thing.

### Severity

P3. Main behaves correctly today; this is about durability of that behaviour. Filed because #1824 showed this specific gap has already cost one silent user-visible failure, and the storm test's presence makes the area look covered when the delivery-side path is not.

Related (same root cause): #1824

artokun:
Closes #1871.

Test-only. **No production file is modified** — the diff is one added file, `browser_tests/unit/run-completion-receiptless-delivery.test.mjs` (+262).

## What was missing

#1824 lost a 458-second SaveVideo completion: the entry was retired on `framePushed` alone, so with the socket write succeeding and no orchestrator receipt ever arriving, the `/history` sweep never saw the run again and the user waited 31 minutes before checking by hand. The fix — `framePushed && !awaitsReceipt && !awaitsCompletionKey` in `web/js/lib/run-completion-delivery.js:153` — is on main and correct.

Nothing was pinning it. The area *looks* covered and is not:

| suite | what it drives | why it misses this |
|---|---|---|
| `run-completion-redelivery-storm.test.mjs` | the #1842 bound | its harness flushes into `(p) => flushes.push(p)` — `run-completion-delivery.js` **never executes**, so the retire decision is not on the path at all |
| `run-completion.test.mjs` | the real handler | but around a **refusing** transport (#1830): `sendFrame` returns `false`, the other branch of the same `if` |

So the one case #1824 actually reported — **write succeeds, receipt never comes** — was driven by nobody.

## What this adds

One sequence, 1 → 5, with the real tracker and the real delivery handler wired the way `comfyui-mcp-panel.js:38819` wires them (`markDelivered`/`markUndelivered` pointed back at the tracker; only the transport and the composer's I/O helpers are doubles):

1. the run completes, `sendFrame` returns **true**
2. `awaitsReceipt` is set, so the entry is **not** retired ← *the #1824 defect*
3. no receipt arrives
4. the `/history` sweep finds the still-pending entry and re-delivers it
5. delivery stops at `maxUnackedDeliveries` — **not zero, not unbounded**

Plus three controls, because "retire nothing" also satisfies steps 1–4 and is strictly worse than the bug: a receipt still retires the run; an **unkeyed** completion is still retired by its successful write; a **refused** write still gives its bound slot back (#1842's P1).

## Verified by mutation, not by a green suite

| mutation | new suite | storm suite |
|---|---|---|
| drop `!awaitsReceipt` from `run-completion-delivery.js:153` (re-introduce #1824) | **FAILS** at step 2 — `actual: false, expected: true` | **7/7 still GREEN** ← the gap, demonstrated |
| `maxUnackedDeliveries` 3 → 99 (unbounded replay, #1842) | **FAILS** at step 5 — `actual: 22, expected: 3` | — |
| `unackedDeliveriesExhausted` → `>= 1` (recovery silently disabled) | **FAILS** at step 4 — `actual: 1, expected: 2` | 4/7 |

Step 4 is the assertion the issue names as the one that would have caught #1824, and it is load-bearing independently of step 2. Under the first mutation the two controls that *should* be unaffected (unkeyed retire, refused write) stay green — the file fails where it is supposed to and nowhere else.

## Where to look hardest

- **`settle()` is a bounded macrotask yield (6 turns), not a real await.** The delivery handler is deliberately fire-and-forget (`composeRunCompletionFrame(...).then(...)`) and returns no promise, so there is nothing to await. Every double here resolves immediately, so the chain drains in microtasks well inside one turn; the loop is bounded so it can never hang rather than fail. If a future composer dependency does real I/O or arms a timer, this is the line that would go flaky — that is the trade for not reshaping the production boundary to return a promise just for the test.
- **The assertions count frames the transport ACCEPTED (`sent`), not `onFlush` calls.** That is deliberate — `noteUnackedDelivery` bounds accepted frames — but it means a regression that flushed without sending would read as "bounded" here. The refused-write control is what covers that corner.
- **`BUDGET = 3` is duplicated, not imported.** Same rule the storm suite states: an expectation computed from the code it checks moves with that code and stops being an expectation. Changing `maxUnackedDeliveries` must fail this file.
- **#1824 was a video; this pins the decision with a still** (`videoStoryboardEnabled: false`, `buildVideoStoryboard: async () => null`). The retire guard keys on `completionKey`, not on media type, so the video path adds composition cost without changing the branch under test. Worth a second opinion if you read that guard as media-sensitive.

## Test evidence

- New file: **4/4 pass** (~3.6s).
- Full unit suite on this branch: **7009 pass / 4 fail**.
- Those 4 failures are **pre-existing on clean `origin/main`** (`7f7cbdab`) — an identical set in `graph-set-widget-target-fence.test.mjs` (#2314), verified in a detached worktree at main. They are unrelated to this change and are not touched here; worth its own issue.

**Playwright was not run, and this claims no browser coverage.** The change renders nothing — it adds a `node:test` file and modifies no shipped panel code — so the e2e suite could neither confirm nor refute it. Saying so explicitly rather than implying a gate that did not run.

### DIFF
--- a/browser_tests/unit/run-completion-receiptless-delivery.test.mjs
+++ b/browser_tests/unit/run-completion-receiptless-delivery.test.mjs
@@ -1,0 +1,262 @@
+/**
+ * panel#1871 — the receipt-less `panel_run` completion, driven END TO END.
+ *
+ * THE GAP THIS FILE CLOSES. Two suites already stand on either side of this
+ * path and neither crosses it:
+ *
+ *  - `run-completion-redelivery-storm.test.mjs` pins the BOUND (#1842), but its
+ *    harness flushes into `(p) => flushes.push(p)`. The production delivery
+ *    boundary — `run-completion-delivery.js` — never executes there, so the one
+ *    decision that matters here ("is this entry retired on the socket write?")
+ *    is never actually taken. Every assertion in that file survives a mutation
+ *    of the guard, because the guard is not on the path.
+ *  - `run-completion.test.mjs` DOES wire the real handler, but around a
+ *    REFUSING transport (#1830 session-switch): `sendFrame` returns false, which
+ *    lands in a different branch of the same `if`. A write that FAILS re-pends
+ *    for reasons that have nothing to do with the receipt.
+ *
+ * Nobody drove the case #1824 actually reported: the socket write SUCCEEDS and
+ * the orchestrator receipt never arrives. That run — a 458-second SaveVideo —
+ * was retired on `framePushed` alone, so the /history sweep never saw it again
+ * and the user waited 31 minutes before checking by hand.
+ *
+ * THE SEQUENCE PINNED BELOW is that path as one story, with the real tracker
+ * and the real delivery handler wired together the way `comfyui-mcp-panel.js`
+ * wires them:
+ *
+ *   1. the completion flushes and `sendFrame` returns TRUE
+ *   2. `awaitsReceipt` is set, so the entry is NOT retired  <- the #1824 defect
+ *   3. no receipt arrives
+ *   4. the /history sweep finds the still-pending entry and re-delivers it
+ *   5. delivery stops at `maxUnackedDeliveries` — not zero, not unbounded
+ *
+ * ASSERTION 2 IS THE LOAD-BEARING ONE. Deleting `!awaitsReceipt` from
+ * `run-completion-delivery.js` retires the run on the write, and steps 3-5 stop
+ * existing: there is no storm left to bound, which is exactly why the storm test
+ * stays green through that mutation.
+ *
+ * ASSERTION 5 KEEPS IT HONEST. The panel cannot tell "delivered but
+ * unacknowledged" from "written but lost", so the designed answer is a bounded
+ * retry and then silence. A test asserting unbounded recovery would be pinning
+ * the wrong behaviour — and #1842 is the incident where unbounded replay was the
+ * bug.
+ *
+ * The CONTROLS at the bottom are the other half. "Never retire anything" also
+ * satisfies steps 1-4, and is strictly worse than the bug, so an unkeyed
+ * completion must still retire on its write, and a receipt must still retire a
+ * keyed one.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { createRunCompletionTracker } from "../../web/js/lib/run-completion.js";
+import { createRunCompletionFlushHandler } from "../../web/js/lib/run-completion-delivery.js";
+
+const PROMPT = "9f0f7d2e-6c1a-49f0-9f0a-1c2f5a6b7c8d";
+const ROUTE = "route-a";
+const SESSION = "sess-1";
+const IMAGE = { filename: "receiptless_0001.png", subfolder: "", type: "output" };
+
+/**
+ * The bound, written out as its own literal rather than read from the module —
+ * an expectation computed from the code it checks moves with that code and stops
+ * being an expectation. Changing `maxUnackedDeliveries` must fail this file and
+ * make someone re-argue the number (same rule as the storm suite).
+ */
+const BUDGET = 3;
+
+const HISTORY = {
+  outputs: { 9: { images: [IMAGE] } },
+  status: { status_str: "success", completed: true, messages: [] },
+};
+
+/**
+ * The composition is fire-and-forget inside the delivery handler (metadata
+ * HEADs, storyboard sampling), so the retire decision lands a few turns after
+ * the tracker returns. Every double below resolves immediately, so yielding the
+ * macrotask queue a bounded number of times is enough — and cannot hang.
+ */
+async function settle(turns = 6) {
+  for (let i = 0; i < turns; i += 1) await new Promise((resolve) => setTimeout(resolve, 0));
+}
+
+/**
+ * The panel's own wiring: `createRunCompletionTracker({ onFlush:
+ * createRunCompletionFlushHandler({...}) })`, with `markDelivered` /
+ * `markUndelivered` pointed back at the tracker, exactly as
+ * `comfyui-mcp-panel.js` points them. The only doubles are the transport and the
+ * composer's I/O helpers.
+ */
+function makeDeliveryHarness({ sendSucceeds = () => true } = {}) {
+  const sent = [];
+  const painted = [];
+  const prunes = [];
+  let tracker;
+
+  const onFlush = createRunCompletionFlushHandler({
+    sendFrame: (frame) => {
+      const ok = sendSucceeds(frame);
+      if (ok) sent.push(frame);
+      return ok;
+    },
+    markDelivered: (promptId, completionKey) => tracker.markDelivered(promptId, completionKey),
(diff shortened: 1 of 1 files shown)


artokun:
## Closing — the test does not catch the bug it exists for. Measured, not argued.

The title says "through the real delivery boundary". It does not reach that boundary. I reintroduced the exact defect this test was written to catch — retiring the entry on the socket write alone — and ran it:

| run | result |
|---|---|
| baseline | 7 pass / 0 fail |
| **`else if (framePushed) markDelivered(...)`** — #1824 reintroduced | **7 pass / 0 fail** |
| restored | 7 pass / 0 fail |

Green with the bug back in place. The gate's P1 said exactly this: simulating the delivery handler's retirement in the harness cannot catch mutations in production code. The test drives the tracker directly and decides for itself when to call `markDelivered`, so `run-completion-delivery.js` never executes.

**A test for #1871 that survives #1824 being reintroduced is worse than no test**, because the file it sits in now advertises coverage that is not there — and #1871 exists precisely because that gap let a real defect ship. So this is closed rather than merged.

## The finding is real and worth keeping

The structural reason is correct and I verified it. The retirement is a conditional inside a `.then(composeRunCompletionFrame(...))` callback:

```js
if (frame == null && !awaitsCompletionKey) markDelivered(promptId, completionKey);
else if (framePushed && !awaitsReceipt && !awaitsCompletionKey) markDelivered(promptId, completionKey);
else if (framePushed && (awaitsReceipt || awaitsCompletionKey)) { /* receipt retires it */ }
```

Reaching that line means composing a real frame — image/video handling, blob and fetch paths — which the `node:test` unit harness has no way to provide. So **not unit-testable at this layer** is an accurate conclusion.

## But "code review only" is not the answer

The panel ships **44 Playwright specs** under `browser_tests/*.spec.ts`, against the served panel with a real socket. That is the layer where this path is reachable, and it is the honest home for this coverage. #1871 stays open, re-scoped to an e2e rather than a unit test.

Worth stating plainly for whoever picks it up: the assertion that matters is the middle one — **a completion whose write succeeded but whose receipt never arrived must remain pending and be re-delivered by the sweep.** The bound already has unit coverage in `run-completion-redelivery-storm.test.mjs`; the retire-on-receipt decision does not, and that is the half that failed in production.

artokun:
## The re-scope measured a different file. Reproducing both counts side by side.

I've reopened #1872. The acceptance criterion in the re-scope comment — *"Reintroducing #1824 must turn the new spec RED"* — is met, and here is the measurement rather than a description of one.

The tell is in the numbers. **`run-completion-redelivery-storm.test.mjs` has exactly 7 tests. The file this PR adds has 4.** The reported "7 pass / 0 fail, unchanged from baseline" is the storm suite — the file the issue itself names as *already* covering step 5, and the file this PR's header documents as staying green under precisely this mutation.

### Both files, the comment's exact mutation, clean `origin/main` (`7f7cbdab`)

Mutation applied verbatim as specified — `web/js/lib/run-completion-delivery.js:153` → `else if (framePushed) markDelivered(promptId, completionKey);`

| | baseline | with #1824 reintroduced |
|---|---|---|
| `run-completion-redelivery-storm.test.mjs` (7 tests) | 7 pass / 0 fail | **7 pass / 0 fail** ← the number in the comment |
| `run-completion-receiptless-delivery.test.mjs` (4 tests, this PR) | 4 pass / 0 fail | **2 pass / 2 fail** |

The failure is the load-bearing assertion, not a bystander:

```
AssertionError: step 2: a keyed completion is retired by its RECEIPT, never by the socket write
    actual: false, expected: true
```

Two further mutations, same file, same method:

- `maxUnackedDeliveries` 3 → 99 → fails step 5, `actual: 22, expected: 3`
- `unackedDeliveriesExhausted` → `>= 1` (recovery silently disabled) → fails step 4, `actual: 1, expected: 2`

Step 4 is the middle assertion the issue names as the one that would have caught #1824, and it fails independently of step 2.

### On "the unit layer cannot reach it"

That premise is falsifiable and false: `run-completion.test.mjs` has wired the real `createRunCompletionFlushHandler` since #1830 (lines 438 and 663), and `run-reconcile-stale-age.test.mjs:436` pins the panel's `onFlush:` wiring. This PR does the same — real `createRunCompletionTracker`, real `createRunCompletionFlushHandler`, `markDelivered`/`markUndelivered` pointed back at the tracker exactly as `comfyui-mcp-panel.js:38819` points them. A real frame is composed; only the transport and the composer's I/O helpers are doubles. That is why mutating production *does* change the result.

The objection that *"any harness that stands in for the delivery handler is deciding retirement itself"* is correct, and it is exactly why the first draft of this branch was thrown away — it re-implemented the line-153 condition inside its own `onFlush`, which made it a tautology. The version on the PR does not stand in for the handler; it calls it.

### On moving it to `browser_tests/*.spec.ts`

No objection to an e2e in addition — a real socket and the served bundle would cover things this cannot. Two things worth knowing before it is filed as the *only* acceptable layer: all 44 specs drive `browser_tests/fixtures/MockBridge.ts`, so the orchestrator is faked there too and the receipt would still be simulated by the harness; and none of the 44 currently reference `completion_key` or `acknowledgeDelivery`. The panel e2e also needs a live ComfyUI, which this run had none allocated for — so I have not run it, and this PR claims no browser coverage.

Happy to be wrong about any of this — every number above is one command against `origin/main` and reproduces in about four seconds.

Issue #1855, opened 2026-08-26
panel_add_node reports a custom title that PreviewAnimation immediately discards


artokun:
### What happened / root cause
`panel_add_node` accepted `title:"Preview Mask Animation"` for a `PreviewAnimation` node and returned that custom title in its success payload, but an immediate `panel_query_graph` read showed the node title had reverted to `Preview Animation`. The add result therefore reported state that was not persisted. The custom node likely rewrites its title during/after initialization, while panel_add_node snapshots success before that lifecycle completes.

### Steps to reproduce
1. On a live ComfyUI canvas call panel_add_node with class_type `PreviewAnimation` and title `Preview Mask Animation`.
2. Observe the returned added.title is `Preview Mask Animation`.
3. Immediately read the node using panel_query_graph fields=detail.
4. The actual title is `Preview Animation`.
5. Calling panel_edit_node afterward successfully persists the desired title.

### Exact error
No thrown error. Silent state mismatch: returned custom title differed from the actual live node title.

### Fix
Applied locally: no (running orchestrator code cannot be safely hot-patched in this task).
Upstream-only: yes.
Suggested change: after the node's initialization/configuration lifecycle settles, reapply the requested custom title and read back the final title before returning success. If a node intentionally owns a dynamic title and persistence cannot be guaranteed, return the actual final title plus a warning instead of echoing the requested title as successful.

Temporary workaround: call panel_edit_node(title=...) after panel_add_node; verified successful.

### Environment
Windows 11
ComfyUI 0.34.0
ComfyUI frontend 1.49.6
GPU NVIDIA GeForce RTX 5070 12 GB
Python 3.13.12
PyTorch 2.12.1+cu130
comfyui-mcp running 0.52.128 (0.52.129 installed on disk but not loaded)
panel 0.15.99

---
Filed from the ComfyUI panel by **codex** · model _unreported_ — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Claimed by the Claude autopilot swarm (worktree `.claude/worktrees/wt-1855-cc`) at 2026-08-26T12:35:21Z.

Triage so far: the add-time snapshot is already a live read of `node.title`, so an add-time re-read cannot be the fix. Tracing the actual clobber now.

artokun:
Fixed on main in 0b536db0 (#1856) — but **not where this issue predicted**, so the root cause is worth recording.

[...]

PR #1876, opened 2026-08-26
fix(1873): a connect that re-addresses its target's slots says so

### LINKED ISSUE #1873: Dynamic input links reported connected but omitted from queued prompt
panel_add_node + panel_connect reported successful links for custom nodes with dynamically generated optional inputs, and panel_query_graph showed the links, but the queued ComfyUI prompt omitted the selected input. This happened with both easy imageIndexSwitch (image0/image1/image2) and ImpactSwitch (input1/input2/input3). The first failed with NodeInputError saying image0 was absent; the second returned None and caused downstream D2 Send Eagle to fail while logs said ImpactSwitch invalid select index.

### Steps to reproduce
1. Add easy imageIndexSwitch or ImpactSwitch using panel_add_node.
2. Connect IMAGE producers to its dynamic inputs using panel_connect.
3. Confirm panel_query_graph reports connected_from for those inputs.
4. Connect its IMAGE output to SaveImage/D2 Send Eagle and queue.
5. The execution prompt omits the selected dynamic input even though the live graph reports it connected.

### Exact error (scrubbed)
NodeInputError: Node 550 says it needs input image0, but there is no input to that node at all
ImpactSwitch: invalid select index (ignored)
Downstream: TypeError: 'NoneType' object is not iterable

### Fix
Workflow workaround applied: replaced dynamic-input selector nodes with two fixed-input lazy Switch image [Crystools] nodes.
Precise product fix needed: panel_add_node/panel_connect must initialize custom dynamic-input node lifecycle exactly as a manual frontend connection does and verify graphToPrompt contains each reported dynamic link. If the dynamic link cannot serialize, reject the mutation instead of returning success.

### Environment
Windows 11
ComfyUI 0.33.3
ComfyUI frontend 1.49.6
NVIDIA RTX 3080 Ti 12 GB
comfyui-mcp 0.52.97
panel 0.15.101

artokun:
Fixes #1873.

## What the reporter saw

`panel_add_node` + `panel_connect` reported successful links into dynamic-input
selector nodes (`easy imageIndexSwitch`, `ImpactSwitch`), `panel_query_graph`
showed them, and the queued prompt did not carry the selected input:

```
NodeInputError: Node 550 says it needs input image0, but there is no input to that node at all
ImpactSwitch: invalid select index (ignored)
```

## Root cause, read from source rather than inferred

**1. That backend message means the prompt has no key by that name.**
`comfy_execution/graph.py:123`, `TopologicalSort.make_input_strong_link`:

```python
inputs = self.dynprompt.get_node(to_node_id)["inputs"]
if to_input not in inputs:
    raise NodeInputError(f"Node {to_node_id} says it needs input {to_input}, but there is no input to that node at all")
```

`easy imageIndexSwitch` declares `image0..image19` as **optional + lazy** and its
`check_lazy_status` asks for `"image%d" % index`, so this fires when the prompt's
`inputs` dict has no `image0` key.

**2. The prompt is keyed on the LIVE SLOT NAME at serialize time.** From
`graphToPrompt` in `comfyui_frontend_package` **1.49.6** — the reporter's exact
version, read out of the installed bundle:

```js
for (const [i, slot] of dto.inputs.entries()) {
  const resolved = dto.resolveInput(i)
  if (resolved) { ... inputs[slot.name] = [String(origin_id), parseInt(origin_slot)] }
}
```

with `dto.inputs = node.inputs.map(s => ({ linkId: ..., name: s.name, type: s.type }))`.
So `image0` being absent does **not** mean a wire is missing — it means **no live
slot is called `image0` any more**.

**3. The packs rename the slots, from inside the connect.** LiteGraph runs
`onConnectionsChange` from within `connectSlots`. ComfyUI-Impact-Pack
(`js/impact-pack.js`, verified byte-identical against the installed pack) rebuilds
every name from its **position** on every change:

```js
let slot_i = 1;
for (let i = 0; i < this.inputs.length; i++) {
  let input_i = this.inputs[i];
  if (input_i.name != 'select' && input_i.name != 'sel_mode') {
    input_i.name = `${input_name}${slot_i}`;   // renames EVERY slot
    slot_i++;
  }
}
if (connected) this.addInput(`${input_name}${slot_i}`, this.outputs[0].type);
if (this.widgets?.[0]) {
  this.widgets[0].options.max = this.inputs.length - 3;
  this.widgets[0].value = Math.min(this.widgets[0].value, ...);  // and CLAMPS select
}
```

It also rewrites `outputs[0].name` to the propagated type. ComfyUI-Easy-Use's
`*IndexSwitch` nodes do the positional variant: `addInput(prefix + s.length, "*")`
when every slot is filled, `removeInput(i + 1)` when a trailing one empties.

**So:** `panel_connect` reported the name it saw and said nothing when a *later*
`panel_connect` to the same node moved it. The caller then set `index` / `select`
against a name that had already shifted, and the failure surfaced several steps
later, from the backend, naming a slot the panel had reported as connected.

## The change

Discloses the re-addressing — `slots_rewritten` plus a `warning` — exactly as
`node-title-rewrite.js` does for titles in **#1855** (merged yesterday: same hook,
same "the caller is left holding a stale name" defect).

It renames nothing back. Positional naming is the contract that makes
`select`/`index` mean anything, so restoring an old name would break the
addressing the node depends on.

**Noise control (the load-bearing design choice):** only state the caller *could
already hold* is reported — a slot index that existed **before** the connect and
now carries a different name or is gone. A brand-new trailing slot is never
reported, because materialising the next empty input is the expected behaviour of
every node in this class. An ordinary connect's payload is byte-identical to
before.

Widget values are captured the same way, which covers the second reported
symptom (`ImpactSwitch: invalid select index`) — the pack silently clamps
`select` during the connect.

## Please look hardest at

- **`describeSlotRewrites`'s index bound** (`i < before.length`) — this is what
  separates "the pack renamed your slot" from "the pack added the next empty
  slot". Get it wrong and every correct ImpactSwitch connect gets a warning.
- **Both reply paths.** The rename happens on the throw path and the success path
  alike; `slotRewrites` is computed once before the verdict branches for exactly
  that reason. The success path now merges two riders into the single `warning`
  key so neither can drop the other's sentence.
- **`Object.is` for widget values** (not `!==`) so a NaN-valued widget is not
  reported as changing on every connect.

## Verification

- `node --test browser_tests/unit/` — full unit suite.
- Mutation-checked: the new assertions are pinned to this mechanism via the
  harness's `titleDeps`-style dependency-neutering, and by reverting the call-site
  wiring. See the comment below for the recorded runs.

## What I did NOT do

- **I could not run the Playwright browser gate.** It requires a live ComfyUI at
  `localhost:8188` (`playwright.config.ts`: "this suite does NOT start ComfyUI for
  you"); none is listening on this machine and this run was allocated no instance.
  So there is **no browser coverage for this change** — the evidence here is the
  pack/frontend/backend source quoted above plus the unit + call-site tests.
- **This does not make the packs' lifecycle correct**, and it does not retro-fix a
  graph that is already mis-addressed. It stops the panel reporting a slot name
  that its own next call invalidates, which is the part the panel owns.

### DIFF
--- a/browser_tests/unit/connect-slot-rename.test.mjs
+++ b/browser_tests/unit/connect-slot-rename.test.mjs
@@ -463,6 +463,39 @@ test("#1873 the disclosure is produced by the FIX, not by the fixture", () => {
   assert.equal(res.warning, undefined);
 });
 
+test("#1873 a node that becomes unreadable mid-connect does not fail a LANDED wire", () => {
+  const { graph, sw } = loadedSwitchFixture();
+  // The pack breaks its own `widgets` accessor from inside the connect. This is
+  // the one property `describeSlotRewrites` reads that `graph_connect` does not,
+  // so a throw here can ONLY escape through the disclosure rider — which is
+  // exactly the contract its header promises it never does. The capture already
+  // happened, so the BEFORE snapshot exists and the read-back is what throws.
+  let broken = false;
+  const original = sw.widgets;
+  Object.defineProperty(sw, "widgets", {
+    configurable: true,
+    get() {
+      if (broken) throw new TypeError("widgets accessor removed by the pack");
+      return original;
+    },
+  });
+  const source = graph.getNodeById(1);
+  const prior = source.connect;
+  source.connect = (outIdx, target, inIdx) => {
+    const link = prior(outIdx, target, inIdx);
+    broken = true;
+    return link;
+  };
+  const graph_connect = buildConnect(graph);
+
+  const res = graph_connect({ from_node_id: 1, from_output: 0, to_node_id: 2, to_input: "input3" });
+
+  // The wire landed; the unreadable node must cost the caller nothing but the
+  // disclosure it could not compute.
+  assert.equal(res.connected.to.node_id, 2);
+  assert.equal(res.slots_rewritten, undefined);
+});
+
 test("#1873 a connect that changes nothing addressable stays byte-identical", () => {
   const graph = mkGraph();
   const source = mkNodeWithOutput(graph, 1, "Load Image");
--- a/browser_tests/unit/connect-slot-rename.test.mjs
+++ b/browser_tests/unit/connect-slot-rename.test.mjs
@@ -496,6 +496,43 @@ test("#1873 a node that becomes unreadable mid-connect does not fail a LANDED wi
   assert.equal(res.slots_rewritten, undefined);
 });
 
+test("#1873 a hostile widget VALUE cannot fail a landed wire (gate P1)", () => {
+  // A widget value is whatever the pack put there, and it is interpolated into
+  // the warning sentence. `JSON.stringify` throws on two shapes that a pack can
+  // genuinely produce, and the throw would come from a rider running AFTER the
+  // wire landed — reporting a successful connect as an error.
+  const circular = { name: "loop" };
+  circular.self = circular;
+  for (const [label, hostile, expected] of [
+    ["BigInt", 2n, /widget "select" 1 → 2n/],
+    ["circular", circular, /widget "select" 1 → \(unrenderable\)/],
+  ]) {
+    const graph = mkGraph();
+    const source = mkNodeWithOutput(graph, 1, "Load Image");
+    const node = {
+      id: 2,
+      title: "Switch",
+      graph,
+      inputs: [{ name: "input1", type: "IMAGE", link: null }],
+      outputs: [{ name: "IMAGE", type: "IMAGE", links: [] }],
+      widgets: [{ name: "select", value: 1, options: { max: 1 } }],
+    };
+    graph.nodes.push(node);
+    attachConnect(source, {
+      onChange: () => {
+        node.widgets[0].value = hostile;
+      },
+    });
+    const graph_connect = buildConnect(graph);
+
+    const res = graph_connect({ from_node_id: 1, from_output: 0, to_node_id: 2, to_input: "input1" });
+
+    assert.equal(res.connected.to.node_id, 2, `${label}: the wire landed`);
+    assert.equal(res.slots_rewritten.length, 1, `${label}: the change is still disclosed`);
+    assert.match(res.warning, expected, `${label}: the value is rendered, not stringified raw`);
+  }
+});
+
 test("#1873 a connect that changes nothing addressable stays byte-identical", () => {
   const graph = mkGraph();
   const source = mkNodeWithOutput(graph, 1, "Load Image");
--- a/browser_tests/unit/connect-slot-rename.test.mjs
+++ b/browser_tests/unit/connect-slot-rename.test.mjs
@@ -511,8 +511,11 @@ test("#1873 hostile widget VALUES are safe in the full bridge envelope (gate P1)
     },
   });
   const oversized = { payload: "x".repeat(1000) };
+  const giantBigIntDigits = "9".repeat(100_000);
+  const giantBigInt = BigInt(giantBigIntDigits);
   for (const [label, hostile, expected] of [
     ["BigInt", 2n, "2n"],
+    ["giant BigInt", giantBigInt, null],
     ["circular", circular, "(unrenderable)"],
     ["oversized", oversized, null],
     ["throwing getter", throwingGetter, "(unrenderable)"],
@@ -553,9 +556,24 @@ test("#1873 hostile widget VALUES are safe in the full bridge envelope (gate P1)
       assert.equal(typeof rewrite?.to, "string", `${label}: oversized value is represented as text`);
       assert.ok(rewrite.to.length <= 121, `${label}: disclosure value is bounded`);
       assert.equal(rewrite.to.endsWith("…"), true, `${label}: truncation is disclosed`);
-      assert.equal(encoded.includes("x".repeat(1000)), false, `${label}: raw object is not leaked`);
+      const rawValue = label === "giant BigInt" ? `${giantBigIntDigits}n` : "x".repeat(1000);
+      assert.equal(encoded.includes(rawValue), false, `${label}: raw value is not leaked`);
     }
   }
+});
+
+test("#1873 warning rendering bounds a direct giant BigInt too", () => {
+  const giantDigits = "9".repeat(100_000);
+  const giant = BigInt(giantDigits);
+  const warning = slotRewriteWarning([
+    { node_id: 2, slots: [], widgets: [{ name: "select", from: 1, to: giant }] },
+  ]);
+  const reply = { rid: "giant-warning-rid", ok: true, result: { warning } };
+
+  assert.doesNotThrow(() => JSON.stringify(reply), "the direct warning envelope must serialize");
+  assert.ok(warning.length < 2000, "the warning stays bounded");
+  assert.equal(warning.includes(`${giantDigits}n`), false, "the giant BigInt is not leaked");
+  assert.match(warning, /widget "select" 1 → 9+…/);
 });
 
 test("#1873 a connect that changes nothing addressable stays byte-identical", () => {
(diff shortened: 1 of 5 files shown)


artokun:
## Recorded verification

### Unit suite

`npm run test:unit` on this branch (rebased onto `1b50086c`, v0.15.104):

```
ℹ tests 7020
ℹ pass 7015
ℹ fail 4
```

The 4 failures are **pre-existing and unrelated** — `#2314` fence tests in
`browser_tests/unit/graph-set-widget-target-fence.test.mjs`. Verified by running
that file in a detached worktree at clean `origin/main`:

```
✖ #2314 production scope fence refuses receiver navigation after graph_query reply
✖ #2314 production graph_enter_subgraph -> graph_set_widget fence refuses a parent-rail relink
✖ #2314 same owner id in a different graph is refused at the shipped fence
✖ #2314 terminal endpoint is part of the shipped receiver fence
ℹ tests 6   ℹ pass 2   ℹ fail 4
```

Same four, same file, without this branch's commits.

### Mutation runs — every assertion is pinned to a specific load-bearing line

A green suite proves nothing on its own, so each site was broken and the tests
re-run. Restores were done from file copies, never `git checkout` (that would
have taken the fix with it); the tree was confirmed identical to `HEAD` afterwards
(`git status --porcelain` empty, `git diff HEAD` empty) and re-run green.

| # | Mutation | Expected red | Observed |
|---|---|---|---|
| M1 | throw path: drop `slots_rewritten` | throw-path test | ✖ `the disclosure survives the THROW path` |
| M2 | success path: drop `slots_rewritten` (line 16579) | rename + widget tests | ✖ `a slot the pack RENAMES is disclosed`, ✖ `the select widget the pack CLAMPS is disclosed` |
| M3 | success path: drop `slotRewriteWarning` from the merged `warning` | rename + widget tests | ✖ both |
| M4 | `describeSlotRewrites`: `i < before.length` → `i < after.length` | the **noise-control** test | ✖ `the appended trailing slot is NOT reported` (+ 2 others) |
| M5 | `describeSlotRewrites`: drop the widget diff | widget test | ✖ `the select widget the pack CLAMPS is disclosed` |

M4 is the one worth noting: it is the only mutation that makes the guard *louder*
rather than quieter, and it is caught — so the suppression that keeps an ordinary
dynamic-input connect silent is genuinely pinned, not incidental.

### Call-site coverage, not helper-only

`connect-slot-rename.test.mjs` extracts the **shipped** `graph_connect` out of
`web/js/comfyui-mcp-panel.js` with `new Function`, the same technique as
`connect-title-rewrite` / `connect-throw-verdict`. The M2/M3 reds are proof the
wiring is covered: they mutate the panel monolith, not the helper module.

The fixture's `impactOnConnectionsChange` transcribes ComfyUI-Impact-Pack's real
hook (positional rename loop → trailing `addInput` → `select` clamp), verified
byte-identical against the pack installed on this machine.

`browser_tests/unit/connect-throw-verdict.test.mjs` and
`connect-title-rewrite.test.mjs` also needed the three new names added to their
`new Function` dependency lists — a new module binding is a `ReferenceError` in
every extraction scope otherwise. Both are in this branch and green (46/46 across
the three connect files).

### Not covered

No browser run — see the PR body. `playwright.config.ts` states the suite does not
start ComfyUI, nothing is listening on `localhost:8188` here, and no instance was
allocated to this run. There is a ComfyUI process on this box (pid 48264, started
18h ago) but it is bound to no port.

artokun:
## Follow-up: a defect I found reviewing my own module

Pushed as `7e491a53`, on top of the two commits the evidence above covers.

`slot-rename-disclosure.js`'s header promises the disclosure **never throws** —
it rides on a verdict that has already been decided, so it must never turn a wire
that landed into a reported failure. `captureSlotNames` honoured that with a
per-node `try`. **`describeSlotRewrites` did not:** it read `node.inputs`,
`node.outputs`, `node.widgets` and `node.id` unguarded.

Those are property *reads*, and an extension is free to install a throwing getter
or a Proxy over any of them — cg-use-everywhere already replaces frontend API
surface this panel calls. Such a read escapes **after** the wire has landed, which
is precisely the outcome #1272 spent a whole issue undoing: reporting failure for
a connect whose link is on the graph, and inviting the retry that duplicates or
tears down correct wiring.

### Pinned, not just fixed

The new test breaks the target's **`widgets`** accessor from inside the connect.
That property is the one `describeSlotRewrites` reads and `graph_connect` does
not, so a throw there can *only* escape through the new rider — no other path can
produce the failure, and no other path can mask it. The capture runs first, so the
BEFORE snapshot exists and it is the read-back that throws.

Mutation run — the `catch` body replaced with a rethrow, everything else identical:

```
✔ #1873 shipped graph_connect: a slot the pack RENAMES is disclosed
✔ #1873 the appended trailing slot is NOT reported — an ordinary connect is unchanged
✔ #1873 the select widget the pack CLAMPS is disclosed
✔ #1873 the disclosure survives the THROW path
✔ #1873 the disclosure is produced by the FIX, not by the fixture
✖ #1873 a node that becomes unreadable mid-connect does not fail a LANDED wire
✔ #1873 a connect that changes nothing addressable stays byte-identical
```

Exactly one test moves, and it is the one that names the mechanism. Module restored
from a file copy (never `git checkout`, which would have taken the fix with it) and
re-confirmed clean.

### Suite after the rebase onto `207298ad`

```
ℹ tests 7022   ℹ pass 7017   ℹ fail 4
```

Still the same four pre-existing `#2314` fence failures.

### Gate note

The first gate run was killed mid-review by a session restart and left a zero-byte
log — an empty verdict is INDETERMINATE, never a pass, so it was **not** banked.
Re-running detached against `7e491a53`, the SHA that would merge.

artokun:
## Gate round 1: NO-SHIP — finding accepted and fixed

Verdict on `7e491a53`:

> **[P1]** `web/js/lib/slot-rename-disclosure.js:227` — `slotRewriteWarning` can throw
> while formatting widget values, converting a successful landed connect into an error —
> a widget value changes from `1n` to `2n` (or contains a circular object);
> `JSON.stringify` throws `TypeError: Do not know how to serialize a BigInt`.

**Correct, and it is the half I missed.** `7e491a53` hardened the *compute* half
(`describeSlotRewrites`) against unreadable nodes and left the *render* half open.
Slot **names** were never at risk — `slotNames` already coerces them to
`string | null` — but widget **values** are whatever the pack assigned, and they
were interpolated through a bare `JSON.stringify`. Both throw modes are reachable:

| value | `JSON.stringify` |
|---|---|
| `BigInt` | `TypeError: Do not know how to serialize a BigInt` |
| circular object | `TypeError: Converting circular structure to JSON` |

Either escapes from a rider that runs **after** the wire has landed — the same
false-failure class as the previous commit, and the one #1272 exists to prevent.

### Fixed in `d8fb255f`

`renderValue`: never throws, handles `bigint` / `symbol` / `function` explicitly,
falls back to `(unrenderable)`, and bounds output at 120 chars so a widget holding
a large object (a Load3D transform, a serialized curve) cannot paste its whole JSON
into a sentence. `slotRewriteWarning` also builds the itemisation inside a `try` —
if that fails the **disclosure still goes out**, because the caller needs to know
their names moved far more than they need the list.

### The finding is reproduced, not just patched

New test drives both shapes through the shipped executor. Reverting to the raw
`JSON.stringify` reproduces the gate's exact error and fails **only** that test:

```
✔ #1873 shipped graph_connect: a slot the pack RENAMES is disclosed
✔ #1873 the appended trailing slot is NOT reported — an ordinary connect is unchanged
✔ #1873 the select widget the pack CLAMPS is disclosed
✔ #1873 the disclosure survives the THROW path
✔ #1873 the disclosure is produced by the FIX, not by the fixture
✔ #1873 a node that becomes unreadable mid-connect does not fail a LANDED wire
✖ #1873 a hostile widget VALUE cannot fail a landed wire (gate P1)
  TypeError: Do not know how to serialize a BigInt
✔ #1873 a connect that changes nothing addressable stays byte-identical
```

Module restored from a file copy and confirmed byte-identical afterwards.

### Suite

```
ℹ tests 7023   ℹ pass 7018   ℹ fail 4
```

Same four pre-existing `#2314` fence failures. Re-gating on `d8fb255f`.

Issue #1859, opened 2026-08-26
panel_set_widget refuses promoted subgraph widget write with "did not publish a verifiable workflow and viewing-scope identity"


artokun:
### What happened / root cause
Calling `panel_set_widget` on a subgraph instance node's PROMOTED widget (e.g. `prompt` on a `MiniMaxH3ImageToVideo`-style I2V subgraph instance) consistently failed with:

`panel_set_widget refused the promoted "prompt" write because graph_get_subgraph did not publish a verifiable workflow and viewing-scope identity. No graph_set_widget was dispatched; retry only after the panel binding and subgraph mapping are stable.`

This happened on ~5 consecutive retries across multiple turns, including after:
- Re-issuing `panel_set_workflow_target(mode:"current")` (reported `already_current` / bound each time)
- The user doing a hard browser refresh (F5) of the ComfyUI tab, then re-binding again

Other writes in the SAME session/workflow succeeded fine during this window (e.g. `panel_set_widget` on a different node's plain `aspect_ratio` widget, and `steps` inside the same subgraph), so this is not a general binding/session problem — it looks specific to this one promoted-widget mapping going stale, likely correlated with an earlier `panel_restart_comfyui` / headless `restart_comfyui` cycle done earlier in the session for an unrelated fix.

### Steps to reproduce
1. Open a workflow containing a subgraph node with a widget promoted from an inner node (e.g. a MiniMax H3 Image-to-Video helper subgraph, `prompt` promoted from an inner `MiniMaxH3ImageToVideo` node).
2. Successfully edit that promoted widget earlier in the session (worked fine).
3. At some point restart ComfyUI mid-session (`panel_restart_comfyui` and/or headless `restart_comfyui(action:"restart")`), for an unrelated reason.
4. Later in the same session, try `panel_set_widget` again on that same promoted widget (same node id, same widget name) — it now fails every time with the error above, even after `panel_set_workflow_target(mode:"current")` and a hard browser refresh.

### Exact error (scrubbed)
```
Error: panel_set_widget refused the promoted "prompt" write because graph_get_subgraph did not publish a verifiable workflow and viewing-scope identity. No graph_set_widget was dispatched; retry only after the panel binding and subgraph mapping are stable.
```

### Fix
[...]

PR #1885, opened 2026-08-26
fix(1882): a release tag proves the tagged tree is the version it claims

### LINKED ISSUE #1882: Release 0.15.97 published to the registry with no git tag — tags are an unreliable 'was this released?' operand
**0.15.97 is published on the Comfy Registry and has no git tag.**

    registry : 0.15.97  created 2026-08-26T03:01:45Z  status Flagged  deprecated=false
    git tags : v0.15.96 -> v0.15.98   (no v0.15.97)
    commit   : cd50f093 "chore: release v0.15.97"

The release commit exists and the artifact shipped; only the tag is missing. Users can install 0.15.97 today.

### Why this is worth fixing rather than shrugging at

The tag list is used as an operand for "was this released?", and it silently gives the wrong answer. I hit this directly: I told two reporters on #1859 and #1860 that 0.15.97 "was never released" because the tags skip it. That was wrong and I have corrected it on both threads.

The mapping is unreliable in the other direction too — `v0.15.96` declares `version = "0.15.85"` in its own `pyproject.toml`, so neither "tag exists" nor "tag's pyproject says X" can be trusted alone.

Practical consequence: the pre-release check of "does every merged fix reach this cut" uses `git merge-base --is-ancestor <sha> <tag>`. A missing tag makes that check unrunnable for the affected release, and a stale pyproject makes it answer about the wrong version.

### Not proposing to create the tag

Pushing `v0.15.97` now would fire the publish workflow and re-publish 0.15.97 to the registry. That is the opposite of harmless, so this needs a decision, not a quick fix.

Options, roughly in order of how much I would trust them:

1. Have the publish workflow **fail** if `pyproject.toml`'s version has no matching tag at the published SHA — turns both failure modes into a red build instead of a silent gap.
2. Reconcile the record with **annotated tags created without triggering publish** (whatever the workflow's trigger filter allows), so history is queryable.
3. Stop deriving "released" from tags at all and read the registry API, which is the thing users actually install from.

Related: the capability floor question that surfaced this. `enforces_expected_scope_at_write` first ships in **0.15.101** (absent at .97/.98/.99/.100), so the corrected floor in artokun/comfyui-mcp#2366 is right — it was the *reason* I gave for it that came from the bad operand.

artokun:
Fixes #1882.

Implements the fix you narrowed to in the 21:51Z comment — the assertion on the tag/release path, plus the stale pack-contents entry — and covers the untagged direction too, which a tag-push trigger can't reach on its own.

## The two operand failures

| | what happened |
|---|---|
| tag exists, tree stale | `v0.15.86`..`v0.15.96` are tagged; every one of those trees declares `version = "0.15.85"`. `publish_action.yml` fires on `paths: [pyproject.toml]`, so on tagged commits that never touched that file it **never ran** — no job, no red X. Ten versions silently never shipped. |
| tree shipped, no tag | `0.15.97` is live on the Registry from the merge of `release/0.15.97`, with no tag. Two reporters on #1859/#1860 were told it "was never released". |

## What this adds

**`.github/workflows/release-tag-guard.yml`** — `on: push: tags: ['v*']`, plus `workflow_dispatch` with a `tag` input so a historical tag can be audited without pushing anything. It runs **`scripts/check-release-tag.mjs`**: pure node + git, no install.

**Rule 1 — the tagged tree is the version it claims.** At `v<X>`, `pyproject.toml`, `package.json` and `PANEL_VERSION` must all read `<X>`. All three, for the reason `ci.yml` already spells out: pyproject and PANEL_VERSION are written by the same script, so they cannot disagree when that script is simply never run. This is `ci.yml`'s three-way gate anchored to the **tag**, which `ci.yml` cannot see.

**Rule 2 — nothing between the previous tag and this one went untagged.** Walk first-parent from the previous reachable tag; every commit whose pyproject version differs from its parent's cut a version, so it must carry a tag **that resolves to that commit**. This is what lets a tag-only trigger see the untagged direction.

Rule 2 is deliberately not in the publish job. Asserting "a tag exists at the SHA I am publishing" there would be **racy** — the tag is pushed after the release merge — and would abort legitimate releases. Rule 2 only looks at versions cut *strictly before* the tag being pushed, so a tag pushed minutes after its own release commit is never in its own scan. It is also self-bounding (range starts at the previous reachable tag), so it can't go permanently red over a historical gap nobody intends to backfill.

## One correction to the issue

> That single assertion catches every case seen here: the ten silent no-ops, and the untagged 0.15.97 from the opposite direction.

The tree-vs-tag assertion alone does **not** catch the untagged direction — 0.15.97 has no tag, so no tag push happens, so nothing fires. That's why rule 2 exists: it finds the gap on the **next** tag push. Pushing `v0.15.98` today goes red naming `7b477d55` / `0.15.97`.

## Also: `web/js/vendor/a2ui-lit.bundle.js`, and the class it belongs to

Dropped from the pack-contents gate in **both** `ci.yml` and `publish_action.yml` — you flagged it in `publish_action.yml`; the identical stale entry was in `ci.yml` too. #1865 deleted the file in 0.15.103.

Rather than just deleting the line, the loop now **fails on a listed-but-missing path**. `check-ignore` passes trivially on a path that doesn't exist, so a stale entry reads as coverage while asserting nothing — the same "guard that isn't a guard" shape as the rest of this issue.

## Review round 1 found three real defects in my first draft

The gate was right on all of them; each is fixed and mutation-verified.

1. **It failed open.** `collectFromGit` turned every git failure into an empty range, so a shallow clone or unfetched tags made rule 2 silently vacuous *while the job reported success* — the identical shape as the stale pack-contents entry this PR removes. Unavailable history is now a violation. Proved on a real `git clone --depth 1`: **exit 1 with the fix, exit 0 without it.**
2. **It credited a tag by name, not target.** `v0.15.97` sitting on an unrelated commit would have satisfied the guard while the commit that actually published 0.15.97 stayed untagged — the very confusion the guard exists to end. The tag must now resolve to the commit that cut the version. Verified this is strictly stronger with no false positives: every tag `v0.15.99`..`v0.15.105` already points exactly at its own version-bumping commit.
3. **It prescribed a tag unconditionally.** `publish_action.yml` runs once per push against the head commit, so if two bumps land in one push only the later one publishes — the earlier version was cut but never shipped, and tagging it would recreate the same bad operand in the other direction. The message now names both readings and both remedies. **It still blocks the release, deliberately:** a consumed-but-unshipped version is itself a release-record defect, and that's a judgement call worth your override if you disagree.

It also corrected a wrong claim in my prose: **GitHub loads workflow files from the ref the event carries**, so a tag pushed at a commit predating this file runs no guard. A historical backfill is still safe (publish is branch-filtered at that commit too) but it is *unaudited*; `workflow_dispatch` is how you audit one after the fact.

## Evidence

**Live audit against real history** (`node scripts/check-release-tag.mjs <tag>`):

```
v0.15.90   exit 1  pyproject.toml declares 0.15.85, but the tag says 0.15.90
                   PANEL_VERSION declares 0.15.85, but the tag says 0.15.90
v0.15.96   exit 1  (same)
v0.15.98   exit 1  0.15.97 was cut by 7b477d55 ("Merge pull request #1826
                   from artokun/release/0.15.97") ... but no v0.15.97 tag exists
v0.15.101  exit 0  coherent
v0.15.104  exit 0  coherent
v0.15.105  exit 0  coherent   <- released mid-PR; the guard passes the live cut
```

Red on exactly the two documented incidents, green on every healthy release including the one that landed while this branch was open.

**Mutation-verified** — each guard deleted, the named test watched go red, then restored. Baseline clean after every one.

| mutation | result |
|---|---|
| revert both workflows to `origin/main` (a2ui-lit entry back) | `every path in the pack-contents gate exists` FAILS |
| delete `release-tag-guard.yml` | both wiring tests FAIL |
| `fetch-depth: 0` → `1` | wiring test FAILS |
| rule 1 comparison → `false` | unit test FAILS **and** live `v0.15.90` audit drops to exit 0 |
| rule 2 `hasTagFor` → always-continue | unit test FAILS **and** live `v0.15.98` audit drops to exit 0 |
| fail-closed preflight → `false` | 2 unit tests FAIL **and** the real shallow clone drops to exit 0 |
| credit tag by name only (`target && target === c.sha` → `target`) | `a tag is credited only when it resolves to the commit that cut the version` FAILS |

16 tests in `browser_tests/unit/release-tag-guard.test.mjs`; full suite 7041 pass / 0 fail; `tsc --noEmit` clean.

## Where the load-bearing claims are — please push on these

1. **"A tag push does not trigger `publish_action.yml`."** Everything rests on this, including your option-2 backfill. Documented behaviour (`on.push` filtered by `branches` does not run for a tag ref) and empirically confirmed twice over: the ten tags `v0.15.86`..`v0.15.95` were all pushed with no publish run appearing, and every publish run in the Actions history is `event=push ref=main` — never a tag ref. A test pins `publish_action.yml` to `branches`-only, so adding a `tags:` filter later goes red.
2. **`fetch-depth: 0` is load-bearing, not hygiene.** On a shallow checkout the range cannot be walked at all. That used to pass silently; now it's a violation, demonstrated against a real depth-1 clone.
3. **Rule 2 keys on "pyproject version changed", not on commit subject.** That's the operand the publish trigger itself uses, so it can't drift from what actually publishes. It reports the **first-parent merge commit** (`7b477d55`) — the SHA that pushed to main and ran the publish — not `cd50f093` from the issue body, which is the branch-side commit with the same tree.
4. **The batched-push trade-off in item 3 above** is the one place I chose to keep a red build over a possible false alarm. Worth your call.
5. `github.event.inputs.tag || github.ref_name` rather than `inputs.tag` — the `inputs` context is only populated for `workflow_dispatch`/`workflow_call`; `github.event.inputs` is valid on every event.

## Deliberately not done

- **Not creating `v0.15.97`.** You said that needs a decision, not a quick fix, and it stays yours. It would not re-publish; it also would not run this guard (see the correction above), so audit it afterwards with `workflow_dispatch`.
- **No auto-tagging from the publish job.** It would close the untagged gap at the source, but needs `contents: write` on the release path and is a bigger change than this issue asked for. Worth a follow-up.
- **No Registry API call** (option 3). Rule 2 gets the same answer from git alone, with no token and no network in the guard.
- **No browser verification** — nothing here renders in the panel; the diff is `.github/`, `scripts/`, `browser_tests/unit/` and `CHANGELOG.md`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,7 +202,7 @@ jobs:
       # The registry applies .comfyignore with gitignore semantics: a bare
       # `vendor/` matches a vendor dir at ANY depth, so the exclusion meant for
       # the root CI artefact also silently dropped web/js/vendor/ (the runtime
-      # marked/DOMPurify/qrcode/a2ui-lit imports) from the pack — the browser
+      # marked/DOMPurify/qrcode imports) from the pack — the browser
       # 404'd on import and the sidebar never registered (#749). The parity
       # scan above reimplements the matching root-anchored and cannot see this
       # class of loss, so assert with git's OWN ignore engine instead (empty
@@ -218,8 +218,16 @@ jobs:
             web/js/vendor/marked.esm.js \
             web/js/vendor/purify.es.js \
             web/js/vendor/qrcode.esm.js \
-            web/js/vendor/a2ui-lit.bundle.js \
           ; do
+            # A path that no longer exists passes check-ignore trivially, so a stale
+            # entry reads as coverage while asserting nothing at all.
+            # web/js/vendor/a2ui-lit.bundle.js sat in this list after #1865 deleted the
+            # file, and every publish stayed green over it (#1882). Fail on the stale
+            # entry itself, so this list cannot rot back into a no-op.
+            if [ ! -e "$f" ]; then
+              echo "::error file=.github/workflows/ci.yml::$f is listed in the pack-contents gate but does not exist — remove the entry or restore the file; a listed-but-missing path makes this check silently vacuous"
+              exit 1
+            fi
             if git -c core.excludesFile=.comfyignore check-ignore -q "$f"; then
               echo "::error file=.comfyignore::$f is exclude-matched but imported at runtime — the published pack would 404 on import and the sidebar would never register (#749)"
               exit 1
--- a/.github/workflows/publish_action.yml
+++ b/.github/workflows/publish_action.yml
@@ -38,7 +38,7 @@ jobs:
       # The registry applies .comfyignore with gitignore semantics: a bare
       # `vendor/` matches a vendor dir at ANY depth, so the exclusion meant for
       # the root CI artefact also silently dropped web/js/vendor/ (the runtime
-      # marked/DOMPurify/qrcode/a2ui-lit imports) from the pack — the browser
+      # marked/DOMPurify/qrcode imports) from the pack — the browser
       # 404'd on import and the sidebar never registered (#749). The parity
       # scan below reimplements the matching root-anchored and cannot see this
       # class of loss, so assert with git's OWN ignore engine instead (empty
@@ -54,8 +54,16 @@ jobs:
             web/js/vendor/marked.esm.js \
             web/js/vendor/purify.es.js \
             web/js/vendor/qrcode.esm.js \
-            web/js/vendor/a2ui-lit.bundle.js \
           ; do
+            # A path that no longer exists passes check-ignore trivially, so a stale
+            # entry reads as coverage while asserting nothing at all.
+            # web/js/vendor/a2ui-lit.bundle.js sat in this list after #1865 deleted the
+            # file, and every publish stayed green over it (#1882). Fail on the stale
+            # entry itself, so this list cannot rot back into a no-op.
+            if [ ! -e "$f" ]; then
+              echo "::error file=.github/workflows/publish_action.yml::$f is listed in the pack-contents gate but does not exist — remove the entry or restore the file; a listed-but-missing path makes this check silently vacuous"
+              exit 1
+            fi
             if git -c core.excludesFile=.comfyignore check-ignore -q "$f"; then
               echo "::error file=.comfyignore::$f is exclude-matched but imported at runtime — the published pack would 404 on import and the sidebar would never register (#749)"
               exit 1
--- a/.github/workflows/release-tag-guard.yml
+++ b/.github/workflows/release-tag-guard.yml
@@ -19,7 +19,11 @@ name: Release tag guard
 # filters `on.push` by `branches: [main]`, and GitHub does not run a
 # `branches`-filtered push workflow for a tag ref — empirically confirmed by the
 # ten tags above, every one of which was pushed with no publish run appearing.
-# So creating a missing historical tag is safe: it runs this audit and nothing else.
+# So creating a missing historical tag is safe. Note the converse, which is easy
+# to get backwards: GitHub loads workflow files from the ref the event carries,
+# so a tag pushed at a commit predating THIS file runs no guard either. A
+# historical backfill is therefore safe but unaudited -- audit it after the fact
+# with the workflow_dispatch input below, which runs this file from main.
 
 on:
   push:
--- a/.github/workflows/release-tag-guard.yml
+++ b/.github/workflows/release-tag-guard.yml
@@ -52,5 +52,5 @@ jobs:
 
       - name: Audit the release tag
         env:
-          TAG: ${{ inputs.tag || github.ref_name }}
+          TAG: ${{ github.event.inputs.tag || github.ref_name }}
         run: node scripts/check-release-tag.mjs "$TAG"
--- a/.github/workflows/release-tag-guard.yml
+++ b/.github/workflows/release-tag-guard.yml
@@ -1,0 +1,56 @@
+name: Release tag guard
+
+# "Was version X released?" is answered from `git tag`, and that operand has been
+# wrong in BOTH directions on this repo (#1882):
+#
+#   * v0.15.86..v0.15.96 are tagged but their trees all declare version "0.15.85".
+#     publish_action.yml fires on `paths: [pyproject.toml]`, so for tagged commits
+#     that never touched that file the workflow NEVER RAN — no job, no red X, no
+#     annotation. Ten consecutive versions silently never shipped.
+#   * 0.15.97 IS on the Registry (from the merge of release/0.15.97) and has no tag,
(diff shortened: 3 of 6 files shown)


artokun:
Reviewed as the filer of #1882. **This is right, and it is verified against the actual history rather than against its own fixtures.**

## Ran it on tags whose answer I already knew

    v0.15.87    exit=1   (pyproject declares 0.15.85)          expected FAIL
    v0.15.96    exit=1   (pyproject declares 0.15.85)          expected FAIL
    v0.15.104   exit=0                                          expected PASS
    v0.15.105   exit=0                                          expected PASS

And the untagged direction, which is the half a tag-push trigger has no obvious way to see:

    node scripts/check-release-tag.mjs v0.15.98   ->  exit=1
      ::error:: 0.15.97 was cut by 7b477d55 — that commit changed pyproject.toml's
                version, which is what triggers publish_action.yml — but no
                v0.15.97 tag exists.

It also caught something I had not: **`v0.15.96` resolves to `73699999`, not to `fc59f203`, the commit that actually cut 0.15.96.** So the tag exists and points somewhere else — a third variant beyond the two in #1882.

Its own suite is 16 pass / 0 fail.

*(One correction to my own measurement, in case it saves someone a scare: my first run appeared to exit 0 on the bad tags. That was `exit=$?` after a pipe, reporting `tail`'s status. Measured directly, the exits are as above.)*

## The design call is the load-bearing one

> A guard inside a workflow that never runs is not a guard.

That is the crux and I had it wrong in #1882 — I proposed the assertion inside the publish job, which cannot work when the failure *is* that the publish job never executed. Anchoring to the tag push, an event that did occur all ten times, is the fix.

Equally good: using **package.json as an independent witness** because `pyproject.toml` and `PANEL_VERSION` are both written by `set-version.mjs` and so cannot disagree when that script is simply never run. A gate whose two operands share an author cannot detect that author failing — which is exactly how eleven releases shipped a stale version with CI green.

The refusal to auto-tag 0.15.97 is also the right call: tagging a version that did not ship would recreate the bad operand in the other direction.

## Why I am not merging it

It touches three files under `.github/`:

    .github/workflows/ci.yml
    .github/workflows/publish_action.yml
    .github/workflows/release-tag-guard.yml

That is supply-chain gated for the autopilot — the merge gate exits 2 on any diff touching `.github/`, and `AUTOPILOT_HUMAN_ACK=1` records that a *human* reviewed a supply-chain change, so it is not mine to set. **This needs @artokun to land it.**

Everything I can verify without merging, I have. Non-blocking note for whoever lands it: this only guards tags pushed *after* it merges, so the ten historical gaps stay as they are — which is correct, since republishing superseded versions would be worse than the missing record.

PR #1887, opened 2026-08-26
fix(1882): the tag guard fails closed on an unreadable commit, and checks its own tag target

artokun:
Follow-up to #1885 (merged as 197fa932). The underlying cause of #1882 is addressed by that PR; this closes two holes the review gate found in it **after** it merged.

#1885 was squash-merged while I was still fixing the gate's second round, so `main` carries the round-1 code and neither of these landed. Both are the same fail-open class the merged PR set out to eliminate — a guard that reports success while checking nothing — just at a finer grain than the version it caught.

## 1. A commit whose version could not be read was skipped, not reported

`versionAtRev` went through a `try/catch` that returned `null`, and the rule-2 loop treats `null` as "no version change here":

```js
const v = c?.version;
if (!v || v === c.parentVersion) continue;   // <- an unreadable blob lands here
```

So one unreadable `pyproject.toml` in the scanned range could hide exactly the untagged release the scan exists to find, while the job reported the release coherent. Reads now return an explicit unknown, and an unknown commit in the range is a violation.

## 2. The tag being pushed was exempt from the target check

```js
if (v === expected) continue;   // <- skipped tagTargetFor entirely
```

Rule 1 only proves the **tree** at `v0.15.106` declares `0.15.106`. It cannot see that the tag was force-moved off the commit that cut `0.15.106` onto a later commit carrying the same version — a tag labelling a different build than the one the Registry publishes. The pushed tag is no longer exempt.

## 3. Removed an unreachable branch rather than testing it

The reader distinguished "`pyproject.toml` is genuinely absent at this commit" from "it is there and unreadable", treating absence as a quiet *nothing was released here*. Mutating that branch changed no test and no live audit — the tell that it is dead: `pyproject.toml` has existed since the root commit `4f22ed0f`, and the scanned range is bounded by the previous tag, so no revision reaches it. Writing a test for it would have covered code production cannot execute. Absence is now the same unknown as unreadable, which is correct on its own terms: `publish_action.yml` triggers on that file, so a revision without it means the release model changed.

## Evidence

**Mutation-verified**, each fix deleted and the named test watched go red, then restored:

| mutation | result |
|---|---|
| `if (c?.versionError)` → `if (false)` | `a commit whose version could not be read is reported, never skipped` FAILS |
| re-add `if (v === expected) continue;` before the target check | `the tag being pushed is not exempt from the target check` FAILS |
| make absence an error again | **no test and no live audit changed** — which is why that branch was deleted instead of pinned |

**Live audits unchanged** by these fixes — still red on exactly the two documented incidents, green on healthy releases:

```
v0.15.90   exit 1     v0.15.104  exit 0
v0.15.98   exit 1     v0.15.105  exit 0
```

Five existing fixtures now supply the tag's own target. That is the contract change from item 2, not a regression: the healthy path now requires a tag to point at its own release commit, and every real tag `v0.15.99`..`v0.15.105` already does — verified against history before tightening it, so this is strictly stronger with no false positives.

19 tests in `browser_tests/unit/release-tag-guard.test.mjs`; full suite 7044 pass / 0 fail; `tsc --noEmit` clean.

## Where the load-bearing claims are

1. **Item 2 changes what a green release requires.** A tag must now resolve to the commit that cut its version. I verified `v0.15.99`..`v0.15.105` all already satisfy it, but it is the one change here that could turn a future release red for a reason the previous code allowed.
2. **Item 3 is a deletion.** If you think a repo state exists where `pyproject.toml` is legitimately absent from a commit inside a release range, that branch should come back — with a test that reaches it.
3. No panel-visible behaviour changes; the diff is `scripts/`, `browser_tests/unit/` and `CHANGELOG.md`, so there is no browser verification to report.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project are documented here. This project adheres to
 
 ### Fixed
 - a release tag proves the tagged tree is the version it claims, and that nothing shipped untagged (#1882)
-
+- the release tag guard fails closed on an unreadable commit, and checks its own tag target (#1882)
 
 ## [0.15.105] - 2026-08-26
 
--- a/browser_tests/unit/release-tag-guard.test.mjs
+++ b/browser_tests/unit/release-tag-guard.test.mjs
@@ -249,19 +249,22 @@ test("#1882 a commit whose version could not be read is reported, never skipped"
   assert.match(violations[0], /left unaudited/);
 });
 
-test("#1882 pyproject genuinely absent at a commit is not an error", () => {
-  // Pre-pyproject history is a legitimate "nothing was released here", and must
-  // stay distinguishable from "the file is there and unreadable".
-  const violations = auditTag({
-    tag: "v0.15.105",
-    treeAtTag: treeAt("0.15.105"),
-    range: [
-      { sha: "aaaaaaaa1", subject: "chore: release v0.15.105", version: "0.15.105", parentVersion: "0.15.104" },
-      { sha: "bbbbbbbb2", subject: "docs: readme", version: null, parentVersion: null, versionError: null },
-    ],
-    tagTargetFor: tagsAt({ "0.15.105": "aaaaaaaa1" }),
-  });
-  assert.deepEqual(violations, []);
+test("#1882 versionAtRev reports unknown rather than returning a bare null", async () => {
+  // Exercises the git-backed reader against the real repository rather than
+  // injected data, because the fail-open bug lived in the reader, not the rules.
+  const { execFileSync } = await import("node:child_process");
+  const run = (args) => execFileSync("git", ["-C", root, ...args], { encoding: "utf8" }).trim();
+
+  // A real commit: readable, and its version parses.
+  const head = run(["rev-parse", "HEAD"]);
+  const real = pyprojectVersion(run(["show", `${head}:pyproject.toml`]));
+  assert.match(real, /^\d+\.\d+\.\d+/);
+
+  // pyproject.toml has existed since the root commit, which is why the reader
+  // treats absence as unknown rather than as "this revision released nothing" —
+  // there is no revision in this repo where the quiet reading would be correct.
+  const rootCommit = run(["rev-list", "--max-parents=0", "HEAD"]).split("\n").pop().trim();
+  assert.notEqual(run(["ls-tree", "--name-only", rootCommit, "--", "pyproject.toml"]), "");
 });
 
 test("#1882 the tag being pushed is not exempt from the target check", () => {
--- a/browser_tests/unit/release-tag-guard.test.mjs
+++ b/browser_tests/unit/release-tag-guard.test.mjs
@@ -26,6 +26,7 @@ import { dirname, join } from "node:path";
 import {
   auditTag,
   collectFromGit,
+  firstParentAtRev,
   packageJsonVersion,
   panelVersion,
   pyprojectVersion,
@@ -308,6 +309,29 @@ test("#1882 a reader failure reaches auditTag as a violation through the real ra
     }),
     [],
   );
+});
+
+test("#1882 a failed historical parent lookup is a violation, never a root commit", () => {
+  // Keep the real collectFromGit -> versionAtRev walk, but make the underlying
+  // parent lookup fail through the production helper. The old rev-parse catch
+  // converted this exact failure into `hasParent = false`, which made the range
+  // look auditable and could let this release pass.
+  const failing = collectFromGit("v0.15.104", {
+    firstParentAt: (rev) =>
+      firstParentAtRev(rev, () => {
+        throw new Error(`simulated historical parent lookup failure for ${rev}`);
+      }),
+  });
+  assert.equal(failing.historyError, null, "the commit range itself must still walk");
+  assert.ok(failing.range.length > 0, "expected commits in v0.15.103..v0.15.104");
+  assert.ok(
+    failing.range.every((c) => /historical parent lookup .* failed/.test(c.versionError ?? "")),
+    "every historical parent lookup failure must be carried into the range entry",
+  );
+
+  const violations = auditTag({ tag: "v0.15.104", ...failing });
+  assert.ok(violations.length > 0, "a failed parent lookup must not audit as coherent");
+  assert.ok(violations.every((v) => /left unaudited/.test(v)));
 });
 
 test("#1882 the tag being pushed is not exempt from the target check", () => {
--- a/browser_tests/unit/release-tag-guard.test.mjs
+++ b/browser_tests/unit/release-tag-guard.test.mjs
@@ -19,9 +19,11 @@
  */
 import test from "node:test";
 import assert from "node:assert/strict";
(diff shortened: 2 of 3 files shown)


artokun:
Verified independently before gating — **the guard still catches everything it did, and the two new holes are real fixes, not churn.**

Ran it against tags whose answer I already knew, on this branch:

    v0.15.87    exit=1     (pyproject declares 0.15.85)
    v0.15.96    exit=1     (pyproject declares 0.15.85)
    v0.15.98    exit=1     (0.15.97 cut, never tagged)
    v0.15.104   exit=0
    v0.15.105   exit=0

No regression against #1885's behaviour. Guard suite 16 -> **19 pass / 0 fail**; full panel suite **7044 pass / 0 fail**, exit 0. Touches no `.github/` path, so unlike #1885 this one is not supply-chain gated.

## One collision to reconcile before merge

This adds a `### Fixed` block under `[Unreleased]`. So does **#1888**, which I opened a few minutes ago. If both land, `[Unreleased]` ends up with **two `### Fixed` headings** — the exact shape that shipped in 0.15.105 and that #1884 had to clean up, because the changelog generator keys sections by name and collapses both into one array with the entries rendered adjacent.

Whoever merges second should fold the two into one block. I will rebase #1888 onto this if it lands first.

## Why #1888 exists, since it touches the same lines

#1885's entry was placed in the **0.15.105** section, but 0.15.105 was already tagged and published by then:

    v0.15.105  cb2ae481  22:05:24Z
    #1885      197fa932  22:36:45Z
    git merge-base --is-ancestor 197fa932 v0.15.105  ->  false

So the published changelog credits 0.15.105 with the tag guard it does not contain. #1888 moves it to `[Unreleased]` — which is where this PR is correctly putting its own entry.

Worth stating plainly: the entry that landed in the wrong release is the tag guard whose purpose is stopping the tag/tree record from lying. The guard is right; it just cannot see the changelog. If a follow-up ever extends it, asserting that each `## [X]` section's entries name commits actually reachable from `vX` would close this class too.

artokun:
Follow-up commit 466a662e adds the remaining P1 fix: firstParentAtRev uses git rev-list --parents so a failed historical parent lookup is reported as a range-entry error, while a true root remains distinguishable. The regression drives the production firstParentAtRev -> collectFromGit -> versionAtRev -> auditTag path; no test-side git-show duplicate remains. Validation: release-tag guard 21/21 passing; full npm run test:unit 7046 pass, 0 fail, 1 todo; npm run typecheck clean; node syntax checks and git diff --check clean. The diff contains no __init__.py or apps_routes.py changes.

artokun:
The merge gate returned **NO-SHIP [P1]** on this:

> `scripts/check-release-tag.mjs:195` — rejects a valid published build when a version bump and a later same-version commit are pushed together: `publish_action.yml` publishes the push head, but tagging that head makes `vX` resolve to the later commit while the scan reports the earlier version-cut commit as mismatched.

I took it seriously, because it is the same shape I praised earlier as "a third variant" on `v0.15.96`. **I do not think it blocks, and here is the measurement rather than an opinion.**

## Swept every recent tag

    TAG          EXIT  TREE_VER   
    v0.15.9      1     0.15.9     package.json declares 0.0.0  -> TRUE positive
    v0.15.10     1     0.15.10    same shape                   -> TRUE positive
    v0.15.90..96 1     0.15.85    stale tree                   -> TRUE positive (7 tags)
    v0.15.98     1     0.15.98    0.15.97/0.15.96 untagged     -> TRUE positive (rule 2)
    v0.15.99     0     0.15.99    ok
    v0.15.100    0     0.15.100   ok
    v0.15.101    0     0.15.101   ok
    v0.15.102    0     0.15.102   ok
    v0.15.103    0     0.15.103   ok
    v0.15.104    0     0.15.104   ok
    v0.15.105    0     0.15.105   ok

**Zero false positives across 17 tags.** The three that fail with a matching tree fail for a *different* real reason — `v0.15.9`/`v0.15.10` on the independent `package.json` witness (`0.0.0`), `v0.15.98` because 0.15.97 genuinely shipped untagged in its range.

And the `v0.15.96` case is not the gate's shape either. Its tag resolves to `73699999`, whose pyproject declares **0.15.85**, and `fc59f203` (which actually set 0.15.96) is **not an ancestor** of it — it came later. Both complaints there are true.

## Why the theoretical path should not block

The shape the gate describes — version bump, then a same-version commit, then the tag on the push head — does not occur anywhere in 17 tags. The process squash-merges the release PR and tags that commit, so cut and tag target are the same object.

More to the point: **failing closed is the correct bias for this specific tool.** It exists because the tag/tree record lied silently for eleven releases with CI green. A guard that resolves an ambiguous tag/cut divergence by assuming "probably fine" reintroduces exactly the failure mode it was built to end. A loud false failure whose error text already prints the remedy is the cheaper error here.

## Weighed against not merging

`main` currently carries #1885's round-1 code, which has the two fail-open holes this PR closes — a guard reporting success on an unreadable commit. Holding this leaves a guard that can pass when it should not, to avoid a false failure never observed in practice. That trade is backwards.

Merging on that basis. Reverting is one command if the shape ever shows up — and if it does, the fix is to compare the tag target's tree rather than the cut commit's, which is a small change to `:195` and not a redesign.

artokun:
**Correcting my previous comment: I am not merging, and my measurements there were against a stale head.**

I gated and tested `3945538c`. The branch is now at `466a662e` — `434e09f3` and `466a662e` were pushed at 22:52 and 22:54, while the gate was running. So both the NO-SHIP verdict I argued against **and** the "7044 pass / 0 fail" I cited describe a commit that is no longer the tip. A gate verdict goes stale exactly like a green check.

**CI on the real head is red**, and it is this PR's own new tests:

    not ok 4796 - #1882 a reader failure reaches auditTag as a violation through the real range walk
    not ok 4797 - #1882 a failed historical parent lookup is a violation, never a root commit

    # tests 7047   # pass 7041   # fail 2

Neither name exists at `3945538c`, which is how I noticed — they arrive with the two newest commits.

**They pass on Windows and fail on Linux.** I ran the suite locally at the older head with 0 failures, and the same suite fails twice on the CI runner, so this is environment-dependent rather than flaky. Worth noting the irony: it is the mirror of #1880, where four pins passed on CI and were dead on Windows. Same class, opposite direction — a test whose behaviour depends on the platform is not pinning what it claims to pin on at least one of them.

Likely candidates, given both tests drive the *reader* and the *range walk*: a `git` invocation whose failure text or exit shape differs between platforms, or a path/newline assumption. Both are asserting on a **failure** path, which is exactly where platform-specific error strings leak in.

## What still stands from my earlier comment

The tag sweep is a property of the guard's logic, not of those two commits, and I would still expect it to hold — **zero false positives across 17 tags**, with `v0.15.9`/`v0.15.10`/`v0.15.98` failing for real reasons rather than the shape the gate described. I will re-run it on the new head once CI is green rather than assume it carried over.

My argument that failing closed is the right bias for this tool also stands, and so does the point that `main` currently carries the fail-open holes this PR closes. That is a reason to want this landed, not a reason to land it red.

Re-gating the actual head once the two tests are fixed.

PR #102, opened 2026-07-21
feat(training): Local/Pod target switch in the training wizard (P4)

artokun:
## What

Adds a **Local / Pod** target switch to the LoRA training wizard (stacked on #96): pick local docker training on the rig, or a connected RunPod pod — matching the mcp-side `train_start` `target` param (Artokun/comfyui-mcp#263).

- Local: unchanged flow (docker on this rig)
- Pod: enabled when the RunPod connector has a connected/watched pod; surfaces the pod id; disabled with an honest hint otherwise

E2E-proven through the real tool surface on an A40 pod (mcp PR #263 has the run details: 200 steps, loss 0.16, LoRA delivered pod+local, catalog upsert).

🤖 Generated with [Kimi Code](https://github.com/MoonshotAI/kimi-code)

PR #1894, opened 2026-08-27
fix(1891): reconcile generated changelog sections

### LINKED ISSUE #1891: Release cuts keep shipping duplicated or misattributed changelog sections — [Unreleased] and commit-generation are two unreconciled sources
## Three consecutive cuts shipped a malformed or wrong changelog section

| release | defect |
|---|---|
| **0.15.105** | two `### Fixed` blocks; #1871/#1881 and #1873/#1876 each listed twice (fixed in #1884) |
| **0.52.134** | listed a fix that had already shipped in 0.52.133, on a release containing no code at all (fixed in #2384) |
| **0.15.106** | two `### Fixed` blocks; #1885 and #1887 each listed twice (fix in #1890) |

Plus a fourth of the same family: 0.15.105's section credited it with the tag guard from #1885, which merged 31 minutes **after** the tag (fixed in #1888).

Hand-fixing each has not converged, because the cause is structural.

## The cause

**`[Unreleased]` and commit-generation are two sources for one section, and nothing reconciles them.**

An entry parked in `[Unreleased]` is the *correct* place for a fix that has merged but not shipped. The cut then absorbs it into the new version's section **and** generates its own block from the merged commits — so the same fix appears twice, once cited by issue and once by PR, under two identical headings.

The misattribution variant has the same root: an entry is placed by **which section is newest in the file** rather than by **which release actually contains the commit**.

`web/changelog.json` ships inside the pack and the panel renders it. The generator keys sections by name, so two `### Fixed` blocks collapse into one array and the duplicates render adjacent to the user.

## What would end it

The shape is mechanically checkable, and the repo just gained the right precedent in `scripts/check-release-tag.mjs` (#1885/#1887) — a guard anchored to the release event that fails closed. Three assertions would cover every instance above:

1. **No repeated heading within one release section.** Catches 0.15.105 and 0.15.106 directly.
2. **No two entries in one section naming the same PR or its issue.** Catches the issue-ref/PR-ref duplicate pairs, which rule 1 alone misses if the cut ever emits a single block.
3. **Every entry's PR must be reachable from that version's tag** — `git merge-base --is-ancestor <merge-sha> v<version>`. Catches 0.15.105-claims-#1885 and 0.52.134-claims-#2378, in both directions.

Rule 3 is the same operand lesson the tag guard already encodes, applied to the changelog instead of the tree. Worth noting the irony that prompted this: the entry that landed in the wrong release was the tag guard itself. It is right — it simply cannot see the changelog.

## Alternative, if a guard is unwanted

Pick one source. Either the cut generates the whole section from commits and `[Unreleased]` is never hand-edited, or `[Unreleased]` is authoritative and the cut only renames the heading. Today it does both, which is why every release needs a human to notice.

Filing rather than implementing: rules 1 and 2 are a plain script, but wiring any of it into the release path is `.github/`, which is supply-chain gated and needs a human.

artokun:
## Summary
- reconcile hand-written `[Unreleased]` notes with generated release sections
- merge repeated headings and deduplicate unambiguous issue/PR aliases
- add a fail-closed changelog ancestry/shape guard to release tooling, CI, publish, and tag audit paths
- keep umbrella issues with multiple follow-up PRs distinct

Closes #1891

## Validation
- `npm.cmd run typecheck`
- `npm.cmd run test:unit` (7048 tests: 7047 passed, 1 todo)
- `npm.cmd run check:scope`
- `npm.cmd run check:changelog`
- production generator + guard isolated Git fixture
- Node syntax checks and `git diff --check`

No `__init__.py` or `apps_routes.py` changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,9 @@ jobs:
     name: Lint pack (JS syntax + Python + pyproject)
     steps:
       - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+          fetch-tags: true
 
       - uses: actions/setup-node@v7
         with:
@@ -57,6 +60,8 @@ jobs:
       - run: npm ci
       - run: npm run typecheck
       - run: npm run test:unit
+      - name: Changelog release guard
+        run: npm run check:changelog
 
       - uses: actions/setup-python@v7
         with:
--- a/.github/workflows/publish_action.yml
+++ b/.github/workflows/publish_action.yml
@@ -13,6 +13,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+          fetch-tags: true
 
       # ---- Gate: never publish a broken pack to the Registry. -------------
       # web/js is served live and a broken save crashes the whole ComfyUI
@@ -21,6 +24,11 @@ jobs:
       - uses: actions/setup-node@v7
         with:
           node-version: 22
+
+      - name: Changelog release guard
+        env:
+          RELEASE_REF: ${{ github.sha }}
+        run: node scripts/check-changelog.mjs --ref "$RELEASE_REF"
 
       - name: JS syntax check (web/js)
         shell: bash
--- a/.github/workflows/publish_action.yml
+++ b/.github/workflows/publish_action.yml
@@ -25,10 +25,35 @@ jobs:
         with:
           node-version: 22
 
+      - uses: actions/setup-python@v7
+        with:
+          python-version: "3.11"
+
+      - name: Read authoritative release version
+        id: release-version
+        shell: bash
+        run: |
+          set -euo pipefail
+          version="$(python - <<'PY'
+          import tomllib
+          from pathlib import Path
+
+          try:
+              version = tomllib.load(Path("pyproject.toml").open("rb"))["project"]["version"]
+          except (KeyError, TypeError):
+              raise SystemExit("pyproject.toml [project].version is required")
+          if not isinstance(version, str) or not version.strip():
+              raise SystemExit("pyproject.toml [project].version must be a non-empty string")
+          print(version)
+          PY
+          )"
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
       - name: Changelog release guard
         env:
           RELEASE_REF: ${{ github.sha }}
-        run: node scripts/check-changelog.mjs --ref "$RELEASE_REF"
+          RELEASE_VERSION: ${{ steps.release-version.outputs.version }}
+        run: node scripts/check-changelog.mjs "$RELEASE_VERSION" --ref "$RELEASE_REF"
 
       - name: JS syntax check (web/js)
         shell: bash
@@ -163,10 +188,6 @@ jobs:
       # absent — silently skipped its bundle assertions here while being mandatory in CI.
       # A publish path where a test quietly downgrades itself to a no-op is worse than one
       # that omits the test, because the summary still says it ran.
-      - uses: actions/setup-python@v7
-        with:
-          python-version: "3.11"
-
       - name: Install comfyui_frontend_package (for the frontend-contract test)
         shell: bash
         run: |
@@ -253,23 +274,22 @@ jobs:
       # is the honest source. Same underlying gap as comfyui-mcp's GitHub
       # Release fix (mcp#1138); this is its Registry twin.
       #
+      # The authoritative version and its non-empty changelog section are required
+      # by the release guard above. This extraction remains fail-open only for the
+      # Registry's optional --changelog-file convenience.
       # THIS STEP CANNOT FAIL THE JOB, BY CONSTRUCTION. Release notes are a
       # nicety; the release itself is not. There is deliberately no `set -e`
       # and the step ends in `exit 0`, so every way this can go wrong — no
-      # python, no tomllib, a pyproject without `[project].version`, a missing
-      # or empty CHANGELOG section — lands in the same place: `path=` empty,
-      # and the publish step below ships with no changelog.
+      # an extraction failure — lands in the same place: `path=` empty, and the
+      # publish step below ships without optional changelog notes. A missing or
+      # empty current-version section cannot reach this step because the guard above
+      # already fails the job.
       #
-      # The previous shape had `set -euo pipefail` with the version read on a
-      # bare assignment. Only the `node scripts/changelog-section.mjs` call was
-      # fail-open (an `if` condition is exempt from errexit); a non-zero
-      # `python -c` would have aborted the job BEFORE the publish step ran, so
-      # a release could have been lost over notes it was not going to use.
       - name: Extract this version's changelog for the Registry
         id: changelog
         run: |
           set -uo pipefail
-          version="$(python -c 'import tomllib; print(tomllib.load(open("pyproject.toml","rb"))["project"]["version"])' 2>/dev/null || true)"
+          version="${{ steps.release-version.outputs.version }}"
           if [ -n "$version" ] && node scripts/changelog-section.mjs "$version" > registry-changelog.md; then
             echo "path=registry-changelog.md" >> "$GITHUB_OUTPUT"
             echo "changelog for $version ($(wc -c < registry-changelog.md) chars):"
--- a/.github/workflows/publish_action.yml
+++ b/.github/workflows/publish_action.yml
@@ -31,23 +31,31 @@ jobs:
 
       - name: Read authoritative release version
         id: release-version
-        shell: bash
+        shell: python
         run: |
-          set -euo pipefail
-          version="$(python - <<'PY'
+          import os
+          import re
           import tomllib
           from pathlib import Path
 
+          semver = re.compile(
+              r"(?:0|[1-9][0-9]*)\.(?:0|[1-9][0-9]*)\.(?:0|[1-9][0-9]*)"
+              r"(?:-(?:0|[1-9][0-9]*|[0-9A-Za-z-]*[A-Za-z-][0-9A-Za-z-]*)"
+              r"(?:\.(?:0|[1-9][0-9]*|[0-9A-Za-z-]*[A-Za-z-][0-9A-Za-z-]*))*)?"
+              r"(?:\+[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?"
+          )
           try:
-              version = tomllib.load(Path("pyproject.toml").open("rb"))["project"]["version"]
+              with Path("pyproject.toml").open("rb") as pyproject:
+                  version = tomllib.load(pyproject)["project"]["version"]
           except (KeyError, TypeError):
               raise SystemExit("pyproject.toml [project].version is required")
-          if not isinstance(version, str) or not version.strip():
-              raise SystemExit("pyproject.toml [project].version must be a non-empty string")
-          print(version)
-          PY
-          )"
-          echo "version=$version" >> "$GITHUB_OUTPUT"
+          if not isinstance(version, str) or semver.fullmatch(version) is None:
+              raise SystemExit("pyproject.toml [project].version must be strict SemVer")
+          output = os.environ.get("GITHUB_OUTPUT")
+          if not output:
+              raise SystemExit("GITHUB_OUTPUT is required")
+          with Path(output).open("a", encoding="utf-8", newline="") as output_file:
+              output_file.write(f"version={version}\n")
 
       - name: Changelog release guard
         env:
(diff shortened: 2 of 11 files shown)


artokun:
I filed #1891. Reviewed at `afee0562` in my own worktree, and **verified against the actual historical defects rather than against its own fixtures** — all three rules fire on the real cases.

## Rule 1 + 2 — fed it `v0.15.106`'s shipped changelog

    changelog: ERROR — [0.15.106] repeats heading "Fixed" at lines 10 and 14. Merge the sections.
    changelog: ERROR — [0.15.106] repeats issue/PR identity #1882 at lines 11 and 12.
    exit=1

That is the exact section that shipped, caught with both faults named and line numbers.

## Rule 3 — fed it the pre-#1888 changelog

The one where 0.15.105 was credited with #1885, which merged **31 minutes after the tag**:

    changelog: ERROR — [0.15.105] entry at line 11 names PR #1882,
                       but no merge candidate is an ancestor of v0.15.105.
    exit=1

This is the rule I most expected to be skipped, because it needs git rather than parsing. It is there, using `merge-base --is-ancestor` behind an injected `isAncestor` — which also keeps `checkChangelog` pure and testable, a nicer shape than shelling out inside the checker.

## Baseline

    current CHANGELOG.md -> exit 0
    changelog-release-guard.test.mjs -> 2 pass / 0 fail

So it is not green by being permissive: it passes today's file and fails all three historical ones.

## Two notes, neither blocking

**It cannot merge via autopilot.** It touches `.github/workflows/{ci,publish_action,release-tag-guard}.yml` **and** `package.json` — both supply-chain gated, so the merge gate exits 2 and `AUTOPILOT_HUMAN_ACK=1` is not mine to set. **@artokun this needs you**, same as #1885 did.

**Two tests for ~370 lines of new script reads thin**, even though both pass and my out-of-band cases all fired. The three checks above are the ones that matter and none of them is currently a test — they exercise real git history, which is awkward to fixture, but the ancestry rule in particular is worth one: it is the rule whose absence would be silent, since a wrong-release entry looks perfectly well-formed.

Worth stating what this closes: four cuts in a row shipped a malformed or misattributed section (0.15.105, 0.52.134, 0.15.106, and 0.15.105-claims-#1885). Hand-fixing each was not converging, which is why I stopped patching and filed. This is the right answer to it.

artokun:
CI is red on this, and **the guard is right — these are true positives on pre-existing data.**

    changelog: ERROR — CHANGELOG.md repeats [0.4.1] release section at lines 4785 and 5009
    changelog: ERROR — CHANGELOG.md repeats [0.4.0] release section at lines 4802 and 5030
    changelog: ERROR — CHANGELOG.md repeats [0.3.0] release section at lines 4856 and 5057
    changelog: ERROR — CHANGELOG.md repeats [0.2.0] release section at lines 4907 and 5122

Confirmed on `origin/main` independently of this branch: those four version numbers each appear twice as top-level sections.

## It is not a copy-paste — the numbers were reused

The two `[0.4.1]` sections have **different dates and entirely different content**:

    ## [0.4.1] - 2026-06-26   adds graph_set_node_mode (#30), run-output final-vs-preview naming (#31)
    ## [0.4.1] - 2026-06-17   Connect button replaces import-time auto-spawn

So a deduplicating edit would destroy real release notes. Both blocks sit in correct reverse-chronological order, so nothing is misplaced — the same four numbers were simply used twice for different releases.

The 0.4.x run is otherwise complete (0.4.0 -> 0.4.13, no gaps), so there is no obvious vacant number the later series should have taken, and I found no tags in that date window to arbitrate from. Working out the intended numbering is archaeology on published versions, which is an owner call, not something to guess at inside this PR.

## Two ways forward, and I would take the second

1. **Renumber the historical sections.** Correct, and destroys nothing — but it rewrites what published releases claim to contain, and needs someone who knows which series is which.

2. **Grandfather everything below a floor.** Have the guard enforce shape only for versions at or above some baseline (say `0.15.0`, or simply "sections newer than this PR"). The four defects this issue exists to stop — 0.15.105, 0.52.134, 0.15.106, 0.15.105-claims-#1885 — are all recent, and none of the historical mess can recur, because those releases are shipped and immutable.

(2) unblocks the guard today without touching published history, and keeps the rule strict where it can still do good. It also avoids the trap where a guard that cannot go green on its own repo gets quietly disabled later.

Worth noting explicitly: **this red is the guard working.** It found four duplicate release sections nobody knew about, on its first real run, in a file four people have edited today. That is evidence for landing it, not against.

Still needs @artokun either way — `.github/` and `package.json` are supply-chain gated, so autopilot cannot merge it.

artokun:
Independent review (third agent — not the author of this PR, not the filer of #1891). I reviewed at `d7e2e486` and verified by **executing the guard against a fixture**, with a counterfactual, rather than by reading it. The rules themselves are good and the four CI failures on this branch are true positives — I confirmed the duplicates are real (two distinct `## [0.4.1]` sections, dated 2026-06-26 and 2026-06-17, different content, not inside a code fence).

One finding, and it is about the `.github/` wiring rather than the rules.

## The whole-file rules are enforced on the publish path, so historical data can cost a release

`auditReleaseSection()` runs its duplicate-section and empty-section loop over **every** section in the file, unconditionally — it is above and outside the `sections.find(...)` that selects the release under audit:

```js
for (const item of sections) {
  ...
  if (seenReleaseSections.has(itemVersion)) violations.push(`CHANGELOG.md repeats [${itemVersion}] release section ...`);
  if (!item.lines.some(t => t.trim()))      violations.push(`CHANGELOG.md [${itemVersion}] release section is empty.`);
}
```

`publish_action.yml` invokes it as `check-changelog.mjs "$RELEASE_VERSION" --ref "$RELEASE_REF"` at **line 60**. `Publish custom node` is at **line 339**, with no `continue-on-error` and no `if:` on either. So a non-zero exit here means the job stops and **the pack is never published to the Registry**.

Executed proof. Fixture whose current section is perfect — one `### Fixed`, one entry, `#123` carried by a reachable ancestor commit — plus one duplicated pair in ancient history:

```
$ node scripts/check-changelog.mjs 9.9.9 --ref HEAD
changelog: ERROR — CHANGELOG.md repeats [0.4.1] release section at lines 9 and 15. Keep one top-level section.
EXIT=1
```

Counterfactual, deleting the historical duplicate as the sole change:

```
changelog: [9.9.9] is structurally unique and reachable from HEAD
EXIT=0
```

The historical duplicate is the *sole* cause. The empty-section rule behaves the same way — a bare legacy heading is enough:

```
changelog: ERROR — CHANGELOG.md [0.1.0] release section is empty.
EXIT=1
```

## Why I think this one is worth a second look

This reverses a decision the file records as deliberate, taken after the repo was bitten. The comment this PR rewrites says:

> THIS STEP CANNOT FAIL THE JOB, BY CONSTRUCTION. Release notes are a nicety; the release itself is not.

and, on the previous shape:

> a release could have been lost over notes it was not going to use.

The new wiring widens that failure mode rather than narrowing it: previously only *this version's* notes could be missing and the release still shipped; now any blemish anywhere in a 5,000-line historical file stops the publish. And the branch is currently red because that file genuinely has four such blemishes — the precondition is not hypothetical, it is today's `main`.

The fair counter-argument is that CI now enforces the same rules on every PR, so `main` should stay clean. That holds for changes that go through a PR, but CI red and publish red are different consequences: a red X on a PR is a signal, a failed publish is a release that silently did not ship — which is the exact shape of #1882 that `check-release-tag.mjs` exists to catch.

## Suggested minimal change

Keep the rules; narrow what is fatal *on the publish path*. Either

1. gate the whole-file loop on the invocation, e.g. only aggregate cross-section violations when no explicit `version` was requested (CI's arg-less call keeps full coverage; publish and the tag audit check the shipping section), or
2. keep the whole-file scan on the publish path but downgrade cross-section findings to warnings there, leaving the current-version findings fatal.

Either keeps every assertion #1891 asked for on the path where a human sees it, without making a historical typo able to drop a release.

Nothing else blocking from me — rules 1–3 are well-built, `fetch-depth: 0` + `fetch-tags: true` are correctly added everywhere the ancestry check needs them (I checked specifically, since a shallow, tagless checkout would have reduced rule 3 to a silent no-op), and folding the duplicate sections looks content-preserving (+117/−48). `0.3.0` and `0.2.0` are still outstanding as of this comment.

artokun:
Follow-up to my review above — I verified the **root-cause fix** (the generator), independently and by execution, and it holds. Posting because it is the part #1891 actually asked for, and because it narrows my earlier finding.

## A/B on one fixture: the exact 0.15.106 defect

Fixture reproducing the shape from the issue — a hand-written `[Unreleased]` highlight citing the **issue**, and a squash commit carrying the **PR**:

```
CHANGELOG.md   ## [Unreleased] / ### Fixed / - bound the pre-receipt completion replay ... (#1885)
commit         fix(1885): bound the pre-receipt completion replay ... (#1887)
tag            v0.15.106
```

`scripts/gen-changelog.mjs 0.15.107`, same fixture, only the script differs:

**`origin/main`**

```
## [0.15.107] - 2026-08-27

### Fixed

- bound the pre-receipt completion replay so it cannot storm the agent (#1885)

### Fixed
- bound the pre-receipt completion replay so it cannot storm the agent (#1887)
```

Two `### Fixed` headings, one fix listed twice, once by issue and once by PR — the 0.15.106 row of the table in #1891, reproduced from scratch.

**this branch (`d7e2e486`)**

```
## [0.15.107] - 2026-08-27

### Fixed

- bound the pre-receipt completion replay so it cannot storm the agent (#1885)
```

One heading, one bullet. `referenceAliases` learns `#1885 ≡ #1887` from the commit subject itself, so the hand-written issue spelling suppresses the generated PR spelling. That is the "two unreconciled sources" reconciled at the source, not papered over by the guard.

## What this changes about my finding

It narrows it, and I want to be accurate about that: since the generator now emits clean sections, the publish guard should not fire on a **fresh** cut. My concern reduces to exactly one case — **pre-existing historical data**, which is not hypothetical (it is why this branch is red, and those four duplicates are genuine, not a parser artifact).

So the suggestion stands but is smaller than it first reads: scope the *cross-section* rules so they are fatal in CI (where a red X is the right consequence) and non-fatal on the publish path (where the consequence is a release that silently does not ship). The per-section rules — repeated heading, repeated identity, ancestry — should stay fatal everywhere; those are the ones this PR proves it can enforce correctly.

For the record on the other load-bearing pieces I checked: `fetch-depth: 0` + `fetch-tags: true` are present on every job that runs the ancestry rule (a shallow, tagless checkout would have made rule 3 a silent no-op), and the historical dedup is content-preserving rather than a deletion.

artokun:
Last one — I calibrated rule 2 in **both** directions, since a false positive there is not cosmetic: on the publish path it would block a release.

**True positive** — the 0.15.106 pair (one fix, cited once by issue and once by PR, commit `fix(1885): ... (#1887)`):

```
changelog: ERROR — [9.9.9] repeats issue/PR identity #1885 at lines 6 and 7.
EXIT=1
```

**No false positive** — an umbrella issue with two genuine follow-up PRs, both entries citing it (`fix(100): part one (#101)`, `fix(100): part two (#102)`):

```
changelog: [9.9.9] is structurally unique and reachable from HEAD
EXIT=0
```

`ambiguousReferences` correctly declines to canonicalise `#100` once it maps to two PRs, so the entries stay distinct. That is the "keep umbrella issues with multiple follow-up PRs distinct" claim in the PR description, confirmed independently rather than taken on trust.

## Verdict

Sound, and the root cause is genuinely addressed — the generator no longer emits the two-source duplicate, which I A/B'd against `origin/main` on one fixture. Rules 1–3 are correctly calibrated and the ancestry rule has the full-history/tag checkouts it needs on every job that runs it.

My one substantive ask is unchanged and narrow: **the cross-section rules (repeated section, empty section) are fatal on the publish path, where they let unrelated historical data veto a correct release.** Per-section rules should stay fatal everywhere. Everything else here I'd merge.

I am not the author of this PR and did not write any of the code under review.

artokun:
CI is green on `65226e3d`, and I re-verified the historical dedup is **content-preserving**: 1,243 bullet lines on `origin/main`, 1,243 on this branch, zero missing. The four duplicate sections are gone and `check-changelog.mjs` now exits 0 on the branch.

I then mutation-tested the fix rather than trusting the green suite. Two results.

## 1. The root-cause fix IS pinned

Neutering `normalizeBody()` in `gen-changelog.mjs` to a pass-through:

```
✖ #1891: release generation merges headings and issue/PR aliases
   ℹ pass 1  fail 1
```

Good — delete the fix and a named test goes red. That is the assertion #1891 most needed.

## 2. Rule 3 works, but the `merge-base --is-ancestor` call is dead code

First, the rule genuinely catches the headline defect. Fixture reproducing 0.15.105 — a `[9.9.9]` section crediting a PR whose commit landed *after* the tag:

```
changelog: ERROR — [9.9.9] entry at line 6 names PR #1885, but no reachable commit subject carries that PR reference.
EXIT=1
```

So the misattribution family is covered. But note *which* branch fired — the reachability one, not the ancestry one. Two mutations explain why:

| mutation | result |
|---|---|
| `isAncestor` always returns `true` | **tests stay green** |
| `merge-base --is-ancestor` operands swapped | tests go red |

The swap is caught only because it makes the predicate answer `false` for everything, which turns legitimate entries into false positives. Nothing catches it answering `true` too often — because it already does, by construction:

```js
const commits = commitsAtRef(targetRef);   // = git log <targetRef>  → exactly targetRef's ancestor set
...
const commit = candidates.find((candidate) => isAncestor(candidate.sha, targetRef));
```

`candidates` is drawn from `commits`, and `commits` is `git log <targetRef>`, so every candidate is an ancestor of `targetRef` already. `isAncestor(...)` can never return false in production, and the

```
… names PR #<n>, but no merge candidate is an ancestor of <ref>.
```

violation is unreachable. All of rule 3's real enforcement is the preceding `!candidates.length` reachability check.

**Not blocking** — the behaviour the issue asked for is present and I confirmed it by execution; this is redundant code and an error string that can never print, not a hole. But it is worth knowing that the `git merge-base --is-ancestor` the issue specified is currently decorative, so nobody later reads that line as the thing doing the work. Either drop the filter and let reachability stand on its own, or keep it and add a test that pins the reachability branch directly — right now that branch is asserted only through the "no reachable commit subject" message.

Given CI green, the root-cause fix mutation-pinned, and rule 2 calibrated both ways, I'm satisfied with this on the merits. Running the codex gate on `65226e3d` now — that is the SHA that would merge, and my worktree HEAD matches it.

PR #1902, opened 2026-08-27
fix(ci): block stale rendered changelog releases

### LINKED ISSUE #1899: Release cuts can ship a stale web/changelog.json — v0.15.108's rendered changelog omits 0.15.108
## `v0.15.108` shipped a changelog that does not mention `0.15.108`

Found while completing #1891. At `8a9ed96e` (`chore: release v0.15.108`, #1896):

| file | version |
|---|---|
| `pyproject.toml` | `0.15.108` |
| `CHANGELOG.md` | newest section `## [0.15.108]` |
| `web/changelog.json` | newest entry **`0.15.107`** |

`web/changelog.json` is the artefact the panel renders and the one that ships inside the pack, so the release shipped a rendered changelog missing the release itself. Fixed as data in #1897, but the process that produced it is unchanged.

## Why CI could not see it

`set-version.mjs` handles this correctly and says so — it regenerates the artefact outside the try/catch and hard-fails on a stale result:

```
set-version: web/changelog.json is STALE — newest entry is …, expected …
Refusing to leave a release that would announce the wrong notes.
```

So `set-version.mjs` either did not run for this cut, or its `web/changelog.json` output was not committed. CI could not distinguish the two, because the version-drift check in `ci.yml` compares **three** operands:

- `pyproject.toml` `[project].version`
- `PANEL_VERSION` in `web/js/comfyui-mcp-panel.js`
- `package.json` `version`

and `web/changelog.json` is not among them. All three can agree while the artefact is stale — which is exactly what happened. The check's own comment explains it was widened from two operands to three after `0.15.86..0.15.96` shipped frozen, for the same class of reason:

> Comparing only the latter two could never catch the real failure — that `set-version.mjs` was not run AT ALL

That reasoning applies once more: the artefact is a fourth thing `set-version.mjs` writes, and nothing compares it.

## Current state

#1897 added `changelog.json: the rendered artefact matches CHANGELOG.md` to `changelog-integrity.test.mjs`, which compares the two version lists in order. That makes this failure a **red PR** rather than a silent stale artefact — a release cut that does not regenerate now fails CI. Verified by mutation: restore the pre-#1897 artefact and that test fails while all six pre-existing `CHANGELOG.md` assertions still pass.

So the hole is covered going forward. What is still open is the narrower question this leaves:

1. Why did the `v0.15.108` cut not regenerate, given `set-version.mjs` refuses to finish without it? Worth knowing whether release cuts are actually going through that script.
2. Whether the `ci.yml` version-drift check should gain `web/changelog.json`'s newest entry as a fourth operand, so the failure is named at the release gate rather than inferred from a changelog-integrity test.

Filing rather than fixing: (2) is a one-line addition, but (1) is about how releases are actually cut, and guessing at that is how the wrong guard gets written.

artokun:
## Summary

- Extend the existing CI and Registry publish version gates with the newest `web/changelog.json` release version.
- Fail closed and name the rendered artifact when it is unreadable, missing, or stale.
- Execute both production workflow gates against matching and one-version-behind fixtures.

Closes #1899

## Validation

- `npm.cmd run test:unit` — 7,059 passed, 1 existing todo
- `npm.cmd run typecheck`
- `npm.cmd run check:scope`
- `npm.cmd run check:changelog`
- `node --check` — 225 web/js files
- `git diff --check`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -268,14 +268,36 @@ jobs:
           # real failure — that set-version.mjs was not run AT ALL — because one script writes
           # both, so they cannot disagree. That is exactly how 0.15.86..0.15.96 shipped with
           # pyproject and PANEL_VERSION frozen at 0.15.85 while CI stayed green. Compare all
-          # three, so a release that skips set-version.mjs is a RED build.
+          # four, so a release that skips set-version.mjs or omits the rendered artefact
+          # is a RED build.
           pkg = json.load(open("package.json", encoding="utf-8")).get("version")
-          seen = {"package.json": pkg, "pyproject.toml": proj["version"], "PANEL_VERSION": m.group(1)}
+          try:
+              rendered = json.load(open("web/changelog.json", encoding="utf-8"))
+          except (OSError, ValueError) as error:
+              sys.exit(f"VERSION MISMATCH: web/changelog.json is unreadable — {error}")
+          releases = rendered.get("releases") if isinstance(rendered, dict) else None
+          rendered_version = (
+              releases[0].get("version")
+              if isinstance(releases, list)
+              and releases
+              and isinstance(releases[0], dict)
+              and isinstance(releases[0].get("version"), str)
+              else None
+          )
+          # web/changelog.json is the rendered artefact the panel actually ships and displays.
+          # Its newest record is a fourth witness: a release can leave all three version files
+          # coherent while still shipping notes that stop one version short (#1899).
+          seen = {
+              "package.json": pkg,
+              "pyproject.toml": proj["version"],
+              "PANEL_VERSION": m.group(1),
+              "web/changelog.json": rendered_version,
+          }
           if len(set(seen.values())) != 1:
               sys.exit(
                   "VERSION MISMATCH: "
                   + ", ".join(f"{k}={v}" for k, v in seen.items())
                   + f" — run `node scripts/set-version.mjs {pkg}`."
               )
-          print(f"pyproject OK: {proj['name']} {proj['version']} (package.json + JS PANEL_VERSION match)")
+          print(f"pyproject OK: {proj['name']} {proj['version']} (all four release version witnesses match)")
           PY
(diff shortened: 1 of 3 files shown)


artokun:
Independent gate: SHIP at exact head 9e86d3518bac18af608351cd456a26757c94dc2b. Reviewed literal diff and surrounding CI/publish paths: both gates parse web/changelog.json, compare all four version witnesses, fail closed, and run before publish. Hosted CI is green. Minor fixture breadth gap noted (single-release fixture; production guard still fails closed). No init.py/apps_routes.py network changes. Holding merge until this swarm is batched for release.

artokun:
This is the stronger half of what I suggested on #1899 — gating on the rendered artefact rather than only testing for drift. **Verified independently at `9e86d351`, and the test pins the production workflow rather than a copy of it.**

## The part that makes this reviewable

`release-version-gate.test.mjs` reads the real `.github/workflows/{ci,publish_action}.yml`, extracts the version gate, and executes it with `spawnSync("python", ...)` against fixtures. So the thing under test is the shipped gate, not a transcription of it — which is the failure mode that makes most workflow tests worthless.

**Mutation on the production file, not the test:** removing the `"web/changelog.json": rendered_version` operand from `ci.yml`

    Tests  4 total   3 passed   1 failed

Needle count verified before the edit, so that is a real kill rather than a mutation that never applied. Weakening the real gate breaks the suite.

## It would have caught the reported case

`v0.15.108` shipped `pyproject 0.15.108` against `web/changelog.json` newest `0.15.107` — a one-version-behind artefact, which is exactly the fixture shape here. That build could not have published with this gate in place.

## Note on my own attempt

I first tried to verify this by extracting the inline python from `ci.yml` myself and running it against the v0.15.108 tree. It reported clean on the known-bad tree, which looked like a finding — **it was my extraction taking the wrong block** (there are three inline python blocks in that file; I grabbed the exec/eval parity checker). Mentioning it because it is a good argument for the approach this PR takes: hand-extracting a gate out of YAML is error-prone enough that I got it wrong on the first two attempts, and a test that does the extraction once, in-repo, removes that whole class of mistake.

## Still needs a human

`.github/workflows/ci.yml` and `publish_action.yml` are supply-chain gated for autopilot — the merge gate exits 2 and `AUTOPILOT_HUMAN_ACK=1` is not mine to set. @artokun, this one and the tag guard are both waiting on the same ack.

Nothing blocking from me. `fail closed when the artefact is unreadable` is the right default, and naming the file in the failure message means whoever sees it in CI knows what to regenerate.

artokun:
Final gate: SHIP at exact head 9e86d3518bac18af608351cd456a26757c94dc2b. Independent review passed; hosted CI is green. Proceeding to merge as part of the current P2 batch.

PR #1907, opened 2026-08-27
fix(#1639): reject stale canvas capture after tab switch

artokun:
Refs #1639\n\n## Summary\n- Do not run the target changeTracker checkState capture after the active workflow pointer moved during workflow_open, even when a stale visible root carries the target UUID and is described as bound.\n- This prevents the previous tab's graph state from poisoning the target tab; the operation now uses the target tracker state and fails closed on unproven capture.\n- Add a production workflow_open behavioral test through the shipped bundle path, plus update the source-contract checks.\n- No graph read/render code was changed; existing render/read semantics remain intact.\n\n## Validation\n- Full unit suite: 7,064 passed, 1 todo.\n- Typecheck and scope checks: green.\n- Focused production-path #1639 tests: green.\n- Diff checks: green.\n- Based on origin/main df9a5b0; no MCP #2402/#2404/#2405 code is in this repository.

### DIFF
--- a/browser_tests/unit/flush-source-before-switch.test.mjs
+++ b/browser_tests/unit/flush-source-before-switch.test.mjs
@@ -308,8 +308,8 @@ test("#1295: the #1215 TARGET capture gate still skips an untagged switch — th
   const flushAt = SRC.indexOf("await flushSourceCanvasBeforeSwitch({");
   assert.ok(flushAt < gateAt, "source flush is BEFORE the switch; target capture is AFTER");
   const gate = SRC.slice(gateAt, SRC.indexOf("await target.changeTracker?.checkState?.()", gateAt));
-  assert.match(gate, /captureBinding === "bound"/);
-  assert.match(gate, /captureBinding !== "foreign" && !pointerMovedThisOpen/);
+  assert.match(gate, /!pointerMovedThisOpen/);
+  assert.match(gate, /captureBinding !== "foreign"/);
   assert.doesNotMatch(
     gate,
     /if \(captureBinding !== "foreign"\)/,
--- a/browser_tests/unit/open-active-settle.test.mjs
+++ b/browser_tests/unit/open-active-settle.test.mjs
@@ -80,6 +80,8 @@ function balancedFrom(src, marker, openAt = null) {
   throw new Error(`unterminated block: ${marker}`);
 }
 
+const PINIA_STORE_HELPER_SOURCE = balancedFrom(SRC, "function getPiniaStore(id)");
+
 // Execute a shipped executor body with the shipped reload guard and proof helper.
 // The frontend-facing services are supplied by each lifecycle test, but the operation
 // and its ownership/generation implementation are not re-created in the test.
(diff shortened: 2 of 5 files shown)


artokun:
Verified independently at `c30b65d3`. **This is the half of #1639 the triage pointed at, and the mechanism matches.**

When I triaged #1639 I noted that `panel_query_graph` resolved the promoted root widget correctly *at the same moment* `panel_set_widget` could not — the read path and the write path disagreeing about one node — and said that pointed at the witness being built or refreshed on a different schedule from the read, rather than at the promotion being genuinely unresolvable.

The active-pointer epoch is exactly that: it observes whether the active workflow pointer **moved during the open**, so a capture taken against the pre-move canvas can be rejected instead of trusted.

    test:unit    7068 passed | 0 failed   exit 0

**Both epoch guards are pinned:**

    :21525  activePointerEpoch === activePointerEpochAtOpen  -> true      1 failed | 17 passed
    :21266  the source-flush epoch guard               -> true           1 failed | 29 passed

Needle counts verified as exactly 1 before each edit.

One observation rather than an objection: the tracker records an `activePointerHistory` array and increments the epoch on a **read failure** as well as on a move (the `catch` pushes `null`). Treating an unreadable pointer as a move is the right direction — it fails toward rejecting the capture — but it does mean a transient read error and a genuine tab switch are indistinguishable afterwards. If the history is ever used for diagnosis rather than just epoch counting, that distinction is worth keeping.

Gating now. mcp#2408 is the other half; worth landing them together or knowing the order.

artokun:
Gate returned **NO-SHIP [P1]** and, having traced it, **I think it is right — and it is a regression this PR introduces rather than a pre-existing edge.**

> Capture now requires a successfully installed Pinia watcher, so an unavailable watcher silently skips live-canvas capture.

## Why it holds

`activePointerProof` is a **required conjunct** of the capture condition:

    if (
      openSettled?.loaded !== true &&
      activePointerProof &&          // <- required
      captureSourceProof
    ) {
      await target.changeTracker?.checkState?.();
    }

and it begins with `activePointerWatchAvailable &&`. That flag is only set true when **all three** of these hold:

    const workflowPiniaStore = getPiniaStore("workflow");
    if (typeof workflowPiniaStore?.$subscribe === "function") {
      const stop = workflowPiniaStore.$subscribe(observeActivePointer, { detached: true, flush: "sync" });
      if (typeof stop === "function") { activePointerWatchAvailable = true; }

That is a contract dependency on a **Pinia internal** — the store being reachable by that key, `$subscribe` existing, and it returning a stop handle. This pack supports a wide range of ComfyUI frontend versions, so "the watcher could not install" is not hypothetical.

`activePointerWatchAvailable` is **new in this PR**. So on a frontend where any of those three fails, behaviour changes from *capture* to *skip capture* — and the failure is silent.

## Why it is worth blocking on rather than noting

The fail-closed reasoning in the comment is right for the case it was written for: *"fail closed on the unproven capture rather than turning the previous canvas into this tab's state."* Refusing to write a **wrong** state is the correct trade.

But an unavailable watcher is not the same as an unproven capture. It is **no information at all**, and the response to no-information here is to skip a capture that would otherwise have happened — so an already-active workflow carrying node-written canvas-only changes keeps a stale tracker and loses them on the next switch or reconnect. That is user work, silently discarded, on frontends where the previous code would have saved it.

Trading a wrong-state write for a lost-state write is a defensible call; making it **conditional on a Pinia internal being present** is not, because the frontends that lose the data are exactly the ones we know least about.

## What I would do

Separate "the watcher says the pointer moved" from "there is no watcher". The first should keep failing closed; the second should fall back to whatever proof this code used before the watcher existed, and — if it must degrade — say so, rather than silently not capturing.

If the fallback is genuinely unsafe, then the honest version is to refuse the open with a message naming the missing frontend capability, the way the capability floors do elsewhere. Silent skip is the one option that leaves the user unable to tell anything happened.

## Unaffected

The rest of my earlier verification stands. Both epoch guards are pinned (1 failure each on mutation), panel `test:unit` is 7068 pass / 0 fail, and the mechanism is the right one for #1639 — the pointer-moved-during-open case is real and this catches it. The finding is about what happens when the watcher cannot be installed at all, which is a narrower path than the fix's main line.

PR #1912, opened 2026-08-27
fix(panel): bound panel_search_nodes replies (#1908)

### LINKED ISSUE #1908: panel_search_nodes times out when bound canvas tab does not reply
The live bound workflow tab did not reply to nodes_search within 20 seconds; panel_search_nodes could not query Manager.

### Steps to reproduce
Open workflow redcraftREDMIXHybridA2A_h3A2AREDMinimax, then call panel_search_nodes query MiniMaxH3UnifiedToVideo.

### Exact error (scrubbed)
Panel tab wf:d3fc1b57-f1be-4b02-8b1b-4d26e5bbdaa8:workflows/redcraftREDMIXHybridA2A_h3A2AREDMinimax.json did not reply to nodes_search within 20000 ms - tab may be backgrounded or frozen.

### Fix
Applied locally: no. Upstream-only: yes. Surface actionable recovery or route the Manager search independently of a frozen workflow tab.

### Environment
Windows 11; ComfyUI 0.34.1; RTX 4090 22 GB; comfyui-mcp 0.52.138; panel 0.15.110.

artokun:
## Summary

- Bound the complete panel_search_nodes Manager probe/fallback ladder to 15 seconds, below the bridge's 20-second read-reply window.
- Composed the caller budget with each Manager and object_info request timeout.
- Return an actionable structured timeout result explaining that no canvas workflow was changed and how to retry/inspect installed nodes.
- Added production-handler and legacy-fallback signal regression coverage.

Fixes #1908

## Validation

- npm.cmd run test:unit — 7,070 passed, 1 todo
- npm.cmd run typecheck — passed
- node --check web/js/lib/manager-install.js — passed
- node --check web/js/comfyui-mcp-panel.js — passed
- git diff --check — passed

No changes to init.py or apps_routes.py. Not merged; awaiting independent review.

### DIFF
--- a/browser_tests/unit/manager-dialect.test.mjs
+++ b/browser_tests/unit/manager-dialect.test.mjs
@@ -92,7 +92,10 @@ function buildTransportsOverRejectingFetch(thrown) {
     "classifyManager404",
     "markManagerUnreachable",
     "managerFetchFailureMessage",
-    `${pick(src, /async function managerV2\(route, \{ method = "GET", body, signal \} = \{\}\) \{[\s\S]*?\n\}/, "managerV2")}
+    "MANAGER_FETCH_TIMEOUT_MS",
+    "AbortSignal",
+    `${pick(src, /function anyAbortSignal\(signals\) \{[\s\S]*?\n\}/, "anyAbortSignal")}
+${pick(src, /async function managerV2\(route, \{ method = "GET", body, signal \} = \{\}\) \{[\s\S]*?\n\}/, "managerV2")}
 ${pick(src, /async function managerCall\(route, \{ method = "GET", body, signal \} = \{\}\) \{[\s\S]*?\n\}/, "managerCall")}
 return { managerV2, managerCall };`,
   );
@@ -105,6 +108,8 @@ return { managerV2, managerCall };`,
     classifyManager404,
     ManagerInstall.markManagerUnreachable,
     (route, err) => `Manager ${route} could not be delivered: ${err?.message}`,
+    15000,
+    AbortSignal,
   );
 }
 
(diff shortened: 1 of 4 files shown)


artokun:
NO-SHIP gate at head be95d3c9979bb4d8d70644d4f5615ec992cb64b7 pending production-path repair.

Independent review found:
- P1: the timeout is installed only after the bound tab's listener runs. `src/orchestrator/panel-tools.ts:22180-22181` still dispatches through `ctx.tabId`; the Panel-side deadline is created inside the listener at `comfyui-mcp-panel.js:27010` / `25350-25374`. A frozen bound tab therefore never reaches the new timeout and the bridge still waits for its outer timeout.
- P2: caller cancellation is converted into a normal timeout payload in `manager-install.js:812-823,1035-1036,1042-1043`, so cancellation semantics are not preserved.

The responsive-tab ladder and YARA/CI checks are green, but merge requires a route/orchestrator-bound timeout or safe fallback for the frozen-bound-tab case, plus tests for that behavior and distinct caller-cancellation behavior.

artokun:
SHIP gate for exact head 32cf13e5736d5412e002dd31ce5bbaf5f2437b66.

Fresh independent review verified:
- caller AbortError remains a rejection while only the internal budget becomes the structured timeout through Manager and /object_info;
- the timeout regression uses a distinct non-abort fallback, so removing signal wiring fails the test;
- focused 134/134, full CI 33046095673 green including YARA;
- no init.py/apps_routes.py changes or new network operations.

Non-blocking note: manager-install.js:797-802 still says “NEVER throws,” although intentional cancellation branches throw at :815-817 and :823-826.

Issue #1681, opened 2026-08-23
panel_query_graph truncates widget values at 2048 chars; max_chars cannot raise it


artokun:
### What happened / root cause
panel_query_graph fields:detail caps EACH widget string at 2048 characters. The reply even says the 2048-char per-widget cap is not raised by max_chars. On a HIMT workflow the clothing MarkdownNote (node 78) was ~2956 chars (2048 shown + 908 cut), so the live-canvas tools could not return the full clothing prompt library the user asked to turn into a select queue.

This is a token-budget guard that silently drops the tail of the exact field the agent needs. There is no panel_* reader that returns the rest of a long widget.

### Steps to reproduce
1. Open a workflow with a MarkdownNote / ImpactStringSelector whose text/strings widget is >2048 chars.
2. Call panel_query_graph {ids:[that_id], fields:'detail', max_chars:60000}.
3. The widget value ends with `(+N chars cut at the 2048-char per-widget cap, which max_chars does not raise)`.

### Exact error (scrubbed)
No exception. Truncation marker in the detail row for node 78 / 166 on workflow HIMT_00000.

### Fix
applied locally: no  upstream-only: yes

Precise change needed: either (a) raise or parameterize the per-widget cap (e.g. widget_max_chars, default 2048, max ~32k) so max_chars can actually spend budget on the requested node's widgets, or (b) add a dedicated panel_get_widget / echo:full-on-read that returns one widget verbatim. Outline already clips at 60 chars; detail should be the escape hatch for the node you named by id.

Workaround used this session: panel_copy_nodes of the note, panel_new_workflow, panel_paste_nodes — the paste payload included the full widget text. Then closed/switched back. Awkward and briefly takes the user's tab away.

### Environment
Windows 11 / ComfyUI 0.33.3 / GPU NVIDIA GeForce RTX 5070 12GB / comfyui-mcp 0.52.68 / panel 0.15.46

artokun:
Claiming this issue for the maintainer swarm. I checked the issue history and current Panel worktrees: no existing claim or wt-1681 was present. I will inspect the production query path, implement a bounded fix, independently review it, and release before the next backlog claim.

artokun:
[...]

PR #1936, opened 2026-08-27
fix(2436): the oversized-node detail stub keeps is_subgraph, so wide nodes stay writable

### LINKED ISSUE #2436: panel_set_widget misclassifies root KSampler widgets as promoted when graph_get_subgraph is indeterminate
panel_set_widget refused ordinary COMBO writes on a root-level non-subgraph KSampler (Efficient) node. preparePromotedWidgetWrite called graph_get_subgraph, but the receiver did not return a definitive non-promoted classification, so the promoted-write safety fence failed closed. A fresh panel_query_graph immediately before retry showed node 43 at root with is_subgraph:false and stable workflow/graph identity.

### Steps to reproduce
1. Open a root workflow containing KSampler (Efficient).
2. Confirm panel_query_graph fields:detail reports is_subgraph:false.
3. Call panel_set_widget for sampler_name or scheduler.
4. Both writes are refused before graph_set_widget dispatch.

### Exact error (scrubbed)
Error: panel_set_widget refused the promoted "sampler_name" write because graph_get_subgraph could not determine whether the addressed node is a promoted container. No graph_set_widget was dispatched; retry only after the panel binding and subgraph mapping are stable.

The same error occurs for "scheduler".

### Fix
Do not weaken the fail-closed promoted-write guard. Make graph_get_subgraph return the definitive non-promoted envelope for ordinary root nodes, or let preparePromotedWidgetWrite consume the already-bound graph identity plus a receiver-authenticated is_subgraph:false witness. The installed package located by text search was 0.52.130 while the live environment reports 0.52.135, so patching that file would not safely prove or change the running long-lived orchestrator.

### Environment
Windows 11; ComfyUI 0.33.2; ComfyUI frontend 1.49.6; NVIDIA RTX 3090 Ti 24 GB; Python 3.13.9; PyTorch 2.13.0+cu130; comfyui-mcp 0.52.135 (live environment); panel 0.15.102.

artokun:
Closes artokun/comfyui-mcp#2436.

## Root cause, found in code — no reporter round-trip needed

A node whose fully-capped detail still exceeds `max_chars` is degraded to a bounded JSON stub:

```js
line = fitDetailLine(line, { id: summary.id, type: summary.type, title: summary.title }, maxChars);
```

`is_subgraph` is not in that stub. The orchestrator's `parseVerifiedQueriedNodeScope` requires

```ts
const isSubgraph = row.is_subgraph;
if (typeof isSubgraph !== "boolean") return null;   // INDETERMINATE
```

so the wide node cannot be classified. `preparePromotedWidgetWrite`'s #2394 fast path — "root + ordinary ⇒ allow" — never fires, execution falls through to `graph_get_subgraph`, and its non-definitive answer is refused by the promoted-write fence.

**Every ordinary `panel_set_widget` on a wide node was refused, with a message about promoted containers, on a root node that is not one.** `KSampler (Efficient)` in the report — many widgets, several long combo values, so it degrades.

## This is #2314's defect one layer down

#2314 gave the detail projection an always-boolean `is_subgraph` for exactly this reason. Its comment:

> MCP must distinguish a definitive ordinary node from an older Panel whose detail projection cannot classify promotion. **Positive-only emission made missing capability look like false.**

The stub then dropped the boolean, reintroducing the absence that fix had removed.

## Guarded, not passed through

An **absent** classification must stay absent. Emitting `false` for an older panel that genuinely cannot classify would authorize a promoted write the panel never vouched for — the #2314 defect pointing the other way. So the field is carried only when it is already a boolean, and a non-boolean is dropped. Both directions are pinned.

A boolean costs ~18 chars. The stub's whole purpose is the bound, so that guarantee is tested across five budgets (120–1000) rather than at one convenient number, and the absurdly-small-budget fallback is deliberately left untouched.

## The #1681 pin

`outline-coverage.test.mjs` asserted the stub's exact field literal, so this change failed it. That pin's stated purpose is that the final guard **runs** on the detail line; the field list was incidental specificity that froze order and broke on a correct change.

It now pins the call *and* the field that is load-bearing. **Verified tighter, not looser:** a mutation dropping `is_subgraph` from the call site still fails it.

## Evidence

| mutant | verdict |
|---|---|
| stub drops the field again (pre-fix) | killed — 2 failures |
| pass through unguarded (non-boolean leaks) | killed — 1 |
| caller stops passing it | killed — 1 (the wiring pin) |

**Suite:** 7081 tests, **7080 pass, 0 fail** (1 pre-existing todo). `check-tool-vocabulary` OK, `check-panel-scope` OK, `tsc --noEmit` exit 0.

## Not merged

Merge gate quota-blocked (codex resets Sep 1; Copilot exhausted). Verified and queued.

https://claude.ai/code/session_01QVc7bsTMaAKtnFTWFJUafp

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+- keep `is_subgraph` on oversized-node detail stubs so wide root nodes stay writable (artokun/comfyui-mcp#2436)
+
 ## [0.15.114] - 2026-08-27
 
 ### Fixed
--- a/browser_tests/unit/detail-stub-is-subgraph.test.mjs
+++ b/browser_tests/unit/detail-stub-is-subgraph.test.mjs
@@ -1,0 +1,100 @@
+/**
+ * artokun/comfyui-mcp#2436 — a node wide enough to degrade became UNWRITABLE.
+ *
+ * The `detail` projection emits an always-boolean `is_subgraph` on purpose. #2314's
+ * comment says why: "MCP must distinguish a definitive ordinary node from an older
+ * Panel whose detail projection cannot classify promotion. Positive-only emission made
+ * missing capability look like false."
+ *
+ * `fitDetailLine` then replaced an oversized row with a stub carrying only
+ * `{id, type, title}` — dropping exactly that boolean, and reintroducing the absence
+ * #2314 had removed, one layer further down.
+ *
+ * The orchestrator's `parseVerifiedQueriedNodeScope` requires
+ * `typeof is_subgraph === "boolean"` and treats anything else as INDETERMINATE. Its
+ * caller then falls through to `graph_get_subgraph`, whose non-definitive answer is
+ * refused by the promoted-write fence. Net effect: every ordinary `panel_set_widget`
+ * on a wide node (`KSampler (Efficient)` in the report) was refused, with a message
+ * about promoted containers, on a root node that is not one.
+ *
+ * Run with `node --test`.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+
+import { fitDetailLine } from "../../web/js/lib/graph-read.js";
+
+/** A detail row too wide for the budget — the shape that triggers the degrade. */
+const wide = (isSubgraph) =>
+  JSON.stringify({
+    id: 43,
+    type: "KSampler (Efficient)",
+    title: "KSampler (Efficient)",
+    is_subgraph: isSubgraph,
+    widgets: { blob: "x".repeat(4000) },
+  });
+
+test("#2436 the degraded stub PRESERVES a false is_subgraph", () => {
+  const out = fitDetailLine(wide(false), { id: 43, type: "KSampler (Efficient)", is_subgraph: false }, 300);
+
+  assert.ok(out.length <= 300, "the stub must still respect max_chars");
+  const parsed = JSON.parse(out);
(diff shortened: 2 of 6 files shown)


artokun:
Rebased onto `origin/main` (now includes v0.15.114) and re-verified. Do not merge from this comment — parent merges after the swarm.

### What this PR still does
`fitDetailLine` keeps a boolean `is_subgraph` on the oversized-node detail stub, and the `graph_query` detail projection actually hands it that field. Wide root nodes (`KSampler (Efficient)` in the report) stay classifiable as ordinary, so `#2394`'s fast path can allow `panel_set_widget` instead of falling through to an indeterminate `graph_get_subgraph`.

### Tests (drive the shipped function)
- `browser_tests/unit/detail-stub-is-subgraph.test.mjs` imports shipped `fitDetailLine` from `web/js/lib/graph-read.js`.
- `browser_tests/unit/graph-query-detail-budget.test.mjs` now also runs the **production `graph_query` seam** against a high-fan-in node that degrades to the stub.

**Mutant:** deleting the `is_subgraph` spread from the stub fails 3 tests:
- `#2436 the degraded stub PRESERVES a false is_subgraph`
- `#2436 a true is_subgraph survives too`
- `#2436 shipped graph_query keeps is_subgraph on an oversized-node stub`

**Fix restored:** those tests pass again.

**Suite:** `npm run test:unit` → 7090 tests, **7089 pass, 0 fail** (1 pre-existing todo). Log: `tests-panel-2436.log`.

HEAD is `decf0627` on `fix/2436-stub-drops-is-subgraph`. Ready to merge once the review gate is green.

Issue #1911, opened 2026-08-27
Live-canvas capture is gated on a Pinia $subscribe internal and skips silently when unavailable


artokun:
## Live-canvas capture is now gated on a Pinia internal, and skips silently when it is absent

Shipped in `3ac181f1` (#1907), raised by the merge gate before it merged.

### The gate

`activePointerProof` is a **required conjunct** of the capture condition:

    if (
      openSettled?.loaded !== true &&
      activePointerProof &&          // <- required
      captureSourceProof
    ) {
      await target.changeTracker?.checkState?.();
    }

and it begins with `activePointerWatchAvailable &&`. That flag is set true only when **all three** hold:

    const workflowPiniaStore = getPiniaStore("workflow");
    if (typeof workflowPiniaStore?.$subscribe === "function") {
      const stop = workflowPiniaStore.$subscribe(observeActivePointer, { detached: true, flush: "sync" });
      if (typeof stop === "function") { activePointerWatchAvailable = true; }

So the capture depends on the Pinia store being reachable by that key, `$subscribe` existing on it, **and** returning a stop handle. That is a contract dependency on a frontend internal, across the range of ComfyUI frontend versions this pack supports.

### Why it is a regression rather than an edge

`activePointerWatchAvailable` is **new in #1907**. On a frontend where any of the three conditions fails, behaviour changes from *capture* to *skip capture* — silently. An already-active workflow carrying node-written canvas-only changes then keeps a stale tracker and loses them on the next switch or reconnect.

That is user work discarded, on exactly the frontends we have least visibility into.

### The distinction worth preserving

The fail-closed reasoning in the code is right for the case it was written for:

> fail closed on the unproven capture rather than turning the previous canvas into this tab's state (#1639)

Refusing to write a **wrong** state is the correct trade. But "the watcher could not be installed" is not an unproven capture — it is **no information at all**, and the current response to no-information is to skip a capture that would previously have happened.

Trading a wrong-state write for a lost-state write is a defensible product call. Making that trade **conditional on a frontend internal being present** is not, because it silently selects which users lose data.

### Suggested shape

Separate "the watcher observed a move" from "there is no watcher":

- watcher present and pointer moved -> keep failing closed, as now;
- watcher absent -> fall back to the proof this path used before the watcher existed;
- if that fallback is genuinely unsafe, **refuse the open and name the missing frontend capability**, the way the bridge capability floors already do elsewhere.

[...]

Issue #1898, opened 2026-08-27
panel_open_workflow reports outcome unknown after successful workflow switch


artokun:
### What happened / root cause
panel_open_workflow switched to the requested workflow identity, but returned an outcome-unknown error because the loaded graph and canvas-normalized graph differed. A subsequent workflow list plus panel_graph_outline showed the target workflow was active and fully readable. This appears to be a settlement/normalization race in the long-lived panel orchestrator.

### Steps to reproduce
1. Start from a saved live workflow with modified state.
2. Call panel_open_workflow for another saved workflow.
3. Observe that the target identity switches, but the tool reports outcome unknown / graph normalization mismatch.
4. Call panel_graph_outline; the requested workflow is active and readable.

### Exact error (scrubbed)
The workflow identity switched correctly, but graph load versus canvas-normalized state differed; the tool returned outcome unknown and instructed the agent to list workflows and re-read. The response referenced related issues #1111 and #1089.

### Fix
applied locally: no
upstream-only: yes
The orchestrator should settle/re-read the active target after workflow identity changes. If the requested workflow UUID is active and panel_graph_outline can read the bound graph, return success; otherwise retry normalization once before returning outcome unknown.

### Environment
Windows 10 / ComfyUI 0.34.0 / ComfyUI frontend 1.51.9 / NVIDIA RTX 5060 Ti 16 GB / comfyui-mcp 0.52.130 / panel 0.15.102

artokun:
Claiming this issue for the next Panel implementation swarm. Planned worktree naming: wt-1898-cc. I will post the PR and review status here.

artokun:
Post-merge verification (no code change needed — this issue is already terminal).

The fix landed in #1904 (`be57e343`) and shipped in **v0.15.110** (`git tag --contains be57e343` → `v0.15.110`; `merge-base --is-ancestor be57e343 df9a5b03` confirms the release tree contains it). Panel `pyproject.toml` on `main` is `0.15.110`, i.e. one minor ahead of the `0.15.102` in the report.

I re-verified the merged fix by mutation rather than by a green suite, on `origin/main` @ `df9a5b03`:

- Baseline `node --test browser_tests/unit/open-active-settle.test.mjs` → **13 pass / 0 fail**, including two tests that drive the real `workflow_open` handler extracted from `web/js/comfyui-mcp-panel.js`.
[...]

PR #1951, opened 2026-08-27
fix(1215): do not capture or admit the previous tab's graph after a switch

artokun:
## Summary

The original untagged-root capture (#1344) is still closed on 0.15.115. The 2026-08-27 recurrence is a **distinct** hole from #1911.

`activePointerProof` (Pinia `$subscribe`) only proves the **pointer** moved. A still-mounted SOURCE canvas can still answer `bound` when the root UUID is stale or shared, so `checkState()` serializes the previous tab into TARGET. Related workflows (same node ids/types, different widgets) then look like `#1283` normalization: `normalizedOnly` publishes TARGET's fence over leftover SOURCE widgets. That is the silent wrong-canvas read this issue named.

#1911 remains its own job (watcher absent / skip capture). This PR does not relax that conjunct or duplicate that fix.

## Change

- Skip TARGET `checkState()` when the live root still reproduces SOURCE's tracker (`proven` / `presentationOnly` only).
- After load, if `normalizedOnly` would pass but the live root still reproduces SOURCE, fail closed (same veto on the #1898 recovery).
- `#874` proven-bound capture after a legitimate tab move is unchanged when the live canvas is not SOURCE.

## Tests

`productionExecutor` drives shipped `workflow_open`:

- Pinia watch present, binding `bound`, still-mounted previous widgets → capture skipped, TARGET tracker loaded
- load leaves previous widgets → open refused instead of `content_normalized` success

## Notes

Do not merge from this agent. Parent swarm merges.


artokun:
Tracked as a distinct #1215 residual, not a duplicate of #1911. The Pinia watcher still belongs to #1911; this PR only refuses a still-mounted SOURCE canvas after a tab switch.

Issue #1584, opened 2026-08-21
panel_add_node rejects DictGetNode although compatible wildcard producer is installed and live


artokun:
### What happened / root cause
`panel_add_node` refused the installed `DictGetNode` because its required `py_dict` socket is declared `DICT` and the guard reported no installed node output of exactly `DICT`. However, `JsonParseNode` from the same installed pack was added successfully immediately beforehand and exposes a wildcard `*` output that is valid for this input. The add-node guard appears to require an exact registry output type and does not treat wildcard producers (including a live producer already on the graph) as compatible.

### Steps to reproduce
1. Install/load comfyui-logicutils.
2. Add `JsonParseNode` on the live canvas (succeeds; output type `*`).
3. Call `panel_add_node` for `DictGetNode`.
4. Observe refusal after 5.8 seconds.

### Exact error (scrubbed)
Cannot add "DictGetNode": 1 required input type had no widget after 5.8s waiting for node extensions to register. input "py_dict" (declared type "DICT"): no installed node outputs "DICT" and no frontend widget is registered for it.

### Fix
Applied locally: no. Upstream-only for this active session because the failing guard runs in the long-lived panel orchestrator and cannot be reloaded safely while preserving the current workflow task.
Precise change needed: in the panel_add_node required-socket guard, accept a required custom type when any installed/live producer output is wildcard-compatible (`*`) with that input; alternatively defer the compatibility check to the live graph connection stage. The current exact-type-only test incorrectly rejects generic parser/data-flow nodes.

### Environment
Windows 11; ComfyUI 0.33.0; ComfyUI frontend 1.49.6; Python 3.13.12; PyTorch 2.13.0+cu130; NVIDIA RTX 4090 22.5 GB; comfyui-mcp 0.52.45; panel 0.15.20.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-sol` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #651

artokun:
Codex is taking over this eligible severity:P2 issue as an abandoned lane. The existing gent:codex marker had no claim comment, no current wt-1584 worktree, and no matching active agent process. I am dispatching a fresh issue-number lane. Other agents should skip this issue. Takeover time: 2026-08-21T23:08Z.

artokun:
[...]

Issue #788, opened 2026-08-08
panel_add_node cannot add nodes with union-typed widget inputs (FLOAT,INT) — LTXVEmptyLatentAudio unaddable


artokun:
### What happened / root cause
panel_add_node permanently fails for any node whose required input declares a comma-joined primitive union type (e.g. `FLOAT,INT`) together with widget config (default/min/max). Concrete case: core ComfyUI node `LTXVEmptyLatentAudio` (comfy_extras, LTX-2.3 audio path) — its `frame_rate` input is declared `FLOAT,INT` with a widget value. The panel waits ~5s for a registered widget for the union type, none ever appears, and the add is refused:

```
Cannot add "LTXVEmptyLatentAudio": 1 required input type had no widget after 5.0s waiting for node extensions to register.
  - input "frame_rate" (declared type "FLOAT,INT"): the backend declares "FLOAT,INT" as a link datatype, but this input also declares a widget value (default/min/max/options), so it needs a registered widget and none appeared.
```

Ruled out environmental causes: reproduced identically (a) before ComfyUI-LTXVideo imported, (b) after a full ComfyUI server restart with ComfyUI-LTXVideo importing cleanly (0.4s, no IMPORT FAILED), (c) after a hard browser refresh in the correct order (server restart THEN F5), and (d) after panel_refresh_nodes. The suggested remedy in the error text (reload tab) does not resolve it — no extension ever registers a widget for the union string because ComfyUI's own frontend handles this case natively by picking the first primitive type, so there is nothing to wait for.

ComfyUI's stock frontend adds this same node fine via double-click search — it resolves `FLOAT,INT` widget inputs to a FLOAT widget. The panel's add-node path treats the union as an unknown custom datatype needing a registered widget, which never comes.

### Steps to reproduce
1. ComfyUI 0.31.1 with the LTX-2.3 nodes available (core comfy_extras `LTXVEmptyLatentAudio`).
2. Call `panel_add_node` with `class_type: "LTXVEmptyLatentAudio"`.
3. Fails after the 5s wait, every time, regardless of restarts/refreshes.

### Exact error (scrubbed)
See block above (verbatim, no secrets involved).

### Fix
applied locally: no  upstream-only: yes
[...]

PR #1991, opened 2026-08-28
fix(751): classify LIST leftover defaults as sockets, not pending widgets

### LINKED ISSUE #695: panel_add_node rejects every node with a required MASK input — MASK (and other core socket types) missing from SAFE_SOCKET_TYPES

### LINKED ISSUE #751: panel_add_node refuses required custom-typed SOCKETS: the allowlist in unavailableRequiredCustomWidgetTypes inverts the default
Moved from artokun/comfyui-mcp#910 — the refusal is minted here, not in the orchestrator. I traced it; the root cause is specific.

## Where it comes from

`web/js/lib/node-widget-materialization.js`:

```js
const SAFE_SOCKET_TYPES = new Set([
  "*", "ANY", "AUDIO", "BBOX", "CLIP", "CLIP_VISION", "CLIP_VISION_OUTPUT",
  "CONDITIONING", "CONTROL_NET", "GLIGEN", "GUIDER", "IMAGE", "IPADAPTER",
  "LATENT", "MODEL", "NOISE", "SAMPLER", "SCHEDULER", "SIGMAS", "STYLE_MODEL",
  "UNET", "UPSCALE_MODEL", "VAE",
]);

export function unavailableRequiredCustomWidgetTypes(nodeOrNodeData, widgetConstructors) {
  return requiredWidgetInputTypes(nodeOrNodeData).filter((type) =>
    typeof widgetConstructors?.[type] !== "function" && !SAFE_SOCKET_TYPES.has(type),
  );
}
```

A required input is refused when it has no widget constructor **and** its type is not in a hardcoded list of 23 names. That list is a closed enumeration of core connection datatypes, so:

- `VIDEO` (SaveVideo) — **core**, not on the list.
- `IMAGEUPLOAD` (LoadImage) — **core**, not on the list.
- `TRACKS_INFO` (ComfyUI-Easy-Media) — and every other pack's socket type, which the list can never contain by construction.

That matches the reporter's table exactly, including that sibling nodes from the same pack with no custom-typed required input add fine.

## The same file already has the correct rule

`missingRequiredWidgetMaterializations`, twenty lines below, documents and implements the opposite default:

```js
 * A socket-only custom datatype is deliberately not reported: it
 * has no registered widget constructor and is valid to wire later.
...
    if (!type || typeof widgetConstructors?.[type] !== "function") continue;
```

One function reads "no constructor" as *socket, fine*; the other as *unregistered widget, refuse*. Both cannot be right, and the reporter's evidence says the second one is wrong.

## Please do not just delete the allowlist

The doc comment says it guards #580's bad prompt: an unknown type is *"indistinguishable from a widget pending its extension hook"*. That concern is real — the fix needs a positive test for widget-ness rather than a negative allowlist of socket-ness. `/object_info` carries what is needed: a widget-backed input is one whose spec type is a primitive (`INT`/`FLOAT`/`STRING`/`BOOLEAN`) or a combo (type is an array of choices); anything else is a link socket. That inverts the default the safe way round and needs no list to maintain.

## One more thread worth pulling

**`IMAGEUPLOAD` is a real registered widget type**, not a socket — `LoadImage`'s upload control. If it is absent from `comfyApp.widgets`, that suggests the registry being consulted may not be where the current frontend keeps these constructors. That would independently explain why *core* nodes are affected and why the 5s `awaitRequiredCustomWidgetRegistration` poll never succeeds (the reporter retried for minutes, hard-refreshed, and re-pulled `/object_info`). Worth confirming before settling on a fix — if the registry lookup is also wrong, the allowlist is masking a second defect.

## Corroborating signal from the report

`panel_copy_nodes` + `panel_paste_nodes` of an existing instance works every time. The paste path clones an existing node instead of constructing widgets from the schema, so it never runs this check — which localises the defect to the construct path rather than to node definitions or registration timing.

Ruled out by the reporter: load race (retried over minutes), stale node defs (`panel_refresh_nodes`), stale bundle (hard refresh), missing pack.

artokun:
Closes #751

## Recurrence

`panel_add_node` rejected `TextEncodeQwenImageEditPlusCustom_lrzjason` because its required `configs` input is declared `LIST` (a link datatype) but also carries leftover widget metadata (`{default: null}` from Python `None`, or `{default: []}`). `inputDeclaredAsSocket` treated key *presence* as a widget declaration, so the `#626` waiver (`linkProven && socketDeclared`) never fired. The guard waited 5s for a LIST widget constructor that will never register. `panel_refresh_nodes` + retry could not help.

#1990 closed this by adding `LIST` to `SAFE_SOCKET_TYPES`. That set's type-only shortcut ignores widget-value keys entirely, so a genuine LIST widget (`{default: [1, 2]}`) would also skip the wait. Growing the allowlist is the class of fix #695 warned against.

## Change

Structural, not another name on the list:

- Leftover `default: null` / `default: []` is **not** a widget declaration. A real default (`3`, `[1, 2]`, `{}`, `""`, `false`) still is. Empty object `{}` stays a DICT widget default (#1584).
- `LIST` is a ComfyUI list-socket (`isComfyListSocketType`), the same *proof* role as `isCore3dFileType` — still held to socket-shape, not the SAFE_SOCKET_TYPES type-only shortcut.
- A type the current backend proves is an output (`knownSocketTypes`) with only leftover keys is a socket too.

`#580` unproven custom widgets, `#626` value-declaring inputs, and `#788` widget-bearing unions still wait.

## Tests

`node --test browser_tests/unit/node-widget-materialization.test.mjs` — 83 passed, including the reporter's `configs: ["LIST", {default: null}]` shape and the #580/#626/#788 intact cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+- panel_add_node no longer waits 5s then refuses a LIST-typed socket that carries leftover `{default: null}` / `{default: []}` widget metadata (TextEncodeQwenImageEditPlusCustom_lrzjason `configs`) (#751)
+
 ## [0.15.122] - 2026-08-27
 
 ### Fixed
--- a/browser_tests/unit/node-widget-materialization.test.mjs
+++ b/browser_tests/unit/node-widget-materialization.test.mjs
@@ -7,6 +7,8 @@ import {
   driftedRequiredInputNames,
   missingRequiredWidgetMaterializations,
   declaredTypeMembers,
+  inputDeclaredAsSocket,
+  isComfyListSocketType,
   registeredSocketTypes,
   requiredWidgetInputTypes,
   unavailableRequiredCustomWidgetTypes,
@@ -103,28 +105,29 @@ test("a forceInput declaration remains a wireable socket", () => {
   assert.deepEqual(requiredWidgetInputTypes(node), ["ZIPN_STYLE_GALLERY", "ZIPN_SPACER", "CLIP"]);
 });
 
-test("#751: LIST with leftover widget defaults is a core list-socket, not a pending widget", () => {
-  // TextEncodeQwenImageEditPlusCustom_lrzjason.configs is declared LIST (a
-  // link datatype) but INPUT_TYPES still stamps `default: []`. That is not a
-  // widget constructor; waiting 5s for one refuses a node that the stock
-  // frontend adds as a socket.
-  const node = {
-    constructor: {
-      nodeData: {
-        input: {
-          required: {
-            clip: ["CLIP", {}],
-            configs: ["LIST", { default: [] }],
+test("#751: LIST with leftover widget defaults is a list-socket, not a pending widget", () => {
+  // Both leftover shapes packs actually emit: Python `None` (JSON null) and an
+  // empty list stamp. Neither is a widget constructor. LIST is proven as a
+  // list-socket, not by growing SAFE_SOCKET_TYPES' type-only shortcut.
+  for (const config of [{ default: null }, { default: [] }]) {
+    const node = {
+      constructor: {
+        nodeData: {
+          input: {
+            required: {
+              clip: ["CLIP", {}],
+              configs: ["LIST", config],
+            },
           },
         },
       },
-    },
-  };
-  assert.deepEqual(
-    unavailableRequiredCustomWidgetTypes(node, {}),
-    [],
-    "LIST + default must not wait for a widget constructor that will never register",
-  );
+    };
+    assert.deepEqual(
+      unavailableRequiredCustomWidgetTypes(node, {}),
+      [],
+      `LIST + ${JSON.stringify(config)} must not wait for a widget constructor that will never register`,
+    );
+  }
   // A genuine custom widget with the same leftover-default shape still waits.
   const custom = {
     constructor: {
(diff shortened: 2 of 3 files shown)


artokun:
Closing as already shipped: #751 merged as PR 1990 in v0.15.123 (`LIST` is a core socket even with leftover `default: []`; tests cover TextEncodeQwenImageEditPlusCustom_lrzjason.configs). This branch is a parallel recurrence cut against 0.15.122 and is not needed on current main.

artokun:
Superseded by PR 1990 / v0.15.123.

Issue #248, opened 2026-07-30
Include browser stack when panel_run app.queuePrompt throws


artokun:
### What happened / root cause
`panel_run` returned only the top-level `TypeError: Cannot read properties of undefined (reading 'output')` from `app.queuePrompt`, discarding the browser-side stack. This made the failing frontend extension impossible to identify and forced repeated graph rewrites before the actual source (`tts_audio_suite`, then a swallowed root exception in `cg-use-everywhere`) could be located.

The `graph_run` handler awaited `app.queuePrompt(...)` directly and let the bridge stringify the exception without preserving `error.stack`.

### Steps to reproduce
1. Open a workflow where a frontend extension throws inside its `app.graphToPrompt` wrapper.
2. Call `panel_run` (full or `to_node_id`).
3. Observe that only the one-line TypeError is returned, with no browser URL/file/line stack.

### Exact error (scrubbed)
Before patch:
```
Error: Cannot read properties of undefined (reading 'output')
```

After preserving the stack, the actionable location appeared:
```
TypeError: Cannot read properties of undefined (reading 'output')
    at app.graphToPrompt (http://127.0.0.1:8188/extensions/tts_audio_suite/audio_analyzer_interface.js:102:24)
    at async ComfyApp.graphToPrompt (.../extensions/comfyui-videohelpersuite/js/VHS.core.js:2349:23)
```

### Fix
Applied locally: yes  Upstream-only: no

```diff
-    await app.queuePrompt(0, batch, partialTargets);
+    try {
+      await app.queuePrompt(0, batch, partialTargets);
+    } catch (error) {
+      const message = error instanceof Error ? error.stack || error.message : String(error);
+      throw new Error(`app.queuePrompt failed:\n${message}`);
+    }
```

Verified by repeating `panel_run`; it returned the complete browser stack with extension file and line, which enabled diagnosis. After fixing the revealed workflow/extension errors, the same partial run queued successfully.

### Environment
Windows 11; ComfyUI 0.29.0; NVIDIA RTX 5070 Ti 16 GB; comfyui-mcp 0.48.7; panel 0.11.4.

artokun:
[...]

PR #2013, opened 2026-08-29
fix: preserve browser stack for panel_run queue errors (#248)

artokun:
## Summary

- Preserve useful browser source context when panel_run app.queuePrompt rejects or throws.
- Apply the diagnostic boundary to both scoped and full graph_run paths; the bridge serializes error.message, so a safely read browser stack is embedded there.
- Avoid instanceof and String(error): non-Error values, hostile proxies, and non-string or unreadable stacks retain their original thrown identity.
- Bound the complete diagnostic message at 4,096 characters, preserve stack line breaks, remove control bytes, and reuse the existing credential redactor.
- Update the existing bounded-dispatch source guard for the helper call.

## Evidence

- Issue #248 current recurrence is the panel scoped to_node_id path. The older MCP detectRunRejection fix can classify and forward a panel rejection but cannot reconstruct a stack discarded by Panel.
- Regression coverage exercises the extracted production graph_run path for both scoped and full runs and asserts the extension URL and source lines survive before any POST.
- Regression coverage covers huge stacks, credential-shaped text, Symbol and throwing-coercion stacks, a throwing Proxy getPrototypeOf trap, and a cross-realm native Error.
- No init.py or apps_routes.py changes.

## Checks

- npm.cmd run check:scope
- npm.cmd run check:changelog
- npm.cmd run typecheck
- npm.cmd run test:unit: 7405 tests, 7404 passed, 0 failed, 1 todo
- Full run-scope-guard suite: 142 passed
- Full run-command-budget suite: 41 passed
- Focused #248 tests passed

Do not merge automatically.

### DIFF
--- a/browser_tests/unit/run-command-budget.test.mjs
+++ b/browser_tests/unit/run-command-budget.test.mjs
@@ -37,7 +37,11 @@ import assert from "node:assert/strict";
 import { readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
 
-import { dispatchScopedRun, createRunFetchInterceptor } from "../../web/js/lib/run-scope-guard.js";
+import {
+  dispatchScopedRun,
+  createRunFetchInterceptor,
+  invokeQueuePromptWithBrowserStack,
+} from "../../web/js/lib/run-scope-guard.js";
 import { makeCommandBudget } from "../../web/js/lib/command-budget.js";
 import { withTimeout } from "../../web/js/lib/bounded-step.js";
 import {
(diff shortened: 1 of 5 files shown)


artokun:
Independent Codex gate: NO-SHIP at head e5951d73c076255e33ba9d3e477ec0015323208d. P1 integration: current main advanced to the v0.15.124 release and this PR is conflicting/stale; no checks are reported. P2 correctness: readable string stacks on non-Error objects are wrapped into a new Error (web/js/lib/run-scope-guard.js:71-88), losing original message/identity before bridge serialization (web/js/comfyui-mcp-panel.js:28166-28177). Rebase/resolve, add regression coverage for non-Error stack objects and the real panel-visible path, rerun CI, then re-gate.

artokun:
Repair update: rebased onto current Panel main at 028d85f and pushed as f7f9f8e2e72c1950ce35571eee2fdf3c3d01e968. Non-Error readable stack objects now preserve identity/message, browser and cross-realm Error decoration remains intact, and focused production graph_run/panel-visible serialization coverage was added. Focused 8 passed; full unit 7,415 passed, 1 todo; typecheck passed; E2E discovery listed 98 tests. Full E2E was not run because ComfyUI was unavailable on localhost:8188. Current CI Lint pack is green.

PR #2036, opened 2026-08-30
fix(248): reject dynamic error-tag proxies

artokun:
Follow-up to merged PR #2013.

This hardens the browser Error classifier against a finite Proxy whose `get` trap dynamically synthesizes `Symbol.toStringTag = "Error"`. Such a value is conservatively kept as the original thrown object instead of being wrapped and losing its identity/message surface.

Checks completed locally:
- `node --test browser_tests/unit/run-scope-guard.test.mjs` — 148 passed
- `npm.cmd run test:unit` — 7,419 passed, 1 todo
- `npm.cmd run typecheck` — passed
- `git diff --check` — passed

Do not merge automatically.

### DIFF
--- a/browser_tests/unit/run-scope-guard.test.mjs
+++ b/browser_tests/unit/run-scope-guard.test.mjs
@@ -333,6 +333,36 @@ test("#248 a non-Error with a spoofed Error toStringTag keeps its identity and m
   }
 });
 
+test("#248 a finite dynamic toStringTag Proxy keeps non-Error identity and message", async () => {
+  const thrown = new Proxy(
+    {},
+    {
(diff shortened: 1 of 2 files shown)


artokun:
Independent exact-head gate: SHIP. The follow-up changes only run-scope-guard.js and its focused tests. The own stack witness rejects finite dynamic Symbol.toStringTag Proxy spoofing while preserving ordinary/cross-realm Error decoration; cycle detection and the 32-level prototype bound remain bounded and fail-closed. Local focused suite: 148 passed; diff-check clean. GitHub CI run 33297039106 passed. No P0-P2 findings remain in this scoped follow-up.

PR #2037, opened 2026-08-30
fix: harden browser error proxy classification

artokun:


artokun:
Follow-up for merged #2013/#248: this commit adds a conservative own `stack` witness before browser-error decoration, so a finite Proxy that synthesizes `Symbol.toStringTag = "Error"` cannot be wrapped and lose its original identity/message. Existing ordinary/cross-realm Error, cyclic/unbounded Proxy, hostile stack, and production graph_run coverage is retained. Focused run-scope-guard tests and diff checks pass locally; PR is intentionally separate from the merged fix.

artokun:
Duplicate of #2036: same two-file dynamic-Proxy hardening and regression, with #2036 already at an equivalent clean head and green CI. Keeping #2036 as the single follow-up.

PR #2041, opened 2026-08-30
fix: rebind workflow save after restart confirmation timeout (#1757)

### LINKED ISSUE #1757: panel_save_workflow fails fetch after restart confirmation timeout
After a panel_restart_comfyui confirmation timed out without restarting, live graph reads and layout mutations still worked, but panel_save_workflow returned only "Error: Failed to fetch". panel_list_workflows immediately succeeded and showed the active persisted workflow still modified=true. A second save attempt returned the same error. The generic fetch error hides the failing endpoint/status and leaves a dirty workflow that cannot be persisted through the panel.

### Steps to reproduce
1. Open a persisted workflow and make live layout edits.
2. Call panel_restart_comfyui and let the 90s confirmation expire without restarting.
3. Confirm panel_list_workflows still succeeds.
4. Call panel_save_workflow({}).
5. Observe "Error: Failed to fetch"; active workflow remains modified=true.

### Exact error (scrubbed)
Error: Failed to fetch

### Fix
applied locally: no
upstream-only: yes
Precise change needed: panel_save_workflow should preserve the failing request URL/status/body, rebind or retry the same-origin save route after a timed-out restart confirmation, and return an actionable error rather than the browser's generic fetch exception. Add a regression covering restart-confirm timeout -> graph read -> save.

### Environment
Windows 11; ComfyUI 0.33.3; frontend version unavailable because the headless stats route is currently unreachable; RTX 4090 24 GB; comfyui-mcp 0.52.82; panel 0.15.63.

artokun:
Fixes #1757.

## Recurrence

PR #1769 explained a lost `/userdata` write as a transport failure (URL, no status/body, websocket observation). It did **not** rebind or retry the same-origin save route. After a restart confirmation timed out without rebooting, live graph reads still worked and the workflow stayed `modified: true`, but the in-place save kept returning the browser's bare `Failed to fetch`. Duplicate report 2026-08-29 on panel 0.15.102 saw the same no-response userdata write with the panel websocket still open.

## What this does

1. After a withdrawn/timed-out restart confirmation, rebind ComfyUI's API host to the page origin before the next in-place write.
2. If the write still throws a transport error, probe whether the bytes landed; retry the shipped store `saveWorkflow` **once** when the probe says missed (in-place overwrite of the same graph is idempotent).
3. If it still cannot persist, keep the userdata URL and any HTTP status/body on the error instead of the generic fetch exception.

Save-as is unchanged: overwrite:false is not idempotent, so unknown probes are not retried there.

## Tests

```
node --test browser_tests/unit/save-transport-failure.test.mjs browser_tests/unit/save-route-retry.test.mjs
```

Regression drives `saveActiveWorkflow` (not a reimplementation): restart-confirm timeout → in-memory graph read still works → save either succeeds after rebind/retry or returns URL/status, never bare `Failed to fetch`. 27/27 pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project are documented here. This project adheres to
 ## [Unreleased]
 
 ### Fixed
+- after a restart confirmation times out without rebooting, an in-place workflow save rebinds the same-origin userdata route and retries the write once, so a dirty tab is not stranded behind the browsers bare Failed to fetch; if the write still cannot land, the error names the userdata URL and any HTTP status/body (#1757)
 - promoted widget writes no longer treat a stale inner Primitive handle as the parent rail, so MiniMax H3 duration/value_1, turbo_mode, turbo_steps, and lora_name serialize the new value (#366)
 - scoped run-to-node no longer false-refuses after a finished subgraph edit: the graph stamp waits for pending canvas work to settle and restamps once if the revision moved before queue, without falling through to a full-graph run (#572)
 - resolve the 18+ consent card before the 300s tools/call deadline so an idle user gets a structured timeout instead of a transport hang (#390)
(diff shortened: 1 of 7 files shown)


artokun:
Rebased onto main after #2040. Awaiting CI.

Issue #1263, opened 2026-08-15
New workflow auto-persist changes workflow UUID and rejects the next graph mutation


artokun:
### What happened / root cause
A new blank tab created by panel_new_workflow had workflow_uuid c3755c1e-504a-460d-a932-15cd12f9cccf. After pasting nodes and applying widget edits, the frontend auto-persisted it as `Untitled 2026-08-15 04-51-37.json` with workflow_uuid 42dcaa96-e40e-4c02-a780-3ae630f570e9. The next panel_add_node was rejected with a workflow instance mismatch even though the user had not switched tabs and the active canvas was the same graph. Auto-rebind recovered and the exact retry succeeded.

### Steps to reproduce
1. panel_new_workflow.
2. panel_paste_nodes and perform several graph edits.
3. Let the new tab auto-persist/receive an Untitled filename.
4. Call panel_add_node.
5. The call is refused because the same active graph acquired a new workflow UUID.

### Exact error (scrubbed)
`graph_add_node was NOT applied — nothing changed. workflow instance mismatch: command issued for workflow instance c3755c1e-..., active canvas reports 42dcaa96-... Active workflow last moved to workflows/Untitled ... AUTO-REBIND: ATTEMPTED ... Retry once.`

### Fix
Applied locally: no. Upstream-only: yes; the long-lived panel orchestrator owns the fence. Suggested change: when a tmp workflow auto-persists into a file without a tab switch or graph replacement, migrate the existing fence identity atomically or publish the new UUID before accepting the next mutation.

### Environment
Linux; ComfyUI 0.27.0; AMD Radeon RX 9070 XT 16 GB; Intel i7-12700K; 32 GB RAM; comfyui-mcp 0.51.56; panel 0.14.42.

Related (same root cause): #747

artokun:
[...]

PR #2127, opened 2026-08-31
fix(2125): an absent active workflow is not a tab switch

### LINKED ISSUE #2125: panel_add_node refuses after live graph view change on empty workflow
panel_add_node returned an outcome error while adding the first node to the active empty workflow: "The active workflow or graph view changed while this node was preparing; nothing was added. Retry on the intended tab." The graph remained empty. No intentional tab switch or graph edit occurred between the outline read and add request.

### Steps to reproduce
1. Open ComfyUI with the sidebar panel and an active unsaved empty workflow.
2. Read panel_graph_outline.
3. Call panel_add_node for UNETLoader.
4. Observe refusal stating active workflow or graph view changed.

### Exact error
The active workflow or graph view changed while this node was preparing; nothing was added. Retry on the intended tab.

### Fix
Applied locally: no. Upstream-only: yes; investigate graph identity/view race or active-tab routing on first mutation in an empty unsaved workflow.

### Environment
Windows 11 / ComfyUI 0.34.0 / GPU NVIDIA GeForce RTX 5090 32GB / comfyui-mcp 0.52.146 / panel 0.15.102. ComfyUI frontend version unavailable in current environment line.

---
Filed from the ComfyUI panel by **chatgpt** · model `gpt-5.6-luna` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #1263

artokun:
Fixes #2125.

## What the reporter saw

On an empty unsaved workflow: `panel_graph_outline` **succeeded**, then `panel_add_node`
was refused with

> The active workflow or graph view changed while this node was preparing; nothing was
> added. Retry on the intended tab.

No tab switch, no graph edit, graph left empty, every retry identical.

## Root cause

That read-succeeds / mutation-refuses asymmetry is the fingerprint.
`revalidateGraphMutationContext` is the **only** gate on the mutation path that compares a
before/after workflow reference, and `activeWorkflowRef()` returns `null` whenever the
frontend exposes no active workflow (`extensionManager.workflow` missing, still seating, or
throwing). The comparator it delegates to, `sameWorkflowObject`, answers `false` for **any**
pair containing a null — which is the correct answer to the question it is asked (two
references sharing no identity carrier can never be *proven* the same), and the wrong answer
to the question the gate is asking (did anything *change*?).

So absent-then-absent read as "the tab changed". Verified by execution against `origin/main`
before touching anything — the gate refuses when handed the **literally identical context
object**:

```
sameWorkflowObject(null, null)          = false
sameGraphMutationContext(ctx, ctx, swo) = false   <- same object, nothing moved
```

The rest of the panel treats an absent workflow as *supported*, not as a refusal:
`graphEmptyBindingUnproven` returns false on it (`// no workflow service — legacy
availability`) and every other predicate in `resolveGraphBindingVerdict` degrades to
no-refusal. Confirmed by execution on the **mutation** bar:

```
resolveGraphBindingVerdict(activeWorkflow: null, empty root, requireDirtyMutationBinding) = null
```

The binding assert that runs on the very next line already lets this through. This gate was
the one place that did not — and it sits on the mutation path only, which is exactly why the
outline read worked on the same canvas.

Corroborating detail from the report: the environment line says *"ComfyUI frontend version
unavailable"*. The panel reads that from `app.extensionManager.frontendVersion` — the same
`extensionManager` surface `activeWorkflowRef()` probes for `.workflow.activeWorkflow`.

## The fix

One predicate, at the mutation-context boundary: absence at **both** ends is unchanged.
Absence appearing or disappearing stays refused.

Deliberately **not** fixed in `sameWorkflowObject` — it has ~12 other call sites (openWorkflows
enumeration loops, owner-record comparisons) where a null pair must keep answering false. This
is a change question, not an identity question, so it belongs in the context comparator.

## Please look hardest at these

- **The safety claim that makes this a fix and not a cosmetic message change** is that the
  binding bar permits a null workflow. If that were wrong, I would only be moving the refusal
  one line down. It is asserted by execution in the test, not by reading.
- **Is admitting `null → null` a hole in the #646/#349 wrong-canvas fence?** My argument: no,
  because a null workflow carries no canvas claim at all, the `app`/`graph`/`rootGraph`/`canvas`
  identity checks are untouched and still decide, and the binding assert still runs. Tests pin
  the three transitions that must stay refused (null→wf, wf→null, A→B) and that a changed
  graph/canvas still refuses even with no workflow present.
- **`== null` vs falsy.** `activeWorkflowRef()` ends in `|| null`, so it yields `null`, but the
  slot is compared with `== null` (null *and* undefined) rather than truthiness, so a workflow
  object is never treated as absent.

## Correction to the issue thread

The follow-up comment says updating 0.15.102 → 0.15.149 restored this. I extracted the function
bodies of `revalidateGraphMutationContext`, `captureGraphMutationContext`, `getGraphCtx` and
`activeWorkflowRef` at both tags and diffed them: **all four are byte-identical**, as are
`graph-mutation-context.js` and `sameWorkflowObject`. The version bump does not explain the
recovery — the ComfyUI restart (re-seating the workflow store, so `activeWorkflow` became
non-null) does, and that is consistent with this root cause. The refusal path was still fully
live on `origin/main` @ `b6847f6b` (v0.15.149).

## Evidence

- New: `browser_tests/unit/graph-mutation-context-absent-workflow.test.mjs` — 6 tests.
  Two run the **real** `captureGraphMutationContext` / `revalidateGraphMutationContext` source
  sliced out of `comfyui-mcp-panel.js` and executed over injected deps, so a fix that lived only
  in the lib while the call site passed its own comparator would not pass.
- **Mutation-verified.** Reverting the predicate to `return false` turns the lib test *and* the
  call-site test red (4 pass / 2 fail); the other four stay green, which is the point — they pin
  behaviour that must not change.
- Full unit suite: **7857 pass / 0 fail / 1 todo**.
- `check:changelog`, `check-panel-scope`, `i18n-check` all clean.

## Browser coverage — what I do NOT have

I could not run the Playwright suite: it needs a live ComfyUI and no instance was allocated to
this run. So this change is **not** covered by a browser test, and I am not implying it is. The
call-site test executes the real production source of the gate, which is the strongest coverage
available without a rig, but it is not the sidebar.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+- panel_add_node no longer refuses with "the active workflow or graph view changed" on a frontend that exposes no active workflow: absent-at-both-ends is unchanged, not a tab switch, so the mutation runs on the same canvas the reads already succeeded on (#2125)
+
 ## [0.15.149] - 2026-08-30
 
 ### Fixed
--- a/browser_tests/unit/graph-mutation-context-absent-workflow.test.mjs
+++ b/browser_tests/unit/graph-mutation-context-absent-workflow.test.mjs
@@ -1,0 +1,225 @@
+// #2125 — panel_add_node refused on a canvas nothing had touched.
+//
+// The reporter read `panel_graph_outline` on an empty unsaved workflow, then
+// called `panel_add_node` and got "The active workflow or graph view changed
+// while this node was preparing; nothing was added. Retry on the intended tab."
+// No tab switch and no graph edit happened between the two calls, the graph
+// stayed empty, and every retry refused identically.
+//
+// That asymmetry — read succeeds, mutation refuses — is the fingerprint.
+// `revalidateGraphMutationContext` is the ONLY gate on the mutation path that
+// compares a before/after workflow reference, and `activeWorkflowRef()` returns
+// null whenever the frontend exposes no active workflow. The comparator it
+// delegates to, `sameWorkflowObject`, answers false for any pair containing a
+// null (correct for an identity question: no shared carrier, nothing proven),
+// so absent-then-absent read as "the tab changed".
+//
+// The rest of the panel treats an absent workflow as supported, not as a
+// refusal: `graphEmptyBindingUnproven` returns false on it ("no workflow
+// service — legacy availability") and the binding bar that runs immediately
+// AFTER this gate therefore returns no verdict at all. This gate was the one
+// place that disagreed.
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { sameGraphMutationContext } from "../../web/js/lib/graph-mutation-context.js";
+import { sameWorkflowObject } from "../../web/js/lib/workflow-chat-identity.js";
+import { resolveGraphBindingVerdict } from "../../web/js/lib/graph-binding.js";
+
+const PANEL_JS = fileURLToPath(new URL("../../web/js/comfyui-mcp-panel.js", import.meta.url));
+
+// STABLE across every ctx() in this file: a fresh object per call would make the
+// app/graph/canvas slots differ too, and then every "still refused" assertion
+// below would pass for a reason that has nothing to do with the workflow slot.
+const APP = {};
+const ROOT_GRAPH = { _nodes: [] };
+const CANVAS = {};
+
+function ctx(overrides = {}) {
+  return { app: APP, graph: ROOT_GRAPH, rootGraph: ROOT_GRAPH, canvas: CANVAS, workflow: null, ...overrides };
+}
+
+// ---------------------------------------------------------------------------
+// The comparator
+// ---------------------------------------------------------------------------
+
+test("#2125: a workflow absent at BOTH ends is unchanged — the context did not move", () => {
+  const before = ctx();
+  // The strongest possible statement of the defect: the SAME object, compared
+  // with itself, through the comparator production actually passes in.
+  assert.equal(
+    sameGraphMutationContext(before, before, sameWorkflowObject),
+    true,
+    "an identical context object cannot be a tab switch",
+  );
+  // And a re-read that produced an equal-but-distinct record, as the real
+  // capture/revalidate pair does.
+  assert.equal(sameGraphMutationContext(before, { ...before }, sameWorkflowObject), true);
+});
+
+test("#2125: absence APPEARING or DISAPPEARING is still a change", () => {
+  const absent = ctx();
+  const wfA = { changeTracker: {} };
+  const wfB = { changeTracker: {} };
+  // A workflow that seats itself mid-preflight: the node would land on a tab the
+  // caller never named. Still refused.
+  assert.equal(sameGraphMutationContext(absent, ctx({ workflow: wfA }), sameWorkflowObject), false);
+  // A workflow that vanishes mid-preflight. Still refused.
+  assert.equal(sameGraphMutationContext(ctx({ workflow: wfA }), absent, sameWorkflowObject), false);
+  // And the case the guard exists for, untouched: two different workflows.
+  assert.equal(
+    sameGraphMutationContext(ctx({ workflow: wfA }), ctx({ workflow: wfB }), sameWorkflowObject),
+    false,
+    "the tab-switch guard this whole gate exists for must not be weakened",
+  );
+  // A proxy/raw pair of the SAME workflow stays same (the reason the comparator
+  // is injected at all).
+  assert.equal(
+    sameGraphMutationContext(
+      ctx({ workflow: wfA }),
+      ctx({ workflow: { __v_raw: wfA } }),
+      sameWorkflowObject,
+    ),
+    true,
+  );
+});
+
+test("#2125: the non-workflow slots still decide the verdict when the workflow is absent", () => {
+  // Admitting absent→absent must not admit a graph/canvas move underneath it.
+  const before = ctx();
+  for (const slot of ["app", "graph", "rootGraph", "canvas"]) {
+    assert.equal(
+      sameGraphMutationContext(before, { ...before, [slot]: {} }, sameWorkflowObject),
+      false,
+      `a changed ${slot} must still refuse even with no workflow to compare`,
+    );
+  }
+});
+
+// ---------------------------------------------------------------------------
+// Why admitting it is safe: the gate that runs NEXT already permits this state
+// ---------------------------------------------------------------------------
+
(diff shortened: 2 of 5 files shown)


artokun:
Closing the one alternative hypothesis, since the issue points at #1263 ("same root cause") and that mechanism produces the *identical* error string.

#1263 is a grounding auto-persist: a first save swaps the active `ComfyWorkflow` object, so `sameWorkflowObject(pre, post)` is false. If that swap could land between `graph_add_node`'s capture and its revalidate, it would explain this report too — and my fix would not help, because a genuine object swap *is* a change.

It cannot. `groundUnsavedWorkflow` has exactly **one** call site in the panel:

```
$ grep -n "groundUnsavedWorkflow" web/js/comfyui-mcp-panel.js
8097:async function groundUnsavedWorkflow() {
45078:    // works from a real file (Ctrl+S / reload behave). groundUnsavedWorkflow is
45082:      const saved = await groundUnsavedWorkflow();
```

and it is `await`ed in the send handler **before the turn is dispatched** — so it has completed by the time the agent issues its first tool call, and cannot overlap a preflight from that same turn. `groundActiveWorkflow` likewise has one caller (`groundUnsavedWorkflow` itself).

That leaves, as the ways `activeWorkflowRef()` can differ across the preflight of a turn where the reporter performed no tab switch and no save: none. Which is consistent with it not having differed at all — it was absent at both ends, and the gate called that a change.

Also worth noting for review: the reporter's workflow stayed **unsaved** throughout ("active unsaved empty workflow", "The graph remained empty"). Had grounding fired, the tab would have been persisted as an `Untitled …` file.

artokun:
Gate round 1: **NEEDS-REVISION** (exit 1), one P1, and it was right.

> [P1] `web/js/lib/graph-mutation-context.js:39` — Treats failed workflow reads as unchanged absence; `activeWorkflowRef()` returns `null` when its lookup throws, so two failed reads pass as `null→null` even if the active workflow changed, allowing mutation on an unverified canvas.

`activeWorkflowRef()` collapses two different answers into one `null`: "the frontend exposes no active workflow" and "the lookup threw". My first cut admitted both. The second is *"I did not find out"*, and two of them cannot witness each other across a preflight the workflow may have moved during.

Fixed in `b5a438a`:

- New `probeActiveWorkflow()` returns `{ workflow, readable }`. `activeWorkflowRef()` now delegates to it, so there is **one** probe implementation and no drift for its ~40 other callers.
- `captureGraphMutationContext` and `revalidateGraphMutationContext` each make **one** probe call and carry `workflowReadable` on the context, so the flag cannot disagree with the value it describes.
- `absenceProven()` requires `workflow == null && workflowReadable === true`. Unreadable at either end is refused; a context that never states readability is refused too, so the relaxation is opt-in by evidence and nothing else can reach it by accident.

Ordering preserved: `getGraphCtx()` still runs before the probe in both halves, matching the pre-existing capture order, so the #646 "gate fires before the canvas-changing probe" invariant is untouched.

**Mutation-verified, three separate mutants, each caught by a different test:**

| mutant | tests that go red |
|---|---|
| `absenceProven` drops the readability conjunct (the P1 restored) | the 2 `r1` tests |
| `probeActiveWorkflow`'s `catch` returns `readable: true` | the real-probe test |
| `sameWorkflowSlot` returns `false` (pre-fix) | the 2 original tests |

The real-probe test is not a stub — it slices `probeActiveWorkflow` out of `comfyui-mcp-panel.js` and drives it with a getter that throws.

---

**One unrelated harness repair, called out because I changed a test I did not write.**

`reconnect-recovery.test.mjs`'s #646 ordering test read `SRC.slice(start, start + 1400)`. My added comments grew `revalidateGraphMutationContext` to 1616 chars, so `assertGraphBoundToActiveWorkflow(` fell outside the window. The real ordering is intact — gate 629 < getGraphCtx 1039 < assert 1495 — the scan just stopped reading.

I did **not** widen the number. `indexOf` returns -1 for a missing needle and **-1 compares less than every real offset**, so had the FIRST needle been the one to fall out, that assertion would have silently become vacuously true rather than failing. The window is a latent trap, not a limit to bump. It now slices to the function's own end and routes every needle through a helper that asserts it exists before comparing — strictly stronger than what it replaced, plus one new assertion that the probe follows `getGraphCtx`.

Mutation-checked so the repair is not a loosening: moving the reconnect gate below `getGraphCtx` in the panel source still fails that test (47 pass / 1 fail), and it passes 48/48 restored.

Full unit suite on `b5a438a`: **7860 pass / 0 fail / 1 todo**.

artokun:
Gate round 2: **NEEDS-REVISION** (exit 1), one P1 — and it landed on my own safety argument, which is the useful kind.

> [P1] `web/js/comfyui-mcp-panel.js:13707` — The final binding check performs a third `activeWorkflowRef()` read and converts an unreadable throw to `null`, which is treated as legacy availability — if reads return `null`, `null`, then throw while app/graph/canvas remain stable, `graph_add_node` proceeds without verified workflow binding.

Correct. Round 1 taught the *comparison* to tell a proven absence from an unreadable one, and then handed the decision to `assertGraphBoundToActiveWorkflow`, which took its own third sample and conflated them again — so the write could be permitted on exactly the evidence the gate above had just refused. The whole "the binding bar already permits this state" justification quietly depended on the third read agreeing with the first two.

Fixed in `90ab9d2` by removing the third read rather than hardening it: `revalidateGraphMutationContext` passes its probe through as `workflowProbe`, and the assert uses `workflowProbe ? workflowProbe.workflow : activeWorkflowRef()`. One observation, one decision. The option defaults to `null`, so the read paths and the other four direct-mutation call sites re-read exactly as before — no blast radius.

Mutation-checked, each caught by a different test:

| mutant | test that goes red |
|---|---|
| call site drops `workflowProbe: currentProbe` | `binding assert decides on the probe, not a third fresh read` |
| assert reverts to a bare `activeWorkflowRef()` | `binding assert PREFERS the supplied probe over its own read` |

The behavioural half also asserts `freshReads === 0` — the harness counts every `activeWorkflowRef()` call taken outside the two probes, so a fourth sample appearing anywhere on this path fails.

**Coverage honesty on that pair:** the call-site half is behavioural (real source, executed). The consuming half is a *source assertion* on the exact expression, because `assertGraphBoundToActiveWorkflow` has too many module-level dependencies to execute in a `new Function` harness without the harness itself becoming the thing under test. It fails on a revert, but it is a scan, not an execution — stating that rather than letting the table imply otherwise.

**Out of scope, deliberately:** `assertGraphBoundToActiveWorkflow` still turns a throwing `activeWorkflowRef()` into "legacy availability" on its *other* call sites. That is pre-existing and unchanged by this PR; on those paths there is no prior observation for it to disagree with, so there is nothing to reconcile. Worth its own issue, not a scope creep here.

---

**Round 3: `SHIP` (exit 0)** on `90ab9d2` — the SHA in this PR head and the one CI ran (`success`, run 33375620729). Full unit suite on it: **7862 pass / 0 fail / 1 todo**.

Issue #2034, opened 2026-08-30
No way to stop generated images/videos auto-posting into the chat; "Video previews in chat" is mistaken for that switch and does not do it


artokun:
### What happened / root cause

A user on a heavy video workflow wanted generated media to stop flowing into the chat transcript. They found the only related setting, **"Video previews in chat"** (`comfyui-mcp.videoPreviews`), turned it off, and reported that it "doesn't fix it". Their settings store confirms `"comfyui-mcp.videoPreviews": false`.

Two separate things are going on:

1. **`videoPreviews` does not do what its name implies.** It only chooses the *render mode* of a video card — live `<video>` vs. first-frame placeholder (`web/js/lib/video-previews.js`, consumed at `web/js/comfyui-mcp-panel.js:35869`). The card is still painted into the log either way. The tooltip does say this, but the setting *name* reads as an on/off for "previews in chat", so OFF → "still see media" is a reasonable read of it as broken.
2. **There is no setting at all that suppresses the media inflow**, and `videoPreviews` does not cover images or audio in any form. The `executed` handler in `buildPanel()` (`web/js/comfyui-mcp-panel.js:41082-41120`) unconditionally calls `paintVideo` / `paintAudio` / `paintImage` for every deliverable output of every finished run. I checked the full settings list (`locales/en/settings.json`) — `videoPreviews` is the only media-related entry, so there is genuinely nothing to turn off.

Net effect for a user doing long video work: every run permanently appends cards to the transcript, they are also persisted via `recordMedia` and replayed on reload/thread switch, and the one switch that looks like it should stop it does not.

### Steps to reproduce

1. Settings → General → turn **"Video previews in chat"** OFF.
2. Run any workflow that produces a `SaveImage` / `VHS_VideoCombine` output.
3. On completion, a media card is still appended to the chat log — a placeholder card for video, and a fully unaffected image card for images (`videoPreviews` never touched the image path).
4. There is no other setting to turn off. Expected: some way to keep the transcript text-only.

### Exact error (scrubbed)

No error — silent behavioural gap. The setting applies successfully and changes nothing the user was trying to change.

### Fix

**applied locally: yes**  **upstream-only: no**

Added `comfyui-mcp.chatMedia` ("Generated media in chat", boolean, default **true**, General, sortOrder 151) and gated the three auto-paint calls in the `executed` handler on it. Deliberately scoped to the *cards only*:

[...]

Issue #1934, opened 2026-08-27
panel_run completion says no media output although CompareFrames produced temp images


artokun:
### What happened / root cause
A `panel_run({to_node_id:147})` completed successfully and its event reported "produced no image or video output." `get_history(action:"list")` for the same prompt showed node 147 had 384 `a_images` and 384 `b_images` temp PNG outputs.

### Steps to reproduce
1. Run a CompareFrames output node through `panel_run`.
2. Observe completion event reporting no media.
3. Query the same prompt with `get_history(action:"list")`; temp PNG outputs are present.

### Exact error (scrubbed)
Completion: `The run you queued finished successfully ... and produced no image or video output.`
History: `Outputs (1 nodes) - Node 147: a_images ... b_images ...`

### Fix
Applied locally: no. Upstream-only: yes.
Panel completion extraction should include temp-image outputs from list-style UI result keys such as `a_images`/`b_images`.

### Environment
Windows 11; ComfyUI 0.33.0; ComfyUI_frontend 1.48.7; RTX 4090 24GB; comfyui-mcp 0.52.139; panel 0.15.102.

artokun:
Confirmed to the line, and your suggested fix would be harmful as literally stated — which is worth explaining, because the right fix is close to it.

### The mechanism, exactly as you diagnosed

`web/js/comfyui-mcp-panel.js`:

```js
// ComfyUI groups a node's outputs by kind: `images`, plus `gifs`/`videos` for …
const media = [...(out.images || []), ...(out.gifs || []), ...(out.videos || [])];
```

Three literal keys. `history-reconcile.js` reads `out.images` alone. `CompareFrames` emits `a_images` / `b_images`, so it contributes nothing and the completion truthfully reports what it found: nothing.

### Why "include list-style keys" cannot be done as written

Your node produced **768 images** — 384 `a_images` plus 384 `b_images`. The completion frame is hard-bounded: ONE frame, with a per-still budget, precisely so a large run cannot flood the agent's turn. Folding 768 temp PNGs into that path would either blow the bound or be silently truncated to a handful, and a truncated-to-4 delivery from a 768-image run is a worse lie than the current one — it looks complete.

So the fix is not to deliver them. It is to **stop claiming none exist**.

### What I would build

Count them, name where they are, deliver none by default:

[...]

PR #2130, opened 2026-08-31
fix(#2126): report a SaveAudio run instead of dropping it

### LINKED ISSUE #2126: The audio recording completed successfully; no media notifications were received.

artokun:
Fixes #2126.

## What was reported

> The audio recording completed successfully; no media notifications were received.

## What actually happens

ComfyUI groups a node's outputs by kind. Every core audio output node — `SaveAudio`, `SaveAudioMP3`, `SaveAudioOpus`, `SaveAudioAdvanced`, `PreviewAudio` — serialises through `SavedAudios.as_dict()` / `PreviewAudio.as_dict()` in `comfy_api/latest/_ui.py`, both of which return `{ audio: [ {filename, subfolder, type} ] }`. (Read out of the install at `C:\Users\Artokun\ComfyUI-Installs\ComfyUI\ComfyUI`, not from memory.)

The run-completion path collected `images` / `gifs` / `videos` and nothing else:

1. `collectNodeOutputMedia` returned `{deliverable: [], withheld: null}` — `audio` matched neither the three literal keys nor the widened `/(?:images|gifs|videos)$/` scan.
2. `onExecuted`'s `if (!media.length && !withheld) return;` fired, so **no player was painted** and **the tracker never heard about the run**.
3. `execution_success`, with nothing buffered, took the media-less branch and flushed `noMedia: true`.
4. `composeRunCompletionFrame` reported:

   > The run you queued finished successfully in 3.0s, and produced no image or video output. … **If this workflow was meant to save a file, no output node produced one**; if its results are text or other non-media outputs, read them from the run's history entry.

   with `metadata: [{outputs: "none", reason: "no_media"}]`.

**Step 4 is the load-bearing part of this diagnosis, and the reason I treated this as a bug rather than a missing feature.** A `SaveAudio` node *did* produce a file. The panel told the agent it did not, in a sentence the agent has every reason to believe — the #710 honesty class, on the run-completion surface. A source comment at `comfyui-mcp-panel.js:36943` called the omission *"silence, not a false claim"* and deferred it to "its own change". The silence half was wrong; this is that change.

## The fix

Audio gets its own channel through the lifecycle, modelled directly on the `withheld` channel added in #1934:

- **Painted** in chat as a player (the panel already has `paintAudio`, audio persistence in `chat-media.js`, and the `panel_show_media` audio disclosure — the only missing wire was the collection).
- **Named** on the completion frame, with the file, the `get_image` arguments to fetch it, and an explicit "there is no way for you to hear it, so do not describe how it sounds".
- **Never attached.** The frame's `images` are inline image blocks; audio delivered as one is a broken picture plus a claim of a perception nobody had (#710).
- **Not folded into `withheld`**, whose note says the outputs "exceed the completion frame's media budget" — which is not why audio is held back.

Covered on every surface a completion can arrive from: the live `executed` path, `execution_success` (both the panel-queued media-less branch and the late-`onQueued`-receipt branch), and the `/history` reconcile — which returns audio on its own key so a recovered audio run is not handed to the agent as an image either.

## Where to look hardest

Named deliberately, because a reviewer told where the load-bearing claim is finds the defect and one handed a diff hunts for it:

- **`comfyui-mcp-panel.js` — kind is settled by PROVENANCE, not by `isAudioOutput`.** Entries from the `audio` bag are painted as audio without consulting the extension regex, because the bag already said what they are. Asking the regex would send a `.flac` variant a future build emits — anything not in `mp3|wav|flac|ogg|oga|opus|m4a|aac` — into `paintImage`. The existing `isAudioOutput` branch is retained and still earns its keep for a descriptor **misfiled** under one of the three image keys.
- **`run-completion.js` — audio does NOT arm the orphan timer, and is NOT counted by `hasBufferedMedia`.** Same rule as `withheld`, same reason: `flush()` deletes an images/videos-empty buffer without emitting, so arming would drop the refs before `execution_success` can report them. This is why the audio is read out of the buffer *before* `flush(k)` in `onExecutionSuccess`, next to `withheld`.
- **The `#356 Bug 2` media-less branch is narrowed, not removed.** A run that genuinely produced nothing still gets its "produced no image or video output" report — pinned by a test, since silently widening the audio note over that branch would re-open the indefinite stall #356 closed.
- **`run-completion-delivery.js` is a two-line pass-through** and would have been invisible to a helper-level test. Test section 6 drives the shipped `tracker + createRunCompletionFlushHandler` composition and asserts on the frame the transport sees.
- **The audio note claims nothing about the chat.** Whether a player was painted depends on the `#2034` chat-media setting and on the surface (a `/history` reconcile paints nothing), so "its player is in the chat" would be false on paths this same note serves.

## Two #2034 assertions changed — read this rather than assuming a loosened test

`chat-media-inflow.test.mjs` asserted `buffered.length === 0` for audio, as a *proxy* for "audio never joins the agent's inline-image delivery". The tracker is called now, so the proxy no longer holds. It is replaced with the invariant it stood for — `output.images` stays empty — plus an assertion that the ref arrives on `output.audio`. That is a tightening, not a loosening: the #710 rule is now asserted directly instead of through a call count.

## Verification

**Unit:** `npm run test:unit` — 7878 tests, 0 fail (was 7862 before the 16 new ones). `npx tsc --noEmit` clean. `check-panel-scope`, the tool-vocabulary gate, i18n and changelog gates all pass.

**Mutation — every wire broken individually, all confirmed RED**, tree restored byte-clean afterwards (`git status --porcelain` empty). A green suite proves a mechanism exists, never that production reaches it:

| # | wire broken | result |
|---|---|---|
| 1 | `node-output-media` stops harvesting the `audio` bag | RED (4 tests) |
| 2 | the panel stops painting/collecting the bag | RED (2) |
| 3 | the panel stops handing `audio` to the tracker | RED (2) |
| 4 | `onExecuted` stops buffering audio | RED (4) |
| 5 | the media-less flush stops carrying audio | RED (3) |
| 6 | the `/history` reconcile flush stops carrying audio | RED (1) |
| 7 | `parseHistoryEntry` drops audio | RED (2) |
| 8 | the delivery handler stops passing `audio` to the composer | RED (1) |
| 9 | the composer never pushes the audio note | RED (3) |

Mutations 2, 3 and 8 are the ones that matter most: they are the call-site and wiring changes a helper-level test cannot see.

**Browser gate: NOT RUN — no ComfyUI is listening on this machine.** Stating what I tried rather than implying coverage I do not have. `curl /system_stats` returned no response on 8188, 8189, 8201, 8202, 8211, 8288, 8000 and 8264; `netstat` shows nothing else in the 8xxx range (8264 is not HTTP, 8384 is Syncthing). The ap-slot route — `git checkout --detach <sha>` in `comfy-instances/ap-N/custom_nodes/comfyui-mcp-panel`, which is a real git worktree — is the route that works here, and it is unavailable because the slots are not running; no instance was allocated for this run and I did not start one.

What that leaves uncovered is narrow and specific: whether the `<audio>` element *renders and decodes* in the sidebar. What is covered without a browser is that production reaches the painter — test section 2 instantiates the **shipped** `onExecuted` closure out of `web/js/comfyui-mcp-panel.js` (the same extraction the #2034 gate tests use) and asserts `paintAudio` receives the `/view` URL and `paintImage` receives nothing. The painter itself is unchanged code that #710 already shipped and that `panel_show_media` drives today.


artokun:
Third commit, added after the PR body was written — flagging it rather than letting a reviewer find it in the diff.

**`3608a3cd` — keep the #1286 preview sweep gated on IMAGE media.**

The first commit had widened `if (media.length)` to `if (media.length || audioBag.length)` around `stripMisattachedExecutionPreviews`, on the reasoning that an `executed` frame is an `executed` frame. Reading what that sweep does made it the wrong call: with `preferNodeId` set it re-homes other nodes' stored image outputs onto that node (`execution-preview-attach.js:387-399`) and records it in the #1374 ownership ledger. Handing it a SaveAudio node means attributing image-preview state to a node that cannot host one.

An audio-only `executed` returned before that line existed on `main`, so leaving the guard alone makes this branch byte-identical there. It is a narrowing, and it is the kind of thing a browser gate would have caught and I cannot run one here — so it was reverted on inspection rather than shipped on optimism.

The mutation table in the description was re-run on `3608a3cd` after this change: all 9 wires still RED, tree restored clean. Unit suite 7879 tests / 0 fail, `tsc --noEmit` clean.

artokun:
## Gate round 1 — NO-SHIP, two P1s, both real, both fixed in `7bff41a1`

Recording the verdict rather than only the outcome, since the findings are the most useful thing on this PR.

`codex-gate.mjs` on `3608a3cd` returned **NO-SHIP**:

> **[P1]** `run-completion.js:730` — `flush()` drops audio-only buffers because it checks only images/videos — with an audio `onExecuted`, missed start/liveness events, then `onExecutingNull()` before success, the agent receives `noMedia` and the false "no output node produced one" report.
>
> **[P1]** `run-completion.js:762` — held completion persistence omits audio — if an image+audio run finishes before delayed prompt identity and the tab reloads, the persisted replay restores the image but drops the audio filename.
>
> Direct repro confirmed the first case; focused and adjacent suites pass but do not cover these states.

Both are correct, and both are **the same defect this PR exists to fix, re-entered through a lifecycle path the first commit did not cover**. My 15 tests were green through both. Reproduced the first before touching anything:

```
onQueued(P); onExecuted(P, {audio:[…flac]}); onExecutingNull(); onExecutionSuccess(P)
→ flushes: 1 · audio on the flush: null · noMedia: true
```

**P1a.** `onExecutingNull()` (ComfyUI's queue-idle signal) swept every inactive buffer through `flush()`. For an images/videos-empty batch `flush()` emits nothing and **deletes** the buffer — which is exactly where `onExecutionSuccess` reads the audio from. A run whose `execution_start` and `executing(node)` frames were dropped is not `active`, so it gets swept.

Fixed at the **call site**, not in `flush()`, because only the call site knows the difference: `onExecutingNull` exists to salvage a stranded batch, and a buffer with nothing to attach has nothing to salvage. Retention is bounded four ways — `onExecuted` puts the key in `pending`, so reconcile or the give-up eviction retires it; the next run's sequential flush in `onExecutionStart` clears it (and *there* the refs are recoverable, since that run reconciles from `/history`); and `execution_success` itself deletes it through `flush()`'s empty-batch path. A companion test pins that the sweep still salvages a real image batch, so the guard cannot quietly disarm what it narrows.

This closes the **identical pre-existing hole for #1934's `withheld` count**, which was destroyed on the same path.

**P1b.** `terminalCompletionMetadataSnapshot` / `restoreTerminalCompletion` omitted `audio`. Restored only from a genuine array, so a persisted record written by a build without the key does not gain an `audio: []` the composer then has to reason about — pinned by a test.

## Updated verification

- Unit: **7883 tests, 0 fail**. `tsc --noEmit` clean. CI (Lint pack) green on `7bff41a1`.
- Mutation table is now **12 wires, every one RED**, spec green again after restore:

| # | wire broken | result |
|---|---|---|
| 1 | `node-output-media` stops harvesting the `audio` bag | RED (6) |
| 2 | the panel stops painting/collecting the bag | RED (2) |
| 3 | the panel stops handing `audio` to the tracker | RED (2) |
| 4 | `onExecuted` stops buffering audio | RED (6) |
| 5 | the media-less flush stops carrying audio | RED (4) |
| 6 | the `/history` reconcile flush stops carrying audio | RED (1) |
| 7 | `parseHistoryEntry` drops audio | RED (2) |
| 8 | the delivery handler stops passing `audio` | RED (1) |
| 9 | the composer never pushes the audio note | RED (3) |
| 10 | **gate P1a** — queue-idle destroys the buffer again | RED (1) |
| 11 | **gate P1b** — the held-completion snapshot drops audio | RED (1) |
| 12 | **gate P1b** — `restoreTerminalCompletion` drops audio | RED (1) |

Round 2 of the gate is running against `7bff41a1`, the SHA that would merge.

PR #2131, opened 2026-08-31
fix(#2128): report a Save3DAdvanced run instead of claiming nothing was saved

### LINKED ISSUE #2128: Run-completion says "no saved output node ran" when a non-image save node (Save3DAdvanced) wrote a real file

A Pixal3D/Trellis2 image-to-3D workflow ending in `Save3DAdvanced` completed successfully and
wrote a real 16 MB `.glb` to `output/3D/`. The run-completion message told the agent the
opposite:

> Run finished, but no saved output node ran — these 7 images are previews (temporary, not a
> final file). Add a SaveImage node to persist the result, or treat the preview as the result
> if that's intended.

All three claims are wrong: a saved output node did run, a final file was written, and adding
a `SaveImage` node could not persist it anyway because the payload is a `FILE_3D_GLB`, not an
image.

**Root cause** — `web/js/lib/run-completion-frame.js`, `buildStillsSegment()` (~line 425-470).
It classifies `bufImages` by ComfyUI's `type` field (`"output"` = final, else preview), then:

```js
} else {
  const previewClause = ... `these ${previewCount} images are previews (temporary, not a final file)`;
  note = `Run finished, but no saved output node ran — ${previewClause}. ` + ...
```

The image classification itself is CORRECT — all 7 attached images genuinely are `type:"temp"`
previews (the PreviewImage taps for base colour, normal, AO, roughness, metallic, UV atlas).
The defect is inferring "no saved output node ran" from the image set alone. `buildStillsSegment`
only ever sees images; a `.glb` never enters `bufImages`, so `finals` is empty and it falls into
the no-output branch. Grepping that file for `glb|gltf|mesh|model_3d` returns nothing — it has
no visibility into non-image outputs at all.

Note this is inconsistent with `panel_query_graph`, which correctly reports `"is_output": true`
on the same `Save3DAdvanced` node. The graph model knows; the completion message does not.

### Steps to reproduce

1. Open any workflow whose terminal output node is `Save3DAdvanced` (e.g. the bundled
   `3d_pixal3d_trellis2_image_to_model` template), with `PreviewImage` taps upstream.
2. `panel_run` (with or without `to_node_id` pointing at the `Save3DAdvanced` node).
3. Let it finish. Observe the completion message.
4. Check `output/3D/` — the `.glb` is there.

Reproduced twice on the current version:

| prompt_id | duration | file written |
|---|---|---|
| `bf7873d8-a7b1-4c06-be50-45463604ec2f` | 9m 28s | `PROP_crate_00001.glb`, 15,659,544 bytes |
| `c4fadda6-4565-41cc-aa2a-931589f34c5a` | 15m 9s | `PROP_crate_00002.glb`, 16,027,192 bytes |

### Exact error (scrubbed)

Not an exception — wrong output. Verbatim message, both runs:

```
Run finished, but no saved output node ran — these 7 images are previews (temporary, not a
final file). Add a SaveImage node to persist the result, or treat the preview as the result
if that's intended.
```

### Impact

Low severity, but it instructs a specific wrong action. An agent that believes it either
re-runs a 9-15 minute job or wires in a `SaveImage` node that cannot help. I avoided both only
by checking the filesystem directly.

### Fix

applied locally: no   upstream-only: yes

Not safely patchable at the call site: `buildStillsSegment` has no channel through which to
learn that a non-image output was produced, so a local edit could only suppress the clause
unconditionally, which would break the genuine preview-only case.

Precise change needed — either:

1. Pass a non-image-output signal into `buildStillsSegment` (a count or flag derived from the
   prompt's `/history` outputs, where a 3D save node's entry appears under a non-`images` key),
   and skip the "no saved output node ran" branch when it is non-zero; or
2. Have the caller decide the clause, since it can see the whole history record, and let
   `buildStillsSegment` report only on images.

Either way the "Add a SaveImage node" advice should be conditional on the graph's terminal
output actually being an image node — it is misleading for 3D/audio/video save nodes.

### Environment

- OS: Windows 11 (build 26200)
- ComfyUI: 0.34.2
- ComfyUI FRONTEND: 1.49.6
- Python 3.13.12 / PyTorch 2.10.0+cu130
- GPU: NVIDIA GeForce RTX 5080, 15.9 GB VRAM
- comfyui-mcp: 0.52.165
- comfyui-mcp-panel: 0.15.149

---
Filed from the ComfyUI panel by **claude** · model `opus[1m]` · effort `high` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #1934

artokun:
Fixes #2128

## What was wrong

A Pixal3D/Trellis2 image-to-3D workflow ending in `Save3DAdvanced` ran for 9m28s, wrote a real
15,659,544-byte `PROP_crate_00001.glb` to `output/3D/`, and the completion frame told the agent:

> Run finished, but no saved output node ran — these 7 images are previews (temporary, not a final
> file). Add a SaveImage node to persist the result, or treat the preview as the result if that's
> intended.

All three claims are false, and the remedy cannot work — `SaveImage` takes an IMAGE and the payload
is a `FILE_3D_GLB`. The reporter avoided a 15-minute re-run only by checking the filesystem.

## Mechanism, from ComfyUI's source

ComfyUI emits 3D outputs in **two unrelated shapes**, neither of which anything read:

| node | outputs bag |
|---|---|
| `SaveGLB` (`comfy_extras/nodes_save_3d.py:605`) | `ui={"3d": [{filename, subfolder, type:"output"}]}` |
| `Save3DAdvanced` / `SaveGaussianSplat` / `SavePointCloud` / `Preview3D(Advanced)` | `{"result": ["3D/PROP_crate_00001.glb", camera_info, model_3d_info]}` |

The second shape is the reported one. All three Save nodes delegate to `execute_save_3d_advanced`
(`nodes_save_3d.py:889`), which serialises through `UI.PreviewUI3DAdvanced.as_dict()`
(`comfy_api/latest/_ui.py:459`) — so `model_file` is a **bare path string** sharing an array with a
camera dict, not a `/view` descriptor.

Neither `3d` nor `result` matches the three literal media keys or the widened
`/(?:images|gifs|videos)$/` scan, so `collectNodeOutputMedia` returned nothing and the `.glb` was
invisible to the whole completion path. **Two** symptoms follow from that one hole:

1. **With `PreviewImage` taps upstream** (the reported run) — `bufImages` holds the 7 `type:"temp"`
   frames, so `buildStillsSegment` takes its no-finals branch and emits the sentence above.
2. **Without them** — the batch is empty, so it lands in the media-less branch and reports "produced
   no image or video output", which is #2126 with a different file extension.

## The fix

3D gets its own channel, modelled on the #2126 audio channel: **named** on the completion frame,
**never attached** (a mesh delivered as an inline image block is a broken picture plus a perception
nobody had — #710), and never folded into `withheld` (that note says the outputs exceed the media
budget, which is not why a mesh is held back). Covered on every surface a completion arrives from:
the live `executed` path, `execution_success` including the media-less and late-receipt branches,
queue-idle, the held-completion replay across a reload, and the `/history` reconcile.

The second half is the issue's own **option 2**: `buildStillsSegment` only ever sees images, so
"no saved output node ran" was a claim about the whole RUN inferred from a strict subset of its
outputs. It is now told which non-image outputs were saved and names them instead.

## Please look hardest at these three

1. **`model3dRefFromResultEntry` reports `type: "output"` for a `result` path string.** Correct for
   the three Save nodes (`_save_file3d_to_output` uses `get_output_directory()` unconditionally) and
   for `Preview3D` whenever it materialises a `File3D`. It is **wrong for one shape**: `Preview3D`
   handed a literal path string passes it through unchanged, so the file may sit under `input/`. The
   two `PreviewUI3D*` shapes are structurally indistinguishable, so this cannot be narrowed from the
   bag alone. Chosen deliberately — it costs a `get_image` call that 404s and can be retried, against
   the status quo of telling the agent to re-run a 15-minute job. Documented at the call site.
2. **Admission from the generic `result` key.** Only a STRING with a 3D extension is admitted, so the
   camera dict, the model-info list, and `PreviewUI3D`'s `"temp/bg_<hex>.png"` are all passed over. A
   `File3DAny` in some format outside the list is under-reported — the fail-safe direction, since a
   missed output is today's behaviour and a false positive would invent an output node.
3. **The `buildStillsSegment` narrowing must not become a blanket suppression.** A genuinely
   preview-only run still gets the original "Add a SaveImage node" advice; pinned by the test
   `a genuinely preview-only run keeps the original advice`.

## It also fixes the #2126 sibling

#2126 gave audio its own channel but only covered the **media-less** path. A `SaveAudio` run **with
preview taps** still produced the reported sentence, because `buildStillsSegment` inferred it from
the image set alone. Reproduced by execution against `origin/main` (`71770ca7`) and pinned by
`the SIBLING defect — audio plus preview taps stopped claiming nothing ran`.

## Verification

**Reproduced by execution before the fix** — both symptoms, verbatim against the issue text:

```
=== reported case: 7 temp previews, .glb written ===
Run finished, but no saved output node ran — these 7 images are previews (temporary, not a
final file). Add a SaveImage node to persist the result, or treat the preview as the result...

=== sibling: SaveAudio + temp previews (post-#2126) ===
Run finished, but no saved output node ran — these 7 images are previews...
```

After, the same input produces:

```
Run finished — these 7 images are previews (temporary, not a final file). But the run DID save a
final result: 1 3D model output (named below in this completion). Do NOT add a SaveImage node —
the saved output is not an image, so SaveImage cannot persist it.

🧊 Also produced 1 3D model output: `PROP_crate_00001.glb`. This is the run's saved result. ...
To get the file itself, call get_image with filename "PROP_crate_00001.glb", type "output",
subfolder "3D" — a 3D model is SAVED TO DISK rather than returned to you inline...
```

- **28 regression tests** in `browser_tests/unit/run-completion-3d.test.mjs`.
- **Mutation table: 22/22 wires RED.** Each wire deleted one at a time, restored from the committed
  blob (never `git checkout`, which would delete the fix), run under `--test-timeout=60000` so a hang
  counts as red. Covers both collection shapes, the Windows separator split, the extension allowlist,
  the panel call-site guard AND its hand-off, the tracker's buffer / flush / media-less branch /
  pre-flush read / reconcile / held snapshot / restore, the `/history` parse, the delivery
  pass-through, and all five composer wires.
- The table's **first run found a real defect in my own change**: the new `3d`/`result` skip in the
  withheld scan was dead code (neither key matches `MEDIA_KEY_SUFFIX`, so no test could kill it).
  Rather than delete it alone — the `audio` skip shipped in #2126 is dead for the identical reason —
  both are folded into one condition with a comment saying plainly what they are. Second commit;
  behaviour byte-identical.
- Full unit suite **7910 pass / 0 fail** (7883 → 7911, +28). `typecheck`, `check:scope` and
  `check:changelog` all clean.

## What I did NOT do

- **No browser verification.** No ComfyUI is listening on this host, so the Playwright suite cannot
  run (it needs one on :8188 and does not start one), and `/history` for the two reported prompt_ids
  was unavailable. Ground truth came from ComfyUI's own source instead. The change is agent-facing
  text on the completion frame rather than a rendered surface — but I am not implying coverage I do
  not have.
- **No chat card for a 3D output.** A mesh has no viewer widget in the panel and the status quo
  paints nothing, so adding one is a separate change. The user still sees ComfyUI's own 3D viewer on
  the node itself.
- **`web/changelog.json` untouched** — it deliberately excludes `[Unreleased]` and is regenerated by
  `set-version.mjs` at release.
- **`metadata` for a run with preview taps stays `[]`**, matching the shipped per-final-still
  convention (and audio's behaviour on that path). The note carries the report there.


artokun:
## Merge-gate response — rounds 1 and 2

Gate: `codex-gate.mjs`, three rounds so far. Round 1 on `6ada6972`, round 2 on `a0e7c2af`.
Both rounds found something; here is what I did with each finding, including the two I am
**not** acting on and why.

> Gate-environment note, since it affects how much weight to give the first attempt: the
> first two invocations returned `NO-SHIP / Review blocked: CreateProcessAsUserW failed: 5`
> with **no code verdict asserted** — that is an INDETERMINATE, not a rejection, and I did
> not treat it as one. Root cause: codex runs its shell commands under a restricted
> sandbox token, and Git-Bash's MSYS binaries (`tail.exe` and friends) die under it with
> `CreateFileMapping … Win32 error 5`. Re-running the gate from PowerShell, whose PATH has
> `Git\cmd` but not `Git\usr\bin`, produced real verdicts. Worth knowing for other agents
> on this machine.

### ✅ Round 1 P1 — a TEMP 3D preview reported as a saved output — FIXED (`a0e7c2af`)

Correct, and my source survey was incomplete. I checked `Preview3D` (which writes to
`output/`) and missed three siblings emitting the identical `{"result": [...]}` shape into
**`temp/`**:

| node | filename | dir |
|---|---|---|
| `Preview3DAdvanced` (`nodes_load_3d.py:174`) | `preview3d_advanced_<32 hex>.<ext>` | **temp/** |
| `PreviewGaussianSplat` (`:244`) | `preview_splat_<32 hex>.<ext>` | **temp/** |
| `PreviewPointCloud` (`:305`) | `preview_pointcloud_<32 hex>.<ext>` | **temp/** |

Reporting one of those as a saved final would have been this defect one node over — the
file is swept with `temp/`, and the `get_image` call my note spells out asks for
`type:"output"` and 404s.

Fixed with a discriminator from ComfyUI's own naming rather than a guess:
`_save_file3d_to_output` builds `f"{filename}_{counter:05}.{ext}"`, so a saved file always
ends in `_<5+ digits>.<ext>`, which a 32-hex uuid cannot produce (the hex is one unbroken
run with no underscore before the dot). Pinned by
`a TEMP 3D preview is never reported as the run's saved result`.

### ✅ Round 2 P1 — `Preview3D` passthrough could still be misreported — FIXED (`cb06761f`)

Also correct, and it is the residual the round-1 fix could not close: a user-supplied path
into `Preview3D` **can** carry a save counter (re-previewing a `PROP_crate_00001.glb` from
an earlier run is an ordinary thing to do), and `PreviewUI3D.as_dict()` /
`PreviewUI3DAdvanced.as_dict()` are structurally identical, so no rule over the bag can
separate them.

So the fix stops asserting what it cannot know. "the run DID save a final result" and
"This is the run's saved result" are gone; it now says the run **produced** the output and
names it. That is true of every shape, and all three reported claims are still retired —
it does not say no saved output node ran, it names the `.glb`, and it still says
explicitly not to add a `SaveImage` node. Pinned by `the note never claims THIS RUN saved
the file`, which asserts the **absence** of all three phrasings as well as the presence of
what replaced them.

### ❌ Round 2 P1 — `File3DAny` formats outside the allowlist are dropped — NOT CHANGING

Accurate as a description, but it is a **deliberate fail-safe**, and acting on it would
make the PR worse.

The stated consequence — "the run is still reported as having no saved output" — is
today's behaviour for *every* 3D output, so it is the status quo rather than a regression;
this PR strictly reduces that set. The alternative is admitting any extension that carries
a save counter, which immediately re-admits `3D/render_00001.png` as a "3D model" — the
exact false positive that `the result siblings are not mistaken for files` pins as wrong.
Trading a silent miss for a fabricated output node is the wrong direction for a defect
whose whole subject is the panel asserting things that are not so.

Note also that the restriction applies only to the bare-`result` path. `SaveGLB`'s `3d`
descriptors are admitted with **no extension filter at all**, because there the bag itself
says what the entries are — so an exotic `File3DAny` format saved through `SaveGLB` is
reported regardless.

### ❌ Round 2 P1 — `terminalNoMediaSuccess` not persisted / `restoreTerminalCompletion` rejects model-only — NOT IN SCOPE

Real, but **pre-existing and exactly symmetric with the shipped `audio` channel** — not
introduced by this PR. Verified by execution against this branch rather than argued:

```
audio-only  (shipped #2126)    held-snapshot entries: 0
3d-only     (this PR)          held-snapshot entries: 0
image+audio (control)          held-snapshot entries: 1

restoreTerminalCompletion(audio-only payload)  -> false
restoreTerminalCompletion(3d-only payload)     -> false
```

The cause is upstream of both: `flush()` early-returns at `run-completion.js:741` when
`images` and `videos` are empty, *before* `terminalNoKeyCompletion.set`, so a non-image-only
run never reaches the held-completion path at all. That contract is load-bearing for
`withheld` (#1934) and `audio` (#2126) — it is what stops a 768-image dump riding the
frame — so changing it is a change to those two channels, not to this one.

Closing it properly means letting a non-attachable batch be held and replayed, and that
belongs in its own change against all three channels at once, with #1934's and #2126's
tests as the guard. Happy to open it as a follow-up; doing it inside a P2 3D fix would
mean touching two shipped channels with no issue describing the failure.

### Evidence recap

- Mutation table re-run **whole** after each re-fix (a narrow re-fix silently disabling an
  earlier wire is the failure mode): **23/23 wires RED**, each restored from an on-disk
  snapshot rather than `git checkout`, under `--test-timeout=60000` so a hang counts red.
- The round-1 re-fix made two earlier wires stop being pinned and the re-run caught both —
  the extension allowlist went GREEN, and the cross-node merge assertion went vacuous.
  Both were repaired rather than accepted.
- 30 regression tests; full suite **7912 pass / 0 fail**; `typecheck`, `check:scope`,
  `check:changelog` clean.

artokun:
## Merge-gate rounds 3–5, and the decision to merge

Five `codex-gate.mjs` rounds returned real verdicts. Rounds 1–4 each found something
genuine and each is fixed — the change is materially better for them. Round 5 introduced
no new information, and this is where I am stopping, per the brief's "do not grind past
~3 rounds" cap. Recording the reasoning so the call is auditable rather than silent.

### Fixed across the rounds

| round | finding | outcome |
|---|---|---|
| 1 | `Preview3DAdvanced` / `PreviewGaussianSplat` / `PreviewPointCloud` write to **temp/** and emit the identical `result` shape — reported as saved outputs | **fixed** `a0e7c2af` — excluded via ComfyUI's own save-counter naming |
| 2 | `Preview3D` passthrough could still be called a saved result | **fixed** `cb06761f` — the note no longer claims *this run* saved anything |
| 3 | `type:"output"` is inferred; get_image args could point at the wrong root | **fixed** `3977b94a`, then superseded |
| 4 | the round-3 hedge never fires when a same-named file exists under `output/` — the wrong mesh is fetched silently | **fixed** `969f57ae` — an inferred ref now gets **no guessed argument list at all** |

Round 4 is the substantive one: `result`-derived refs are marked `typeInferred` and the
note sends the agent to `get_history`, which carries ComfyUI's own reference verbatim,
while `SaveGLB`'s `3d` descriptors — which *state* their type — still get the full
get_image argument list. Advice is now proportional to evidence, so the completion cannot
point at the wrong file.

### Round 5 — no new information

**(a) `terminalNoMediaSuccess` not persisted.** Re-raised verbatim from round 2, which I
answered [above](#issuecomment-5478216238) with execution evidence: it is pre-existing and
*exactly symmetric* with the shipped `audio` channel, because `flush()` early-returns at
`run-completion.js:741` before `terminalNoKeyCompletion.set` for any non-attachable batch.

```
audio-only  (shipped #2126)    held-snapshot entries: 0
3d-only     (this PR)          held-snapshot entries: 0
image+audio (control)          held-snapshot entries: 1
restoreTerminalCompletion(audio-only) -> false      (3d-only) -> false
```

That contract is load-bearing for #1934 and #2126 — it is what stops a 768-image dump
riding the frame — so changing it is a change to those two channels. Out of scope for a P2
3D fix; I will open it as a follow-up against all three channels with their tests as guard.

**(b) `Preview3D` previewing `input/mesh_00001.glb`.** The fifth statement of the same
inference, with a new consequence each round. It is genuinely unclosable from the outputs
bag: `PreviewUI3D.as_dict()` and `PreviewUI3DAdvanced.as_dict()` are structurally identical
to what `Save3DAdvanced` emits, and narrowing by root instead would break a real save —
`get_save_image_path` permits a `filename_prefix` of `input/foo`, which legally writes to
`output/input/foo_00001.glb`.

After the round-2 and round-4 fixes its remaining consequence is small and bounded: for a
`Preview3D` node handed a literal path whose basename happens to match `*_NNNNN.<3d-ext>`,
the completion says the run *produced* a 3D output and points at `get_history` — where the
agent reads the truth — instead of advising a `SaveImage` node, which for a 3D preview
workflow was not correct advice either. No file is claimed to have been saved by the run,
and no guessed fetch argument is handed over.

### Why merge

The defect being fixed is unconditional and reproduced: **every** workflow ending in
`Save3DAdvanced` is currently told "no saved output node ran — Add a SaveImage node" while
a 16 MB `.glb` sits in `output/3D/`, costing a 9–15 minute re-run. The residual is a
contrived configuration whose worst outcome is differently-phrased advice plus a pointer to
the authoritative record. Shipping the fix is strictly better than not shipping it, and the
residual is documented in the code, pinned by tests, and stated here.

Evidence standing behind the merge, independent of the gate:

- **25/25 mutation wires RED**, table re-run *whole* after every re-fix — which is how the
  two wires that stopped being pinned by the round-1 fix were caught (the extension
  allowlist went GREEN, and the cross-node merge assertion went vacuous). Both repaired.
- Both reported symptoms **reproduced by execution** before the fix, verbatim, and re-run
  after.
- 31 regression tests; full suite **7913 pass / 0 fail**; `typecheck`, `check:scope`,
  `check:changelog` clean; CI green.

### Known follow-ups (not done here)

1. Persist and replay held completions for non-attachable batches — `withheld` (#1934),
   `audio` (#2126) and `models3d` together.
2. Reject drive-lettered / absolute `result` paths, which `get_save_image_path` can never
   produce. Strictly correct but marginal; left out rather than re-open the gate cycle for it.
3. No browser verification: no ComfyUI is listening on this host, so the Playwright suite
   cannot run and `/history` for the reported prompt_ids was unavailable. Ground truth came
   from ComfyUI's source. Stated plainly rather than implied.

Issue #1580, opened 2026-08-21
Save guard refuses a save whose only drift is presentation-only (node geometry) — the guard discards its own presentationOnly signal


artokun:
### What happened

`saveWouldPersistStaleSnapshot` (shipped in #1567 / `03da405d`, fixing #1563 and #1564) refuses a save when two conjuncts hold: upstream positively reports it skipped the pre-save capture, **and** the live root does not reproduce the state about to be written.

The second conjunct is read as `graphRootReproducesStateContent(...)?.proven !== true`. That expression is also satisfied by a difference the panel's **own classifier already vouches for as losing nothing authored** — `presentationOnly: true`. So a canvas that differs from its snapshot only by node geometry refuses the save.

The returned object carries `presentationOnly` and the guard discards it.

`web/js/comfyui-mcp-panel.js:3787`

### Measured

Driving the real conjunct-2 expression against `graph-binding.js` unmodified, four canvas/snapshot pairs:

| Case | `surfaces` | `proven` | `presentationOnly` | refuses? |
| --- | --- | --- | --- | --- |
| canvas == snapshot | `[]` | `true` | false | no ✅ |
| **node dragged (`pos` only)** | `["nodes"]` | `false` | **`true`** | **YES** ⚠️ |
| widget value changed | `["nodes"]` | `false` | false | yes ✅ |
| group added | `["groups"]` | `false` | false | yes ✅ |

Rows 3 and 4 are the defect #1567 exists to stop and must refuse. Row 2 is a save that was fine.

### Why it is reachable

The suppression window `#1567`'s own browser spec uses is `_restoringState`, held for the whole of an undo/redo restore (`await loadGraphData`) — and what an undo most often restores is node geometry. So "a suppression window is open" and "the only drift is `pos`" are correlated by the undo path rather than independent. The window is wide on a large graph; #1564's was 83 nodes.

User-visible effect: `workflowService.saveWorkflow` **throws** — on autosave and Ctrl+S too, since the guard wraps the whole funnel — with a message saying *"the file would be missing changes the canvas already has"* and directing the user to reload the tab and re-open the workflow. True of a node position, but it reads as data loss and the named recovery is far heavier than the drift.

### Severity: minor

[...]

Issue #878, opened 2026-08-09
An in-place save writes stale bytes: values a node set are lost on disk


artokun:
Found while fixing #874, and it is the same root cause writing to **disk** rather than to the canvas — so strictly worse.

## Measured

Fresh canvas, one `EmptyLatentImage`, saved once so it has a real path. Then a value changed the way a **node** does — directly, no user input event — and `panel_save_workflow({})` (in-place, no name):

```
live canvas width : 1337
changeTracker     : [512, 512, 1]
ON DISK after save: [512, 512, 1]     <- read back through ComfyUI's /userdata API
```

The user's real file now holds the default. The 1337 is gone.

## Why

`workflow-save.js` says it plainly at line ~394:

> EVERY save route here ultimately persists `wf.activeState`: the in-place branch writes it through `saveWorkflow`, and both copy routes read it in `saveAs`.

And ComfyUI's ChangeTracker fills `activeState` on **user input**. So any value a node wrote itself is missing from it.

The **copy** routes already handle this. `saveAs` is preceded by a guarded flush, with the reasoning spelled out at ~913:

```js
// #708 — REFRESH THE SOURCE, NEVER THE COPY. ... that state can legitimately lag the
// canvas by a capture (ComfyUI snapshots on user-input events; bridge edits are
// snapshotted after the fact). Flushing the canvas into the tracker before the copy
// is taken is what keeps a save current.
if (normalizeCanvasBinding(canvasBinding, wf) === "bound") {
  try { wf?.changeTracker?.prepareForSave?.(); } catch (err) { /* refuse */ }
}
```

`saveInPlace` (~1325) has **no equivalent**:

```js
async function saveInPlace(svc, wf, { readSaveFailureCause, path } = {}) {
  try {
    if (typeof svc.saveWorkflow === "function") return await svc.saveWorkflow(wf);
    if (typeof wf.save === "function") return await wf.save();
  ...
```

So the route that overwrites the user's existing file is the one route that does not refresh first.

## Reachability

The panel's command dispatch calls `checkState()` **after** each completed command, so a panel edit followed by a save is fine — the tracker caught up in between. The exposed window is a value written by a **node** with no panel command after it:

- an `ImpactWildcardEncode` populate,
- a `control_after_generate` roll,
- a subgraph's promoted widgets,
- anything an extension writes on `graphChanged`.

…then a save. Which is an ordinary sequence: run once, then save.

## Severity

[...]

PR #2137, opened 2026-08-31
fix(#2133): refuse a save whose snapshot upstream never refreshed

### LINKED ISSUE #2133: Workflow save reports success but post-restart reload loses newly added nodes
What happened: a saved workflow was extended with YTDLLinksInput, YTDLDownloader, MarkdownNote and a group using panel tools. panel_save_workflow returned saved:true for VR180 Restoration - 1s Trim Proof. After a ComfyUI restart, panel_graph_outline of that same workflow reported only the original 7 nodes and 2 groups; the 3 added nodes and group were absent. Steps: open saved workflow; add nodes/group; save successfully; restart; inspect same workflow. Evidence: before save 10 nodes and 3 groups including ids 8,9,10 and group 00 Optional YouTube Source Staging; save reply routing key wf:workflows/VR180 Restoration - 1s Trim Proof.json; after restart 7 nodes, 2 groups. This is workflow data loss after a reported successful save. Fix: upstream-only; investigate panel save persistence/flush before restart, especially newly installed custom node types. Environment: Windows 11; ComfyUI 0.33.0; frontend 1.49.6; RTX 3090 24GB; comfyui-mcp 0.52.161; panel 0.15.147.

artokun:
Fixes #2133

## The report

A saved workflow was extended with `YTDLLinksInput`, `YTDLDownloader`, `MarkdownNote`
and a group through panel tools. `panel_graph_outline` read the canvas back as **10
nodes / 3 groups**. `panel_save_workflow` answered **`saved:true`**. After a ComfyUI
restart the same workflow came back as the original **7 nodes / 2 groups**.

`graph_outline` reads the **live** graph (`graph._nodes` / `graph._groups`), not the
tracker snapshot — so the report says the canvas had the work and the file did not.
That is the #1563 shape: ComfyUI serializes `changeTracker.activeState`, never the
graph on screen.

## Why the existing guard did not catch it

`saveWouldPersistStaleSnapshot` (panel#1563) refuses exactly this, on two conjuncts.
Conjunct 1 asks *"is one of ComfyUI's suppression flags set **right now**"*. Those
flags only describe an **open** window. A capture swallowed a moment earlier — the
deferred post-command snapshot exhausting its bounded ~1.4s retry chain, or
`prepareForSave()` returning without capturing — leaves every flag back to `false` by
the time the write funnels through the wrapper, so the guard allowed it.

Measured against the shipped bundle
(`comfyui_frontend_package-1.49.6/static/assets/settingStore-CwkLtSKP.js`),
`captureCanvasState` has **five** early returns, not the three the flag model reads:

```js
captureCanvasState(){
  let e=this._restoringState,t=this.changeCount>0;
  if(!$.graph||t||e||ChangeTracker.isLoadingGraph)return;
  if(!isActiveTracker(this)){reportInactiveTrackerCall(`captureCanvasState`,this.workflow.path);return}
  ...
}
```

and, in the same file,

```js
prepareForSave(){isActiveTracker(this)&&this.captureCanvasState()}
```

— so the refresh the panel requests before every overwrite (`workflow-save.js`, #878)
can be a **total no-op with nothing at all to observe**, and neither `!app.graph` nor
`!isActiveTracker(this)` is visible to `trackerCaptureSuppressed`. The panel source
asserted the opposite in three places ("the three fields `captureCanvasState`'s own
early return reads"); those comments are corrected here.

## The fix

Chasing each condition by name is a race against a frontend this panel does not own,
so conjunct 1 gains a second reading asked in **upstream's own vocabulary**:

> `ChangeTracker.graphEqual` — the comparator `captureCanvasState` itself uses to
> decide whether to replace `activeState` — still reports a difference **after** a
> refresh was requested.

A capture that actually ran would have replaced the snapshot. Every path that reaches
this wrapper asks for that refresh first (`workflowService.saveWorkflow` and
`saveWorkflowAs` both call `prepareForSave()` immediately before
`workflowStore.saveWorkflow`; autosave and Ctrl+S funnel through the service; the
panel's own in-place and copy routes call it directly), so a difference that *survives*
that request is not the "lag `prepareForSave` would resolve a microsecond later" that
conjunct 1 was scoped around — the microsecond has passed.

New: `trackerSnapshotBehindCanvas(tracker, rootGraph, state)` in
`web/js/lib/change-tracker-snapshot.js`, reached off `tracker.constructor` exactly the
way `isLoadingGraph` already is. **Positive evidence only** — an absent, throwing or
undecided comparator (`undefined`/`null`/`0`), an unserializable root, a root that
answers `null`, and an absent snapshot all answer `false`, because a guard that cannot
read its evidence must never invent a refusal (#1563 r2).

Conjunct 2 is **unchanged**, so the two questions still have to agree.

## Where the load-bearing claims are (please attack these)

1. **"A surviving `graphEqual` difference means the capture was swallowed."** This
   rests on every caller of `workflowStore.saveWorkflow` / `saveAs` having already
   requested a refresh. I enumerated them from the bundle (offsets 1502595–1574358);
   all of them route through `workflowService`. If some path calls the store directly
   with no `prepareForSave()`, this conjunct fires on ordinary lag there.
2. **Widening conjunct 1 gates autosave and Ctrl+S, not just `panel_save_workflow`.**
   A false positive is "the user cannot save". The mitigation is that upstream's
   comparator is *more tolerant* than the panel's shape comparison on exactly the axes
   that worried me — it does not compare `config` at all, and arrays inside `nodes`
   compare as **sets**, so a `widgets_values` reorder or a `config` drift answers
   "equal" and never reaches conjunct 2. Both are pinned as tests.
3. **The mechanism is inferred, not reproduced on the reporter's rig.** See below.

## Verification

- `npm run test:unit` — **7929 pass / 0 fail** (1 todo), including
  `check-panel-scope` (the gate that catches a new module binding resolving in the
  wrong scope).
- **Mutation-verified, both halves, after committing the fix:**
  - remove the new disjunct at the call site → `#2133 THE REPORTED CASE`,
    `#2133 the Save-As COPY point`, and the wiring pin go **red** (42/45); every
    pre-existing #1563 and #1580 test stays green, so the new tests are not just
    re-testing the old behaviour.
  - `return false` at the top of `trackerSnapshotBehindCanvas` → the same two
    behavioural tests plus two helper tests go **red** (68/72).
- The fixture uses a **transcription of the real `graphEqual`** (quoted from the
  bundle in the test), with its own guard test — so the cases are not being decided by
  a comparator ComfyUI does not have.
- Wiring is pinned at the **call site**, not just the helper: a source assertion that
  conjunct 1 is a disjunction, read off the same frozen snapshot and the same `state`
  the write serializes, and that conjunct 2 still gates the refusal after it.

## Not done, stated plainly

- **The Playwright suite was NOT run.** This change is panel-visible and
  `browser_tests/stale-snapshot-save-refused.spec.ts` is the spec that covers this
  funnel end to end, but no ComfyUI was allocated to this run. The only listener in
  range (`:8264`) belongs to another agent, and that spec calls `graph_clear` and
  writes into the instance's workflows dir — running it there would destroy live work.
  I have not implied coverage I do not have. **The single highest-value follow-up is
  to run that spec** (it already asserts the guard *clears* once the window closes,
  which is the regression this widening could plausibly cause).
- **The reporter's workflow does not exist on this machine** (no `VR180*` under any
  `user/default/workflows`), so the mechanism is derived from the shipped frontend
  source plus the panel's own control flow, not from the failed file. The alternative
  explanation I could not exclude is a save that wrote correctly and a *post*-restart
  read that lost the nodes; I rejected it because the missing **group** can only come
  from older serialized bytes, and because `graph_outline` reads the live graph.
- The refusal's recovery advice is rewritten: it no longer tells the user to re-open
  the workflow from disk, which would discard the canvas holding the only copy of the
  lost work. It now names the non-destructive repairs (wait; then any panel graph
  mutation or a click/drag, both of which re-run the capture).


artokun:
### Second commit — a cost regression the first one introduced

Reading upstream's comparator needs the live root serialized, and this guard sits on
ComfyUI's **whole save funnel**. Until #2133 the flag read short-circuited before any
of that work, so the first commit silently added a `rootGraph.serialize()` to every
autosave and Ctrl+S — and #581 records that serializing a large nested subgraph is slow
enough to matter here.

`a277aa09` splits the cheap half out (`trackerExposesCaptureComparator`) and asks it
first: with no flag set and no comparator to ask, neither reading of conjunct 1 can
establish anything, so it stops at **exactly the pre-#2133 cost**. A suppressed tracker
still reaches conjunct 2 with no comparator at all, which is the #1563 path unchanged
and is what the existing `#1563 WRAPPER: the save that loses the group is refused` test
covers.

The wiring pin was tightened with it, so the early-out cannot quietly become the whole
gate:

```
/if \(!suppressed && !trackerExposesCaptureComparator\(tracker\)\) return false;/
```

### Verification on `a277aa09`

- `npm run test:unit` — **7930 pass / 0 fail**, `[check-panel-scope] OK — every name in
  web/js/comfyui-mcp-panel.js resolves.`
- CI: **green** on `a277aa09` (run 33414219858).
- Mutation-verified again on the restructured code, after committing:
  - drop the comparator half of the early-out (`if (!suppressed) return false;`) →
    3 red / 42 green
  - drop the disjunct itself → 3 red / 42 green

  Every pre-existing #1563 / #1580 test stayed green under both, so the new tests are
  not standing in for the old behaviour.
- Codex merge gate: **PASS (exit 0, SHIP)** — run twice, once per head, because a
  verdict names a SHA: `86f44036`, then again on `a277aa09` after the restructure.

The "not done" section of the description still stands as written — in particular the
Playwright suite was **not** run, and `browser_tests/stale-snapshot-save-refused.spec.ts`
remains the highest-value follow-up.

Issue #2031, opened 2026-08-29
panel_set_widget silently reverts COMFY_DYNAMICCOMBO_V3 child values at queue time (mode.scale runs with the OLD value)


artokun:
### What happened / root cause

`panel_set_widget` on a dotted child of a `COMFY_DYNAMICCOMBO_V3` input reports success, and a follow-up `panel_query_graph` read confirms the new value — but the prompt that `panel_run` queues executes with the OLD value. The change is silently lost between the read-back and serialization.

Node: `MinimaxH3LatentUpscaler3D` (Deno MiniMax H3 pack)
Widget: `mode.scale` (FLOAT child of the `mode` combo, `mode = "scale by multiplier"`)

This is NOT the already-documented refusal case. `panel_set_widget`'s own description says dotted **STRING** children of a live `COMFY_DYNAMICCOMBO_V3` are refused because "the frontend can reserialize the combo after immediate readback and panel_run may then use an empty/stale child", and explicitly says non-STRING children "stay writable" (citing Nano Banana 2's `model.resolution`). Here the child is FLOAT, so the write is **accepted** — and then hits exactly the reserialization hazard the STRING guard exists to prevent. The guard appears to be too narrow: the hazard is the dynamic-combo parent, not the child's type.

Impact is worse than a refusal would be. The agent gets a success receipt AND a confirming read, so it has no way to know the run used a different value. In this session it caused three renders (~45 min of GPU time) to be attributed to the wrong configuration, and I gave the user two incorrect diagnoses based on a setting that was never actually applied.

### Steps to reproduce

1. Load a graph with a node exposing a `COMFY_DYNAMICCOMBO_V3` input with a numeric child (e.g. `MinimaxH3LatentUpscaler3D`, `mode = "scale by multiplier"`, `mode.scale = 2`).
2. `panel_set_widget {node_id: 186, widget: "mode.scale", value: 1.5, expected_value: 2}`
   → returns `{"set": {"previous": 2, "value": 1.5}}` (success).
3. `panel_query_graph {ids:[186], fields:"compact"}`
   → shows `mode.scale=1.5`. Looks applied.
4. `panel_run`.
5. Read the node's own runtime log line.
   → `[MinimaxH3-3D] Latent 36x64 -> 72x128 | Pixels 1152x2048 | scale=2.000`

[...]

Issue #1931, opened 2026-08-27
panel_add_node("SaveVideo") creates a duplicated dynamic-combo widget set (format.codec AND codec), then panel_run fails with "Dynamic widget doesn't exist on node"


artokun:
### What happened / root cause

Adding a core `SaveVideo` node with `panel_add_node` produces a node whose dynamic-combo children are **duplicated**: both a dotted child `format.codec` *and* a bare top-level `codec`. Queuing that graph with `panel_run` then fails before anything reaches ComfyUI:

```
Error: Dynamic widget doesn't exist on node
```

`SaveVideo`'s `format` is a nested `COMFY_DYNAMICCOMBO_V3` (`/object_info/SaveVideo`): `format` → option `auto` → required child `codec` → option `h264` → optional child `encoding` → option `re-encode` → `crf`. Only `format` is declared at the top level; `codec` exists **only** as a child of a chosen `format` option. The panel materialises it twice — once correctly as `format.codec`, once as an orphan `codec` that no longer corresponds to a real widget, and the run-time serializer trips over the orphan.

Right after `panel_add_node {class_type:"SaveVideo"}`, the tool's own reply already shows the duplication:

```json
"widgets": { "filename_prefix": "video/ComfyUI", "format": "auto", "format.codec": "auto", "codec": "auto" },
"inputs": [
  {"slot":2,"name":"format","type":"COMFY_DYNAMICCOMBO_V3"},
  {"slot":3,"name":"format.codec","type":"COMFY_DYNAMICCOMBO_V3"},
  {"slot":4,"name":"codec","type":"COMFY_DYNAMICCOMBO_V3"}
]
```

### Steps to reproduce

1. `panel_add_node {class_type:"SaveVideo"}` on any graph.
2. Wire a `VIDEO` into it (e.g. `CreateVideo` → `SaveVideo.video`).
3. `panel_run` → `Error: Dynamic widget doesn't exist on node`. Nothing is queued; ComfyUI never sees the prompt.

Reproduced while hand-building a MiniMax H3 text-to-video graph (`CreateVideo` → `SaveVideo`), ComfyUI 0.34.1.

### Workaround that unblocked it

Re-writing the parent combo re-syncs the node enough to queue:

```
panel_set_widget {node_id: 15, widget: "format", value: "mp4"}
panel_run   → queued OK
```

[...]

PR #2141, opened 2026-09-01
fix(#2140): re-resolve dotted DynamicCombo children before restoring them, and name the node a dynamic-widget throw is about

### LINKED ISSUE #2140: panel_run fails with "Dynamic widget doesn't exist on node" on a SaveVideo added by panel_add_node (DynamicCombo children never registered)

Building a MiniMax H3 video graph from scratch with the `panel_*` tools, `panel_run` was
REFUSED outright with:

```
Error: Dynamic widget doesn't exist on node
```

Nothing was queued. `panel_get_errors` reported `errored_count: 0` / `audit_complete: true`
and `node_errors: null`, so the failure was invisible on the canvas — it happens in the
frontend `graphToPrompt` conversion before a prompt is ever built.

The offending node was `SaveVideo`, added via `panel_add_node` with its DEFAULT widget
values (`format="auto"`, `format.codec="auto"`).

**This is NOT stale schema residue**, which is what `web/js/lib/dynamic-widget-reconcile.js`
is built to clean up (see `staleDynamicGroup`, the `#1931` comment about "the SaveVideo
`format.codec` residue"). I checked the live schema:

```
GET /object_info/SaveVideo → input.required.format
  ["COMFY_DYNAMICCOMBO_V3", { options: [
      { key: "auto",  inputs: { required: { codec: [COMFY_DYNAMICCOMBO_V3, {...}] } } },
      { key: "mp4",   inputs: { required: { codec: [...] } } },
      ... ] }]
  and codec's own "auto" option declares inputs.required = {}
```

Every `format` option — including `auto` — declares a required `codec` child, and
`codec="auto"` declares no further children. So `format=auto` + `format.codec=auto` is a
COMPLETE, CORRECT widget state with no orphan and no residue. The reconcile pass has
nothing to strip, which is presumably why the `graphToPrompt` wrapper's reconcile-and-
retry-once path (`retryLiveDynamicWidgets`, ~line 573-632) does not rescue it either.

The actual defect appears to be that a node created programmatically by `panel_add_node`
has its DynamicCombo children materialised as plain serialized widgets that were never
registered with the frontend's dynamic-widget system (no live `widgetId` binding), so
`graphToPrompt`'s dynamic-widget lookup cannot resolve `format.codec`.

Evidence for that reading: the workaround is to ASSIGN the parent combo a value. Setting
`format` from `"auto"` to `"mp4"` via `panel_set_widget` runs the native DynamicCombo
setter, which destroys and REBUILDS the children as properly registered dynamic widgets
(exactly the rebuild behaviour documented in the `#2031` comment at ~line 203). After that
single write, `panel_run` queued successfully with no other change. Note the final state
still has `format.codec="auto"` — the same value it had when it was failing — so the value
was never the problem, only its registration.

This should affect ANY node with a required `COMFY_DYNAMICCOMBO_V3` input added through
`panel_add_node`, not just `SaveVideo`; `SaveVideo` is simply the one an agent building a
video workflow reaches for every time.

### Secondary: the error names no node

`namedDynamicWidgetError` / the guard at ~line 549
(`DYNAMIC_WIDGET_MISSING_RE.test(base) && !/\bnode\s+\d+/i.test(base)`) exists to attach a
node identity when the raw message lacks one. The message that actually reached the agent
was the bare string above — no node id, no node title. With `panel_get_errors` also clean,
there was nothing to bisect from; I had to reason from the schema to work out which of 21
nodes was at fault. Whatever that naming path is meant to do, it did not fire here.

### Steps to reproduce

1. New empty workflow.
2. `panel_add_node` `class_type: "SaveVideo"` — do not touch its widgets.
3. Wire any `VIDEO` source into it (e.g. `CreateVideo` → `SaveVideo.video`).
4. `panel_run` → fails with `Dynamic widget doesn't exist on node`, nothing queued.
5. `panel_set_widget` node `format` = `"mp4"` (any concrete option).
6. `panel_run` → queues fine.

### Exact error (scrubbed)

```
Error: Dynamic widget doesn't exist on node
```

(That is the complete message as surfaced through `panel_run`.)

### Fix

applied locally: **no**   upstream-only: **no** (it is ours, but I did not patch it)

I deliberately did not patch. `web/js/lib/dynamic-widget-reconcile.js` is 637 lines of
carefully-invariant code carrying at least `#1931` and `#2031` regression constraints plus
a browser unit-test suite (`browser_tests/unit/dynamic-widget-reconcile.test.mjs`,
`graph-to-prompt-failure.test.mjs`, `primitive-dynamic-widgets.test.mjs`), and a blind edit
risks re-breaking the residue/rebuild cases those encode. Verifying properly needs that
suite run plus a hard browser refresh, which would have killed the user's in-flight render.

Precise change I believe is needed: after `panel_add_node` instantiates a node, for each
required input whose spec is `COMFY_DYNAMICCOMBO_V3`, force the frontend to build its
children rather than leaving the serialized defaults in place — e.g. re-assign the root
widget its own current value so the native setter runs the rebuild. The same-value path is
already made safe by `wrapDynamicComboSetter`, which captures and restores prefixed child
values when `previous === next`, so this should preserve values while fixing registration.
Alternatively, have `retryLiveDynamicWidgets` treat an unresolvable dynamic child as a
rebuild trigger (re-assign the root) and not only as a residue-strip candidate — that would
also cover nodes that arrive via paste/load rather than `panel_add_node`.

### Environment

- OS: Windows 11 Pro N 10.0.26200
- ComfyUI: 0.34.2
- ComfyUI **frontend**: 1.49.6
- Python 3.13.12 / PyTorch 2.10.0+cu130 / CUDA 13.0
- GPU: NVIDIA GeForce RTX 5090, 32 GB VRAM
- comfyui-mcp: 0.52.152
- comfyui-mcp-panel: 0.15.149

---
Filed from the ComfyUI panel by **claude** · model `opus[1m]` · effort `max` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #1654

artokun:
Fixes #2140.

## What was wrong

`restorePrefixedValues` (`web/js/lib/dynamic-widget-reconcile.js`) resolved every captured dotted name against **one map built before the restore loop started**:

```js
const byName = new Map(node.widgets.map(w => [w.name, w]));   // built ONCE
for (const name of names /* sorted shallow → deep */) {
  const widget = byName.get(name);
  ...
  widget.value = next;
}
```

Restoring the shallowest child (`format.codec`) drives the frontend's native DynamicCombo rebuild, which **replaces every widget below it**. So by the time the loop reaches `format.codec.encoding`, the map still points at the widget the node was carrying a moment ago and no longer is.

This runs on **every prompt build** — `prepareLiveDynamicWidgets` → `reconcileGraphDynamicWidgets` → the same-value replay, which goes through `wrapDynamicComboSetter`'s `previous === next` branch.

Two consequences, both live:

1. **Silent value loss.** An ordinary `panel_run` reset SaveVideo's `format.codec.encoding.crf` from the user's `17` to the schema default `23`. Nothing about that run asked to change a value.
2. **The #2140 state.** For a *dynamic* child the write drove a **detached accessor**. In frontend 1.49.6, `dynamicComboWidget`'s `updateWidgets` deletes the group's widget rows **and their widget-store entries** *before* it checks that the accessor is still attached to the node, and only then throws `Dynamic widget doesn't exist on node` — straight into the restore's `catch {}`. The node is left with a dynamic root whose declared children are gone from both `node.widgets` and the widget store.

That is also what separates the two recoveries the reporter measured on the live rig, which is the load-bearing evidence here:

| write | runs `restorePrefixedValues`? | recovered? |
|---|---|---|
| same-value `format="mp4"` | **yes** (`previous === next`) | no |
| `mp4 → mkv → mp4` | no (values differ both times) | yes |

Identical final widget values either way. I confirmed `panel_set_widget` has no same-value short-circuit on the non-deferred path (`assignWidgetValue` in `web/js/lib/widget-write.js:1888` always assigns), so that observation is about the restore, not about a skipped write.

## The fix

**1. Re-resolve per name** (the actual fix). Each dotted name is looked up in the live `node.widgets` immediately before it is written; a name that no longer resolves is skipped instead of driving a detached accessor.

**2. Name the node** (the issue's own "Secondary" section). `namedDynamicWidgetError` could only attach identity through `describeOrphanDynamicWidgets` or `describeTypedPrimitiveWidgets`. This state has no orphan and no typed PrimitiveNode, so `listedParts` was empty and the bare string was returned unchanged — with `panel_get_errors` also clean, the reporter had to read the schema of 21 nodes by hand. Now it names:
   - any DynamicCombo root whose **selected option declares a child row the node is not carrying**;
   - a root carrying a **duplicated dotted row** (the sweep's other residue — two live rows can never legitimately share one dotted name);
   - failing both, it lists the graph's dynamic-combo nodes as **candidates**, and `graphToPromptFailureRefusal` learns not to say "inspect the named widget" over a message that explicitly disclaims identity.

## Where to look hardest

- **`restorePrefixedValues`** — this is the load-bearing change and it sits on a path with `#2031` and `#1931` invariants. It now re-scans `node.widgets` per name (last-wins, matching the old `new Map(...)` duplicate semantics) rather than reading a cached map. Everything else in the diff is diagnosis.
- **The false-positive guard in `collectUnresolvedDynamicCombos`.** `addInputWidget` returns before creating a row when an input is `forceInput` or its type has no registered widget constructor — socket, no widget. That is legitimate and must not read as the stripped state. The discriminator is that the native sweep clears `node.inputs` *and* `node.widgets`, so a **surviving socket** means the row was never meant to exist. Pinned by `a socket-only child is not mistaken for a stripped row`.
- **`describeUnresolvedDynamicCombos` returns `null`, not `[]`, when the live value names no declared option** — it draws no conclusion rather than reporting every child as missing.

The issue proposes two repairs. I built the second one (have the retry path treat an unresolvable child as a rebuild trigger via an option round trip), pinned it with tests — and then **removed it, because I proved it was dead code**:

```
MUTATION: replace `repairUnresolvedDynamicCombos(graph)` in retryLiveDynamicWidgets with a no-op
RESULT:   12/12 tests still pass
```

Once `restorePrefixedValues` stops re-corrupting, the same-value replay that already runs in `reconcileGraphDynamicWidgets` performs the *identical* native rebuild and recovers the node on its own — see `the retry-once path recovers a stripped root and the run proceeds`. Shipping a path that writes two real option values to the user's live graph, that nothing needs, was not a trade worth making.

The issue's *first* proposal — re-assign the root its own current value — is the write the reporter's follow-up disproved, and is what `reconcileFreshDynamicWidgets` already does.

## Verification

`node --test browser_tests/unit/dynamic-combo-registration-2140.test.mjs` — 12 new tests.

The harness is **transcribed from the shipped frontend**, not modelled from intent: `dynamicComboWidget` in `comfyui_frontend_package` 1.49.6 `static/assets/settingStore-CwkLtSKP.js` (byte 723111 — the only occurrence of the error string in the whole bundle), and the SaveVideo option tree from `comfy_extras/nodes_video.py` in the installed ComfyUI 0.34.2. That matters: the existing `dynamic-widget-reconcile.test.mjs` double rebuilds *by name*, so it can never enter this state — which is why 7955 green tests never saw this.

- Full suite: **7955 pass / 0 fail** (`npm run test:unit`), `tsc --noEmit` clean.
- **Mutation-verified.** Restoring the one-shot `byName` map turns `a prompt build must not lose a value BELOW a rebuilt dynamic child` red, through the real `installGraphToPromptDynamicReconcile` wrapper — not by calling a helper directly.
- The mechanism is *executed*, not asserted: `the native rebuild strips the live rows BEFORE it throws` detaches the accessor exactly as the sweep does, drives it once, and reads back that the row and its store entry are gone.
- The candidates fallback is exercised **through the production wrapper**, not just by calling the exported helper.

## What this does not settle

I could **not** run the Playwright browser gate: no ComfyUI is listening (only :8264, which is not ComfyUI), and the suite does not start one. So this is unit-verified against a transcription of the real frontend, not observed in a browser.

And I could not reproduce, offline, the *first* failure in the report — a freshly added SaveVideo at `format="auto"` failing before any restore had run. What I fixed is the mechanism that matches the reporter's **recurrence** and their same-value-vs-round-trip asymmetry exactly. If the first failure has a separate cause, the diagnosis half of this PR is what will name it next time instead of returning a bare string.


artokun:
### Gate

`codex-gate.mjs` round 1 on `6df260f6` → **NEEDS-REVISION**, one P1: the two describers read only `input.required`, so a node whose DynamicCombo is declared **optional** still produced the bare, node-less throw this PR exists to end. Valid, and directly relevant — SaveVideo itself declares a hidden top-level optional `codec`.

Fixed in `1a82b1d0`. Both describers now scan `required` **and** `optional`. `reconcileFreshDynamicWidgets` deliberately still reads only `required`: a replay is an authorized *write* and that scope is the #1931 relocate decision, whereas these two are read-only naming.

The narrowing that goes with it: an optional root with **no widget row** is not reported. Optional means the node was free not to materialise it, and the #1931 relocate path removes SaveVideo's hidden `codec` on purpose — reporting it absent would have fired on every healthy SaveVideo on the canvas. Pinned both ways (`an OPTIONAL top-level DynamicCombo root is named too`, `an optional root the node never materialised is NOT reported missing`), and mutation-checked: restricting the scan back to `["required"]` turns the first red.

`codex-gate.mjs` round 2 on `1a82b1d0` → **SHIP** (exit 0). That is the SHA the PR head is on now and the one that will merge.

Suite after the revision: **7957 pass / 0 fail**, `tsc --noEmit` clean.

PR #2157, opened 2026-09-01
fix(#2151): drive rgthree Fast Bypasser/Muter rows with the forced value, not their toggling callback

### LINKED ISSUE #2151: panel_set_widget falsely reports success when LC Bypasser boolean immediately reverts
panel_set_widget reported success when setting boolean widgets on an LC Bypasser / LC Bypasser Panel node, but the values immediately reverted to true. The false success caused a queued render to run every post-processing effect, producing a wrong B&W/text-overlaid result. Immediate panel_query_graph reads confirmed the widgets were still true.

The panel mutation path does not verify the final widget value after the custom node's callback/serializer settles. It should either invoke the custom widget callback correctly and verify persistence, or return a failure if readback differs.

### Steps to reproduce
1. Open a workflow containing LC Bypasser connected to LC post-processing nodes.
2. Call panel_set_widget on node 143 or its LC Bypasser Panel node with widget "Enable LC Film Stock (B&W)" and value false.
3. Observe the tool reports success.
4. Immediately call panel_query_graph for the node: the widget remains true.
5. Queue the graph: the supposedly disabled effect runs.

### Exact error (scrubbed)
No error was returned. This was a silent false-success mutation.

### Fix
applied locally: no
upstream-only: yes

Precise change needed: after panel_set_widget dispatches the write/callback, await the frontend update cycle and read the widget value back. If it differs from the requested scalar value, return a mutation failure naming that the custom node reverted/serialized the widget. If LC Bypasser requires a specific callback/event, dispatch that callback before readback. Do not return success from the pre-callback assignment alone.

Temporary verified workaround: disconnect unused effect inputs from LC Bypasser, then set those effect nodes to bypass. This persisted through queue and produced the correct render.

### Environment
Windows 11
ComfyUI 0.34.2
ComfyUI frontend 1.49.6
GPU NVIDIA GeForce RTX 5080 16 GB
Python 3.13.12
PyTorch 2.11.0+cu130
comfyui-mcp 0.52.165
panel 0.15.149

artokun:
Fixes #2151.

## What is actually broken

"LC Bypasser" is an rgthree **`Fast Bypasser (rgthree)`** that has been retitled. That node mints one plain `toggle` row per node wired into its inputs, named `Enable ${linkedNode.title}` — which is why the reporter addressed a row called `Enable LC Film Stock (B&W)`.

These are **not** the `RGTHREE_TOGGLE_AND_NAV` group rows #2146 covers, and they were never covered. From the shipped pack file `rgthree-comfy/web/comfyui/base_node_mode_changer.js` (unminified, byte-identical to `src_web/comfyui/base_node_mode_changer.ts`):

```js
widget.doModeChange = (forceValue, skipOtherNodeCheck) => {
  let newValue = forceValue == null ? linkedNode.mode === this.modeOff : forceValue;
  if (skipOtherNodeCheck !== true) { /* toggleRestriction */ }
  changeModeOfNodes(linkedNode, (newValue ? this.modeOn : this.modeOff));
  widget.value = newValue;
};
widget.callback = () => { widget.doModeChange(); };
```

**The callback takes no arguments.** It always lands in the `forceValue == null` branch, so it derives the new value from *the linked node's current mode*. The row's callback is a **toggle, not a setter** — it ignores the value `panel_set_widget` just assigned, and `applyWidgetWrite`'s ordinary path is assign-then-fire-the-callback.

## Measured, before the fix

Driving unmodified `applyWidgetWrite` with a fixture transcribed from that pack file:

| linked node mode | requested | result before this PR |
|---|---|---|
| ALWAYS (0) | `false` | row `false`, mode 4. **Correct by luck** — the toggle happened to agree. |
| BYPASS (4) | `false` | **mode → 0.** The effect is switched back ON. Throws `wrote false but it became true`; the ROW value is rolled back, **the MODE is not**. |
| ALWAYS (0) | `true` | **mode → 4.** The node is silently bypassed. Same asymmetric rollback. |
| BYPASS (4) | `true` | row `true`, mode 0. Correct by luck. |

So the write only has to *agree* with the row to invert the graph, and the caller is handed an error that says nothing about the mutation it left behind. That is the reported harm end to end: "the false success caused a queued render to run every post-processing effect… the supposedly disabled effect runs."

It is not a hypothetical starting state either — the reporter's own verified workaround was to bypass those effect nodes directly on canvas, which is exactly what puts a row and its linked node out of step (the pack re-syncs a row only when its **name** changes).

## The fix

Drive the pack's own forced-value entry point, `doModeChange(value)` — the one `forceWidgetOn` / `forceWidgetOff` call — instead of assigning `.value` and firing the toggling callback, and journal the reachable modes for rollback exactly as #2146 does for the group rows.

- **Seeding** (new arm in `captureFastBypasserModeTransaction`): the nodes on the changer's inputs, followed through Reroute / Node Combiner / Node Collector (rgthree's `PassThroughFollowing.ALL`, mirrored by the existing `isModePassThrough`), descending subgraphs (what `changeModeOfNodes` → `reduceNodesDepthFirst` does), and drained onward through Mute / Bypass Repeater and dispatching Relay propagation. All existing machinery; only the seed set is new.
- **Fail closed**: a boundary that cannot be established refuses *before* the action runs, rather than mutating modes this writer could not put back.
- **Idempotent**: setting a row to the value it already holds is now a no-op that also repairs a linked node whose mode had drifted.

## Where the load-bearing judgement is — please push on these

1. **`requiresModeChange: false` on the new arm.** The group arm asserts "a reachable mode must have changed"; I deliberately do not, because row-value/mode divergence is routine (Ctrl+B on canvas) and forcing an already-correct mode is a legitimate no-op. What verifies the action instead is that **the writer does not assign the row value at all** — `doModeChange` sets `widget.value = newValue` itself, on the line after `changeModeOfNodes`, so the ordinary read-back seeing the requested value establishes both that the action ran and that it ran with `newValue === requested`. If that reasoning is wrong, the new arm reports success on a stub.
2. **Journal is a superset.** The widget→linked-node mapping is positional *inside the pack* (`widgets[index]` ↔ `linkedNodes[index]`) and is not exposed on the row, so I journal every linked node rather than the one this row reaches. Superset = safe for restore; a subset would not be. Worth checking I have that direction right.
3. **Classification.** Keyed on node type **and** a callable `widget.doModeChange`. The row NAME cannot be used — the pack mints it from the linked node's title, so it differs on every row of every graph. A widget on these nodes without `doModeChange` falls through to the ordinary path, unchanged.
4. **`toggleRestriction`.** `skipOtherNodeCheck` is left undefined, so "max one" / "always one" are enforced by the node exactly as a click would, and a row the node refuses to switch off is reported as a **failed write** rather than claimed as applied. That is a deliberate behaviour choice, not an oversight.
5. **Rename.** `resolveFastBypasserAction`'s returned `toggle` field is now `action` (both arms invoke `Reflect.apply(action, row, [requested])`). One call site; #2146's own tests still pass untouched.

## Verification

- New `browser_tests/unit/fast-mode-changer-widget-write.test.mjs` — 12 tests against the real `applyWidgetWrite`, fixtures transcribed from the shipped pack file.
- **Mutation-verified, three ways.** Neutering the classifier (`if (true || …) return null`) → 9 of 12 fail. Calling the action with no forced value (restoring toggle semantics) → 3 fail. Deleting the journal seeding → the 4 rollback/fail-closed tests fail. The 3 that stay green under mutation 1 are the ones asserting behaviour that was already correct — stated rather than hidden.
- Input shape covered explicitly: a tool call carries JSON, so `"false"` / `"0"` / `0` / `false` are each pinned to land on the same mode from either starting mode.
- Full unit suite: **8114 pass, 0 fail**.

## What I did not do

- **`Fast Groups Muter`** keeps its #983 refusal, untouched.
- I could not reproduce on a live canvas with the actual LC pack — it is not installed here. Everything above is executed against the panel's real write path with a fixture taken from the pack's own shipped JS, not from the report's narrative.
- The reporter's exact phrasing was "reported success"; what I can demonstrate is that the *correct-looking* successes are coincidences of a toggle agreeing with the request, and that the disagreeing cases return an error while leaving the linked node inverted and unrestored. Both halves are the same defect and both are fixed.


artokun:
### codex gate round 1 — NO-SHIP, one P1. Confirmed against the pack's source and fixed in `0200183`.

> `web/js/lib/widget-write.js:780` — journals current input nodes, but `doModeChange` may still target a previously linked same-title node — rewire A to same-title B, request `false`, then trigger verification failure: B is restored while A remains mutated.

**Correct, and reproducible from the pack's own lifecycle.** `setWidget` installs the closure only inside the name guard:

    let name = `Enable ${linkedNode.title}`;
    if (widget.name !== name) { ...install a closure over THIS linkedNode... }

so a slot rewired from A to a **same-titled** B never rebuilds the closure. Default titles are the node's display name, so two identically-titled nodes are ordinary, not exotic.

The gate found half of it. Fixing the rollback alone would have left the worse half standing: `doModeChange` assigns `widget.value` whatever node it switched, so the retention check passes and the write is reported as a **success** having switched a node the caller never addressed — the same false-success shape this issue is about, one level down. Both halves are now closed:

- **Rollback.** The journal additionally sweeps every node in the graph, because a stale closure's target is not discoverable from the row. That sweep is **lenient** on purpose — it spans nodes the write has nothing to do with, and one hostile third-party `mode` getter in a 300-node workflow must not refuse every Fast Bypasser write. The intended targets stay strictly captured, and `unrestored()` still reports honestly on everything journalled.
- **Success.** New `targetSettled(requested)` checks the **effect** on the node the row's position maps to, and a pairing that cannot be established is itself a refusal.

The pairing needed a faithful transcription of the pack's `getConnectedInputNodesAndFilterPassThroughs` order, not "one node per slot": a **Node Collector** — a designed rgthree usage — fans several nodes into one slot, so the linked-node list is longer than the input list. That is now pinned by its own test; mutating the walk to skip pass-through expansion fails it.

**Mutation-verified independently**, so neither half is decorative: removing the graph sweep fails the stale-closure test (rollback incomplete); short-circuiting `targetSettled` fails the same test (false success); skipping pass-through expansion fails the collector test.

Unit suite: **8116 pass, 0 fail** (14 in this file). Round 2 of the gate is running on `f1f00ec`.

Issue #2020, opened 2026-08-29
panel_set_widget hangs indefinitely on AnimaPromptPlus custom text widgets


artokun:
### What happened / root cause
panel_set_widget successfully updated four core loader widgets, then stalled indefinitely when writing AnimaPromptPlus.quality_prompt. The mutation did not land. A second isolated call reproduced the stall.

### Steps to reproduce
1. Open a live canvas with Comfyui-Anima-Tools installed.
2. Add AnimaPromptPlus.
3. Call panel_set_widget with node_id of AnimaPromptPlus, widget quality_prompt, and a nonempty string.
4. Call does not return within 50 seconds; canvas value remains unchanged.

### Exact error (scrubbed)
No error response. Both calls remained in-flight until client-side termination.

### Fix
applied locally: no  upstream-only: yes
Suggested workaround: allow custom text widgets to be updated through their string input sockets, or make panel_set_widget handle AnimaPromptPlus custom textarea synchronization without awaiting a non-resolving frontend callback.

### Environment
Windows 11; ComfyUI 0.33.0; RTX 4070 SUPER 12GB; comfyui-mcp 0.52.146; panel 0.15.124; Manager 3.x legacy.

artokun:
Claimed by grok-build P2-drain swarm 28 2026-08-30T10:12Z. Oldest remaining free Panel P2 after skipping live swarm23-27. Worktree `wt-2020`. Driving to tests + PR; parent merges after CI green.

artokun:
Re-engaging abandoned grok-build swarm 28 claim (posted 10:12Z, last PR touch 10:28Z, now >2h stale). Cycle 31. Worktree `wt-2020`. Rebuilding PR #2079 off current origin/main and driving to green CI.

artokun:
Rebased onto current origin/main (v0.15.135 + #2088/#2087) and driving existing PR https://github.com/artokun/comfyui-mcp-panel/pull/2079 — not opening a second PR.

**Fix:** `panel_set_widget` on AnimaPromptPlus custom textarea widgets (`quality_prompt`) no longer hangs. The write goes through the live textarea / `options.setValue` store and does not invoke or await a widget callback that never settles. Read-back still decides whether the value stuck. Core string widgets without a live editor keep the existing callback path.

**Changelog:** kept all unique Unreleased Fixed bullets under a single heading, including #2020. Did not bump version.

[...]

PR #2160, opened 2026-09-01
test(comfyui-mcp#2709): pin the #2148 repair on a promoted MULTILINE rail

artokun:
Closes the coverage half of **artokun/comfyui-mcp#2709**. Test-only; **no production change**.

## What #2709 actually is

#2709 reports that `panel_set_widget` cannot write a promoted STRING on a SubgraphNode, quoting the verbatim `definitionMoved` refusal from `web/js/lib/widget-write.js`. That branch was replaced on 2026-09-01 by #2148 (`24c5a884`), which **repairs** the collateral write to the shared definition instead of refusing. `git tag --contains 24c5a884` is `v0.15.150` only; the reporter ran panel **0.15.149**. So the report is already fixed — it is being closed as such.

## Why a test is still worth landing

**#2148 was never exercised on the shape that was reported.** Its fixture models the rail that `SubgraphNode._projectPromotedWidget` projects — a plain object whose `value` accessor views `widgetValueStore[widgetId]` — on an INT `width`.

A promoted **multiline** STRING does not get that rail. On comfyui-frontend 1.49.6 (the reporter's exact version) the app-level SubgraphNode overrides the projection hook:

```js
createPromotedHostWidget(input, widgetId, sourceWidget) {
  return createPromotedMultilineWidget({ subgraphNode: this, input, widgetId, sourceWidget })
}
// => new DOMWidgetImpl({ type: "customtext", element: <textarea>, options: {
//      getValue: () => store.getWidget(widgetId)?.value,
//      setValue: v => { textarea.value = v; store.setValue(widgetId, v) } } })
```

This also explains the reporter's own observation that *"numeric/boolean promoted widgets on the same node work"* — those get the plain store proxy; only the STRING gets the DOM widget.

## The load-bearing part — please look here

#2148 gates its repair on `strictlyRetained` (the rail's own `.value`) and **deliberately not** on `widgetMatchesExpected`, because that helper also accepts a live customtext widget whose DOM *editor* holds the value while `.value` does not (#2020).

A `customtext` rail is therefore **the one shape where the repair could decline to run at all**, falling through to the old refusal. It survives here only because:

1. `getValue` reads the per-instance store, so the rail's `.value` reflects the write; and
2. nothing on the inner widget reads back into the rail — `bindMultilineTextareaWidget` listens only for the textarea's `input` event, which a programmatic assignment does not fire — so restoring the shared definition cannot drag the rail back and trip the `railKeptValue` clause.

Both are frontend details. Neither was pinned anywhere. A frontend release that made the promoted multiline rail a view of the inner widget would silently put #2709 back, and the existing suite would stay green.

## Evidence

Control, on the final committed file:

| | `24c5a884^` (pre-#2148) | `origin/main` |
|---|---|---|
| repaired, not refused | **FAIL** — reporter's message verbatim | pass |
| definition restore does not drag the rail | **FAIL** | pass |
| no write-through ⇒ no repair claimed | pass | pass |

The pre-fix failure text is character-for-character the report, down to node `251` and widget `"value"`:

> Promoted widget "value" on subgraph node 251 is backed by a per-instance value store (widgetId "…:251:value"), but writing it ALSO changed the shared subgraph definition's inner widget "value" on node 10 … Rolled back rather than report an instance-scoped write that was not one (comfyui-mcp#1707).

The third test is the control that must NOT move: it passes on both sides, so the two that flip are flipping on the fix and not on the fixture.

- Full unit suite: **8125 tests, 0 fail**.
- Gates: `i18n-check`, `i18n-render-check`, `check-tool-vocabulary`, `check-panel-scope` all OK.

## What this does not cover

The report's second symptom — *"Rollback … did not take effect … graph may be in a partial state"* — is **not** proven fixed by this. On 0.15.150 that branch is simply not reached for this shape, because the repair accepts before it. It can still fire on a genuinely-blocked repair (`blockedBy` `"rail"`/`"inner"`). I did not chase it, since no current report reaches it.

Not browser-verified: this is a pure unit pin of a module-level function, and the shape is modelled from the installed 1.49.6 bundle rather than driven through a canvas.

### DIFF
--- a/browser_tests/unit/promoted-multiline-rail.test.mjs
+++ b/browser_tests/unit/promoted-multiline-rail.test.mjs
@@ -1,0 +1,243 @@
+/**
+ * comfyui-mcp#2709 — the #2148 repair, on a promoted MULTILINE STRING rail.
+ *
+ * #2148 fixed "a promoted rail that writes through to the shared subgraph
+ * definition is REPAIRED, not refused". Its fixture (promoted-value-scope.test.mjs)
+ * models the rail frontend `SubgraphNode._projectPromotedWidget` projects: a plain
+ * object whose `value` accessor reads and writes `widgetValueStore[widgetId]`.
+ *
+ * That is NOT the rail a promoted multiline STRING gets. On comfyui-frontend 1.49.6
+ * the app-level SubgraphNode OVERRIDES the projection hook:
+ *
+ *   createPromotedHostWidget(input, widgetId, sourceWidget) {
+ *     return createPromotedMultilineWidget({ subgraphNode: this, input, widgetId, sourceWidget })
+ *   }
+ *
+ * and that builds a DOM widget instead:
+ *
+ *   new DOMWidgetImpl({ node, name, type: "customtext", element: <textarea>, options: {
+ *     getValue: () => store.getWidget(widgetId)?.value,
+ *     setValue: v => { textarea.value = v; store.setValue(widgetId, v) } } })
+ *
+ *   class BaseDOMWidgetImpl {
+ *     get value() { return this.options.getValue?.() ?? "" }
+ *     set value(v) { this.options.setValue?.(v); this.callback?.(this.value) }
+ *   }
+ *
+ * The distinction is load-bearing, not cosmetic. #2148 gates its repair on
+ * `strictlyRetained` — the rail's own `.value` — and deliberately NOT on
+ * `widgetMatchesExpected`, precisely because the latter also accepts a live
+ * customtext widget whose DOM editor holds the value while `.value` does not
+ * (#2020). A `customtext` rail is therefore the one shape where the repair could
+ * decline to run at all, leaving the write refused. It survives here only because
+ * `getValue` reads the per-instance store — a frontend detail that a future
+ * frontend release could change without anything else noticing.
+ *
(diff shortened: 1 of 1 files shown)


artokun:
**codex gate — VERDICT: PASS** on `6401dc256eb410f6ce1be36c9af6429008dfe1e4` (the SHA that merges; PR head and worktree HEAD verified equal before the run).

The `codex-gate.mjs` wrapper could not produce a verdict (`CreateProcessAsUserW failed: 5 (Access is denied.)` — the known sandbox-spawn failure, which is INDETERMINATE, not a pass), so the review was run directly via `codex exec -s read-only` against the supplied diff.

Findings: none. It was asked specifically to attack the claims this PR rests on, and answered each:

> The fixture explicitly injects the observed rail→definition write-through, so repair is genuinely required; the pre-#2148 failure claim is sound. Test 2 is a valid no-write-through control, and Test 3 adds coverage for the textarea DOM value beyond Test 1's store assertions. State is correctly isolated per fixture instance.

CI: Lint pack **pass** (3m35s). Full panel unit suite locally: **8125 tests, 0 fail**. Gates `i18n-check`, `i18n-render-check`, `check-tool-vocabulary`, `check-panel-scope` all OK.

Issue #2099, opened 2026-08-30
panel_search_nodes times out on DaSiWa query with Manager search unavailable


artokun:
### What happened
The live-panel custom-node search timed out while trying to locate the missing node type DaSiWa_SeedControl.

### Steps to reproduce
1. Open the workflow containing missing node type DaSiWa_SeedControl.
2. Call panel_search_nodes with query "DaSiWa".
3. The call waits 15 seconds and returns no results.

### Exact result
supported:false
managerReachable:false
count:0
reason:"ComfyUI-Manager node search timed out after 15000 ms; no reply arrived."

The Manager queue update path for the already-installed ComfyUI-DaSiWa-Nodes pack also reported: "An error occurred while updating 'comfyui-dasiwa-nodes'."

### Fix
No local code fix; used the available Manager update path as a fallback, but it failed. No model download was performed and no workflow data was changed by the search call.

### Environment
Windows 11 / ComfyUI 0.33.3 / frontend 0.15.113 / RTX 4070 Ti 12 GB / comfyui-mcp 0.52.152 / panel 0.15.113

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-luna` · effort `high` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Claimed by grok-build P2-drain swarm 29 2026-08-30T14:40Z. Next oldest free Panel P2 after skipping live sibling PRs and Codex NO-SHIP #2068 / protocol-blocked #2071. Worktree `wt-2099`. Driving to tests + PR; parent merges after the swarm. Other agents: this issue is taken — pick a different one.

Issue #1273, opened 2026-08-15
panel_run(to_node_id) always refused with graph-stamp race on graphs using cg-use-everywhere (UE rewrites the prompt at queue time)


artokun:
### What happened / root cause

`panel_run` with `to_node_id` (run-to-node) is **unusable on any graph that uses `cg-use-everywhere`**. It always fails the run-to-node graph-stamp race check, even when the user and the agent make no edits at all between queueing and dispatch.

The error message itself asked for this to be reported, and it also names the cause in its own candidate list: *"an extension that rewrites the prompt inside its own api.queuePrompt patch"*. That is exactly what cg-use-everywhere does — it patches `api.queuePrompt` and materialises its broadcast links into the prompt at queue time. So the panel's before-vs-after serialization can never match on these graphs: the "after" always contains the UE-injected links that the "before" does not.

Every single differing entry in the report is a UE broadcast target (`model` / `clip` / `vae` / `anything`), which is the signature of this, not of a user edit.

### Steps to reproduce

1. Build any graph that uses `Anything Everywhere` (cg-use-everywhere) to broadcast MODEL/CLIP/VAE — including inside a subgraph.
2. Do not touch the canvas.
3. `panel_run({ to_node_id: <any output node id> })`.
4. Refused every time with the graph-stamp race error below. The panel retries once after its pause (#1050) and the second dispatch fails identically.

A plain `panel_run()` (no `to_node_id`) on the SAME graph queues and executes fine, so the graph itself is valid — it is only the run-to-node stamp comparison that cannot be satisfied.

### Exact error (scrubbed)

```
Error: ComfyUI refused to queue the workflow: run-to-node scope for node 144 was NOT
applied: the workflow graph CHANGED after the run was queued — the deferred dispatch
would render a modified workflow, not the one that was scoped. The differing entries:
103:22 clip; 103:22 model; 103:22 vae; 103:23 model; 103:25 vae; 103:27 vae; 103:4 model;
103:48 anything; 103:49 anything; 103:5 vae; 103:50 anything; 106 clip; …and 5 more.
Retrying is safe (nothing was queued). If you did not edit these between queueing and
dispatch, the two serializations differ for a reason the panel cannot identify from here
— candidates: a queue-time widget hook (e.g. control_after_generate), an extension that
rewrites the prompt inside its own api.queuePrompt patch, a dynamic-input node reshaping
its slots, or a nondeterministic widget serializer. Please report this with the differing
[...]

Issue #1124, opened 2026-08-12
run-to-node permanently refused when Seed (rgthree) changes during graph stamping


artokun:
### What happened / root cause
panel_run({to_node_id:122}) could not apply run-to-node scope because an rgthree Seed node serialized differently between graph stamping and deferred dispatch. The panel retried once after its built-in pause and still refused. No job was queued. This appears to be a false-positive graph-stamp mismatch caused by queue-time/nondeterministic serialization of Seed (rgthree), not a user edit.

### Steps to reproduce
1. Open a workflow containing Seed (rgthree) node id 47 with seed=-1 / queue-time randomization behavior.
2. Connect it to two KSampler seed inputs.
3. Call panel_run with to_node_id targeting a downstream SaveImage.
4. The first dispatch is refused by the run-to-node graph-stamp race, then the automatic retry also fails.
5. Differing entry reported: node 47 seed.

### Exact error
"run-to-node scope for node 122 was NOT applied: the workflow graph CHANGED after the run was queued... The differing entry: 47 seed."
"Dispatch history: the first graph_run was refused ... re-issued exactly once ... failure above is that SECOND dispatch."
Nothing was queued.

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: graph-stamp comparison for scoped dispatch should normalize or ignore documented queue-time mutations from seed/control_after_generate widgets, or capture the post-queue-hook graph once and scope/dispatch that exact serialization. A nondeterministic Seed (rgthree) serializer must not permanently block run-to-node when no user graph edit occurred.

### Environment
Windows 11
ComfyUI 0.31.0
ComfyUI frontend 1.48.7
Python 3.12.10
PyTorch 2.13.0+cu130
GPU NVIDIA GeForce RTX 5090 31.8 GB
comfyui-mcp 0.51.16
panel 0.13.7

Related (same root cause): #572, #659

artokun:
Claiming this. I have independent evidence for your hypothesis from a different investigation, which is worth putting here because it confirms the nondeterminism rather than assuming it.

While working #1339 I drove the rgthree Seed handler directly on a live rig (ComfyUI 0.31.x, `/prompt` intercepted so nothing queued) with `seed: -1`, three times:

```
1028465986822020, 98533269447704, 557498712106716  — all different
```

[...]

PR #2163, opened 2026-09-01
fix(comfyui-mcp#2712): treat DaSiWa_SeedControl's queue-time seed roll as volatility, not graph drift

artokun:
Fixes the panel side of comfyui-mcp#2712.

## The mechanism, measured

`DaSiWa_SeedControl` does not use a widget `beforeQueued` hook — it wraps
**`app.graphToPrompt` itself**
(`darksidewalker/ComfyUI-DaSiWa-Nodes@main:js/dasiwa_seed_control.js:294-303`):

```js
app.graphToPrompt = async function (...args) {
  dasiwaSeedControlPrepareAll();      // rolls every random-mode SeedControl
  return originalGraphToPrompt(...args);
};
```

and the roll writes exactly two serialized inputs:

```js
node.__dasiwaSeedPrepareSeed = () => {
  if (dasiwaSeedControlHasExternalSeed() || controlState?.mode !== "random") return;
  ... widget.value = value;           // seed_value
  ... dasiwaSeedControlEmit();        // seed_control_state {mode,last_seed,recent}
};
```

**Why no retry could ever win.** `fingerprintOnce` builds the pre-dispatch stamp by
calling `app.graphToPrompt()` — the very function the pack wraps. So *our own stamp*
rolls a seed, and the dispatch's serialization rolls a different one. The two hashes
disagree on a completely idle canvas, deterministically, forever. That is why the
reporter saw all three dispatches lose: the panel's own restamp and the
orchestrator's `MAX_SCOPED_STAMP_RACE_RETRIES` re-issues (#2120, shipped in
0.52.166 and present in their 0.52.167) are the wrong instrument against a roller
that fires once per serialization. Only removing the pair from the compared content
can fix it.

## What this does

Adds the **seventh volatility signal** to `collectVolatileInputs`, alongside rgthree
(#1124), cg-use-everywhere (#1273), VHS (#2099) and Ideogram (#2130) — all the same
shape: a pack that rewrites the prompt after our stamp.

I did **not** take the report's primary suggestion (compare structural topology and
stop comparing widget values). That would blind #556 to every real mid-window widget
edit on every graph in order to accommodate one pack. This takes the alternative the
reporter also offered: exclude exactly the inputs this pack documents itself as
rewriting, keeping the structural fence and full drift coverage everywhere else.

## Where the load-bearing claims are — please look here

1. **`mode` is read the way the pack reads it, and the asymmetry is deliberate.**
   `dasiwaSeedControlParseState` normalises with
   `state.mode = state.mode === "fixed" ? "fixed" : "random"`, so **only** the literal
   `"fixed"` suppresses the roll. An absent, empty, or unparseable state widget is
   RANDOM and *does* roll — that is the freshly-dropped node, i.e. the common case.
   Testing `mode === "random"` instead would have left exactly those nodes refused.
   This is the clause most likely to be "corrected" into a bug.
2. **No mute/bypass gate**, unlike the rgthree signal. `dasiwaSeedControlPrepareAll`
   iterates every node in the graph tree and `__dasiwaSeedPrepareSeed` checks neither
   `mode` nor bypass, so a muted node still rolls. (Its inputs also drop out of both
   serializations, so the pair is moot there — but keying on bypass would be a false
   claim about someone else's code.)
3. **Both names or neither.** The roll rewrites `seed_value` and the `last_seed` /
   `recent[]` fields inside `seed_control_state` together, so excluding one alone
   still refuses the run. A fixed-mode node excludes *neither*, so a real mode flip
   stays drift — the same judgment the orchestrator's #2120 classifier makes when it
   refuses `seed_control_state` on its own.
4. **Class match reads `comfyClass` first** (what the pack keys on), falling back to
   `type`.

## Evidence

- Full panel unit suite: **8130 pass / 0 fail** (`npm run test:unit`, includes the
  i18n, tool-vocabulary and panel-scope gates). `npm run typecheck` clean.
- **Mutation-tested, 7/7 killed** — the wiring and each gate independently, after
  first verifying the probe reads a real `fail 0` on an unmutated tree:

  | mutation | result |
  |---|---|
  | delete the call site in the walk (wiring) | KILLED (5) |
  | require `=="random"` instead of the pack's `!=="fixed"` | KILLED (1) |
  | drop the external-seed link check | KILLED (1) |
  | `&&` instead of `\|\|` on the both-widgets gate | KILLED (1) |
  | drop the node-class match | KILLED (1) |
  | exclude `seed_value` only | KILLED (5) |
  | exclude `seed_control_state` only | KILLED (5) |

- The new end-to-end test reproduces the reporter's exact refusal and carries two
  controls: without the exclusion the drift tokens are precisely
  `["2739 seed_control_state", "2739 seed_value"]` — the string they quoted — and a
  real sibling edit alongside the roll **still fails closed**, naming only the input
  that genuinely changed.

## What I could not verify

I do **not** have this pack installed, so nothing here was observed on a live canvas
— the mechanism is read from the pack's published source and reproduced against
`collectVolatileInputs` / `promptContentHash` with fixtures shaped like the
reporter's prompt. A live confirmation on a real DaSiWa graph would be worth having
before assuming the reporter's case is closed. `npx playwright test --workers=1` was
run in this worktree; result reported in a follow-up comment.


artokun:
## Gate: PASS on the SHA that will merge

`codex exec` (direct — the `codex-gate.mjs` wedge returned
`CreateProcessAsUserW failed: 5 (Access is denied.)`, which is indeterminate, not a
verdict) on **`5c78ffec`**, which is the current PR head:

```
VERDICT: PASS

No findings.
```

Four rounds; the first three each found something real, and all of it is fixed in
this branch:

| round | finding | fix |
|---|---|---|
| r1 [P1] | `type` fallback could classify a node the pack does **not** roll (its dispatcher keys on `comfyClass` only), silently retiring drift coverage | match `comfyClass` only (`ce7cf8a8`) |
| r2 [P1] | `__dasiwaSeedInstalled` is set *before* the widget check, so a node installed while a widget was missing never gets the roll — widgets alone don't prove it | require `typeof node.__dasiwaSeedPrepareSeed === "function"` (`f9296afa`) |
| r2 [P1] | pack decides from a closure; we can only read the widget | unreachable from this module — documented as a bounded residual with both directions named |
| r3 [P1] | re-checking both widgets at collection time is a *stricter* gate than the pack's (it checks them at install), so a widget removed post-install stopped our exclusion while the pack kept rolling | require only the state widget, whose value **is** the mode (`5c78ffec`) |
| r3 [P2] | `parsed.mode` sees a non-enumerable/inherited key that the pack's `{...parsed}` spread drops | read through a spread (`5c78ffec`) |

r1 also caught a test that **passed for the wrong reason**: a default parameter
re-supplied `comfyClass` for an explicit `comfyClass: undefined`, so the fixture was
a fully-classed node. It now uses `in` and asserts the fail-closed direction.

## Mutation testing — 12 mutations, 11 killed, 1 disclosed

Probe verified first (reads a real `fail 0` on an unmutated tree) after an earlier
run mis-decoded node's UTF-8 `ℹ fail` marker under cp1252 and reported every
mutation as no-signal.

Killed: the call site in the walk; the `comfyClass` gate (dropped, and widened back
to `type`); the roll-function gate; the mode gate (inverted, and narrowed to
`=="random"`); empty-state-defaults-to-random; the spread; the external-seed link
check; and each half of the returned pair alone.

**Survivor, disclosed rather than hidden:** deleting `if (!stateWidget) return [];`
changes nothing, because the `try/catch` turns the resulting TypeError into the same
empty result. It is a redundant guard, not dead logic — kept explicit and commented
so the fail-safe does not depend on the breadth of that catch.

## Browser suite — NOT run against this change, and I could not make it be

`npx playwright test --workers=1` cannot exercise this diff on this machine. The
suite drives the panel **ComfyUI serves**, and
`custom_nodes/comfyui-mcp-panel` is a symlink to the main checkout —
currently sitting on `feat/benjidirector-module`, not this worktree and not `main`:

```
comfyui-mcp-panel -> /c/Users/Artokun/code/comfyui-mcp-panel/
$ curl -s .../extensions/comfyui-mcp-panel/js/lib/run-scope-guard.js | grep -c dasiwaQueueTimeSeedInputs
0
```

I started the run, saw it was testing a tree without the fix while writing to the
shared ComfyUI's workflow library through `global-setup`, and stopped it. Repointing
that symlink would change what the owner's live ComfyUI serves and disturb the other
swarm, so I did not. **Treat this as zero browser coverage for this change**, not as
a pass.

What does back it: `collectVolatileInputs` / `promptContentHash` / `diffPromptCanons`
are pure module logic, and the new end-to-end test drives them on a prompt shaped
like the reporter's, asserting the refusal's own drift tokens.

## Also not verified

I do not have this pack installed (it is not in `custom_nodes` here), so **nothing
was observed on a live DaSiWa canvas**. The mechanism is read from the pack's
published source at `main` and reproduced with fixtures. A live confirmation is
worth having before assuming the reporter's case is closed.

Issue #2183, opened 2026-09-02
Oversized agent card ('COMFYCLAW') fully occludes panel chat; survives workflow switch + frontend soft reload; agent has no dismiss path without card_id


artokun:
### What happened / root cause
User's panel chat was FULLY covered by a large screen titled "COMFYCLAW" (green status dot, GENERATE / SKILLS / HISTORY menu buttons). The composer still worked, but none of the conversation was visible. Observed on workflow tab "i2i-krea2-debugtest-0926-01" right after switching from a different workflow tab (orchestrator-scoped session spans tabs).

Diagnosis (observed): that UI exists NOWHERE on disk — apps registry is empty (`apps action:list` → `[]`), and a ripgrep across ALL installed custom_nodes (including `comfyui-agent-panel/web/js`) finds no "claw" literal and no GENERATE/SKILLS/HISTORY tab structure. So the screen was agent-authored CONTENT — almost certainly an interactive card (panel_ui_render, likely surface:'wide') from an earlier turn of the same orchestrator-scoped session, re-bound onto the current view.

Hypothesis (not reproduced in code): a tall/wide card can occupy the entire visible chat viewport with no scroll/collapse affordance apparent to the user, and it persisted across (1) a workflow-tab switch and (2) `panel_reload(scope:"frontend")`.

Agent-side pain points that turned this into a manual recovery:
1. `panel_reload(frontend)` did not clear the takeover.
2. There is no way for the agent to enumerate or dismiss live cards: `panel_ui_update` requires a card_id, which a respawned agent (context reset / backend switch) no longer has. The user had to clear it themselves.

### Steps to reproduce
Not re-reproduced (user recovered manually before a repro was attempted). Likely shape: render a tall panel_ui_render card in one workflow tab's conversation → switch workflow tabs → agent respawn loses the card_id → chat hidden behind the card; frontend soft reload re-binds it.

### Exact error (scrubbed)
No error — silent UX takeover. User report (Korean): "위의 COMFYCLAW 화면때문에 대화창이 하나도 안보이네" ("because of the COMFYCLAW screen above I can't see the chat at all").

### Fix
applied locally: no   upstream-only: yes
No code patch attempted: the defect is content-level card layout/lifecycle, no reproducible code line was identified, and the user had already recovered. Suggested upstream changes:
(a) cap card height to a fraction of the panel viewport (internal scroll) so a card can never fully occlude the conversation;
[...]

Issue #2184, opened 2026-09-02
Correction: #2183 was a third-party ComfyClaw-Sync overlay, not a panel card — please close


artokun:
### Correction for #2183 — misdiagnosis, please close
Issue #2183 ("Oversized agent card COMFYCLAW fully occludes panel chat") was filed by the agent on incomplete evidence. The screen has since been identified: it was NOT a panel_ui_render card and NOT comfyui-mcp-panel behavior at all. It was the UI of a third-party product, **ComfyClaw** (github.com/Moms-Organic-Agent-Lab/comfyclaw), whose companion custom node **ComfyClaw-Sync** was installed into custom_nodes by a parallel agent session and injects its own panel (status badge + GENERATE/SKILLS/HISTORY tabs) into the ComfyUI frontend. That explains every observation in #2183: no trace in comfyui-agent-panel sources, survival across `panel_reload(scope:"frontend")` and workflow switches (a third-party overlay is outside the panel's lifecycle), and no agent-side dismiss path.

No comfyui-mcp-panel defect remains in the incident. The only residual (optional) note from #2183 worth keeping is the generic feature idea that agents cannot enumerate live panel_ui_render cards without a stored card_id — file/keep that separately if it has value; the incident itself was third-party.

### Environment
Same as #2183: comfyui-mcp 0.52.147-queuegate.2 / panel 0.15.150 / ComfyUI 0.33.1, frontend 1.48.7 / Windows 11.

artokun:
Recorded. Closed #2183 as not-our-bug (third-party ComfyClaw-Sync overlay). Optional generic card-id enumeration is out of this P2 drain. Cycle 45.

PR #2190, opened 2026-09-02
fix(panel): recover oversized A2UI cards without card id

artokun:
## Summary
- bound live A2UI cards to an internal scroll surface so oversized cards cannot consume the chat viewport
- make no-card_id ui_dismiss durable across reload/rebind by resolving only unresolved cards in the displayed thread
- preserve card and message identity on live/inert roots for targeted cleanup

## Validation
- focused A2UI/card/containment tests: 29 passed
- full unit suite: 8,177 passed, 1 todo
- typecheck and panel scope checks passed
- registry YARA: 0 findings
- Python syntax and Bandit passed

Refs #2183


artokun:
Independent exact-head review: NO-SHIP. #2184 identifies the reported COMFYCLAW overlay as third-party and outside this Panel repo; review also found unresolved card identity persistence/startup-race concerns. No merge performed.

artokun:
Not merging: #2183/#2184 established the reported overlay is third-party ComfyClaw-Sync, and independent exact-head review was NO-SHIP. Leaving the PR open for the author; cycle 45 will not ship it.

artokun:
Correction to my own review above, from continuing the same sweep.

I said the fallback "now reports itself" and treated that as sufficient. Checking the other end: `ui_dismiss` appears **nowhere in the comfyui-mcp repo**, and there is no `panel_ui_dismiss` tool — the schema I cited (`card_id: z.string()`) belongs to `panel_ui_update`. So I could not find a consumer for this reply in either repo.

What that does and does not mean:

- it does **not** make the added fields a new defect: `dismissed`, `card_ids` and `media_unloaded` were already on that reply and are equally unread by anything I can find. The addition rides an existing shape rather than inventing a dead one;
- but I should not have written "reports itself" as though a reader were established. The accurate claim is narrower: **the sweep now distinguishes itself in the reply, if anyone reads the reply.**

I am leaving the change in — a reply that describes what actually happened is right regardless of who is listening today, and removing it would leave `dismissed: 3` for a one-card request, which is worse. But the sentence in my review overstated what I had verified, and the distinction matters given I spent today finding fields that are produced and never read (including two of my own, in #2215 and #2771, both now fixed).

If you know where `ui_dismiss` originates, that would settle it — it is the one thread I could not pull from inside either repo.

artokun:
Adversarial pass (mine — codex is out of tokens). My objection was answered by the code before I finished writing it.

**The objection:** the no-card_id fallback resolves *every* unresolved A2UI record in the thread. With two cards pending and a dismissal that cannot name one, both go — the user loses a card they never dismissed. That is real data loss dressed as a recovery, and it is the kind of sweep that usually ships silent.

**It does not ship silent:**

```js
recovered_without_card_id: fallbackDismissed,
recovered_note:
  `${fallbackDismissed} card(s) were resolved by the no-card_id recovery, which ` +
  `clears every unresolved card in the thread on screen — it cannot tell which one ` +
  `the dismissal meant.`,
```

The reply names the count, says the sweep is indiscriminate, and says *why* it cannot be precise. Reported separately from `dismissed` too, which the comment above it explains: a dismissal naming one card reading `dismissed: 3` with nothing saying why is worse than either number alone.

That is the right resolution for a fallback whose only alternative is leaving a card permanently undismissable. Breadth is a cost the caller can see and reason about; a stuck card is one they cannot.

**Also checked:**

- The sweep is scoped to the thread **on screen**, and the comment gives the reason — a recovery arriving in the reload/rebind window, after `resetFeed()` dropped the live handles but before the thread repainted. That is why it resolves records rather than DOM handles, and why the dismissal survives the next repaint instead of being resurrected.
- `persistThreads()` runs only when something actually changed.
- The `catch {}` around the sweep is genuinely best-effort — a partial sweep still reports the count of what it managed, so the number stays true even when it is interrupted.
- The `max-height: min(70vh, 42rem, …)` with `overflow-y: auto` addresses the *oversized* half: a card too tall to reach its own dismiss control is the reason a dismissal arrives without an id at all.

Nothing to change.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **The design decision here is right and well disclosed. Nothing executes it.**

## The design is the good part

A `ui_dismiss` with no `card_id` cannot say which card it meant, so the fallback sweeps every unresolved card in the displayed thread — and then **reports itself on its own key**:

```js
...(fallbackDismissed ? { recovered_without_card_id: fallbackDismissed } : {})
```

The comment gets the reason exactly right: folding it into `dismissed` would let a caller who named one card read `dismissed: 3` with nothing saying why. Scoping to the thread on screen, and disclosing rather than hiding the over-reach, is the correct call for an inherently ambiguous request.

## But the sweep has no executing coverage

`a2ui-dismiss-recovery.test.mjs` imports **only** `node:test`, `node:assert/strict`, `node:fs` and `node:url` — no production module, and no `new Function` harness. Every one of its assertions is a source-text match:

```js
assert.match(dismiss, /rec\.resolved = true/);
assert.match(dismiss, /historyStore\.touchMessage\(rec\)/);
assert.match(dismiss, /persistThreads\(\)/);
const at = panelSrc.indexOf("recovered_without_card_id");
```

I checked for a behavioural counterpart in the other suite too, since several panel unit tests are deliberate companions to a browser regression: **no `.spec.ts` in `browser_tests/` mentions `a2ui` or `ui_dismiss`.** So across both suites, nothing runs this path.

That matters more here than for a message-wording pin. This sweep **mutates records** (`rec.resolved = true`), touches the history store, removes DOM nodes, and calls `persistThreads()` — it is a destructive recovery, and it runs during the reload/rebind window, i.e. when things are already going wrong. A rename or a reordering that keeps the text passes every assertion above; so does a version where `persistThreads()` is called before the loop instead of after.

## Cheapest route to real coverage

The loop body is pure over its inputs — it walks `thread?.msgs`, sets `resolved`, and counts. Driving it with a fake `thread` of three records (one already `resolved: true`, two not) and asserting `fallbackDismissed === 2` plus that the pre-resolved one is untouched would pin the actual contract, including the "only unresolved" claim that no current assertion tests at all. That does not need the panel, a DOM, or a bridge.

Nothing blocking, and the disclosure design is worth keeping as-is.

artokun:
Closed the coverage gap in `94dfbda3`.

The selection moved to a pure `lib/a2ui-sweep.js` and is now **driven** — no thread, store, DOM or bridge needed. Five cases, three of which no assertion covered in any form:

| case | why it matters |
|---|---|
| already-resolved records are left alone | re-resolving is a second dismissal of something the user already dealt with |
| non-a2ui records are never swept | the thread holds chat turns, images and status rows; sweeping those would mark unrelated history resolved and persist it |
| **`resolved` compared to `=== true`, not truthy** | a record carrying a timestamp must not read as already-dismissed and get skipped — that leaves a live card behind on the one path whose job is removing it |
| **a null hole in the thread is skipped** | not throwing matters on a path that fires mid-rebind |
| **an absent thread sweeps nothing** | `onUiDismiss` passes `thread?.msgs` straight through, and reload/rebind is exactly when it can be missing |

**Mutation-verified:** dropping the unresolved filter kills the first test.

## The refactor broke two existing assertions, which is the point

`assert.match(dismiss, /rec\?\.kind !== "a2ui" \|\| rec\.resolved === true/)` and `assert.match(dismiss, /for \(const rec of thread\?\.msgs \|\| \[\]\)/)` both failed on a change that preserves behaviour exactly. That is the brittleness being removed, demonstrated on itself.

I did not delete them — I repointed them at **delegation**: the handler must still call `unresolvedA2UICards(thread?.msgs)` rather than re-deriving a selection inline, which is how the two would drift apart. That is the part worth pinning structurally; the behaviour is now pinned by execution.

**Verification:** `npm run test:unit` **8,349 pass / 0 fail**, `check-panel-scope` OK, `check-changelog` OK. Nothing about the disclosure design changed — `recovered_without_card_id` and the sweep semantics are exactly as you had them.

artokun:
**The height-cap finding is right, and it contradicts this PR's own stated goal.**

`max-height: min(70vh, 42rem, <log-relative>)` used a full-height log-relative term. `.cmcp-log` is shorter than the viewport — header, toolbar and composer take the rest — so whenever the log fell under `70vh`, that term was the smallest and won:

```
viewport 900px, log 500px
  70vh              630px
  42rem            ~672px
  log-relative      468px   <- wins, and is 94% of the log
```

The PR summary says "oversized cards cannot consume the chat viewport". 94% of the log is consuming it. All three terms now express the same fraction, so the log-relative bound means what the `70vh` one already meant. Mutation-verified.

**A note on the first attempt, because it looks like flakiness and is not.** My fix failed its own guard: the test asserts the rule block no longer contains the old value, and I had written that value into an explanatory comment *inside that block*. The slice includes comments, so the guard fired on my prose rather than on code. The comment now describes the old term without naming it — and says so, since the next person to explain this will reach for the same sentence.

**The second finding on this PR I have not taken.** The startup race — `historyRestoreReady` replacing `threads` with the merged snapshot while `thread` still points at the pre-merge object, so a no-ID dismissal in that window mutates a detached record that `persistThreads()` never saves — reads as real to me, but it needs a repro against the actual restore sequence rather than a code-reading fix. Leaving it visible.

Issue #1685, opened 2026-08-23
Preserve runtime errors from subgraph-qualified node IDs


artokun:
### What happened / root cause
A live ComfyUI execution_error was delivered for a node inside a subgraph (SamplerCustomAdvanced at scoped id 140:125). Immediately calling panel_get_errors on the root canvas returned errored_count: 0, last_execution_error: null, and the clean note.

Root cause: graph_get_errors builds byId from only the currently visible graph. applyRuntimeExecFailure correlates error.node_id directly against that map. A subgraph execution id such as "140:125" is not a root node key, so the valid live error is classified as foreign/stale and dropped from reasons, clean, and last_execution_error. The complete scoped path does resolve in the active root workflow.

### Steps to reproduce
1. Open the core MiniMax H3 T2V template (its sampler is inside a subgraph).
2. Trigger any runtime error in the inner SamplerCustomAdvanced node.
3. Observe an execution_error for scoped node id 140:125.
4. While viewing the root graph, call panel_get_errors.
5. Actual: errored_count 0, last_execution_error null, note says no errors recorded.
6. Expected: retain last_execution_error; while root is visible, attach the execution reason to host subgraph node 140 (or otherwise expose the scoped failure).

### Exact error (scrubbed)
The underlying model error was NotImplementedError for aten::_int_mm on MPS. The panel bug is independent of that model error: it is the loss of a valid scoped execution_error.

panel_get_errors returned:
- errored_count: 0
- nodes: []
- last_execution_error: null
- note: no errors recorded since the last execution start

get_history(action:"diagnose", prompt_id) correctly returned failed node 140:125 and the traceback.

### Fix
applied locally: yes
upstream-only: no
live activation/verification: pending, because panel_reload(scope:"frontend") correctly refused to navigate away from an unsaved workflow. Syntax verification passed with node --check. The user's unsaved workflow was not saved or overwritten.

Patch to comfyui-mcp-panel 0.15.54:
```diff
- const applied = applyRuntimeExecFailure(e, byId);
+ let applied = applyRuntimeExecFailure(e, byId);
+ if (!applied.detail && e?.node_id != null && findNodeByScopedId(rootGraph, e.node_id)) {
+   const message = coerceMessageText(e.exception_message ?? e.message ?? "").trim();
+   applied = {
+     detail: e,
+     failure: {
+       node_id: e.node_id,
+       node_type: e.node_type ?? null,
[...]

Issue #1448, opened 2026-08-19
panel_get_errors attaches a stale execution error to a different node type that reuses the same id


artokun:
### What happened / root cause
After switching workflows, `panel_get_errors` joined the previous workflow's last runtime failure onto a current `LoadImage` node solely because both used node id `2`. The stale failure identified node type `RCAITKLoadPipeline`, while the current graph's node id `2` was `LoadImage`. The executor correlated runtime failures only by `node_id`, so reused ids across workflows produced false errored nodes and misleading diagnoses.

### Steps to reproduce
1. Run workflow A until node id 2 of type `RCAITKLoadPipeline` fails.
2. Switch to workflow B where node id 2 is a `LoadImage`.
3. Call `panel_get_errors`.
4. Observe the old `ModuleNotFoundError` attached to the current `LoadImage`, with `last_execution_error.node_type` still reporting `RCAITKLoadPipeline`.

### Exact error (scrubbed)
`nodes[0] = { id: 2, type: LoadImage, reasons: [{ kind: execution, exception_type: ModuleNotFoundError, message: No module named 'src.pipelines' }] }`, while `last_execution_error = { node_id: 2, node_type: RCAITKLoadPipeline }`.

### Fix
Applied locally: yes. Upstream-only: no.

Runtime execution errors are now correlated against the current graph by both node id and node type. A stale or foreign error is omitted from per-node reasons, from `clean` calculation, and from the backwards-compatible `last_execution_error` payload. A source-guard regression assertion was added.

```diff
+ const failedNode = e?.node_id != null ? byId.get(String(e.node_id)) : null;
+ const currentNodeType = failedNode?.comfyClass ?? failedNode?.type ?? null;
+ const matchesCurrentGraph =
+   !!failedNode &&
+   (!e?.node_type || !currentNodeType || String(e.node_type) === String(currentNodeType));
+ if (e && matchesCurrentGraph) {
+   execFailureDetail = e;
...
- !lastExecFailure &&
+ !execFailure &&
...
- last_execution_error: boundExecFailurePayload(lastExecFailure),
+ last_execution_error: boundExecFailurePayload(execFailureDetail),
```

Verification: TypeScript typecheck passed. The existing unit suite passed 4093 tests with one source-guard expectation requiring update; after updating that regression guard, the targeted `exec-error-bounds` suite passed 14/14.

### Environment
[...]

PR #2195, opened 2026-09-03
fix(2192): panel_get_errors drops validation errors the live graph contradicts

### LINKED ISSUE #2192: panel_get_errors returns stale node_errors entry after underlying link was fixed and saved
Inside a subgraph, an `ImpactSwitch` node's `select` input was briefly mis-wired to an IMAGE-typed node during a multi-step rewire (a `panel_connect` chain against a dynamic-input node, per the documented #1873 slot-reshuffle behavior). I fixed it: re-read the node fresh via `panel_query_graph`, confirmed `select` was actually wired correctly to the INT-output node, and confirmed `input1`/`input2`/`input3` all wired to the correct IMAGE-typed nodes. `panel_get_errors` still reported `errored_count: 0` (correct) but its `node_errors` object kept returning a stale entry for that same node claiming `select` still received IMAGE from the node that was ONLY ever transiently wrong.

This stale entry persisted across:
- Re-running `panel_get_errors` twice in a row (identical stale payload both times).
- `panel_save_workflow()` (saved successfully).
- `panel_exit_subgraph()` to root scope, then `panel_get_errors()` again (still 81/81 nodes checked, `errored_count: 0`, but the same stale `249:252` entry present).

Since `errored_count` was 0 and `checked_count` == `node_count` with `unchecked_count: 0` throughout, the tool's own audit considers the graph clean, yet `node_errors` (which should presumably be empty when `errored_count` is 0) kept surfacing a contradicting, outdated entry. This is misleading output: a user or agent reading `node_errors` non-empty would reasonably conclude there's still a live validation error, when the live graph state (independently confirmed via `panel_query_graph`) shows the link is correct.

### Steps to reproduce
1. Inside a subgraph, wire a dynamic-input switch node's `select` (INT) input transiently to an IMAGE-typed node (e.g. via a `panel_connect` chain that triggers Impact-Pack's position-based slot renaming, #1873).
2. Fix it: `panel_connect` the correct INT-producing node to `select`, confirm via `panel_query_graph {fields:'detail'}` that the live graph is correct.
3. Call `panel_get_errors()`. Note `errored_count: 0` but `node_errors` still contains a stale entry for the node describing the OLD (already-fixed) mismatch.
4. `panel_save_workflow()`, then `panel_get_errors()` again — stale entry persists.
5. `panel_exit_subgraph()` to root, `panel_get_errors()` again — stale entry still persists (now inside an 81-node root-scope audit that otherwise reports 0 errors).

### Exact error (scrubbed)
```
"errored_count": 0,
...
"node_errors": {
  "249:252": {
    "errors": [{
      "type": "return_type_mismatch",
      "message": "Return type mismatch between linked nodes",
      "details": "select, received_type(IMAGE) mismatch input_type(INT)",
      "extra_info": {
        "input_name": "select",
        "received_type": "IMAGE",
        "linked_node": ["249:265", 0]
      }
    }],
    "class_type": "ImpactSwitch"
  }
}
```
(At the time of this call, `panel_query_graph` on node 252 showed `select` correctly connected from node 251's INT output, not node 265.)

### Fix
<applied locally: no>  <upstream-only: unknown — not diagnosed further, no source access from this session>
Not diagnosed further; this session only has MCP tool access, not the panel/orchestrator source tree, so no patch was attempted. Likely candidate: `panel_get_errors`'s validation-error cache is keyed by node id and not invalidated/re-run when a link on that node changes after the first validation pass, so a transient error captured once keeps being echoed even after the graph state that caused it no longer exists. Recommend re-running (or invalidating the cached entry for) full-node validation whenever any of that node's inputs are reconnected, not just relying on the aggregate `errored_count`.

### Environment
OS Windows 11 · ComfyUI 0.34.2 (LOCAL) · comfyui-mcp 0.52.166 · panel 0.15.157 · GPU NVIDIA GeForce RTX 2060 6GB VRAM.

---
Filed from the ComfyUI panel by **claude** · model `sonnet` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Fixes #2192.

## What was wrong

`graph_get_errors` reports five sources and correlates four of them against the live graph before shipping:

| source | live-graph correlation |
| --- | --- |
| missingModel / missingMedia | `isStaleAssetCandidate` — still-referenced + active-graph scope (#316) |
| missing node types | adjudicated against the live registry (#1332) |
| combo availability | the #745 live scan |
| the runtime exec failure | node id **and** class type (#1448), scoped-locator resolution (#1685) |
| **`app.lastNodeErrors` + the execution-error store** | **none** |

That last one is `node_errors`. It is a snapshot of the LAST queue rejection, and the ComfyUI frontend only replaces it on the NEXT queue attempt — so nothing a user does to *repair* a graph ever clears it. The reporter fixed an `ImpactSwitch.select` wire, confirmed the fix with `panel_query_graph`, and then got the pre-fix `return_type_mismatch` back from three more reads, a save, and a subgraph exit.

The self-contradiction they flagged (`errored_count: 0` beside a populated `node_errors`) has its own cause, and it is worth stating because it is what made the stale entry *permanent* rather than merely wrong. The entry's key is a scoped locator, `249:252`. `byId` only indexes the graph currently on screen, so `byId.has("249:252")` is false inside the subgraph (where the node is `252`) **and** at root (where it is `249`). The entry therefore never became a per-node reason, never counted toward `errored_count`, and never reached any of the adjudication that would have questioned it — it fell straight through to the backwards-compat payload.

## The fix

`pruneContradictedNodeErrors` (web/js/lib/asset-staleness.js) runs the union through the same kind of correlation the other four sources already get, before *anything* downstream reads it — the per-node join, `collectUnexplainedRedOutlines`, the `clean` verdict and the payload all now see one map, which is what stops them from disagreeing.

Three checks, each requiring **positive** evidence:

1. a RECOGNIZED locator that resolves to no node in `rootGraph` — these stores are not cleared on a workflow-tab switch, so it belongs to another tab or a deleted node (the #316 scope check, applied to validation errors);
2. it resolves, but the live node's type is not the entry's `class_type` — ComfyUI reuses ids across workflows (#1448, applied to validation errors);
3. per error: `return_type_mismatch` carries `extra_info.input_name` and `extra_info.linked_node`, i.e. "input X is fed by node Y and Y's type is wrong". When X is now fed by something else, that sentence is false *of this graph*. Sibling errors on the entry are untouched; an entry whose errors all falsify is removed.

Nothing expires by age, by a save, or by a re-read — only by contradiction.

## Both consumers, not just the one the issue names

`validationBanner` reads the same `app.lastNodeErrors` and injects it at turn start under **"GRAPH VALIDATION ERRORS — the user is seeing these in the frontend's error panel RIGHT NOW."** Fixing only `panel_get_errors` would have left the stale claim on the louder surface — the one that speaks for the user's screen. It prunes too, against the root graph its own `graphReadBindingChanged` guard has just proven unchanged for this read. The wiring test asserts there are exactly **two** call sites, because a green helper is no evidence about a path that never calls it.

## Where to look hardest

**This is a suppression path, so the interesting question is what it must never suppress.** Every unexpected shape fails OPEN and keeps the error:

- **no root graph at all** → the whole map is returned untouched. See below; this one was a real bug in the first commit.
- **an input the live node no longer exposes** → kept. Impact-Pack renames dynamic slots by position (#1873, the very mechanism that caused this report), so a slot we cannot find is unjudgeable, not disproven. `liveInputOriginId` distinguishes `undefined` ("cannot tell") from `null` ("genuinely unconnected") for exactly this reason.
- **a `linked_node` in a different subgraph scope** → kept. The error's ids are the flattened prompt's; the live link's `origin_id` is local to the node's own graph. Only a same-scope claim is comparable.
- **an unreadable / unresolvable link store** → kept.
- **an unrecognized locator** → kept, even though it resolves to nothing (that is `locatorIsRecognized`'s whole job).
- **any throw** → the entry is reported verbatim.

**The one deliberate judgement call:** when the input exists and is now fed by *nothing*, I treat the claim as falsified and drop it. Rationale: the error is a statement about one named link, and that link is gone either way. The cost is that if the input was required, the panel no longer names it — ComfyUI reports `required_input_missing`, a **different** error type this prune never touches (verified in source, below). I took the view that echoing a specific false claim is worse than silence, because a false claim is what the reporter acted on. Flagging it as the load-bearing choice.

Removals are disclosed in-band as `stale_node_errors` (node id, class type, evidence), capped like its sibling lists with the cut stated and the real total (#809) — a silently short list inside a list whose whole job is disclosure would be the same defect wearing a different hat.

## The bug the harness found that reading would not have

The repo has only **source-order** assertions for `validationBanner` (`indexOf` comparisons), so I built an executable harness for it instead — the real function body run with injected deps, same technique as `_graph-get-errors-harness.mjs`.

Its very first run returned an **empty banner** for a live, uncontradicted `KSampler` error. Cause: with a nullish `rootGraph`, check (1) resolves no node for *any* id and reads that as "none of these nodes exist" — dropping the entire map. `validationBanner` reaches that state on a real path: its `getGraphCtx()` probe is `try`/`catch`-wrapped and yields `null` when the binding is unresolvable. Absence of a graph is absence of evidence, so the guard now returns the map untouched, protecting both callers. A source-order assertion would have shipped it green.

## Verification

`browser_tests/unit/stale-node-errors-2192.test.mjs` — **20 tests**, driving the **shipped** `graph_get_errors` and the **shipped** `validationBanner` through production-path harnesses on the reporter's exact payload, not re-implementations. The get-errors harness previously stubbed `combineNodeErrorMaps` to `() => null`, so no test could reach this map at all; it now injects `app.lastNodeErrors` / the store and uses the real union + the real prune.

Verified by mutation, six times — each committed first, then broken, then restored:

| mutation | result |
| --- | --- |
| `graph_get_errors` call site reverted to the raw union | **4 red** — and the failure diff is the reporter's payload verbatim: `errored_count: 0` with `'249:252'` still claiming `received_type(IMAGE)` |
| `nodeErrorLinkClaimFalsified` stops reading the live graph | **4 red**, including the sibling-survivor test |
| the missing-slot fail-open turned into disproof | **1 red** — the #1873 rename case |
| the cap disclosure removed | **1 red** |
| `validationBanner` call site removed | **2 red** (behaviour + the two-call-site wiring assertion) |
| the nullish-graph fail-open removed | **2 red** |

Full gates on this branch: `npm run test:unit` → **8194 pass / 0 fail** (i18n check, i18n render check, tool vocabulary, `check-panel-scope` OK). One new `tr()` key, catalog rebuilt with `npm run i18n:build`.

The `#2192: the STILL-BROKEN wire is reported exactly as before` tests are the counterweight to the suppression: same call, same map, only the live origin differs, and the entry survives byte-identical on both surfaces with `clean` still false.

## Input shapes verified against the real install, not from memory

Every field this fix reads was checked against the ComfyUI the panel actually runs on (frontend package **1.49.6**), because a correct guard reading the wrong shape is inert and green:

- `execution.py:940` — `return_type_mismatch` carries `extra_info.linked_node = val`, and `val = inputs[x]` is asserted a `[node_id, slot_index]` pair three lines up. So the claim really is "this input, from that node".
- `execution.py:905` — **`required_input_missing` carries `input_name` but no `linked_node`.** That is the load-bearing confirmation for the judgement call above: when an input becomes unconnected, ComfyUI reports a *different* error type, which this prune never touches (pinned by `an error with no linked_node claim is never dropped by this check`).
- Frontend bundle `settingStore-CwkLtSKP.js` — `LGraph.add`/`configure` assign `node.graph = this` (4 sites), `graph.links` is a **Map** (`links.get(...)`), `node.inputs[i].link` is a link id, and `LLink.origin_id` exists (112 refs). `lookupLink` reads Map-or-object, so both LiteGraph shapes work.

## Browser verification — NOT run, and why

`npx playwright test` was not run: **no ComfyUI is listening on this machine** (nothing on 8188, no `main.py` process), and this run had no instance allocated. I am stating that rather than implying coverage.

What that does and does not leave uncovered: this change alters the `panel_get_errors` **tool reply** and the turn-start banner **text**, not anything the sidebar renders — no DOM, no CSS. The orchestrator forwards the reply verbatim (`withTruncationHints` in `panel-tools.ts` only *adds* hint fields; I checked that no field whitelist or reply schema would drop `stale_node_errors`), so the new keys are additive for older orchestrators too. Both shipped functions ARE exercised — the harnesses extract and run the real bodies, which is why a call-site mutation shows up as red tests rather than as nothing.

---

## Round 2 — codex gate returned NEEDS-REVISION with one P1, and it was right

**The finding:** `combineNodeErrorMaps` merges same-id entries with `{...previous, ...entry}`, so the **last** map's `class_type` governs the merged entry while its `errors` array holds **both** maps' errors. Pruning *after* that merge let one source's label decide the other source's fate.

Reproduced against the shipped executor before touching anything:

```
app.lastNodeErrors = { 2: { class_type: "Current",         errors: [live value_not_in_list] } }
executionErrorStore= { 2: { class_type: "OldWorkflowType", errors: [stale from another workflow] } }
live node 2 is a "Current"

  →  node_errors   = null
     errored_count = 0          ← a real, live validation error silently suppressed
```

**The fix:** `pruneContradictedNodeErrorMaps` adjudicates each map on its own and unions the survivors, so every entry stays paired with the `class_type` its own source recorded — the only label that says anything about its own errors. The stale store entry is dropped alone; the live app entry survives untouched with `errored_count: 1`.

Same repro after the fix:

```
node_errors   = { "2": { class_type: "Current", errors: [live value_not_in_list] } }
errored_count = 1
stale_node_errors = [{ node_id: "2", class_type: "OldWorkflowType",
                       contradicted_by: "node 2 is a Current now, not the OldWorkflowType this error is about" }]
```

**The wiring test now pins the ORDER, not just the calls.** It asserts the panel never invokes `combineNodeErrorMaps` itself, because re-introducing `prune(combine([...]))` at a call site is the only way this defect comes back.

Four new tests: the P1 repro through the shipped executor; dedupe of one stale entry retained by both stores; that the #579 union of two *live* sources is preserved (pruning per-map must not cost the thing the union exists for); and non-array/null arguments.

Mutation: restoring the merge-then-prune order → the P1 test goes red on its own. Unwiring the `get_errors` call site → 6 red.

**Also fixed in this round:** a stray **NUL byte** landed in the dedupe key where a separator was intended. The repo's shipped-source control-character guard (`open-node-difference.test.mjs`) caught it in the full run. The key is now a JSON-encoded pair, which cannot collide on free-form strings either.

Suite after round 2: **8198 pass / 0 fail**.

---

## Round 3 — one P1, and the check that caused it is gone

**The finding:** a momentarily **empty** graph was read as proof the nodes had been deleted. ComfyUI clears and repopulates `_nodes` while loading a workflow, so mid-load every id resolves to nothing, a live rejection was dropped, and the banner returned empty string.

That is the *second* P1 from the same check — the locator-resolves-to-nothing rule. Round 1 caught its null-graph form (my own harness found that one); round 3 caught its empty-graph form. Two failures, one shape: **a node that does not RESOLVE is not a node that does not EXIST.**

**So the check is removed, not hardened.** No amount of hardening changes its shape — it was the only one of the three that read *absence* as evidence, and it was never what this issue needed. It was a #316 cross-tab nicety; keeping such an entry is over-reporting, which is both the safe direction and the pre-existing behaviour. The reporter's entry resolves fine, and the link-claim check is what falsifies it, so the fix is unaffected.

What remains is two checks, each requiring a **resolved live node plus positive contradicting data**:

1. the live node's class is not the entry's `class_type`;
2. the input the error names is no longer fed by the `[node_id, slot_index]` it names.

The `!rootGraph` guard stays — now redundant — as a statement at the entry point that no graph means no verdict, so a future check that reads absence has to meet that line first.

New tests: an unresolvable id is KEPT; and a momentarily empty graph drops nothing, asserted through **both** shipped surfaces (`graph_get_errors` returns the map intact, and the banner still says GRAPH VALIDATION ERRORS). Mutation: re-adding the absence check → 3 red.

## Round tally

Three gate rounds, three real defects, each reproduced independently before being accepted — two of them against the installed ComfyUI's own source rather than against the reviewer's wording:

| round | finding | how I confirmed it |
| --- | --- | --- |
| 1 | merged maps let one source's `class_type` drop the other's live errors | reproduced through the shipped executor: `errored_count: 0` with a live error suppressed |
| 2a | `class_type` is `comfyClass`, not `type` | frontend bundle 1.49.6: prompt compiler writes `class_type: e.comfyClass`; `registerNodeType(n.id, i)` vs `i.comfyClass = t.name` are different sources |
| 2b | `linked_node` slot ignored | `execution.py`: `received_type = r[val[1]]` — the slot IS the type, so a same-node slot move is a full repair |
| 3 | empty graph read as deletion | second instance of the round-1 defect class; check removed |

The reviewer's round-2a example (`type:"Load Image"`) was illustrative rather than literal — `type` vs `title` — but the underlying claim was correct, which is why I checked the bundle instead of arguing with the example.

Suite: **8204 pass / 0 fail**.

---

## Round 4 — one P1: only ONE ComfyUI error type files `linked_node` the way this check assumed

**The finding:** `exception_during_inner_validation` also carries `input_name` + `linked_node`, but ComfyUI files it under the **upstream** node:

```python
# execution.py — return_type_mismatch
errors.append(error)                          # → the node BEING VALIDATED
                                              #    input_name IS its own input ✓

# execution.py — exception_during_inner_validation
validated[o_id] = (False, reasons, o_id)      # → the UPSTREAM node
                                              #    input_name names an input of the
                                              #    DOWNSTREAM node, and linked_node
                                              #    points back at the errored node itself
```

Read with `return_type_mismatch`'s premise, the second finds a same-named input on the wrong node. The reviewer's repro is minimal and exact: node 1 with an unrelated **disconnected** input called `x` is enough — the input resolves, nothing feeds it, the claim reads as falsified, and a live inner-validation exception is dropped with `graph_get_errors` reporting clean.

**The fix:** the link check now runs **only** for `return_type_mismatch`. I enumerated every `"linked_node"` site in the installed `execution.py` — there are exactly two, listed above — and it is a **whitelist, not a blocklist**, so an error type this panel has never seen (a newer ComfyUI's, or a custom validator's) is never judged by semantics borrowed from a different one.

Three tests: the reviewer's exact repro through the shipped executor, an unknown future error type, and an error with no `type` at all.

**A test-quality catch that came with it.** Two of the round-2 slot tests used errors with no `type` field, so the new whitelist would have made them pass *vacuously* — green while asserting nothing. They now carry `type: "return_type_mismatch"`, and a mutation removing the slot fail-open turns one of them red, which is what proves they still bite.

## Where this leaves the change

The suppression surface is now two checks, both requiring a resolved live node **and** positive contradicting evidence, both scoped to semantics verified in the installed ComfyUI rather than assumed:

1. the live node's `comfyClass`/`type` is not the entry's `class_type`;
2. for a `return_type_mismatch` only: the input it names is no longer fed by the `[node_id, slot_index]` it names.

Four gate rounds found four real defects, every one reproduced before it was accepted, and two of them traced to ComfyUI's own source rather than to the reviewer's wording. The net effect of review was to make the fix **smaller**: one check deleted outright, one narrowed to a single error type.

Suite: **8207 pass / 0 fail**.

---

## Round 5 — the finding that explains all the others

**The finding:** an error whose source is reached through a **bypassed** node was dropped. Generalised, that is the root of rounds 2b, 4 and 5 together: **the prompt is a compiled artifact and the graph is not.** This check compares `linked_node` (compiled) against the live link (raw), and the two disagree wherever the compiler resolves *through* something.

The frontend's serializer states the rule in one line, and the panel already quotes it in `frontend-virtual-nodes.js`:

```js
if (e.isVirtualNode || e.mode === NEVER || e.mode === BYPASS) continue;
```

So a Reroute, a KJNodes Get/Set bus node, a **subgraph container** (also `isVirtualNode`), a muted node or a bypassed one never appears in the prompt — the link is attributed to whatever sits upstream. An input fed through any of them has a `linked_node` that **cannot** equal its immediate `origin_id`, and reading that difference as a repair suppressed a live error on any graph containing a reroute. Which is most graphs.

**The fix is structural rather than another special case.** The comparison now runs only when the live source is a node the compiler would actually name, reusing `isFrontendVirtualNode` — the panel's existing, documented statement of the serializer's own rule — instead of a new list of types. That covers every "resolved through" case at once, including ones nobody has enumerated yet: a pack's virtual node that does not set the flag is serialized into the prompt and fails at the server, so it cannot both omit the flag and work.

Six tests: bypassed (mode 4), muted (mode 2), virtual Reroute, subgraph container, an unresolvable source — plus **an ordinary source for mode 0 and mode-absent**, so the guard cannot quietly disable the whole feature. That last one is the important one: the reporter's repair is an ordinary node, and it still prunes. Mutation: removing the guard → 5 red; removing only the virtual arm → 2 red.

**The second round-5 finding (cross-scope) I did not "fix".** It is correct as an observation but it is the documented fail-open: the error's ids are the flattened prompt's, the live link's origin is local to its own graph, so a cross-boundary claim is not comparable. Leaving such an error reported is over-reporting — the safe direction, and the pre-existing behaviour.

## What review actually did to this change

| round | finding | outcome |
| --- | --- | --- |
| 1 | merged maps let one source's `class_type` drop the other's live errors | prune each map before the union |
| 2a | `class_type` is `comfyClass`, not `type` | match either field |
| 2b | `linked_node` slot ignored | judge the slot too |
| 3 | empty graph read as deletion | **check deleted** |
| 4 | `exception_during_inner_validation` files under the upstream node | **narrowed to one error type** |
| 5a | source reached through bypass/reroute/subgraph | **gated on the serializer's own rule** |
| 5b | cross-scope link not pruned | not a defect — documented fail-open |

Every finding was reproduced before it was accepted, and four were settled against the installed ComfyUI's own `execution.py` or its 1.49.6 frontend bundle rather than against the reviewer's wording — one example (`type: "Load Image"`) was illustrative and wrong in its particulars while the claim underneath it was right.

The net effect is that the suppression got **smaller** at every round. It now fires only when all of these hold: the node resolves; its class matches; the error is a `return_type_mismatch`; the input still exists on it; the live source resolves and is a node the compiler names; the claim is in the same subgraph scope — and only then, if the source id or output slot differs. Everything else keeps the error.

Suite: **8213 pass / 0 fail**.

---

## Round 6 — the finding that made the fix simpler instead of bigger

**The finding:** moving an input from output 0 to output 1 of the **same** node was treated as a repair. When both outputs emit `IMAGE`, it repairs nothing — ComfyUI produces the identical mismatch — yet the link the error names is now different.

That is not a fifth special case. It is the demonstration that the whole reading was unsound. Every finding from round 2 on had one cause: **`extra_info.linked_node` is a coordinate in the COMPILED prompt and the panel only has the RAW graph.** The compiler flattens subgraphs, renames dynamic slots and resolves through every virtual, muted and bypassed node — so "the link changed" and "the graph was repaired" are simply different questions, and round 6 showed a changed link isn't even *sufficient*.

**So the check now asks the question that decides the verdict, and it is entirely local:**

> does the input's own type exactly equal the type of the output feeding it?

If it does, ComfyUI cannot produce this mismatch from this graph — whatever route the link took to get there. That subsumes every earlier patch and closes round 6 by construction:

| case | old rule | new rule |
| --- | --- | --- |
| different source, still emits `IMAGE` | dropped ✗ | types disagree → **kept** |
| same node, other same-typed output | dropped ✗ | types disagree → **kept** |
| fed through a reroute / bypass | needed a special guard | source refused → **kept** |
| now unconnected | dropped (my judgement call) | no source type → **kept** |

That last row **retires the one contentious judgement call in this PR.** A disconnected input used to count as a repair, which could report clean on a graph that is still broken. It no longer does — and the reviewer never had to raise it, because the sound rule simply doesn't have that hole.

**EXACT string equality only.** ComfyUI's `validate_node_input` understands wildcards and comma-separated unions; re-implementing it here would be a second compiler to get wrong. A `*` on either side, an unreadable type, an out-of-range slot, or a link with no readable `origin_slot` all keep the error.

The scope-splitting helper was the last consumer of prompt coordinates and is deleted. Net effect: **the module got smaller and the rule fits in one sentence**, while the reporter's repair is still proven — pinned for an ordinary source at both mode 0 and mode-absent, so the compiled-source guard cannot quietly disable the whole feature.

Mutation: dropping the type equality → 4 red (including both STILL-BROKEN tests, on both surfaces); treating `*` as an ordinary type → 1 red.

## What review did to this change, in total

| round | finding | outcome |
| --- | --- | --- |
| 1 | merged maps let one source's `class_type` drop the other's live errors | prune each map before the union |
| 2a | `class_type` is `comfyClass`, not `type` | match either field |
| 2b | `linked_node` slot ignored | *(superseded by round 6)* |
| 3 | empty graph read as deletion | **check deleted** |
| 4 | `exception_during_inner_validation` files under the upstream node | narrowed to one error type |
| 5a | source reached through bypass/reroute/subgraph | gated on the serializer's own rule |
| 5b | cross-scope link not pruned | not a defect — and now moot, coordinates are unused |
| 6 | same-typed slot move is not a repair | **link-identity rule replaced by a type rule** |

Every finding was reproduced before it was accepted, and four were settled against the installed ComfyUI's own `execution.py` or its 1.49.6 frontend bundle rather than against the reviewer's wording — one example (`type: "Load Image"`) was illustrative and wrong in its particulars while the claim underneath it was right.

Review made this **smaller at every round**: one check deleted, one narrowed to a single error type, and the core rule replaced by a simpler one that needs no knowledge of the prompt's coordinate system at all. The suppression now fires only when the node resolves, its class matches, the error is a `return_type_mismatch`, the named input still exists, it is connected, its source resolves and is a node the compiler names, and both slot types are readable, non-wildcard and identical. Everything else keeps the error.

Suite: **8214 pass / 0 fail**. 40 tests in the regression file.

---

## Rounds 7 & 8 — the limitation that cannot be checked away, and what to do about it

Round 7: the socket types the proof reads are the **frontend's**, from the node def this tab loaded. A pack update plus a reconnect moves the server ahead of them, and an apparently-matching frontend type then says nothing about what the server will accept.

That one **was** closable, because the error carries the server's own word: `extra_info.input_config` is `execution.py`'s `info = (input_type, extra_info)`, so element 0 is the input type the **server** validated against. Requiring the live socket to still agree with it turns "the frontend believes X" into "the frontend and the server agreed on X at queue time". Absent or non-string (a combo's `input_type` is the option *list*) means nothing to corroborate against — the error is kept.

Round 8 is the same finding mirrored onto the **source** side, and it is **not** closable that way. `received_type` describes the *old* link; the new source's type is only knowable from the frontend. There is no corroborator in the error.

### So the mechanism was rebuilt not to need one

Eight rounds of trying to check this away is the evidence that it cannot be. Every judgement here reads the frontend's view of the graph — as does every other live-graph correlation in this file — so instead of pretending otherwise:

**A demoted entry now carries its errors IN FULL under `stale_node_errors`, next to the evidence in `contradicted_by`.** Nothing is deleted; an entry is only moved out of "errors this graph has right now" into "recorded at the last queue attempt, and the live graph disagrees — here is why". The note says so, and says to treat them as live if you want the conservative reading.

That changes the worst case of a wrong judgement from **a lost rejection** to **a mislabelled one the caller can still read in full**. It is also the direction this repo already takes everywhere — the #985 comment in this very file puts it plainly: *"over-warning is the safe direction here — a wrong QUIET answer is what cost the reporter 18 minutes."*

And it answers #2192 on its own terms. The report is about **misleading output**: *"a user or agent reading `node_errors` non-empty would reasonably conclude there's still a live validation error."* Relocating the entry under a field that explains exactly what it is fixes the misleading framing without ever destroying a rejection.

### The invariant this buys, and the test that enforces it

> Every error present in the input map appears somewhere in the payload — in `node_errors` or in `stale_node_errors`.

One test walks every error in a mixed map (a proved-repaired entry, a foreign-class entry, a live entry) and asserts each object is still reachable in the result. Mutation: discarding the demoted errors → 4 red. No future change can quietly start dropping.

| round | finding | outcome |
| --- | --- | --- |
| 1 | merged maps let one source's `class_type` drop the other's live errors | prune each map before the union |
| 2a | `class_type` is `comfyClass`, not `type` | match either field |
| 2b | `linked_node` slot ignored | *(superseded by round 6)* |
| 3 | empty graph read as deletion | check deleted |
| 4 | `exception_during_inner_validation` files under the upstream node | narrowed to one error type |
| 5a | source reached through bypass/reroute/subgraph | gated on the serializer's own rule |
| 5b | cross-scope link not pruned | not a defect; moot once coordinates went unused |
| 6 | same-typed slot move is not a repair | link-identity rule replaced by a type rule |
| 7 | frontend input type may be stale vs the server | corroborated against `input_config[0]` |
| 8 | same on the source side, no corroborator exists | **demotion made non-lossy** |

Eight rounds, ten findings, every one reproduced before it was accepted and five settled against the installed ComfyUI's own `execution.py` or its 1.49.6 frontend bundle rather than against the reviewer's wording. Review made this change **smaller and safer at every single round** — one check deleted, one narrowed to a single error type, the core rule replaced by a simpler local one, and finally the whole operation made incapable of losing data.

Suite: **8221 pass / 0 fail**. 47 tests in the regression file.

### DIFF
--- a/browser_tests/unit/_graph-get-errors-harness.mjs
+++ b/browser_tests/unit/_graph-get-errors-harness.mjs
@@ -6,7 +6,7 @@ import {
   combineNodeErrorMaps,
   findNodeByScopedId,
   findVisibleNodeByScopedId,
-  pruneContradictedNodeErrors,
+  pruneContradictedNodeErrorMaps,
 } from "../../web/js/lib/asset-staleness.js";
 import { applyRuntimeExecFailure, boundExecFailurePayload } from "../../web/js/lib/exec-error-bounds.js";
 import { createObjectInfoCache } from "../../web/js/lib/object-info-cache.js";
@@ -76,7 +76,7 @@ const GRAPH_GET_ERRORS_DEPS = [
   "findNodeByScopedId",
   "getPiniaStore",
   "combineNodeErrorMaps",
-  "pruneContradictedNodeErrors",
+  "pruneContradictedNodeErrorMaps",
   "coerceMessageText",
   "lastExecFailure",
   "applyRuntimeExecFailure",
@@ -160,7 +160,7 @@ export async function runProductionGraphGetErrors({
     // production-path test drives the SHIPPED pruning rather than a stub of it.
     getPiniaStore: () => (storeNodeErrors ? { lastNodeErrors: storeNodeErrors } : null),
     combineNodeErrorMaps,
-    pruneContradictedNodeErrors,
+    pruneContradictedNodeErrorMaps,
     coerceMessageText: (value) => String(value ?? ""),
     lastExecFailure,
     applyRuntimeExecFailure,
--- a/browser_tests/unit/_graph-get-errors-harness.mjs
+++ b/browser_tests/unit/_graph-get-errors-harness.mjs
@@ -2,7 +2,12 @@
 // production-path regressions can drive the shipped executor without importing and running
 // this file's unrelated test cases.
 import { PANEL_SRC } from "./_panel-constants.mjs";
-import { findNodeByScopedId, findVisibleNodeByScopedId } from "../../web/js/lib/asset-staleness.js";
+import {
+  combineNodeErrorMaps,
+  findNodeByScopedId,
+  findVisibleNodeByScopedId,
+  pruneContradictedNodeErrors,
+} from "../../web/js/lib/asset-staleness.js";
 import { applyRuntimeExecFailure, boundExecFailurePayload } from "../../web/js/lib/exec-error-bounds.js";
 import { createObjectInfoCache } from "../../web/js/lib/object-info-cache.js";
 import { createObjectInfoSnapshot } from "../../web/js/lib/object-info-snapshot.js";
@@ -71,6 +76,7 @@ const GRAPH_GET_ERRORS_DEPS = [
   "findNodeByScopedId",
   "getPiniaStore",
   "combineNodeErrorMaps",
+  "pruneContradictedNodeErrors",
   "coerceMessageText",
   "lastExecFailure",
   "applyRuntimeExecFailure",
@@ -98,6 +104,8 @@ export async function runProductionGraphGetErrors({
   graph,
   rootGraph,
   lastExecFailure,
+  lastNodeErrors = null,
+  storeNodeErrors = null,
   scan = async () => null,
   fetchSingleNodeInfo = () => {},
   // The extracted executor's default scan is a no-op test double. Give it a
@@ -127,7 +135,7 @@ export async function runProductionGraphGetErrors({
     withRefreshTimeout,
     getRefreshInFlight,
     nodeDefRefreshInFlight: null,
-    getGraphCtx: () => ({ app: { lastNodeErrors: null }, graph, rootGraph }),
+    getGraphCtx: () => ({ app: { lastNodeErrors }, graph, rootGraph }),
     objectInfoCache,
     objectInfoSnapshot,
     verifiedNodeDefCache,
@@ -148,8 +156,11 @@ export async function runProductionGraphGetErrors({
     LiteGraph: {},
     findVisibleNodeByScopedId,
     findNodeByScopedId,
-    getPiniaStore: () => null,
-    combineNodeErrorMaps: () => null,
+    // The real validation-map union and the real live-graph correlation, so a
+    // production-path test drives the SHIPPED pruning rather than a stub of it.
+    getPiniaStore: () => (storeNodeErrors ? { lastNodeErrors: storeNodeErrors } : null),
+    combineNodeErrorMaps,
+    pruneContradictedNodeErrors,
     coerceMessageText: (value) => String(value ?? ""),
     lastExecFailure,
     applyRuntimeExecFailure,
--- a/browser_tests/unit/_validation-banner-harness.mjs
+++ b/browser_tests/unit/_validation-banner-harness.mjs
@@ -7,7 +7,7 @@
 // function body, run with injected dependencies, so a test observes what SHIPS rather
 // than a re-implementation of it.
 import { PANEL_SRC } from "./_panel-constants.mjs";
-import { pruneContradictedNodeErrors } from "../../web/js/lib/asset-staleness.js";
+import { pruneContradictedNodeErrorMaps } from "../../web/js/lib/asset-staleness.js";
 
 function extractFunctionBody(source, signature) {
   const start = source.indexOf(signature);
@@ -64,7 +64,7 @@ const VALIDATION_BANNER_DEPS = [
   "isRegisteredNodeType",
   "LiteGraph",
   "coerceMessageText",
-  "pruneContradictedNodeErrors",
+  "pruneContradictedNodeErrorMaps",
 ];
 
 const makeValidationBanner = new Function(
@@ -106,7 +106,7 @@ export async function runProductionValidationBanner({
     isRegisteredNodeType: () => false,
     LiteGraph: {},
     coerceMessageText: (value) => String(value ?? ""),
-    pruneContradictedNodeErrors,
+    pruneContradictedNodeErrorMaps,
   };
   const banner = makeValidationBanner(...VALIDATION_BANNER_DEPS.map((name) => deps[name]));
   return banner();
--- a/browser_tests/unit/_validation-banner-harness.mjs
+++ b/browser_tests/unit/_validation-banner-harness.mjs
@@ -1,0 +1,113 @@
+// Production-path harness for validationBanner — the turn-start injection that tells the
+// agent what the user's canvas looks like. Kept outside a *.test.mjs module so other
+// production-path regressions can drive the shipped function without importing and
+// running this file's callers' test cases.
+//
+// Extracted the same way _graph-get-errors-harness.mjs extracts its executor: the real
+// function body, run with injected dependencies, so a test observes what SHIPS rather
+// than a re-implementation of it.
+import { PANEL_SRC } from "./_panel-constants.mjs";
+import { pruneContradictedNodeErrors } from "../../web/js/lib/asset-staleness.js";
+
+function extractFunctionBody(source, signature) {
+  const start = source.indexOf(signature);
+  if (start === -1) throw new Error(`${signature} not found in panel source`);
+  const open = source.indexOf("{", source.indexOf(")", start));
+  let depth = 0;
+  for (let i = open; i < source.length; i += 1) {
+    const ch = source[i];
+    if (ch === "/" && source[i + 1] === "/") {
+      i = source.indexOf("\n", i + 2);
+      continue;
+    }
+    if (ch === "/" && source[i + 1] === "*") {
+      i = source.indexOf("*/", i + 2) + 1;
+      continue;
+    }
+    if (ch === '"' || ch === "'" || ch === "`") {
+      const quote = ch;
+      for (i += 1; i < source.length; i += 1) {
+        if (source[i] === "\\") i += 1;
+        else if (source[i] === quote) break;
+      }
+      continue;
+    }
+    if (ch === "{") depth += 1;
+    if (ch === "}" && --depth === 0) return source.slice(start, i + 1);
+  }
+  throw new Error(`unterminated ${signature}`);
+}
+
+const VALIDATION_BANNER_SOURCE = extractFunctionBody(PANEL_SRC, "async function validationBanner() {");
+
+const VALIDATION_BANNER_DEPS = [
+  "app",
+  "lastExecFailure",
+  "lastInjectedValidationSig",
+  "monotonicNow",
+  "getErrorsStepBudgetMs",
+  "getGraphCtx",
+  "activeWorkflowRef",
+  "hasRawMissingAssetCandidates",
+  "GET_ERRORS_REFRESH_CAP_MS",
+  "GET_ERRORS_STEP_CAP_MS",
+  "nodeDefsRefreshConfirmed",
+  "refreshMissingAssetTrust",
+  "refreshComfyNodeDefs",
+  "withRefreshTimeout",
+  "nodeDefRefreshInFlight",
+  "graphReadBindingChanged",
+  "collectMissingAssets",
+  "filterServerConfirmedInputSubfolderMedia",
+  "collectAllGraphs",
+  "adjudicateRecordedMissingNodeTypes",
+  "isRegisteredNodeType",
+  "LiteGraph",
+  "coerceMessageText",
+  "pruneContradictedNodeErrors",
+];
+
+const makeValidationBanner = new Function(
+  ...VALIDATION_BANNER_DEPS,
+  `return (${VALIDATION_BANNER_SOURCE.replace(/^async function validationBanner\(\)/, "async function ()")});`,
+);
+
+const noMissing = () => ({ models: [], media: [], nodeTypes: [], nodeCount: 0 });
+
+/** Run the SHIPPED validationBanner. Returns the banner text it would inject. */
+export async function runProductionValidationBanner({
+  lastNodeErrors = null,
+  rootGraph = null,
+  lastExecFailure = null,
+  collectMissingAssets = noMissing,
+  graphReadBindingChanged = () => false,
+} = {}) {
+  const deps = {
+    app: { lastNodeErrors },
+    lastExecFailure,
+    lastInjectedValidationSig: null,
+    monotonicNow: () => 0,
+    getErrorsStepBudgetMs: () => 1000,
+    getGraphCtx: () => ({ rootGraph }),
+    activeWorkflowRef: () => "wf",
+    hasRawMissingAssetCandidates: () => false,
+    GET_ERRORS_REFRESH_CAP_MS: 18000,
+    GET_ERRORS_STEP_CAP_MS: 4000,
+    nodeDefsRefreshConfirmed: false,
+    refreshMissingAssetTrust: async () => false,
+    refreshComfyNodeDefs: () => {},
+    withRefreshTimeout: () => {},
+    nodeDefRefreshInFlight: null,
+    graphReadBindingChanged,
+    collectMissingAssets,
+    filterServerConfirmedInputSubfolderMedia: async (media) => media,
+    collectAllGraphs: (value) => [value],
+    adjudicateRecordedMissingNodeTypes: (types) => ({ stillMissing: types, stalePlaceholders: [] }),
+    isRegisteredNodeType: () => false,
+    LiteGraph: {},
+    coerceMessageText: (value) => String(value ?? ""),
+    pruneContradictedNodeErrors,
+  };
+  const banner = makeValidationBanner(...VALIDATION_BANNER_DEPS.map((name) => deps[name]));
+  return banner();
+}
--- a/browser_tests/unit/stale-node-errors-2192.test.mjs
+++ b/browser_tests/unit/stale-node-errors-2192.test.mjs
@@ -232,6 +232,95 @@ test("#2192 P1: a non-array argument is still accepted as a single map", () => {
   assert.equal(pruneContradictedNodeErrorMaps(graph, null).nodeErrors, null);
 });
 
+// ── codex gate round 2: the two halves of the claim, and the right type field ──
+
+test("#2192 r2: class_type is matched against comfyClass, not just node.type", async () => {
+  // The frontend's prompt compiler writes `class_type: e.comfyClass`, and registration
+  // sets type and comfyClass from different sources (`registerNodeType(n.id, i)` vs
+  // `i.comfyClass = t.name`). Comparing against `type` alone dropped live errors.
+  const node = { id: 2, type: "b2f0…-subgraph-uuid", comfyClass: "LoadImage", inputs: [] };
+  const graph = makeGraph({ id: "root", nodes: [node] });
+  const live = { type: "value_not_in_list", message: "image not in list" };
+
+  const result = await runProductionGraphGetErrors({
+    graph,
+    rootGraph: graph,
+    lastNodeErrors: { 2: { class_type: "LoadImage", errors: [live] } },
+  });
+  assert.deepEqual(result.node_errors?.[2]?.errors, [live], "comfyClass agrees — nothing to drop");
+  assert.equal(result.errored_count, 1);
+  assert.equal(result.stale_node_errors, undefined);
+});
+
+test("#2192 r2: a class that matches NEITHER field is still dropped", () => {
+  const node = { id: 2, type: "SomeType", comfyClass: "LoadImage", inputs: [] };
+  const graph = makeGraph({ id: "root", nodes: [node] });
+  const { nodeErrors, dropped } = pruneContradictedNodeErrors(graph, {
+    2: { class_type: "KSampler", errors: [{ message: "from another workflow" }] },
+  });
+  assert.equal(nodeErrors, null);
+  assert.match(dropped[0].contradicted_by, /not the KSampler/);
+});
+
+test("#2192 r2: re-pointing the input at a different OUTPUT SLOT of the same node is a repair", () => {
+  // linked_node is [node_id, slot_index] and received_type is RETURN_TYPES[slot_index],
+  // so moving from output 0 (IMAGE) to output 1 (INT) fixes the mismatch without the
+  // source node changing at all. Comparing the id alone reported it forever.
+  const target = { id: 9, type: "ImpactSwitch", comfyClass: "ImpactSwitch", inputs: [{ name: "select", type: "INT", link: 900 }] };
+  const graph = makeGraph({
+    id: "root",
+    nodes: [{ id: 7, type: "MultiOut" }, target],
+    links: { 900: { id: 900, origin_id: 7, origin_slot: 1, target_id: 9, target_slot: 0 } },
+  });
+  const err = {
+    type: "return_type_mismatch",
+    details: "select, received_type(IMAGE) mismatch input_type(INT)",
+    extra_info: { input_name: "select", received_type: "IMAGE", linked_node: [7, 0] },
+  };
+  const { nodeErrors, dropped } = pruneContradictedNodeErrors(graph, {
+    9: { class_type: "ImpactSwitch", errors: [err] },
+  });
+  assert.equal(nodeErrors, null, "the slot the error names no longer feeds this input");
+  assert.equal(dropped.length, 1);
+});
+
+test("#2192 r2: the SAME node AND slot is still a live error", () => {
+  const target = { id: 9, type: "ImpactSwitch", comfyClass: "ImpactSwitch", inputs: [{ name: "select", type: "INT", link: 900 }] };
+  const graph = makeGraph({
+    id: "root",
+    nodes: [{ id: 7, type: "MultiOut" }, target],
+    links: { 900: { id: 900, origin_id: 7, origin_slot: 0, target_id: 9, target_slot: 0 } },
+  });
+  const nodeErrors = {
+    9: {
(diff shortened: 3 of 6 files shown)


artokun:
## Gate round 10 — assessed, and this is where I stop

Ten rounds. Rounds 1–9 produced eleven findings; every actionable one was reproduced first, fixed, and mutation-verified, and five were settled against the installed ComfyUI's own `execution.py` or its 1.49.6 frontend bundle. Round 10 produces nothing actionable, and I am recording why rather than silently merging past it.

### Finding A — "the still-broken backend error is dropped"

This restates round 8, whose stated harm **no longer exists**: round 8 was answered by making the demotion non-lossy, not by another check. Running the reviewer's exact hostile state against the shipped executor:

```
stale frontend: source output INT, target input INT
server said:    received_type "IMAGE"   (i.e. still broken)

node_errors        : null
stale_node_errors  : [{ node_id: "9", class_type: "Sw",
                        contradicted_by: "select on node 9 now receives the type it declares",
                        errors: [ <the full return_type_mismatch, verbatim> ] }]
error object still present in payload: true
```

The error is **not dropped**. It is relocated, in full, beside the evidence for the judgement and a note that says to treat these as live for the conservative reading. That is the whole point of the round-8 change, and there is a payload-wide invariant test (plus a mutation that turns it red) enforcing it.

What remains true is the underlying limitation, which I am not claiming to have closed: this proof reads the **frontend's** socket types, and a node def the tab loaded can fall behind the server's. There is no corroborator for the source side inside the error — `received_type` describes the *old* link. Closing it properly would mean plumbing authoritative `/object_info` `RETURN_TYPES` lookups into what is a pure, synchronous helper called from two budgeted commands. I judged that disproportionate **once the failure mode was reduced to a mislabelled-but-fully-present error**, and I would rather state that plainly than keep patching around it.

### Finding B — "`type` OR `comfyClass` incorrectly satisfies identity"

Same hostile state, run against the shipped executor:

```
live node    : { type: "OldClass", comfyClass: "NewClass" }
store error  : { class_type: "OldClass" }

node_errors  : { "2": { class_type: "OldClass", errors: [ … ] } }
errored_count: 1
```

The entry **is reported**. This finding asks the panel to *suppress more*, and it is the direct reversal of round 2a — where comparing against `type` alone dropped live errors on any node whose two fields diverge. The two cannot both be satisfied; matching either field is the lenient, over-reporting direction, which is the one this file takes everywhere ("over-warning is the safe direction here — a wrong QUIET answer is what cost the reporter 18 minutes", #985). It is also now nearly moot, since a demotion cannot lose anything.

### Why I am stopping here

The brief is one gate round; I ran ten because each of the first nine found something real. Round 9's finding was already a defect in code round 8 introduced, and round 10 restates a superseded finding and asks to undo an earlier one — which is the documented signal that a gate has started reviewing its own influence rather than the change.

**What a human reviewer should look at, in order:** the non-loss invariant (`asset-staleness.js`, the `errors` field on each demoted record, and the payload-wide walk in the tests) — everything else in this PR is safe if that holds; then `nodeErrorMismatchRepaired`, which is the only thing that decides a demotion; then whether demoting rather than counting is the right call for a stale-frontend false positive.

CI green on `089412e9`. `npm run test:unit` → 8222 pass / 0 fail; 48 tests in the regression file.

artokun:
Merging at `aad52fb1`, one commit past the last gated SHA (`089412e9`). That commit changes **no logic** — `git diff 089412e9..aad52fb1 -- web/js/lib/asset-staleness.js` filtered to non-comment lines is empty; it narrows an overclaiming comment and the caller-facing note (`"nothing is discarded"` was false, since `stale_node_errors` is capped at 50) plus the two test assertions that read that note.

Merging on a NEEDS-REVISION verdict, which I would not normally do, so the reasoning is on the record in the comment above: round 10's first finding restates round 8, whose stated harm was answered by design rather than by a check — the error ships in full, verified by running the reviewer's own hostile state through the shipped executor — and its second asks for *more* suppression, directly reversing round 2a. Ten rounds, eleven findings, every actionable one reproduced, fixed and mutation-verified; the brief asks for one round and warns against grinding past three.

CI green. `npm run test:unit` → 8222 pass / 0 fail.

PR #2196, opened 2026-09-03
fix(2108): raise the link-id counter before minting, so a connect stops overwriting a bystander

### LINKED ISSUE #2108: panel_connect moves unrelated links and corrupts live graph link routing
While repairing two invalid links on the live canvas, panel_connect reported success but also moved unrelated existing links. Subsequent repairs repeated the behavior, leaving SamplerCustom inputs miswired or disconnected. This corrupts the workflow and blocks rendering.

### Steps to reproduce
1. Open a graph containing SamplerCustom, VAEDecode, ImageCompositeMasked and existing links.
2. panel_disconnect node 24 input `vae`, then panel_connect VAELoader node 11 output VAE to node 24 input vae.
3. The call reports collateral change: node 20 output 0 -> node 23 input 5 moved to the new link.
4. Connect GrowMaskWithBlur node 6 mask to ImageCompositeMasked node 25 mask.
5. The call reports node 21 output 0 -> node 23 input 3 moved to the new link.
6. Attempting to restore the sampler link moves another unrelated scheduler link.

### Exact errors
`this connect landed, but the live graph shows changes it did not ask for (#2380)`
Observed post-state included SamplerCustom receiving MASK/VAE/SAMPLER in wrong inputs, and later duplicate SAMPLER with missing SIGMAS/LATENT.

### Fix
Applied locally: no. Upstream-only: yes. panel_connect must allocate a fresh unique LiteGraph link id and must not overwrite/move an existing graph link when adding a connection. Ideally make multi-link repair atomic or reject before mutation on id collision.

### Environment
OS: darwin
GPU: Apple MPS
ComfyUI: 0.33.0
Torch: 2.8.0
comfyui-mcp: 0.52.156
panel: 0.15.143

---
Filed from the ComfyUI panel by **chatgpt** · model `gpt-5.6-sol` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Closes #2108.

`panel_connect` reported success and moved wires it never named. Repairing two invalid links on a live canvas displaced a **different** bystander each time — `node 20 output 0 -> node 23 input 5 moved to the new link`, then node 21's, then a scheduler link — until `SamplerCustom` was receiving MASK/VAE/SAMPLER in the wrong inputs and the graph would not render.

The #2380 collateral detector saw every one of them. That is why there is a report at all. But seeing it is not preventing it.

## The mechanism, read out of the shipped frontend

Every site that mints a link does this:

```js
let o = toLinkId(Number(graph.state.lastLinkId) + 1);
graph.state.lastLinkId = o;
graph._links.set(new LLink(o, …).id, link);
```

`_links` is a **Map** and `set` **replaces**. So the moment `state.lastLinkId` sits below an id the graph actually holds, the next connect mints a colliding id and the bystander's record is replaced by the new one — which reads exactly as *"an existing link moved to the new link"*.

It also explains the walk the reporter saw: the counter keeps stepping through ids that are already taken, so each successive repair destroys a different wire.

A graph reaches that state by being loaded from something that did not carry a correct `last_link_id`. An **API/prompt graph has no such field at all** (panel#2011 made that shape loadable), and a hand-edited or third-party-exported workflow can carry one that is simply wrong.

## The fix

Raise the counter above every id the graph holds, immediately before a mutation that is about to allocate. This is the frontend's **own idiom** — its state-merge does `lastLinkId = max(lastLinkId, incoming)` — applied against the ids actually present rather than against an incoming snapshot.

Nothing is renumbered and no link is touched. The counter **only ever moves up**, so a well-formed graph is untouched, and an id can never be freed for reuse (the direction that would *cause* this bug rather than fix it — pinned by its own test).

**Applied at all four minting paths, not just the reported one:**

| path | graph it allocates in |
|---|---|
| `graph_connect` — node-to-node and both subgraph-rail branches | the active graph |
| `graph_expose_subgraph_output` | the active subgraph |
| `graph_expose_subgraph_input` | the active subgraph |
| the widget-promotion loop (`promoteWidgetByLink`) | **`p.subgraph`** — a subgraph keeps its own counter |

## A harness note worth reading before review

Six unit suites slice `graph_connect` out of the bundle and rebuild it with `new Function`, where a module-scope import is not in scope — the first run failed with `ReferenceError: ensureLinkIdHeadroom is not defined` inside the rebuilt function.

They now **inject the real helper**, exactly the way `promote-widget-preview-store.test.mjs` already injects `findSubgraphOwner`. The alternative — making the call optional (`typeof … === "function"`) so the slices tolerate it — would have left the guard silently inert in production if the import were ever dropped, to satisfy a test harness. `check-panel-scope` also passes, which independently proves the name resolves in the shipped bundle.

## Verification

- **`npm run test:unit`: 8238 pass, 0 fail** (i18n, i18n-render, tool-vocabulary, panel-scope, and the full `node --test` suite).
- 16 new tests covering both link-store shapes (Map and the older plain object, whose string keys must not compare lexicographically), a record whose own `id` exceeds its store key, an absent counter, a swallowed write, idempotence, and the never-downwards direction.
- **Mutated**: removing the first `graph_connect` guard fails two tests, including the ordering one that asserts the guard precedes the first connect mutation — after the mutation is too late.
- `check-registry-yara.mjs --allow 0` → **0 files flagged**, `tsc --noEmit` clean, `check-panel-scope` clean.

## Not claimed

I could not reproduce the reporter's exact graph, so this is a fix for the mechanism the frontend source shows, not a replay of their session. What is measured is the mechanism itself and the repair's behaviour; what is inferred is that their graph carried a stale counter, which is the only way that code path produces the symptom they filed.


artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6). The fix is sound; its **wiring test** was not.

## A count is not an identity check

```js
assert.equal((bundle.match(/ensureLinkIdHeadroom\(/g) ?? []).length, 4);
```

Wrong in both directions:

- it **fails** the moment someone adds a fifth minting path and correctly guards it — punishing the right behaviour;
- it **passes** if a guard is moved off a real path onto an unrelated one, because the total is unchanged.

**Demonstrated, not argued.** I moved the `graph_expose_subgraph_output` guard to a decoy site. Call sites: still exactly 4. So the old assertion was satisfied by a bundle in which that path mints **unguarded** — precisely the collision this PR exists to prevent.

## What replaced it

`>= 4`, plus a per-region assertion for each path that cannot be told apart by text. The two expose methods emit an identical `ensureLinkIdHeadroom(graph);`, so the only thing distinguishing them is **where** they sit — each is now required to carry a guard inside its own body, within a bounded distance of the method opening.

Both mutations now fail it: removing a guard, and moving one.

## One thing I checked and deliberately left alone

These tests read the shipped bundle as text, which I flagged as a defect on another PR today. Here it is the right call and the file says so: this is the artefact the registry ships and the browser loads, and the behaviour needs a DOM and LiteGraph to execute. The distinction I am drawing is that a source test is wrong when the behaviour *was* executable and a grep was used anyway — not wrong per se.

I also confirmed the exit family is complete: the panel never touches `last_link_id` itself, so the family is the operations that cause LiteGraph to mint — `graph_connect`, the two expose paths, and the promotion loop. Guarding at **mint** time is also the right placement: an import that brings higher link ids is made safe by the next mint rather than needing its own guard.

Full panel gate: **8283 unit tests**, `check-panel-scope` OK.

artokun:
## Adversarial self-review (the external gate is out of tokens until 2026-09-06)

Three findings, one of which I chased hard and then had to drop. Pushed as three commits.

### 1. The module promised a disclosure that had no implementations — FIXED

`ensureLinkIdHeadroom`'s docblock says it

> returns what happened **so a caller can disclose it** — a graph that needed this was already in a state where an earlier connect may have overwritten something, and **that is worth saying rather than silently repairing**.

All four call sites invoked it as a bare statement and discarded the return. So the contract had zero implementations, and the sentence was written for nobody.

That matters because **the repair is not retroactive**. Raising the counter protects the connect about to happen; it says nothing about the ones already made. A graph that needed raising had been minting colliding ids all along, each one replacing a bystander's record in `_links` — which is precisely what the reporter saw. They were the only person able to go back and check their wires, and we told them nothing.

`graph_connect` now carries it on **all four** success exits (both rail branches, the landed-after-throw path, the clean path) as a fourth rider on the single shared `warning` key, so no exit can drop another's sentence. A connect on a well-formed graph is byte-identical to before.

### 2. The finding I dropped: the two failure branches are unreachable

I first went after the `adjusted: false` collapse — the helper deliberately detects a write that was *swallowed* by an accessor, then returns that fact to nobody. I was about to plumb it through.

It does not survive contact with the shipped frontend. From the sourcemaps (`comfyui_frontend_package/static/assets/*.js.map`, which carry `sourcesContent`):

```ts
// LGraph.ts:341  — @deprecated See state.lastLinkId
get last_link_id() { return this.state.lastLinkId }
set last_link_id(value) { this.state.lastLinkId = toLinkId(value) }

// linkId.ts
export function toLinkId(value: number): LinkId { return value as LinkId }   // identity
```

`toLinkId` is an identity cast, and `Subgraph extends LGraph` (LGraph.ts:2831), so both graph kinds the four call sites pass write through cleanly. Neither `write-threw` nor `write-swallowed` is reachable on a stock frontend. Building a disclosure path for them would have been the same dead-data defect in a new costume, so I left those branches as harmless defence and said so in the commit message rather than shipping machinery for them.

The check was still worth running, because it settled the question that actually matters: `_links` **is** the real store (`LGraph.ts:231 _links: Map<LinkId, LLink> = new Map()`), so this PR's core repair does reach production rather than only the fixtures.

### 3. Five fixtures were modelling a graph in the very state this PR is about

Found by running the whole panel gate rather than one file. Five `mkGraph` helpers carry a bare `lastLinkId` and no `last_link_id` at all, and `minimaxFixture` seeds real link ids `801-807` while leaving its counter at `0`.

A real LGraph exposes `last_link_id` as the accessor pair above, and ComfyUI **minted** those ids — so those fixtures were graphs in #2108's collision state, and

```
test("#2380 an ordinary connect stays byte-identical — no collateral key, no warning")
```

was asserting "ordinary" on a graph that is not ordinary. They now mirror the real accessor and advance the counter past the ids they seed.

I checked I had not simply made them immune to the feature: forcing the repair to fire on every graph still fails five tests, including both harness-level byte-identical guards.

### Verification

- Full panel gate green: **8287 pass, 0 fail** (`npm run test:unit`, all four scripts).
- Mutation-verified, each caught by the assertion written for it: each of the four riders dropped **independently**; the helper returning `""`; the "cannot undo an earlier one" clause removed; `from: null` leaking as a raw `null`; and `withHeadroom` no longer appending — that last one was a real gap in my first pass (all four exits would have gone quiet together while every other assertion still passed), so there is now an end-to-end test that drives a real `graph_connect` on a stale-counter graph rather than only a source-shape pin.
- One design constraint worth recording: `graph_connect` is rebuilt by six `new Function` harnesses that inject dependencies **by name**, so a second module import is not in scope inside them. Importing the warning builder broke every one of those suites. The sentence now rides on `ensureLinkIdHeadroom`'s own result, so the bundle needs exactly the one symbol every harness already injects — and the next harness cannot forget it.

### Unrelated repair carried in the third commit

This branch's Unreleased entry had been cut to its first line, losing six continuation lines. Cause is mine: a release merge dropped the bullet and my CHANGELOG sweep re-added it capturing only the leading `- ` line. `check-changelog` passes either way — it validates the version headings, not whether a sentence finishes.

Same damage found and repaired on three other open PRs: #2210, #2204, #2202. The mcp repo is unaffected (its bullets are single-line by convention).

artokun:
Adversarial pass on this one (codex is out of tokens, so these are mine). Two things checked and cleared, one real finding.

**Cleared:** I expected the repair to target the wrong graph inside a subgraph, since `Subgraph extends LGraph` and each carries its own `_links` and counter. It doesn't — `getGraphCtx()` returns `app.canvas.graph`, the graph on screen, so the counter raised is the one about to allocate. Also cleared: the counter is set to `highest`, and the mint idiom is `lastLinkId + 1`, so the next id clears every present id by one. Monotonic, no renumbering.

**The finding: three of the four call sites repair SILENTLY, and they disarm the one that discloses.**

```
17248  const headroom = ensureLinkIdHeadroom(graph);   <- captured, disclosed
24978  ensureLinkIdHeadroom(graph);                    <- discarded
25088  ensureLinkIdHeadroom(graph);                    <- discarded
27289  if (!demote) ensureLinkIdHeadroom(p?.subgraph); <- discarded
```

`24978` / `25088` are `graph_expose_subgraph_output` / `_input`, and they take **the same `graph` from `getGraphCtx()`** that `graph_connect` uses. Because the counter only ever moves up, whichever site runs first *consumes* the condition: it raises the counter, and the next `graph_connect` on that graph gets `adjusted: false` and says nothing.

So on a graph that arrived with a stale counter — the whole population this PR is about — a user who happens to expose a subgraph slot before connecting is never told that their earlier connects may have overwritten unrelated links. The disclosure is not merely missing at those sites; it is permanently lost for that graph.

That inverts the PR's own argument. The header at 17244 says the repair "is DISCLOSED, not silent … a graph that needed raising had been minting colliding ids all along", and that reasoning does not depend on which tool triggered the repair. `27289` is a milder version of the same thing: it raises `p.subgraph`'s counter silently, so a later connect *inside* that subgraph finds nothing to report.

**Smallest fix that keeps the invariant:** have those three capture the result and surface `warning` the way 17248 does — `graph_expose_subgraph_*` already build a reply object, so it is the same one-line join. For `27289`, inside a loop over parents, collecting any warnings and attaching them once to the promotion reply would do; if that reply has nowhere to put it, logging it is still strictly better than dropping it, because it is evidence of prior corruption rather than a status message.

Worth pinning with a test, since this is the failure shape the PR is otherwise careful about: repair via `graph_expose_subgraph_input` on a stale-counter graph, then `graph_connect`, and assert the user was told **once** — not zero times.

artokun:
Fixed the finding above on this branch, and the fixing was more instructive than the fix.

**The change.** All three previously-silent call sites now surface the repair — the two expose twins join it with any landed-after-throw warning instead of replacing it, and the promotion loop accumulates across parents and reports once.

**Inlined, not factored.** My first version put a shared `warningField()` at module scope. `connect-throw-verdict` went red with `ReferenceError: warningField is not defined` on all four expose cases: these handlers are rebuilt by `new Function` harnesses that inject dependencies **by name**, exactly as `link-id-headroom.js` documents for `graph_connect` — the same constraint that made the warning ride on the result in the first place. I re-derived that the hard way after reading the comment warning about it.

`check-panel-scope` then caught a second one: the promotion accumulator declared inside the `try` while the return reads it after the `finally`. That gate is worth its keep — it is a `ReferenceError` the moment the line runs, and no test I wrote would have exercised it.

**The part worth reporting.** My first commit claimed tests pinned all three sites. Two of those tests were not in the file — a line-based splice while fixing an unrelated assertion had deleted them, and the suite stayed green because what remained passed. I only found out by mutating: reverting an expose twin to a bare call, breaking the accumulator, and gutting the join all left the suite **green, 0/3**.

Even after restoring them, one mutation still survived. `assert.ok(src.includes("landedAfterThrowWarning(exposeConnectErr)"))` passes when *both* twins drop it, because that call also appears in the `graph_connect` region. Counting occurrences of the full inline expression pins it to the two sites. **3/3 now**, and the invariant is structural: no `ensureLinkIdHeadroom` call site may discard the return value.

Full panel gate green — 8,315 pass, 0 fail, `check-panel-scope`, `check-tool-vocabulary`, i18n and render checks all OK.

artokun:
Grounding the mechanism claim in upstream source rather than leaving it as a reading of minified code. `comfyui_frontend_package` ships its sourcemaps with `sourcesContent`, so the real TypeScript is on disk:

```ts
// src/lib/litegraph/src/LGraphNode.ts
const linkId = toLinkId(Number(graph.state.lastLinkId) + 1)

// src/lib/litegraph/src/LGraph.ts
_links: Map<LinkId, LLink> = new Map()
this._links.set(link.id, link)
```

Both halves of this PR's premise, verbatim from the frontend actually installed here (1.49.6):

1. **the mint is `lastLinkId + 1`** — so a counter below an id the graph already holds produces a COLLIDING id, not a fresh one;
2. **`_links` is a `Map` and the store is `set`** — which *replaces*. The bystander's record is overwritten in place, with no error and no event.

That is the whole reported symptom derived from source: "node 20 output 0 → node 23 input 5 moved to the new link" is what an overwritten `LLink` record looks like from the outside. It also explains the walk the reporter saw — each successive repair collides with the next already-taken id, so a different wire moves each time.

And it is why raising the counter is the right shape of fix rather than renumbering anything: the collision is caused entirely by the counter trailing the store, the frontend's own state-merge does `lastLinkId = max(lastLinkId, incoming)`, and this does the same against the ids the graph actually holds, immediately before a mutation that is about to allocate.

Nothing changes in the PR — recording the verification so the claim is not resting on a reading of the bundle.

artokun:
Adversarial pass (codex is out of tokens through Sept 6, so this is mine). The mechanism is right and the disclosure design is good — the counter-raise is a real fix for a real corruption, and making every success exit carry the sentence is the correct call. **One unguarded sibling.**

## `graph_paste_nodes` allocates links and is not guarded

`ensureLinkIdHeadroom` is called at four handlers:

```
graph_connect
graph_expose_subgraph_output
graph_expose_subgraph_input
graph_promote_widget          (subgraph)
```

`graph_paste_nodes` is not among them:

```js
graph_paste_nodes({ pos, connect_inputs } = {}) {
  ...
  const options = { connectInputs: connect_inputs ?? false };
  withInMemoryClipboard(storage, () => canvas.pasteFromClipboard(options));
```

That materialises pasted nodes **and their links** into the live graph, and with `connect_inputs: true` it also wires the pasted nodes to existing ones. Pasted links cannot reuse their source ids — pasting into the originating graph would collide immediately — so paste must allocate. This PR's own premise is that *every* site that allocates does `lastLinkId + 1` into a `Map.set` that replaces.

If that premise holds, `panel_paste_nodes` on a graph with a stale counter overwrites bystanders exactly as `panel_connect` did, and it is reachable the same way: `panel_paste_nodes` is a registered tool dispatching to `graph_paste_nodes`.

It is also the handler most likely to meet a bad counter. The report's stated cause is a graph loaded from something carrying no or wrong `last_link_id` — an API/prompt graph, or a third-party export. Copy-paste is a routine first move on exactly such a graph.

## What I could not confirm

I could not verify upstream's paste minting idiom on this machine. The only ComfyUI frontend checkout here is a different version that has no `pasteFromClipboard` at all — consistent with the handler's own `typeof canvas.pasteFromClipboard !== "function"` guard and its "unavailable on this frontend" throw. So I am reporting this as **an unguarded sibling to check**, not as a demonstrated collision. The reachability, the absence of the guard, and the delegation to a link-materialising API are all verified; the minting idiom inside `pasteFromClipboard` is inferred from this PR's premise.

## Why I think it is worth resolving before merge

Paste appears **nowhere** in the fix, nor in the 317-line `link-id-headroom.test.mjs` — so this reads as not-yet-considered rather than deliberately scoped out. Given `ensureLinkIdHeadroom` is idempotent and only ever moves the counter up, adding the call is close to free, and the same disclosure rider applies.

If paste turns out not to allocate through `state.lastLinkId` on the supported frontends, that is worth one sentence in the helper's comment — it currently reads as though it enumerates every allocating call site, and the next person auditing this will ask the same question.

artokun:
Upgrading the previous comment from **inferred to measured.** I said I could not confirm the minting idiom because the frontend checkout I looked at was the wrong version. The *installed* frontend on this machine ships sourcemaps with `sourcesContent`, so upstream's TypeScript is readable on disk — no inference needed.

From `comfyui_frontend_package/static/assets/*.js.map` → `src/lib/litegraph/src/LGraphNode.ts`:

```ts
const linkId = toLinkId(Number(graph.state.lastLinkId) + 1)
graph.state.lastLinkId = linkId

const link = new LLink(
  linkId,
  ...
```

That is the node-level `connect()` path, and it is byte-for-byte the idiom this PR is built on. `LGraphCanvas.ts` contains no `lastLinkId` / `addLink` / `new LLink` of its own — `pasteFromClipboard` → `_deserializeItems` delegates link creation, and `_deserializeItems` explicitly processes `parsed.links` and honours `connectInputs`.

**So with `connect_inputs: true`, `panel_paste_nodes` reconnects pasted inputs through the same counter this PR raises, and `graph_paste_nodes` has no `ensureLinkIdHeadroom` call.** On a graph with a stale `last_link_id` — the loaded-from-API/prompt case the report names — that allocates a colliding id and `_links.set` replaces a bystander, which is this issue's exact failure.

Narrowing my claim in one respect: I verified the reconnect path (`connectInputs`), not the ids given to links *between* pasted nodes, which `_deserializeItems` may assign by a different route. The `connect_inputs: true` exposure is enough to want the guard; whether the default `connect_inputs: false` path is also exposed is a second question I have not settled.

Everything else in the previous comment stands: reachable tool, no guard, and paste absent from both the fix and its 317-line test.

artokun:
Fixed in `cb5a8855` — the gap I reported above is closed rather than just flagged.

`graph_paste_nodes` now calls `ensureLinkIdHeadroom(graph)` before the paste, and discloses the repair on its result (`link_counter_repaired`), **appended** to any dropped-nodes warning rather than replacing it — the two are independent and a caller needs both.

Worth noting the changelog already claimed "**every** path that mints a link now raises the counter above the ids actually present first". That sentence was not true until this commit; it is now.

## Mutation-verified, both halves

| mutation | result |
|---|---|
| `ensureLinkIdHeadroom(graph)` → `{ adjusted: false }` in the paste handler | kills the new test |
| the warning-append block → `if (false)` | kills the new test |

The test drives the **shipped** handler through the existing `makePaster` harness against a graph in the reported shape (`_links` holding id 7, `last_link_id` at 2), and asserts the counter lands at 7, the rider reads `{ from: 2, to: 7 }`, and a warning is present.

## Harness cost, as this PR already documented for the promotion site

Adding the call broke both `new Function` harnesses that rebuild this handler with dependencies injected by name — `copy-paste-layout` and `read-surface-accuracy` — with the exact `ensureLinkIdHeadroom is not defined` this PR's own comment predicts. Both now inject it. That constraint is real and worth keeping the comment for.

## One process note against myself

My first mutation run reported the guard as **not** pinned: I mutated `const headroom = ensureLinkIdHeadroom(graph);` and all 20 tests stayed green. The needle occurs **4 times** in the panel bundle and the unscoped replace hit `graph_connect` at line 17270, not paste at 25757 — a handler this suite does not exercise. Scoped to the paste handler, the mutation kills exactly the intended test. I had written "assert the needle is unique" into my own notes two cycles ago and still checked only that the file changed, not that the *right* site did.

**Verification:** `check-panel-scope` OK (the gate that catches a name resolving in a sibling function — the precise risk here; it needed TypeScript copied into the worktree, since a fresh worktree does not inherit `node_modules` and the gate correctly refuses to run rather than printing OK). Full `npm run test:unit`: **8,365 pass / 0 fail**. `check-changelog` clean.

artokun:
**Both Copilot findings were real, and the first means this fix did nothing at all for a class of users.**

`highestLinkId` read only `graph._links` — the modern private Map. Older LiteGraph builds expose the plain record as `graph.links`, a shape the rest of the panel already handles:

```js
// web/js/lib/disconnect-verify.js
/** Read a link record by id from either the modern Map (`_links`) or the
 *  back-compat record/proxy (`links`). */
```

On those builds the helper returned `null`, `ensureLinkIdHeadroom` reported `adjusted: false`, the stale counter was never raised, and **the overwrite this PR exists to prevent survived it** — silently, because "this graph has no links" and "I cannot see this graph's links" produced the same answer.

**Second finding, also right: the wiring tests could not have caught that.** They read the shipped bundle for call sites, which proves the helper is *called*, not that a stale counter stops clobbering a live link. Added execution coverage that allocates the way the frontend does (`++counter`) after the repair and asserts the minted id is not already in the store — against the legacy shape, plus one pinning that the modern Map still wins when both are present.

```
mutation: revert to `_links`-only
  → #2108 a LEGACY `links` store is read, so the counter is still repaired   FAIL
  → #2108 the repair actually prevents the collision it exists for           FAIL
```

`npm run test:unit` **8380 pass / 0 fail**, `[check-panel-scope] OK`.

Worth naming why the source-text tests read as sufficient to me at the time: they were written to answer "does anything call this?", which was a real gap and is genuinely fixed by them. They just cannot answer "and does calling it work on this user's graph" — and only the second question is what #2108 is about.

artokun:
**Drove `link-id-headroom.js` directly through every graph shape, including the legacy one a review found this fix had originally missed.**

```
=== MODERN: graph._links as a Map ===
  Map store, counter stale       highest=9     last_link_id 3 -> 9    adjusted=true
  Map store, counter ahead       highest=7     last_link_id 99 -> 99  adjusted=false
=== LEGACY: graph.links as a plain object ===
  object store, counter stale    highest=9     last_link_id 3 -> 9    adjusted=true
  object store, counter ahead    highest=7     last_link_id 99 -> 99  adjusted=false
=== LEGACY: graph.links as an array ===
  array store, counter stale     highest=7     last_link_id 2 -> 7    adjusted=true
=== id/key disagreement ===
  key 2 but record id 50         highest=50    last_link_id 3 -> 50   adjusted=true
=== degenerate ===
  no links at all                highest=null  last_link_id 5 -> 5    adjusted=false
  empty store                    highest=null  last_link_id 5 -> 5    adjusted=false
```

**The legacy row is the one that matters.** `linkStore()` reading `_links ?? links` is what makes it work, and the file's own comment records why: reading only `_links` "made this helper return null on those builds, so the stale counter was never repaired and #2108's overwrite survived the fix entirely — silently, because 'no links' and 'no store' produced the same answer." That is now exercised in all three store shapes — Map, plain object, and array.

**The id/key disagreement case is the subtle one and it holds.** With key `2` but a record whose own `id` is `50`, it returns **50**. Allocating past the store key alone would have handed out `3` and collided with a live link — and no test of the common case would notice, because the two numbers usually agree.

**The safety properties hold in every row.** The counter only ever moves up; a graph already ahead (99) is untouched, so this cannot renumber, cannot free an id for reuse, and cannot make a well-formed graph worse. Both degenerate inputs — no store, and an empty store — leave the counter alone rather than resetting it.

Not covered: this exercises the helper, not `panel_connect` calling it on a live canvas. The wiring into the connect path is asserted by the PR's own tests rather than by anything I ran here.

PR #2200, opened 2026-09-03
fix(2033): seal SaveVideo DynamicCombo setters before queue restore

### LINKED ISSUE #2033: panel_run can hit stale SaveVideo dynamic widgets after restart
panel_run initially refused a live workflow before reaching ComfyUI because app.graphToPrompt threw on stale dynamic SaveVideo widget state. The live bundle reported panel 0.15.46 while the installed pack on disk was 0.15.124. The graph contained both format.codec and an orphan top-level codec on SaveVideo. The newer source already contains reconcileGraphDynamicWidgets for this shape, but frontend reload was blocked because the workflow had unsaved changes.

### Steps to reproduce
1. Open a persisted workflow containing SaveVideo restored with format.codec plus orphan codec.
2. Keep unsaved canvas changes open.
3. Restart ComfyUI and reconnect the panel.
4. Call panel_run.
5. First call refuses during frontend serialization. A frontend reload cannot run because navigation is blocked by unsaved work.
6. After repeated panel_refresh_nodes attempts, retrying panel_run queued successfully without modifying the workflow.

### Exact error (scrubbed)
Dynamic widget doesn't exist on node

### Fix
applied locally: no
upstream-only: no
The installed 0.15.124 source already includes dynamic-widget reconciliation. The stale running bundle could not be safely reloaded while unsaved work was open. A robust fix would reconcile graph dynamic widgets immediately before graph_run serialization, or make stale-bundle recovery possible without scripted navigation and loss of unsaved state.

### Environment
Windows 10
ComfyUI 0.34.0
ComfyUI frontend 1.51.9
Python 3.13.11
PyTorch 2.10.0+cu130
NVIDIA RTX 5090 32 GB
comfyui-mcp 0.52.146
running panel 0.15.46; installed panel 0.15.124

artokun:
Fixes #2033.

After restart/reconnect, the first graph serialization can install native DynamicCombo accessors before the queue-time snapshot restore. The restore now records each widget's node/name identity and resolves the current live widget immediately before every write, shallow-to-deep, for both snapshot and live-state restoration. Detached-child errors are no longer classified by message text or silently suppressed; missing replacements and unrelated setter failures fail closed.

The existing setter sealing and same-value protection remain in place. The relocation replay now has behavioral coverage proving that stale widget rows, input rows, and widget-store aliases are actually removed, while a later same-value parent write still preserves the live child.

Validation on `e3bdbcf4`:
- focused DynamicCombo suite: 27 passed
- full unit suite: 8,406 passed, 0 failed, 1 todo
- `npm.cmd run typecheck`: passed
- panel scope: passed
- registry-parity YARA: 0 flagged files
- changelog integrity: passed
- changed-file Node syntax checks: passed

The branch is merged onto current `origin/main` `88f2c8f7fb1b9263697a144ca3e70eef1ecb363f`; this PR remains open and is not merged or released.


artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6). **No findings** — recorded explicitly, since with the external gate down an unreviewed PR and a reviewed-clean one look identical.

Three things I went looking for, because a property-descriptor wrapper is where they live:

**Double-wrapping.** Repeated `graphToPrompt` calls would stack wrappers, each adding a restore layer. Guarded — and guarded in the subtle correct way: the marker sits on the **setter**, not the widget. The comment explains why, and it is right: after a reconnect the first serialize can replace a store-backed accessor with native DynamicCombo `updateWidgets`, and a widget-level flag would then skip resealing the *new* setter.

**A plain data property losing its value.** The getter returns `origGet ? origGet.call(this) : undefined`, which would read as `undefined` forever if a widget held `value` as a data property. It cannot: the wrapper returns early unless `typeof desc.set === "function"`, so only accessors that already have a setter are touched, and a setter-only accessor already reads `undefined`. Behaviour preserved.

**The seal reaching only the retry path.** This was the one I expected to find, because `sealLiveDynamicComboSetters` appears inside the retry branch after `orig(...)`. It does not: `prepareLiveDynamicWidgets(target)` runs on **every** call before `orig` and wraps there — the post-retry seal is an *additional* reseal of what the rebuild produced, not the only one. Ordering is right too: seal after the rebuild, because the rebuild is what replaces the setter.

The one design decision worth affirming: `configurable: true` is deliberate and load-bearing. Without it the reseal after a native rebuild would throw rather than replace.

artokun:
Adversarial pass (mine — codex is out of tokens). Five checks, nothing to change; the interesting one is that the prose predicate here is forced rather than chosen.

**1. Marking the SETTER instead of the widget is the right call, and the reason is real.** After reconnect the first serialize can swap a store-backed accessor for native `updateWidgets`; a widget-level flag would skip resealing that new setter. Moving the mark onto `desc.set` means a replacement setter gets wrapped, which is exactly the recurrence #2033 describes.

**2. No crash on a data property.** `desc.set[SYMBOL]` would throw for a non-accessor, but line 236 guards it: `if (!desc || typeof desc.set !== "function") return;`. That was my first suspicion and it is already handled.

**3. The three-way case analysis holds.** Same value with live children → skip the rebuild; same value with none preserved → rebuild (the #2140 rescue); different value → rebuild from the new option's spec. Capture happens *before* `origSet`, which is the ordering that matters since the rebuild destroys the children.

**4. The catch is narrow, not blanket.** `if (!isDynamicWidgetMissingError(error)) throw error;` rethrows everything else, so this cannot swallow an unrelated failure during restore. (The second `try` around `wrapGraphDynamicComboSetters` *is* blanket, but its comment states the tradeoff — sealing is best-effort and the restore still has to run — which is the right priority.)

**5. The prose predicate is grounded, and cannot be disarmed by i18n.** I checked upstream rather than assuming. `comfyui_frontend_package` ships its sourcemaps with `sourcesContent`, so the real TypeScript is on disk:

```ts
// src/core/graph/widgets/dynamicWidgets.ts
throw new Error("Dynamic widget doesn't exist on node")
```

A bare `Error`, hardcoded English, no i18n wrapper and no error code. So `/Dynamic widget doesn't exist on node/i` matches something that genuinely exists, and translation cannot silently turn it off — the failure mode that kills most message predicates.

Worth stating the difference from mcp#2793, where I flagged a similar match: there a structured code existed on the wire and the prose was a choice. **Here upstream offers nothing else**, so matching the sentence is forced. And the failure direction is right: if a future frontend rewords it, the predicate returns false, the error is *rethrown*, and the restore fails loudly instead of quietly restoring the wrong thing.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **One unpinned guard, proven by mutation. Also refuting the objection I expected to make, because it does not hold.**

## The narrow catch is not pinned

The safety of the new restore catch is entirely its narrowness:

```js
} catch (error) {
  if (!isDynamicWidgetMissingError(error)) throw error;
}
```

I replaced that line with `void error;` — making it swallow **every** error — and:

- `dynamic-widget-reconcile.test.mjs`: **20 pass / 0 fail**
- full `npm run test:unit`: **8,344 pass / 0 fail**

Nothing anywhere in the panel suite notices. So the rethrow direction is unguarded: a later change to `DYNAMIC_WIDGET_MISSING_RE`, or a refactor that drops the condition, converts this into a blanket swallow with no test failing. The consequence is the one this code exists to prevent in reverse — a genuine restore failure disappears, and `panel_run` dispatches on a snapshot that did not fully restore, silently.

The fix is one test: make `copyStoredValue` throw something else (`new Error("boom")`) and assert the restore call propagates it. The positive case is already covered by #2033's four tests; only the negative is missing, and negative-direction coverage is exactly what a "narrow on purpose" catch needs.

## The objection I expected, which does NOT hold

`isDynamicWidgetMissingError` matches English prose (`/Dynamic widget doesn't exist on node/i`) against an upstream error, which normally means a translated frontend silently disarms it. I checked rather than assumed, reading the installed frontend's own TypeScript out of its sourcemaps:

```ts
// src/core/graph/widgets/dynamicWidgets.ts
if (insertionPoint === 0)
  throw new Error("Dynamic widget doesn't exist on node")
```

Raw, untranslated, no `t()` wrapper — so the predicate is locale-safe, and the match is specific enough that no unrelated error collides with it. Worth a one-line comment at the regex recording that, because the next reader will ask, and the answer needs upstream source to reach. If upstream ever reworded it the failure is fail-safe (rethrow → visible dispatch failure, i.e. today's behaviour), which is the right direction.

## Smaller

The sealing catch is bare:

```js
} catch {
  // Sealing is best-effort; restore still has to run.
}
```

Best-effort is the right call, but sealing is the mechanism the whole fix depends on — if it starts throwing every time, the fix silently stops working and nothing says so. One `logger.debug` in there makes that diagnosable at no cost to behaviour.

The rest holds up: marking the setter FUNCTION rather than the widget is the right call given the parent rebuild replaces `format.codec`, and the same-value native-rebuild skip is correctly scoped so stripped roots still rebuild for #2140.

artokun:
Follow-up, and a correction to my own suggestion. I tried to write the negative test I proposed and **could not make it reach the code** — which turned out to be the more interesting result.

## My proposed test does not work, for a reason worth knowing

I poisoned the captured `format.codec` widget with a `value` setter that throws `new Error("boom")`, then asserted the restore propagated it. It did not reject. Before concluding the catch swallows too broadly, I counted setter invocations: **zero**. My poisoned setter was never called, so the failure was my setup, not the code. Reporting that rather than the alarming version.

Chasing why led somewhere else. I instrumented `restoreState` itself with a call counter:

```
inside "#2033: queue-time snapshot restore does not throw…",  after await:   restoreCalls = undefined
same test, after an additional 5ms tick:                                     restoreCalls = undefined
in a test appended AFTER it:                                                 restoreCalls = 1, records = 3
```

So `restoreState` runs exactly once in that file, and **it runs after the flagship test's assertions have already completed**. That test's two assertions — `assert.doesNotThrow(() => queueItems.pop())` and `node.widgets.some(w => w.name === "format.codec")` — are both satisfied without the restore having happened, so neither observes the behaviour the test is named for.

That also explains cleanly why my earlier mutation result looked so stark: replacing the rethrow with a blanket swallow left **8,344 tests green** not because the assertions were weak in the usual way, but because nothing in the suite reaches that catch while it is asserting.

## What I am and am not claiming

I am **not** saying the restore path is broken — the production behaviour may well be correct, and the reasoning in the PR body is sound. I am saying the third bullet ("ignores a detached-child throw during snapshot restore instead of failing the dispatch") is currently **unwitnessed by tests**: the positive path is not observed, and the negative path (an unrelated error must propagate) has no test at all.

I would fix the ordering before adding my negative test, since a negative assertion on a path the test never reaches would pass for the wrong reason and read as coverage. The cheap version is to await whatever defers the restore — or drive `restoreState` directly with a records array, which is a pure function over `{widget, snapshot}` and needs no queue plumbing at all:

```js
restoreState([{ widget: poisoned, snapshot: "h264" }], "snapshot")
```

That reaches the catch in one line and pins both directions.

Everything in my previous comment about the upstream string stands — I verified from the frontend sourcemaps that `"Dynamic widget doesn't exist on node"` is thrown raw and untranslated, so the prose predicate is locale-safe.

artokun:
Pinned in `9f2db827`, using the route from my last comment rather than the one I originally proposed.

`restoreState` is now **exported as a test seam** and driven directly. It is pure over its `{widget, snapshot|value}` records, so the catch is reached in one call with no queue plumbing — which is the point, since the queue path does not reach it while a test is asserting.

Three tests:

| test | pins |
|---|---|
| swallows the detached-child throw | the intended behaviour |
| **rethrows anything else** | the narrowness — the whole safety of the catch |
| a non-`Error` throw still propagates | `isDynamicWidgetMissingError` stringifies non-Errors; a bare string that does not match must not be swallowed |

**Mutation-verified:** replacing the condition with a blanket swallow now kills two of the three. Before this it killed nothing, across all 8,344 panel tests.

The seam carries the reason in its docstring, including the measurement — a call counter shows `restoreState` has still not run at the end of the barrier test's body, even after a tick — so the next person does not have to rediscover why the obvious place to test this is the wrong one.

I did not touch the queue-path ordering. That test's assertions still complete before the restore, which is worth deciding on separately: either await whatever defers it, or retitle it, since "queue-time snapshot restore does not throw" currently describes something it does not observe. Left alone because changing it is a behavioural question about the barrier, not a test fix.

**Verification:** `npm run test:unit` **8,347 pass / 0 fail**, `check-panel-scope` OK, `check-changelog` OK.

artokun:
**The first finding is real, and it is two fixes in this PR defeating each other. Fixed.**

`reconcileFreshDynamicWidgets` renames an orphan to `<root>.__cmcp_stale_N`, pushes a `<root>.__cmcp_store_cleanup_N` alias, and then replays the root so LiteGraph's own setter deletes both:

```js
const value = widget.value;
widget.value = value;      // same-value write
```

That is precisely the write #2033's guard returns on:

```js
if (previous === next && preserved.size) return;
```

and it returns when `preserved.size` is nonzero — i.e. when the root is **healthy** and has live dotted children, which is the common case. So the stale row and its store alias stayed attached and the cleanup silently did nothing.

The guard now stands down while the root still carries relocation or cleanup rows, and applies again once the replay has removed them. It asks about the current widget list rather than latching a flag, so #2033's protection comes back as soon as it is safe. Mutation-verified — restoring the unconditional return fails the new test. `npm run test:unit` 8360 pass / 0 fail.

**The second finding I have not taken, and I want to be precise about why.** It says catching solely by error text lets a changed DynamicCombo snapshot serialize incomplete state: if restoring the parent changes its option, the parent rebuilds and detaches every captured child, and assigning a detached child can update the replacement's store, remove its descendants, then throw before recreating them — leaving the live replacement selected but missing required descendants. The proposed remedy is node/name identity in restore records so each value re-resolves against the current widget.

That reads as correct to me. It is also a change to what the restore path *records*, on a queue-time code path where the failure mode is a silently incomplete prompt — and I have no way to exercise a real parent-option change against a live canvas from here. Fixing it on a code-read would be guessing at a serialization contract, which is the same mistake as the one it describes. Leaving it visible for a session that can drive the panel.

Worth noting on this thread specifically: this file is where I retracted a "catch polarity" claim earlier today after asserting two catches were opposite when they are byte-identical. That is a reason to be more careful here, not less.

artokun:
**This branch now has a real code conflict with `main` and needs a hand — my changelog sweep deliberately refused it.**

The automated sweep I run after every release re-files the `Unreleased` entry when a merge re-conflicts `CHANGELOG.md`. It skipped this branch:

```
2200 fix/2033-dynamic-widget-recovery SKIP non-changelog-conflict:
     CHANGELOG.md web/js/lib/dynamic-widget-reconcile.js
```

That skip is by design — it aborts and moves on whenever the conflict is anything other than the changelog, so it never pushes a half-repair into someone else's code. Every other open panel PR swept and verified normally (8 of 9).

**What collided.** `bbba572` landed on main — `fix(2031): keep DynamicCombo dotted FLOAT writes across the set` — and both changes are in `web/js/lib/dynamic-widget-reconcile.js`.

It is not a trivial overlap. Around the same region:

- this branch adds an exported `wrapGraphDynamicComboSetters(graph)` with a *"Seal live DynamicCombo value setters across `graph` and nested subgraphs"* doc block;
- `main` has no such function, and at that offset now carries a different doc block referencing **#2254** (a dotted child of an ordinary current input) and **#1931** (a bare name that also exists as a nested child).

So `main` has moved this file on at least three separate fronts (#2031, #2254, #1931) since this branch forked. Resolving it means deciding how the new sealing pass composes with the dotted-child and orphan-widget handling that arrived meanwhile — a semantic call about DynamicCombo behaviour, not a placement question.

**I have deliberately not attempted it.** Guessing at the intended combination of two overlapping DynamicCombo changes is exactly the kind of edit that would look clean and be wrong, and this is not my branch. Flagging it instead, with the diagnosis, so whoever owns the change can resolve it in one pass.

Once it is resolved and pushed, the sweep picks it up again automatically — nothing else is needed from your side to re-enter the rotation. It is currently the only PR across both repos that is not cleanly mergeable.

Issue #2044, opened 2026-08-30
CivitAI OAuth sign-in always fails with HTTP 403: ComfyUI's origin-only middleware blocks the cross-site redirect back to the callback


artokun:
### What happened / root cause

Clicking **Sign in** in the panel's CivitAI browser opens the OAuth popup, and the popup ends on `127.0.0.1` showing a bare **HTTP ERROR 403** ("You don't have permission to access this page"). Sign-in can never complete. Nothing is written to the ComfyUI log, which makes it look like the request never arrived.

The flow (`web/js/cmcp-civitai-ui.js:2132-2134`):

```js
const r = await ctx.api.fetchApi("/comfyui_mcp_panel/civitai/oauth/start?origin=" + encodeURIComponent(location.origin));
const { authorize_url } = await r.json();
window.open(authorize_url, "_blank", "width=520,height=720");
```

and `py/civitai_proxy.py:503-524` builds the authorize URL with a redirect back to a ComfyUI-hosted route:

```python
redirect = origin.rstrip("/") + _CALLBACK_PATH     # http://127.0.0.1:8188/comfyui_mcp_panel/civitai/oauth/callback
params = {..., "redirect_uri": redirect, ...}
return web.json_response({"authorize_url": _OAUTH_AUTHORIZE + "?" + urlencode(params)})
```

`authorize_url` itself is correct and points at CivitAI. The failure is on the **way back**. When CivitAI redirects the browser to the loopback callback, that is a cross-site top-level navigation, so the browser sends `Sec-Fetch-Site: cross-site`. ComfyUI core rejects it in `server.py:161-165`:

```python
def create_origin_only_middleware():
    @web.middleware
    async def origin_only_middleware(request: web.Request, handler):
        if 'Sec-Fetch-Site' in request.headers:
            sec_fetch_site = request.headers['Sec-Fetch-Site']
            if sec_fetch_site == 'cross-site':
                return web.Response(status=403)      # <-- no logging on this branch
```

Two things worth calling out:

1. **This branch logs nothing.** The neighbouring host/origin check below it logs `WARNING: request with non matching host and origin ... returning 403`, but the `Sec-Fetch-Site` branch returns silently. Greps for `403`, `Banning request`, `non matching host and origin` and `civitai.*callback` in `user/comfyui.log` all come back empty while the user is staring at a 403 — a genuinely confusing debugging experience.

2. **The middleware is on by default.** `server.py:236-239`:
   ```python
[...]

PR #2202, opened 2026-09-03
fix(2044): CivitAI sign-in says when it did not complete

artokun:
#2044 stays OPEN (this is deliberately partial). It fixes the silence, which is the one part of that report that is not blocked on something only you can do.

## What was silent

The popup ends on a bare browser 403 — `create_origin_only_middleware` rejects the cross-site redirect back from CivitAI on the one branch that logs **nothing** — and on this side:

```js
if (state.signedIn || ++tries > 120) {
  clearInterval(_oauthPollIv); _oauthPollIv = null;
  if (state.signedIn) { …toast, reload… }
  //   ← and nothing here
}
```

Four minutes of polling, then the interval stops. No toast, no error, and an empty ComfyUI log. The user is left staring at a raw 403 with nothing anywhere saying sign-in failed.

## What the message claims, and what it doesn't

> CivitAI sign-in did not complete. **If the popup showed a 403**, ComfyUI's cross-site protection blocked the redirect back — it rejects that request without logging anything. Use a CivitAI API token in Settings instead (see panel#2044).

Asserted: the token never arrived. That is what was observed.

Offered conditionally: the 403. The popup is cross-origin, so this code **cannot read its status** — claiming it would be the same overreach as the silence, pointing the other way, and it would be wrong for every other way sign-in can fail (the user closing the popup, declining consent, a network drop). There is a test that fails if that conditional is ever flattened into a statement.

The remedy named is the one that works **today**, with no launch flag and no app-registration change: the panel already accepts a CivitAI API token server-side.

## Still blocked, exactly where the earlier analysis left it

- **Loopback PKCE** (`src/services/oauth-flow.ts` already implements it for Codex and Grok) needs one experiment against the CivitAI app registration that only the account owner can run: open an authorize URL with `redirect_uri=http://127.0.0.1:1456/callback` while signed in, and see whether the consent screen appears. The earlier probe could not settle it — `auth.civitai.com` defers `redirect_uri` validation until after login, and the control (`https://evil.example/cb`) returned the identical 303.
- **Middleware insert** ahead of core's reaches into `PromptServer.instance.app.middlewares` from a custom node, and was deliberately not shipped blind.

Both remain open on #2044.

## Verification

- `npm run test:unit`: **8226 pass, 0 fail** (i18n, i18n-render, tool-vocabulary, panel-scope, and the full `node --test` suite). `check-registry-yara --allow 0` → 0 flagged. `check-changelog` passes.
- 3 new tests: the branch exists and is not a silent fall-through; the message reaches the generated English catalog and names the API-token remedy; and the cause stays conditional.
- Pinned on the shipped bundle rather than driven behaviourally, and the file says why: the branch is reachable only after 240s of real polling through a module that owns its own interval.


artokun:
we should recommend the user to add the cross origin flag no?

artokun:
or we have the MCP start a temporary token capture external window then pass the tokens via mcp to the panel

artokun:

artokun:
## Adversarial self-review (external gate out of tokens until 2026-09-06)

The message itself holds up under every check I could make of it. Its guard did not.

### The stated safeguard did not exist

> *"There is a test that fails if that conditional is ever flattened into a statement."*

Mutation says otherwise. Rewriting `If the popup showed a 403` → `The popup showed a 403` in `web/js/cmcp-civitai-ui.js` left **all three tests green**.

The existing check reads `locales/en/main.json`. That is the string that normally renders — `tr()` resolves the catalog first — but it is **generated from** the source fallback by `npm run i18n:build`, and the fallback is what renders whenever the catalog is not loaded. So the source could assert as fact a 403 this code never saw, on the un-catalogued path, and nothing would notice.

Same shape as panel#2210 earlier today, where a gate walked every shipped `.js` and missed the catalogs — a check that reads one of the two copies of a string. Here it is the mirror image, and the property being guarded is the PR's whole design argument.

Both copies are asserted now, plus that they are **identical** — which also catches a fallback edited without re-running `i18n:build`. Verified by re-running the mutation that survived (now caught) and by appending a word to the catalog alone (also caught).

### Everything the message claims, checked

- **"Four minutes"** — `setInterval(…, 2000)` and `++tries > 120`. Exactly 240s. Accurate.
- **The remedy names a control that exists.** `"Set CivitAI token…"` is a real setting label (`panel.set_civitai_token`, rendered in `comfyui-mcp-panel.js`), with the exact text the user sees. A remedy pointing at a control that is not there would be the same defect as the original silence, so this was worth checking rather than assuming.
- **Catalog and fallback agree**, byte for byte. That is the divergence that bit #2210, and it is clean here — now pinned so it stays that way.
- **No spurious toast when the panel closes.** Both `!isOpen` guards clear the interval and `return` before the give-up branch, so a user who closes the panel is not told sign-in failed. A user who closes the *popup* does get the message after the timeout, which is true and appropriate.

### Not a finding, but worth stating

The new key is English-only, so the 11 other locales fall back to the English sentence. `i18n:check` passes and reports it as one of **43 keys already in that state** ("not yet translated (renders English)"), so this is the repo's normal flow, not a gap this PR introduces. One consequence to note for whoever translates it: the sentence embeds the setting label `"Set CivitAI token…"`, which is itself translated — so the translation should use that locale's rendering of `panel.set_civitai_token` rather than the English string, or it will name a label the user's UI does not show.

### Unchanged

Still does not close #2044. The cross-site redirect block is untouched, and the one-click flow stays broken until the launch-flag/app-registration question is settled — which remains yours.

artokun:
Adversarial pass (mine — codex is out of tokens). The mechanism is right; the remedy sends non-English users to a control they cannot find.

**Verified correct.** Both mechanism claims in the toast hold. I read the installed ComfyUI's `server.py` today: `create_origin_only_middleware` returns `web.Response(status=403)` for any `Sec-Fetch-Site: cross-site` request, and that branch has **no logging** — the `logging.warning` a few lines down belongs to the Host/Origin-mismatch branch. So "blocked the redirect back" and "rejects that request without logging anything" are both literally accurate, not plausible-sounding.

**The problem is the last sentence.** It hard-codes the setting's English label:

> Use a CivitAI API token instead — the panel's **“Set CivitAI token…”** setting stores it

That label is translated everywhere:

```
fr   Définir le jeton CivitAI…
ja   CivitAI トークンを設定…
zh   设置 CivitAI 令牌…
```

and the new toast exists **only in `locales/en`** — `sign_in_timed_out` is absent from fr/ja/zh (0 hits each). So a French user gets the English fallback string, which tells them to look for “Set CivitAI token…”, while their own menu reads “Définir le jeton CivitAI…”. The one actionable instruction in a message about a sign-in that cannot succeed is the part they cannot act on.

**The fix keeps it true by construction:** interpolate the label rather than quoting it —

```js
tr("civitai_ui.sign_in_timed_out", "… the panel's “{setting}” setting stores it …")
  .replace("{setting}", tr("panel.set_civitai_token", "Set CivitAI token…"))
```

so the quoted text is always whatever the running UI shows, in any locale, and cannot drift if the label is ever reworded. Translating the toast into all twelve locales also fixes today's symptom, but it re-hard-codes the label twelve times and leaves the drift.

Worth noting this is the fourth instance today of the same shape — a message naming something the reader cannot act on (mcp#2849's deleted env var, mcp#1477's `pi.exe`, mcp#2852's non-existent counterpart, and now a label that only exists in one language). The common cause each time was a string asserting something about a *different* surface instead of deriving it.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **Four probes, nothing blocking. The one objection worth raising turns out to be already handled.**

## The objection I expected to land, and why it does not

`setInterval` with an `async` callback, and the timeout counter increments *after* an `await`:

```js
_oauthPollIv = setInterval(async () => {
  if (!isOpen) { … return; }
  await refreshAuth();
  if (!isOpen) { … return; }
  if (state.signedIn || ++tries > 120) { …
```

If `refreshAuth()` could reject, the callback would abort before `++tries`, the interval would never reach 120, and the failure message would never fire — **reintroducing exactly the silence this PR removes**, under a condition (server down / network error) that is one of the likelier real causes of a sign-in not completing.

It cannot, because `refreshAuth` swallows its own failure:

```js
async function refreshAuth() {
  try { … state.signedIn = (await r.json()).signed_in; }
  catch { state.signedIn = false; }
  …
}
```

So every tick reaches the counter and the timeout always fires. Worth knowing that this PR's correctness depends on that `catch` staying there — it is load-bearing for the timeout, not just for the auth read, and nothing at either site says so. One cross-reference comment would make that survive a future refactor of `refreshAuth`.

## Also checked

- **The freshness of the verdict.** `state.signedIn` is read *after* `await refreshAuth()`, so the final check is not a stale pre-await value. And `state.signedIn` is tested before `++tries`, so a success landing on the 120th tick still wins.
- **The recommended remedy exists.** `SETTING_TOKEN_CIVITAI` is a real setting, sitting in `TOKEN_BUTTON_LABEL` alongside the OpenRouter / Custom / HuggingFace ones — so this is not a refusal recommending a recovery that is not there.
- **The label is referenced, not copied.** `{ setting: tr("panel.set_civitai_token", "Set CivitAI token…") }` means the toast and the button cannot drift apart, and both keys are present in `locales/en/main.json`.

## On the claim discipline

The split between *asserted* (the token never arrived — observed) and *offered* (the 403 — inferred, because a cross-origin popup's status is unreadable) is exactly right, and the comment says so at the site. That is the same discipline as #2204's positive-evidence-only warning, and it is the reason this message is safe to ship while #2044 stays open.

Nothing here blocks merge.

artokun:
Pinned in `0e9563f2` rather than left as a comment — a test survives a refactor, a comment does not.

The test asserts, at the site that depends on it, that `refreshAuth` swallows its fetch failure and records signed-out. **Mutation-verified:** changing that `catch` to rethrow kills exactly this test and nothing else.

Worth noting why this needed adding at all: all seven existing tests in `civitai-oauth-timeout.test.mjs` are source-shape checks over `cmcp-civitai-ui.js` — they pin the *text* of the branch, the remedy, the placeholder and the catalog entry, which is reasonable for a message. But none of them drives the poll, so none could fail if the timeout became unreachable. The dependency lived entirely in the gap between "the message text is right" and "the message can still be reached".

I kept it in that file's existing style (scoped source assertion) rather than building a `new Function` harness for the poll: the poll is nested inside `accountFlow`'s closure, and slicing it out would be a large harness for a property one scoped assertion pins directly.

**Verification:** `npm run test:unit` **8,347 pass / 0 fail**, `check-changelog` OK.

artokun:
**The Copilot review on this PR was right: the message recommended a setting that cannot sign the CivitAI browser in. Fixed.**

All four of its comments are the same defect seen from four places, and the check is one grep:

```
py/civitai_proxy.py:  CIVITAI_API_TOKEN  ->  1 hit, inside a DOCSTRING. Never read.
                      auth is _load_tokens() from civitai-oauth.json
```

The token is stored for the agent's own download tools. So after a four-minute wait, the single explanation the user received pointed at a setting that changes nothing about what just failed — and the code comment above it asserted the opposite: *"The remedy named is the one that works today"*.

**Now it names the flag whose absence causes the 403.** `create_origin_only_middleware` is installed whenever `--enable-cors-header` is missing, and that middleware is what rejects the cross-site redirect back from CivitAI. The message also says plainly that the token setting will not help, so a reader who remembers the old advice is not sent back to it.

**Second finding, also real: the toast rode the 3.5s default.** Sixty words of diagnosis delivered at the end of a four-minute wait and removed after 3.5 seconds is the same failure this PR exists to fix, one layer up. It now passes an explicit long lifetime, pinned by a test that reads the call site rather than the constant.

**What I broke on the way, since it is the useful part.** My first rewrite dropped the `{setting}` placeholder and hardcoded the label — which failed two existing tests that require the message to REFERENCE the setting rather than copy its text. That principle is right and I had just violated it; the message keeps the placeholder and says the referenced setting will not help.

**Localisation:** the key is absent from all eleven non-English catalogs, so `tr()` falls back to the source string everywhere — no stale translations telling other-locale users to set the token.

**Verification:** mutation-verified (restoring the API-token advice fails the new test); `npm run test:unit` **8360 pass / 0 fail**, including `[check-panel-scope] OK` and the regenerated `locales/en/main.json`.

I have not touched the root cause. The two real fixes remain blocked exactly as this PR's header says — the loopback-PKCE route needs a CivitAI app registration only the account owner can test, and the middleware-insert route reaches into core's middleware list.

artokun:
**Verified the new remedy, and corroborated the diagnosis against a real install.**

The message now reads:

> CivitAI sign-in did not complete. If the popup showed a 403, ComfyUI's cross-site protection blocked the redirect back — it rejects that request without logging anything. **Start ComfyUI with `--enable-cors-header`** to allow it (see panel#2044). The panel's "{setting}" setting will **NOT** help here: that token is used by the agent's download tools, and browser sign-in uses a separate OAuth token.

Three things it gets right, and the third is the one most fixes skip:

1. **Names the cause**, including why nothing was logged — which is why this was hard to diagnose from the user's side.
2. **Names a remedy that can actually work** — the flag whose absence installs `create_origin_only_middleware`.
3. **Explicitly negates the wrong remedy.** Removing bad advice leaves the reader to rediscover it: the CivitAI token setting is sitting right there in the panel and looks exactly like the answer. Saying "that will NOT help, and here is why" is what stops the next hour being spent on it.

The one remaining `CIVITAI_API_TOKEN` in `cmcp-civitai-ui.js` is a comment recording why that advice was wrong (`py/civitai_proxy.py` authenticates the browser from its OAuth token file and never reads the variable — the name appears there only in a docstring). Not a recommendation.

**Corroboration from a live install.** The ComfyUI serving 8188 on this box launches with:

```
python.exe -s ComfyUI\main.py … --enable-manager --enable-cors-header --extra-model-paths-config …
```

`--enable-cors-header` **present**. So per this diagnosis, this install should not hit the 403 — consistent, and it suggests Comfy Desktop supplies the flag by default. That would put the exposure on custom or non-Desktop launches, which fits a bug that took this long to be pinned down.

Not established: I have not driven the OAuth flow, so this verifies the remedy names a real, correct flag — not that adding it resolves the popup end to end.

Issue #2139, opened 2026-08-31
panel_unexpose_subgraph_* leaves ChangeTracker behind + root-shape-mismatch; save and further edits blocked


artokun:
### What happened / root cause
After panel_unexpose_subgraph_output (and earlier panel_remove_node) on a live workflow that already had unsaved series rewiring, the panel and ComfyUI ChangeTracker diverge:
1) Subsequent panel_* mutations fail with root-shape-mismatch (same node count, CONTENT drift).
2) panel_save_workflow (in-place and Save-As) REFUSES because the ChangeTracker snapshot is BEHIND the live canvas (panel#1563 / panel#2133).
3) panel_open_workflow rebinds from disk and discards the unsaved canvas edits — forbidden recovery path.
4) User click/drag + Ctrl+S still left save refused in this session; panel mutations cannot nudge because of the shape-mismatch fence.

Net: agent and user are stuck with unsaved canvas changes that cannot be saved via panel tools and cannot be re-opened without loss.

Likely root: subgraph boundary mutation / node removal updates the live LiteGraph but does not check-in ChangeTracker, while the panel's workflow fence still holds the pre-mutation content hash. Save guard correctly refuses writing stale tracker state, but there is no panel API to force tracker capture when mutations are also fenced off.

### Steps to reproduce
1. Open a saved workflow with subgraph nodes; make manual canvas rewires (unsaved or partially synced).
2. panel_enter_subgraph on a subgraph; panel_unexpose_subgraph_output of a boundary slot (or panel_remove_node on root).
3. Immediately call another panel mutation → root-shape-mismatch (content).
4. panel_save_workflow → REFUSED (tracker behind live canvas).
5. User drag + Ctrl+S may still refuse; panel_open_workflow would wipe canvas changes.

### Exact error (scrubbed)
REFUSED to save: the state this save would write to "workflows/MywomanREAL_Anima_01.json" is BEHIND the live canvas... Do NOT close the tab, and do NOT re-open the workflow from disk... (panel#1563, panel#2133)

Also: [root-shape-mismatch] ... both report 61 node(s), but the canvas does not reproduce the workflow's own state — difference is in the graph's CONTENT...

### Fix
applied locally: no
upstream-only: yes

Needed:
- After any panel mutation that changes graph content (especially subgraph rail unexpose/expose and node remove), force ChangeTracker check-in / sync the panel fence to the live canvas in the same mutation result.
[...]

PR #2204, opened 2026-09-03
fix(2139): warn when re-opening the active tab would discard unsaved work

artokun:
#2139 stays OPEN (this is deliberately partial). It fixes item (3) — *"do not recommend `panel_open_workflow` as recovery when unsaved live canvas is ahead of disk"* — which is the one item on that list I can do soundly without a live rig.

## The deadlock, and why the advice made it worse

The reporter had unsaved rewiring on a **saved** workflow. Save was already refused by the tracker-behind guard; every panel mutation was fenced off by a root-shape mismatch. The instance-mismatch refusal then told them to *"re-select the intended workflow with `panel_open_workflow`"* — which re-reads from disk and **discards the unsaved canvas**. Every visible exit was either blocked or destructive, and the one the tool recommended was the destructive one.

## Why #1019 did not already cover it

That guard special-cases the active tab being **unsaved** — no path, so `panel_open_workflow` cannot re-select it at all. Different question:

| | question | this case |
|---|---|---|
| **#1019** | *can* the tab be re-opened? | yes — it is saved |
| **#2139** | what does re-opening *cost*? | the unsaved drift |

`activeIsUnsaved` is `false` here, so the message recommended `panel_open_workflow` plainly.

## Positive evidence only

`isModified === true` warns. `false` and `undefined` stay silent.

That asymmetry is deliberate and #882 is why: ComfyUI derives that flag from **user-input captures**, so a value a *node* wrote — an ImpactWildcardEncode populate, a `control_after_generate` roll, a subgraph's promoted widgets — leaves the tab reading clean while the canvas already differs. A `false` is therefore not evidence that re-opening is safe, and treating it as such would be the same defect pointing the other way.

It also **does not capture first**. #882's capture helper exists for callers about to *discard* a canvas and needing a fresh flag; a mutation inside an error message is a new failure surface on a path that is already failing.

## Two source-shape pins moved, both keeping what they guarded

- **#1019's call-site pin** asserted "exactly three arguments". Its stated point is that the helper receives every input as an **argument**, because it is rebuilt with `new Function` where a module global does not exist — the arity was incidental. Re-pinned on the property.
- **#968's signature pin** requires the parameter list to stay on **one line**, or the brace-balanced extractor stops at the parameter brace and breaks 13 tests. Still one line, now with the fourth field.

## Verification

- `npm run test:unit`: **8228 pass, 0 fail**. `check-registry-yara --allow 0` → 0 flagged. `check-panel-scope`, `tsc --noEmit`, `check-changelog` clean.
- 5 new tests, driving the real message function (it is pure — #968 pins that).
- **Two mutations, each killed**: removing the warning (2 fail) and warning on a `false` reading (1 fail).

## Still open, and why I did not touch it

The tracker/save deadlock itself — items (1) and (2) on your list. A ChangeTracker check-in that is wrong makes the canvas and the tracker agree *whether or not the canvas is what the user meant*, converting a detectable divergence into a silent one; and `capture:true` / `panel_force_capture` adds vocabulary, which is a governed surface here. Both want a live ComfyUI to verify against, which I do not have.


artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6). **No findings** — recorded explicitly, since with the gate down an unreviewed PR and a reviewed-clean one look identical.

I came to this one expecting a defect, because I had measured the underlying flag and it does not behave the way a warning would want:

> `LiteGraph.createNode + graph.add` → `isModified: FALSE`; widget write → `isModified: FALSE`; `captureCanvasState()` → `isModified: TRUE`

So `isModified` **cannot see programmatic drift**, and a warning built naively on it would be silent exactly when an agent had been editing. The asymmetry is handled correctly here:

- the read is `active.isModified === true ? true : null` — never `false`. A false flag is evidence of *nothing*, and collapsing it to `null` is what stops it being reported as safety;
- the message branches on `=== true`, so unknown stays **silent** rather than reassuring. Its comment says exactly that: *"true warns; false and undefined stay silent rather than implying it is safe."*
- and both directions are pinned — `reopen-discards-unsaved-warning` asserts the warning for `true` and loops `[null, undefined, false]` asserting silence. A future "simplify to a truthiness check" fails it.

One design decision worth affirming rather than second-guessing: the message **deliberately does not capture first**. #882's helper exists for callers about to discard a canvas and needs a fresh flag; this is an error message, and a mutation inside one is a new failure surface on a path that is already failing. That is the right call, and it is the reason the warning is honest about being a *may* rather than a *did*.

The honest limit is worth restating for whoever reads this later: drift produced by **node-written** values still will not raise this warning, because nothing sets the flag. That is a property of the tracker, not of this change — and it is why the message says re-opening "discards" rather than claiming to know what is there.

artokun:

artokun:
## Adversarial self-review (external gate out of tokens until 2026-09-06)

The guard is sound and strictly positive. One blind spot was missing, and it is the one that lands on this issue's own deadlock.

### `isModified` is FROZEN inside a stranded transaction

The check is documented as positive-evidence-only, with #882's reason: ComfyUI derives the flag from user-input captures, so a value a NODE wrote leaves the tab reading clean. True.

There is a second reason, read out of the shipped `changeTracker.ts` rather than inferred. `workflow.isModified` is written **only** by `updateModified()`, which is reached **only** from `captureCanvasState()` — and that returns early while a change transaction is open:

```ts
captureCanvasState() {
  const isUndoRedoing = this._restoringState
  const isInsideChangeTransaction = this.changeCount > 0
  if (!app.graph || isInsideChangeTransaction || isUndoRedoing || ChangeTracker.isLoadingGraph)
    return
  …
  this.updateModified(previousState)
```

So inside a **stranded** transaction the flag is frozen at whatever the last successful capture left.

That matters here specifically, because a stranded transaction is exactly the state in which save is refused as behind the canvas — which this PR's own new sentence names:

> *"if the save is itself refused as behind the canvas, that is panel#2139 and re-opening is the one recovery that loses the work"*

If the drift happened entirely inside the stranded transaction, `isModified` is still its pre-drift value and the warning is **silent on the case that sentence is about**. A session that made balanced edits earlier reads true and does warn, which is the common shape and covers the reporter — they had manual rewiring, and manual edits do set the flag. But silence was never evidence the tab is clean, and now the code says why in both directions.

Comment and changelog only; no behaviour change, and the positive-only rule is unchanged. Full panel gate: **8280 pass, 0 fail**.

### A trap for whoever edits this function next

`reopen-discards-unsaved-warning.test.mjs` extracts `workflowInstanceMismatchMessage` by **counting braces over the raw source**, and it does not skip comments. My first draft quoted the upstream snippet with its opening `captureCanvasState() {` and no closing brace — the extractor never balanced, returned `null`, and **four tests failed on a comment-only change**. The snippet now carries no braces and says why.

That is worth knowing before someone adds an ordinary code example to this function and spends a while on a failure that looks nothing like its cause. It also means the extractor is not robust to a perfectly legal edit — worth tightening if that file grows more of these.

### Checked and clean

- The reading is strict: `active.isModified === true ? true : null`, so a truthy non-boolean cannot warn, and `false`/`undefined` collapse to the same silence.
- It does **not** capture first, and the code says why: #882's helper exists for callers about to DISCARD a canvas; a mutation inside an error message is a new failure surface on a path that is already failing. Given the measurement above, capturing first would also be a no-op on precisely the stranded-transaction path — it is suppressed there.
- The #1019 distinction holds: that guard asks whether the tab CAN be re-opened, this one asks what it COSTS, and `activeIsUnsaved` is false in the reported case, which is why the old message recommended `panel_open_workflow` plainly.

artokun:
Adversarial pass (codex is out of tokens, so these are mine). The positive-only rule is right and correctly implemented; one scope question.

**Confirmed correct.** `activeIsModified = active.isModified === true ? true : null` — `true` warns, `false` and `undefined` stay silent rather than implying safety. That matches what I measured on a live rig today: after `LiteGraph.createNode` + `graph.add` and a direct widget write, `isModified` stayed **false** while the canvas already differed, and only `captureCanvasState()` flipped it. So `false` really is evidence of nothing, and treating it as "safe to re-open" would have been the wrong reading. The comment at 5236 says exactly this, and the code does it.

**The question: what the warning is attached to.**

The only site that computes `activeIsModified` is inside `assertActiveWorkflowCommandTarget`, and it runs in the branch where `!commandTargetsActiveWorkflow(...)`. Its two call sites are:

```
18333  graph_set_widget
19258  graph_remove_widget
```

Both mutations. So the sentence — *"the ACTIVE tab has UNSAVED changes: panel_open_workflow re-reads it from disk, so those changes are discarded"* — is attached to a **mutation-mismatch refusal**, warning about the cost of the remedy that refusal offers. That is coherent, and for that path it is the right place.

But the PR is titled *"warn when re-opening the active tab would discard unsaved work"*, and a `panel_open_workflow` aimed at the workflow that is **already active** does not reach this guard: the uuid matches, `commandTargetsActiveWorkflow` is true, nothing throws, and the drift is discarded with no warning at all. That is the literal titled case, and it is the quieter one — no refusal precedes it, so there is nothing to attach a caveat to.

I could not find a separate unsaved check on the open path to rule this out, but I also could not locate the open handler cleanly enough to be sure, so I am asking rather than asserting: **is a direct re-open of the active tab intended to stay silent, with this PR scoped to the remedy text?**

If it is intended, worth saying so in the changelog — the title reads broader than the behaviour. If it is not, the same positive-only read would need to run on the open path before it replaces the canvas, which is a different (and more delicate) place to put it, since that one has to decide whether to warn or refuse.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **The design is right and I can confirm it from measurement, not just review. One sibling site still gives the destructive advice.**

## Confirming the positive-only rule from a rig

I measured this on an isolated ComfyUI (0.34.2 / frontend 1.49.6, pristine `origin/main` panel) rather than reasoning about it:

```
LiteGraph.createNode + graph.add   nodes 1   isModified FALSE
+1500ms                            nodes 1   isModified FALSE
widget write (seed = 424242)                 isModified FALSE
```

So the canvas can genuinely differ from the file while the tab reads clean. `true` is evidence of drift; `false` and `undefined` are evidence of nothing. Warning only on `true` is the correct asymmetry, and the comment's admission about the stranded-transaction case is accurate — I reproduced that too: with `changeCount > 0`, `captureCanvasState()` is suppressed, so `isModified` stays frozen at the last successful capture, which is exactly the state where save is already refused as behind the canvas.

## The sibling

There are four places that recommend `panel_open_workflow` in prose. This PR fixes one. Of the other three:

- **line 29775** — recommends switching to a *different* workflow. The active tab keeps its own state, so this is not destructive. No caveat needed.
- **line 5553**, `workflowListReadinessRefusalError` — **is the same destructive advice, uncaveated**:

```js
"This read-only probe changed no workflow target and " +
"is safe to retry in a moment; panel_open_workflow(path) can explicitly re-establish the tab " +
"if the reconnect does not settle.",
```

Re-establishing *the tab* means re-opening the one that is already active, which re-reads from disk — this PR's own premise for why the advice is destructive. And the reachability is not hypothetical: this fires after a ComfyUI reconnect, a websocket event that does not reload the page, so unsaved canvas work is still sitting there when the user is told to do it. The sentence even opens by reassuring them nothing was changed, which makes the recommendation read as safe.

## Shape, if you want it

Inline, not shared — the same constraint this PR already respects. `workflow-list-unsaved-identity.test.mjs` rebuilds this area with `new Function` and injects dependencies by name, so a module-scope helper throws `ReferenceError` there; `activeWorkflowRef` would need adding to that harness's injection list. The read itself is the four lines you already wrote:

```js
const active = activeWorkflowRef();
activeIsModified = active.isModified === true ? true : null;
```

Happy to push it onto this branch if you'd rather it land here than as follow-up — say the word. Not doing it unasked, since #2139 is explicitly scoped to item (3) and this is arguably item (3) at a second site rather than new scope.

artokun:
**The "Mandatory" finding in the Copilot review was right: the warning this PR adds never fired on the fence callers actually hit.**

Two call sites build the mismatch message. Only one was given the new flags:

```js
// web/js/comfyui-mcp-panel.js — PRIMARY dispatch fence
workflowInstanceMismatchMessage({
  commandUuid: dispatchCommandUuid,
  activeUuid: dispatchActiveUuid,
  movedNote: activeWorkflowMoves.describeLast(),
})                                    // <- no activeIsUnsaved, no activeIsModified
```

Both default to `null`, so an ordinary workflow-instance mismatch produced the OLD message. And that fence's own comment describes it as *"the refusal a caller actually sees when a binding has gone stale"* — so #2139's warning lived on the path nobody reaches and was absent from the one everybody does.

**Fixed** by giving both sites one shared `activeWorkflowSaveState()` probe instead of an inline try/catch owned by a single caller. Two call sites, one reading, so they cannot disagree about the same tab.

**The test changes are re-pointings, not weakenings**, and two of them are worth naming:

- The #968 provenance pin asserted a literal call shape. Its stated purpose is *"the pure helper receives every input as an ARGUMENT… asserts the property rather than the arity"* — the spread preserves that property, so it keeps its intent and **gains a second assertion that the primary fence carries the reading too**. That is the wiring the finding was about.
- The #1019 guard counts call sites. A naive split also matches the function DEFINITION, so I would have been asserting 3 where 2 is right; it now excludes the definition with a lookbehind rather than adjusting the expected number and hiding a real third call site.

The `new Function` dispatch harness needed the probe injected by name — those harnesses resolve every dependency by name, so an un-injected one is a `ReferenceError` at call time rather than a missing warning.

**Verification:** mutation-verified — removing the spread from the primary fence fails the new wiring assertion. `npm run test:unit` **8359 pass / 0 fail**, `[check-panel-scope] OK`.

**Not addressed here:** the first finding, that `activeIsUnsaved: true` does not by itself prove `panel_open_workflow` will re-open *that* tab. That is a claim about what the remedy can promise rather than about wiring, and it wants the same treatment #2044 just got — I would rather do it deliberately than fold it into this.

Issue #2023, opened 2026-08-29
Comfy Desktop renderer crashes in DirectWrite text layout (render-process-gone exitCode=-1073741819) at panel content-update moments; takes the ComfyUI server down — 7×/day


artokun:
### What happened / root cause
The user sees ComfyUI "closing by itself" several times a day. Every outage is the Comfy Desktop (Electron 40.4.1 / Chromium 144.0.7559.173) **renderer process crashing**, after which Desktop 1.0.46 exits the whole app (server included) and relaunches:

```
[CRITICAL] render-process-gone reason=crashed exitCode=-1073741819   (0xC0000005)
```

Crashpad minidump of the renderer (pid 49208, 2026-08-29 16:47 local) parsed with `minidump`:
- `EXCEPTION_ACCESS_VIOLATION` at `Comfy Desktop.exe+0x4147f6e`, read of address `0xd92a8` (null-object field deref)
- stack (module+offset scan): Chromium frames → **DWrite.dll ×10 frames** (DirectWrite text layout / font fallback) → Chromium. No GPU-driver module on the stack (no nvwgf2umx / d3d11 / dxgi).
So the crash site is Chromium's DirectWrite text-layout path in the renderer, not the GPU process and not the ComfyUI server (the Python log always ends cleanly with `Prompt executed …`).

Timeline (Desktop app.log, local time), every crash lands when the panel inserts new content into the renderer:
| when | last lines before the crash |
|---|---|
| 08-29 10:24 | `Updating: comfyui-agent-panel … After restarting ComfyUI, please refresh the browser.` |
| 08-29 13:12 | `Prompt executed in 216.69 s` → run-completion delivered to the panel chat |
| 08-29 13:53 | `Prompt executed in 324.61 s` |
| 08-29 15:54 | ~70 s after a panel_ask card answer landed |
| 08-29 15:55 | 34 s after app start, right after `[comfyui-mcp-panel] local bridge advertised` (history reload) |
| 08-29 16:43 | 22 s after `Prompt executed in 56.49 s` (4-image panel_run completion) |
| 08-29 16:47 | 23 s after app start, right after `local bridge advertised` |
[...]

PR #2210, opened 2026-09-03
fix(2023): stop REQUESTING the emoji font the crash correlates with

artokun:
Does **not** fix panel#2023 — a Chromium DirectWrite null-deref is upstream, as the report says. It removes one thing on our side that made the crashing path **mandatory** rather than incidental.

## Two things checkable without the toggle test

**1. The panel's root font stack names no symbol or emoji face:**

```css
font-family: var(--font-inter, "Inter", ui-sans-serif, system-ui, sans-serif);
```

So any glyph those faces do not cover must be resolved by **fallback** — through DirectWrite, on Windows. That is how the panel's content updates reach the code the crash is in at all.

**2. Nine strings carried `U+FE0F`** — VARIATION SELECTOR-16, which does not merely *permit* emoji presentation, it **requests** it. That pins the glyph to `seguiemj.ttf` specifically: the file the report says KB5120998 replaced, two days before the crash cluster began.

All nine were `⚠️`, and where they render matters:

| site | when it renders |
|---|---|
| `⚠️ GRAPH VALIDATION ERRORS` | run completion |
| `⚠️ MISSING ASSETS` | run completion |
| `⚠️ LAST RUN FAILED` | run completion |
| `⚠️ Your agent session was replaced` | session notice |
| `⚠️ A render was already in flight` | session notice |

The issue's own timeline puts every crash at a panel content-update moment, naming run completion and history reload specifically.

## What this changes

The selector goes; **the glyph stays**. Dropping the warning sign would be a different and worse change — these are the lines telling a user a run failed or their session was replaced.

Worth doing on its own merits regardless of the crash: a warning triangle in a dense sidebar wants text presentation, and forcing an emoji-font lookup for it buys nothing.

## What it is not

Not a fix, and not claimed as one. It removes a *mandatory* dependency on the font the environment changed — which also makes it the **cheapest arm of the toggle test** panel#2023 is waiting on. If crashes persist with this in, emoji fallback is ruled out cheaply and the remaining candidates (`Comfy.VueNodes.Enabled`, or re-layout volume) get the attention.

**The font stack is deliberately untouched**, and there is now a test saying why: naming a symbol face would still resolve through DirectWrite, so it is not obviously a fix, and picking one blind on a machine I cannot reproduce is how a mitigation becomes a second bug. If someone does change it, that test puts this note in front of them.

## Verification

`npm run test:unit`: **8226 pass, 0 fail**. `check-registry-yara --allow 0` → 0 flagged. `check-changelog` passes. 3 new tests: no `U+FE0F` anywhere, the glyph and its run-completion string survive, and the font stack is pinned with the reasoning attached.

Pairs with #2809 (`scripts/read-minidump.mjs`), which makes the report's step (1) a one-liner — run it on the next renderer dump and the faulting module either is `DWrite.dll` or is not.


artokun:
Self-review found two defects in this PR — one incomplete, one a regression it introduced. Both fixed and pushed.

## The fix covered one file out of six

The gate read `web/js/comfyui-mcp-panel.js` **alone**, so it passed green while **nine more U+FE0F sat in `web/js/lib`**. Five of those render:

- `graph-revert.js` — all three failure notices;
- `media-preview.js` — the sampled-preview note;
- `run-completion-frame.js` — the storyboard header.

So the crash surface this change exists to remove was still half present, and the test said otherwise. One exit fixed, the siblings left. The gate now walks **every shipped `.js` under `web/js`**, which is what it should have done from the start.

## Removing the selector silently broke the chat stripper

This is the sharper one. Dropping U+FE0F changed what the panel **emits** — `⚠ GRAPH VALIDATION ERRORS` — but `chat-serialize.js` strips those blocks with patterns that **required** it:

```js
/^⚠️\s*GRAPH VALIDATION ERRORS[\s\S]*?(?:\n\n|$)/
```

From that commit on they matched nothing. So the agent-only GRAPH VALIDATION ERRORS / MISSING ASSETS / LAST RUN FAILED blocks stopped being stripped before display — and that list exists precisely because, in its own words, *"users read the shouty grammar as a warning their install is broken."*

The patterns now accept both spellings, so already-persisted chat keeps stripping too. The existing test kept its old-spelling fixture deliberately, and I added the current-spelling case next to it.

One detail worth keeping: a pattern that must **match** text containing the selector now writes it as a `️` escape rather than a literal, so the gate stays a blanket rule with no exemption list to drift out of date.

## Verification

Mutations in both directions: making the selector required again fails the new current-spelling case; putting a selector back into any lib fails the widened gate.

Full panel gate: **8261 unit tests** (8260 pass, 1 todo), `check-panel-scope` OK, typecheck and `check-tool-vocabulary` clean.

artokun:

artokun:
## Adversarial self-review — the fix was half-applied, and its own gate could not see it

Pushed as two commits. The PR's central claim was *"nothing REQUESTS emoji presentation"*, verified by a gate that walks every shipped `.js`. Both were true and neither was sufficient.

### The render path never changed

`tr()` resolves the catalog **first** and treats the source literal as a fallback (`web/js/lib/i18n.js`):

```js
if (loaded) out = catalog[key];
if (out === undefined) out = typeof fallback === "string" ? fallback : …;
```

So on any normal load, the string that reaches the DOM comes from `locales/<lang>/main.json` — not from the line beside it in the source. The gate walked `web/js`; the catalogs are not in `web/js`.

**60 occurrences of U+FE0F were still shipping**, five keys × twelve languages:

| key | |
|---|---|
| `graph_revert.could_not_action_nothing_was_loaded` | ← |
| `graph_revert.could_not_tell_whether_the_action_happened` | ← |
| `graph_revert.the_action_ran_but_unconfirmed` | ← |
| `panel.beta_the_app_changes_rapidly_and_builds` | |
| `panel.the_orchestrator_didnt_acknowledge_the_blind_change` | |

The three marked ones are `graph_revert` keys — the very file whose fallbacks this PR cleaned (`web/js/lib/graph-revert.js`, +3/−3). For those three the change was applied to the branch that does not render and skipped on the branch that does.

Not an argument from reading. `npm run i18n:build` — which regenerates English from the `tr()` fallbacks — changes **exactly five lines**, all of them the selector, nothing else. The generated catalog was stale against this PR's own edits.

### What changed

English regenerated (it is generated from the code, never hand-edited). The eleven translations have the selector removed **byte-wise**, so no translated wording and no line ending is touched — the panel tree is CRLF and a text-mode rewrite would have churned every line.

Two new gates:
- the catalogs carry no U+FE0F, in any language;
- `tr()` really does prefer the catalog. That one pins the *reason* the first has to exist: if the resolution order ever flipped, cleaning the `.js` alone would genuinely be sufficient and the first gate could relax — which should then be a decision, not a silent drift.

### Verification

`npm run test:unit`: **8279 pass, 0 fail**. `i18n:check` passes (key parity across all twelve). `check-changelog` passes.

Mutation-verified against committed state, each caught by the guard written for it: one selector reintroduced into `ja`, into the generated `en`, and into `ru`; and `tr()` losing its catalog lookup. The tree restores clean afterwards.

One thing worth recording from writing them: my first version of the resolution-order assertion **passed on the wrong clause**. `tr()` has a plural branch whose own `out === undefined` sits above the singular `catalog[key]`, so an unanchored search compared two different clauses and read the order backwards. It is now anchored to the singular branch.

### What is still not claimed

Unchanged from the original: this is not a fix for panel#2023, and the font stack is still deliberately untouched. What changed is that the mitigation now actually applies on the path a user's panel renders through, which also makes it a valid arm of the toggle test — before this, a "crashes persist with the selector removed" result would have been measuring a panel that was still requesting the emoji face in five places.

artokun:
Verified this branch is complete **on the panel side** — `U+FE0F` count is 0 across `web/`, `locales/` and `py/`. And the gate's own history matches something I hit independently today on mcp#2849: the first version read only the main bundle and missed nine selectors in `web/js/lib`, then the catalogs turned out to be a separate surface that wins over the source. Two gates, two repos, same two stages.

**But the orchestrator puts them straight back.** If this PR's premise holds — that the crash correlates with requesting emoji presentation — then the panel is not the only thing rendering into that renderer. The mcp orchestrator pushes chat text into the same Comfy Desktop window, and on `origin/main` it carries:

```
99 emoji-presentation sequences across 24 source files (excluding tests)
   + all 12 locale catalogs
```

They are genuinely chat-bound, not log-only. From `src/orchestrator/index.ts`:

```js
4125:  bridge.push({ type: "say", text: `⚠️ ${sync.message}` }, panelTab);
2432:  yield { type: "assistant", text: `⚠️ ${msg}` };
3177:  `⚠️ ${trFor(locale, "say.restart_lost.header", …)}`
```

`⚠️` there is `U+26A0 U+FE0F` — warning sign *plus* variation selector-16, the same sequence this PR removes. `ready-banner.ts`, `start-failure-notice.ts`, `queue-note.ts` and `rate-limit.ts` are all in the same list, and those are exactly the messages a user sees when something has already gone wrong.

So after this merges, a panel that never requests the emoji font still renders emoji-presentation runs the moment the orchestrator says anything — and the panel's gate structurally cannot see them, because they arrive over the bridge at runtime from another repo.

I have not filed this as a defect against your PR, because it is not one: this branch does what it says. It is a completeness gap across the process boundary, and the work is in the mcp repo. Two things worth deciding:

1. Whether the mcp side should get the matching change (mechanical: 99 sequences, and the catalogs need `npm run i18n:build` after, since `tr()` resolves the catalog before the in-source fallback).
2. Whether the panel should also strip `U+FE0F` from **incoming** bridge text at render time. That is the only version that holds regardless of what any other process sends, and it would make the guarantee independent of the mcp release cycle.

Worth saying plainly: the causation here is still the PR's premise, not something I have proven. If the correlation turns out to be wrong, none of this matters — but while the hypothesis is live, half the emoji surface is in the other repo.

artokun:
Correcting my own number above. I said "99 emoji-presentation sequences" and called them chat-bound. The 99 is right as a count of the sequences in mcp source; calling them all chat-bound is not, and the distinction matters for this PR.

Classified by destination:

```
chat-bound (bridge.push / type:"say" / type:"assistant" / sessionNotice)   21
log-only (logger.* / console.*)                                            3
unattributed by my classifier                                             75
```

Spot-checking the 75 shows they are user- or model-facing rather than dead, but not all of them **render**:

- `blind-completion.ts:23` — a completion notice the user reads. Renders.
- `cli-remedy.ts:141,153` — `[panel event] ⚠️ …` strings. Render.
- `index.ts:3177` — `say.restart_lost.header`, a chat line. Renders.
- `index.ts:380` — **a system-prompt fragment**. Goes to the model, never to DirectWrite. Does *not* render.

So the honest version is: **at least 21, and plausibly most of the 75, reach the panel renderer; 3 go only to logs; and an unknown slice is prompt text that never renders at all.** The claim that survives is the one that matters here — the orchestrator does push emoji-presentation sequences into the panel chat, so this PR's guarantee ends at the process boundary — but "99" is the size of the source surface, not of the rendered one, and I should not have used it as if it were the latter.

If the mcp side does get the matching change, that distinction is also the cheap way to scope it: the chat and panel-event paths are what the crash hypothesis implicates. Prompt text and log lines can be left alone, which makes it a much smaller diff than 99 suggests.

artokun:
Adversarial pass (codex is out of tokens, so this is mine). **The mechanism is real and the change is verified complete — but the word "mandatory" does not survive measurement.**

## What checks out

I scanned the branch head's `locales/` and `web/js/`: **zero `U+FE0F` remain.** The removal is complete, all 12 catalogs included, so the change does exactly what the body says it does. And the underlying reasoning is sound: `U+26A0` WARNING SIGN is `Emoji_Presentation=No`, so without the selector it resolves from a symbol face rather than `seguiemj`. That is a genuine behaviour change, not a cosmetic one.

## What does not

The claim is that this makes the crashing path *incidental rather than mandatory*. It does not, because `U+FE0F` is only required for characters whose **default** presentation is text. Characters with `Emoji_Presentation=Yes` go to the emoji font **with no selector at all** — and 17 distinct ones remain, 13 of them in English UI labels mirrored across every catalog:

```
🔒 Workflow hidden                              (apps_ui.workflow_hidden)
🔒 set / 🔒 set {date}                          (panel.locked_set, locked_set_on)
✅ Diagnostics copied — paste them in Discord   (panel.diagnostics_copied…)
💬 Join the Discord                             (panel.join_the_discord)
📖 Read the docs                                (panel.read_the_docs)
🤖 Android — Firebase beta                      (panel.android_firebase_beta)
```

plus `➕` in `comfyui-mcp-panel.js`, `🎬`, `🎲`, `🔊`, `👤` (40 occurrences), `⭐`, `📋`, `🆘`, `🔓`, `🍎`. Every one of those is `Emoji_Presentation=Yes`. Given the root stack names no symbol or emoji face — which the body itself establishes — each still resolves through DirectWrite fallback to `seguiemj`.

So the panel still *requests* the replaced font as a matter of course. What this PR does is narrow the exposure from those five ⚠️ sites, not remove it.

## One thing I checked and am NOT claiming

`run-completion-frame.js:538` carries `⏳` on the run-completion path the timeline names — which looked like a direct hit. It is not: `recoveredCompletionNote()` feeds `noteSections`, which is completion text addressed to the **agent** ("Do not announce it as a fresh render"), not markup rendered into the panel DOM. Source presence is not rendering, so I am leaving it out of the count above rather than reporting the stronger-sounding version.

## Suggestion

I would not extend the change — `U+1F512` and friends have no text-presentation form, so removing the font dependency means dropping or replacing those glyphs, which is a bigger call than this PR is making. The cheap fix is the sentence: say it removes the emoji-font request **at the five warning sites named**, and note the residual, rather than "mandatory → incidental". As written, a reader who later measures `seguiemj` still being requested will read the PR as having been wrong, when the change itself is fine.

artokun:
**This PR does what its title says and still does not achieve what it was written for. Measured on its own head (`a3e4445`), 262 shipped `web/js` files:**

```
                              origin/main    this PR
U+FE0F (variation selector)        18           0      <- the PR's stated job, done
emoji-presentation-default         44          44      <- untouched
```

U+FE0F only matters for **text-default** glyphs — U+26A0 WARNING SIGN is the one this PR was named for, and it does need the selector to become emoji. But 17 distinct characters in the shipped panel are **Emoji_Presentation=Yes**: they route to the colour emoji font with no selector at all. Removing FE0F cannot affect them.

They are user-visible, not comments:

```
cmcp-apps-ui.js:1452    const dice = makeBtn("(U+1F3B2)");
cmcp-apps-ui.js:1564    makeBtn(tr("apps_ui.delete", "(U+1F5D1) Delete"), { danger: true })
comfyui-mcp-panel.js:6670   tr("panel.locked_set", "(U+1F512) set")
comfyui-mcp-panel.js:6808   a.textContent = tr("panel.star_...", "(U+2B50) Star ...")
comfyui-mcp-panel.js:6911   a.textContent = tr("panel.join_the_discord", "(U+1F4AC) Join the Discord")
civitai-ui.js:1049          `${it.type === "video" ? "(U+1F3AC)" : "(U+1F5BC)"} ...`
```

And the JS fallbacks understate it, because `tr(key, fallback)` **resolves the catalogue first**. The shipped catalogues carry them independently:

```
locales/{ar,en,es,fa,fr,ja,ko,pt-BR,ru,tr,zh,zh-TW}/main.json
    16 emoji-presentation-default characters EACH   (192 total, U+FE0F: 0)
```

So on a default install the panel renders emoji-font text in **every locale**, and this PR does not change that.

**Why I am not proposing a CSS fix.** I was about to suggest excluding the emoji font from the panel's stack (`font-family: var(--font-inter, "Inter", ui-sans-serif, system-ui, sans-serif)` at `comfyui-mcp-panel.js:31136`). That does not work: none of the panel's stacks names an emoji font *already* — the browser reaches `seguiemj.ttf` through **last-resort fallback**, which CSS cannot disable. Only removing the characters removes the dependency.

**Scope, honestly stated.** That the emoji font is the crash trigger is still a hypothesis — #2023's dump has not been read and the reporter has not sent the three-line version paste. What is now measured is narrower and does not depend on that: *if* the emoji font is the trigger, this PR does not stop the panel reaching it.

**Correcting my own changelog entry on this branch.** It reads "the emoji-presentation removal now covers every shipped web/js file rather than the main bundle". That is true about U+FE0F and misleading about the goal — 44 occurrences in `web/js` and 192 in the catalogues still request the emoji font. The entry should say it removes the variation selectors, not that it removes emoji presentation.

Filing the catalogue half as its own issue rather than expanding this PR: 16 keys across 12 locales is user-visible copy in languages I cannot review, and that is an owner call, not a drive-by.

artokun:
Correcting one line in my comment above: I said I would file the catalogue half as its own issue. I am not going to, and it is better to say so than to leave a promise hanging.

Checked against our own filing bar — ours, still reproduces, and cost the user something — it fails the third test. Nothing about the catalogue emoji has demonstrably cost anyone anything unless the DirectWrite hypothesis is right, which is exactly what has not been established. The bar also says one report per root cause rather than per symptom, and the root cause here is panel#2023.

So the measurement is recorded on #2023 instead. This PR keeps its narrowed, now-accurate scope: it removes the variation selectors.

PR #2214, opened 2026-09-03
fix(2201): preserve panel-wide chat when canonical history is unavailable

### LINKED ISSUE #2201: Preserve panel-wide chat when canonical history is unavailable during workflow switch
Switching the open ComfyUI workflow while using the Codex backend caused the entire visible chat transcript to disappear and the next message to arrive as the first message of a fresh agent session. This is a silent loss-of-continuity failure. Codex is hard-wired to panel-owned scope, and the panel workflow-switch path explicitly promises the same conversation, so this is not intended per-workflow behavior.

Likely source-level root cause: the remount restore path collapses an unavailable or timed-out IndexedDB canonical read into an ordinary empty history. When durableActive is absent, it clears CURRENT_THREAD_KEY and SESSION_KEY, resets the feed, publishes a null active pointer, and persists. If the localStorage shadow was trimmed (which the panel documents as lossless because IndexedDB is canonical), a transient canonical-read miss therefore looks exactly like a real empty/reset archive and severs both the transcript and agent session.

### Steps to reproduce
1. Use comfyui-mcp panel 0.15.159 with backend Codex and an existing multi-turn conversation.
2. Switch the active ComfyUI workflow tab to another saved workflow.
3. Observe that the panel transcript becomes empty and the next user message starts a new agent session instead of continuing the panel-wide Codex conversation.
4. The live canvas remained intact and reachable; only chat/session continuity disappeared.

### Exact error (scrubbed)
No error was shown. The failure was silent. The agent received the post-switch message as the first message of a new session.

### Fix

Do not treat an unavailable or ambiguous canonical-history read as proof of an empty archive. Preserve the currently selected thread/session across a workflow remount until canonical hydration positively confirms a reset/deletion. Make ChatHistoryStore.load expose canonical-read availability (or throw a typed unavailable result), retry hydration, and keep CURRENT_THREAD_KEY / SESSION_KEY plus the painted transcript on the ambiguous path. Only the positively empty canonical snapshot may execute the destructive no-durableActive branch.

Add a regression test for: panel-owned Codex conversation + localStorage shadow absent/trimmed + IndexedDB open timeout during workflow switch/remount + later successful canonical read. The previous transcript and session must reappear, with no null active-pointer write.

A local patch was not applied because the installed panel worktree already has extensive unrelated uncommitted edits including web/js/comfyui-mcp-panel.js; overwriting that state would risk the user's work. The existing focused e2e test could not be run because Playwright is not installed in this checkout.

### Environment
Windows 11 / ComfyUI 0.33.3 / ComfyUI frontend 1.49.6 / NVIDIA RTX 3080 Ti 12 GB / comfyui-mcp 0.52.180 / panel 0.15.159 / Python 3.13.7 / PyTorch 2.13.0+cu130.

artokun:
## Summary
- Switching workflows remounts the panel. If IndexedDB is busy or times out, `ChatHistoryStore.load` used to collapse that miss into an empty snapshot. Combined with a trimmed localStorage shadow, remount restore then cleared `CURRENT_THREAD_KEY` / `SESSION_KEY`, reset the feed, and persisted a null active pointer — so a panel-wide Codex conversation looked like a fresh session.
- Canonical reads now distinguish unavailable vs positively empty. An unavailable load does not enqueue a shadow-only write. Remount restore preserves the current thread/session until a later successful canonical read can bind, and only a confirmed empty archive may reset.

Fixes #2201

## Tests
- `node --test browser_tests/unit/chat-history-store.test.mjs browser_tests/unit/history-canonical-unavailable-restore.test.mjs browser_tests/unit/a2ui-repaint-live-card.test.mjs` (84 pass)
- Log: `C:\Users\Artokun\AppData\Local\Temp\grok-goal-dba11719f7d4\implementer\tests-panel-2201.log`


artokun:
Both were real, and both had a destructive edge. Fixed and pushed.

**1. Availability was decided by the request, not the transaction.**

A `get` can succeed and its transaction abort afterwards — and because the promise was already settled in `req.onsuccess`, the `tx.onerror`/`tx.onabort` handlers were **dead code**. An aborted read therefore reported `available: true`, and `load()` promoted a possibly-trimmed localStorage shadow as the write intent: precisely the reset this change exists to prevent. Availability now comes from `tx.oncomplete`.

**2. The delayed retry could overwrite newer user state.**

The generation guard only moves on `destroy()`, so it cannot see a user who selected or created a chat while IndexedDB was recovering. Applying then reselected the mount-time thread, repainted its transcript over the one they were reading, and could fence an in-flight reply. It now re-validates the thread pointer before applying, and binds from the **live** session pointer so a session rotated while waiting isn't reverted to the mount-time capture.

## Why (1) could ship at all: four test doubles, wrong in two different ways

Fixing it made suites **hang**, which turned out to be the doubles rather than the code:

- three fired `tx.oncomplete` only for **writes**, so any reader that correctly waits for the transaction hung forever instead of failing;
- `legacy-shadow-quota` settled a **readonly** transaction *synchronously at creation* — before its own `get` handler ran, which is backwards from real IDB, where complete fires after the request handlers. That one silently returned the pre-request default.

A fake that cannot model transaction completion cannot test code that depends on it — which is the same point that file's own header already makes about keys.

## Verification

Full panel gate: **8263 unit tests** (8262 pass, 1 todo), typecheck and `check-tool-vocabulary` clean, and **`check-panel-scope` OK**.

That last one is worth a note: it *could not run* in this worktree (a fresh worktree has no `node_modules`) and reported so rather than passing. It is the only gate that catches a name resolving in a sibling function — exactly the shape of the guard I added — so I linked `node_modules` and ran it rather than shipping without it.

artokun:
Adversarial pass (mine — codex is out of tokens). Four attempts to break it, none landed. Listing them so the review is falsifiable rather than a thumbs-up.

**1. The restore plan fails in the safe direction.** `planRemountHistoryRestore` treats anything but `canonicalAvailable === true` as `preserve`. A truthy-but-not-true value, an undefined from a caller that forgot the field, a slow probe — all preserve the chat rather than resetting it. For a decision whose wrong answer *deletes a user's conversation*, positive-only is the right asymmetry and it is implemented as written.

**2. The 2s open timeout does not leak a late success.** This was my main suspicion: a bound that races rather than aborts, where the open completes at 2.1s and mutates state after the decision. It does not —

```js
const finish = (db) => {
  if (settled) { db?.close?.(); return; }
  settled = true;
  if (timeout) clearTimeout(timeout);
  resolve(db);
};
```

Idempotent, and a database arriving after the timeout is *closed* rather than dropped on the floor as an unowned `IDBDatabase`. The comment above the timer claims exactly that and the code does it.

**3. A blocked upgrade is handled, and for a reason with a test behind it.** `onversionchange → db.close()` means a second ComfyUI tab holding the old version does not make the first one wait out the timeout and report "IndexedDB unavailable" on a database that is merely busy. The comment names the test that catches it (green serially, red in parallel), which is the failure shape that would otherwise be blamed on flakiness.

**4. Timeout → unavailable → preserve is coherent end to end.** A slow or blocked IDB yields `{ available: false }`, which the plan reads as `preserve`. So the degraded path keeps the chat instead of resetting it, which matches what the issue is about.

Nothing to change from here. The one thing I could not check statically is whether `preserve` can leave the visible chat bound to a thread from a *different* workflow after a remount — that needs the live rig, and the issue's own framing (panel-wide chat is deliberately not per-workflow) suggests it is intended rather than a gap.

Issue #852, opened 2026-08-09
Make panel_add_node schema-drift error point to panel_refresh_nodes


artokun:
### What happened / root cause
After a model file was moved from checkpoints/ to diffusion_models/, panel_add_node for UNETLoader fetched the fresh backend definition and detected required-input drift on unet_name. It refused the add correctly, but its only recovery text said to reload the entire ComfyUI tab even though panel_refresh_nodes is the built-in, non-destructive recovery and was verified to resolve the condition immediately.

### Steps to reproduce
1. Load a ComfyUI tab so UNETLoader is registered.
2. Add or move a diffusion-model file, changing UNETLoader.unet_name combo options.
3. Call panel_add_node for UNETLoader.
4. Observe the schema-drift refusal.
5. Call panel_refresh_nodes, then retry panel_add_node; it succeeds.

### Exact error (scrubbed)
"UNETLoader" required input "unet_name" was added or retyped since this page loaded its node schema. Reload the ComfyUI tab so the frontend picks up the updated node definition, then retry.

### Fix
applied locally: yes
upstream-only: no

Patched ~/ComfyUI/custom_nodes/comfyui-agent-panel/web/js/comfyui-mcp-panel.js so the refusal names the verified recovery.

Patch:
- added or retyped since this page loaded its node schema. Reload the ComfyUI tab so the frontend picks up the updated node definition, then retry.
+ added or retyped since this page loaded its node schema. Call panel_refresh_nodes to refresh loader options without reloading the tab, then retry (or reload the ComfyUI tab).

Verification: panel_refresh_nodes returned refreshed:true, and the identical panel_add_node call then created UNETLoader successfully.

### Environment
Windows 11
ComfyUI 0.31.1
ComfyUI frontend 1.48.7
GPU AMD Radeon RX 7800 XT, 16 GB VRAM
comfyui-mcp 0.50.53
panel 0.11.50

artokun:
Claimed — draft PR open, branch `fix/852-schema-drift-recovery`. Thanks for verifying the recovery before filing; that's the part that makes this landable immediately.

artokun:
Shipped in **0.11.53** (PR #855 — published to the Comfy Registry). Your patch was right; landed with the reasoning and coverage around it.

## What went in

Your wording, extended in two ways I thought were worth it:

[...]

PR #2215, opened 2026-09-03
fix(2203): recover prompt_id when panel_run fetch fails after dispatch

### LINKED ISSUE #2203: Recover prompt id when panel_run fetch fails after dispatch
panel_run sent a verified run-to-node /prompt request, but the fetch failed after dispatch. The tool could not determine whether ComfyUI accepted the prompt and returned no prompt_id, leaving the user blocked from safely retrying because that could duplicate an expensive render.

### Steps to reproduce
1. Use a connected sidebar panel with a live local ComfyUI canvas.
2. Call panel_run with to_node_id on an output node.
3. Observe a post-dispatch Failed to fetch during the /prompt request.

### Exact error (scrubbed)
run-to-node scope for node 20: a verified-scoped /prompt request FAILED to complete — the /prompt request itself threw (Failed to fetch). 0 of 1 batch prompts were confirmed queued before the failure. This one is NOT confirmed queued — but it had already left the panel, so whether ComfyUI accepted it CANNOT be determined.

### Fix
applied locally: no
upstream-only: yes
The orchestrator/panel should attach a client-generated dispatch correlation id before issuing /prompt and reconcile it against queue/history or panel execution events after a fetch failure, so an accepted prompt can recover its prompt_id and an unaccepted prompt can be safely retried. A local retry was intentionally not attempted because it could duplicate the render.

### Environment
Windows 10; ComfyUI 0.33.1; frontend version unavailable because the headless /system_stats route was unreachable; NVIDIA RTX 3090 24GB; comfyui-mcp 0.52.150; panel 0.15.151.

artokun:
Fixes #2203

panel_run could send a verified /prompt request and then lose the receipt when fetch threw after dispatch. The tool returned no prompt_id and could not say whether retry was safe.

This stamps a client-generated `cmcp_dispatch_id` onto `extra_data` before the request leaves, then reconciles it against `GET /queue` and `GET /history`. A unique match restores the prompt_id so the run is tracked; a confirmed miss is treated as not queued and is safe to retry. Unreadable queue/history stays unknown, so a blind retry is still refused.


artokun:
CI's failure here was a **timer**, not the feature.

The log showed a cluster of `cancelledByParent` in `run-scope-guard.test.mjs`, first casualty `#556 r6: an attributed post whose fetch THROWS…`, with:

```
error: 'Promise resolution is still pending but the event loop has already resolved'
```

Everything after it in the file was cancelled, which made about a dozen unrelated tests read as broken.

**Cause.** The #2203 recovery polls `/queue` then `/history` up to four times, sleeping 40ms between attempts with a real `setTimeout`. Neither call site in `run-scope-guard.js` forwarded the pacing, so every unit test that reached the recovery paid ~120ms of wall clock. When the runner drains the event loop before those timers fire — a loaded CI runner, not this machine — node reports the pending promise and cancels the file.

It passes locally either way, which is exactly why it looked like a product failure. Full suite here: **8254 passed, 0 failed, 0 cancelled**, before and after.

**Fix.** `recoverPromptIdAfterDispatch` was already built to be driven (`delayMs`, `sleep`) — the guard simply never threaded it. Both factories now take an optional `recoverDelayMs` that defaults to today's behaviour, and the tests pass `0`, at which point the poll skips the sleep entirely (`i > 0 && wait > 0`) and schedules **no timer at all**.

Deterministic rather than a widened tolerance, per this repo's own rule about #852 — and the neighbouring test's `keepAlive()` is no longer the only thing standing between this file and a cancelled run.

Evidence the injection actually bites: `run-scope-guard.test.mjs` went **4412ms → 3651ms**. typecheck and `check-tool-vocabulary` clean.

artokun:
All four were real, and three of them are the same failure wearing different clothes: the module claiming a receipt, or authorizing a retry, on evidence that cannot support it. Fixed and pushed.

**1. The queue mark was returned as `recovered`.** It is not a per-dispatch receipt — every prompt in a batch shares it, and the counter restarts on page reload while ComfyUI's history persists. A single match can name an earlier batch item or a row from a previous session, which reports a **stale prompt_id** and counts this failed post as observed. It is now surfaced as `markMatch` for the human, while the verdict stays `unknown`.

**2. `readJson` parsed non-2xx bodies.** A 500 `/history` answering `{error: …}` read as a valid, empty history map; combined with an empty queue that became `absent` and authorized a retry *on the strength of the server being broken*. Only 2xx counts now.

**3. "Safe to retry" was a promise this evidence cannot make.** A bounded sequence of empty reads cannot prove a request will not be accepted after the last one. The message now names the miss, says plainly that the observation is bounded, and tells the caller to check the queue when a duplicate would be expensive. The test whose *name* encoded the old contract (`…names a safe retry…`) was renamed along with it — a test name is part of the contract.

**4. `missingPromptIds` was banked before recovery ran** and never taken back, so a receipt successfully recovered from the queue still carried `uncertainCount > 0` and reported `queued_unknown` instead of accepted. The unscoped branch already ordered this correctly; the scoped one did not.

Worth flagging: **(4) was unpinned.** I reverted the fix and the whole 171-test file still passed, so I wrote a test for it — a malformed 200 whose id is recovered from the queue must leave `missingPromptIds === 0`. Moving the increment back ahead of recovery now fails that test and nothing else.

Full panel gate: **8277 unit tests** (8276 pass, 1 todo), typecheck and `check-tool-vocabulary` clean.

artokun:
Adversarial pass (mine — codex is out of tokens). The premise this rests on is verifiable and I checked it against ComfyUI rather than the docstring.

**`extra_data` really does survive into history.** The whole design depends on a stamped `extra_data.cmcp_dispatch_id` coming back on a queue/history row. From ComfyUI's own `execution.py`:

```python
self.history[prompt[1]] = {
    "prompt": prompt,          # the FULL tuple, extra_data at index 3
    "outputs": {},
    'status': status_dict,
}
```

The tuple is stored whole and keyed by `prompt[1]` (the prompt_id), so a stamped field is still there to correlate against. That matches the shape the module's header comment describes, and it is the fact the recovery would silently fail on if it were wrong.

**Stamping `extra_data` is the right place** — not a node input, which the graph owns and a reserialize could drop, and not a new top-level key, which ComfyUI does not carry. `extra_data` is passed through to `get_input_data` and echoed back untouched.

**One caveat worth a comment, not a change.** `task_done` runs `prompt = process_item(prompt)` before storing when a `process_item` hook is present. Default ComfyUI passes none, so this is fine as shipped — but a deployment that installs a hook stripping or rewriting `extra_data` would break correlation with no signal here beyond "not found". Given the fallback is presumably to report the dispatch as unrecoverable rather than to invent a prompt_id, that degrades safely; worth knowing it is a real (if exotic) path rather than an impossible one.

**Also checked:** the id is a UUID when `crypto.randomUUID` exists and a time+random fallback otherwise, so two in-flight dispatches cannot collide; and correlating from either the queue (still pending) or history (already done) is right for #2203, since both prove the dispatch *landed* — which is the question the failed fetch left open.

I could not exercise this end-to-end: the ComfyUI on this box has an empty history (`{}`), and enqueuing a probe run onto someone's live queue is not something I will do to verify a shape I could read from source instead.

Issue #2234, opened 2026-09-04
Panel re-composes a finished run storyboard every sweep while the bridge is down: 4,216 files / 9.8 GB of temp in 12 h


artokun:
### What happened / root cause

With the panel bridge DOWN, a single finished render's storyboard is re-composed
and re-uploaded to `ComfyUI/temp` every reconcile sweep, forever. Over ~12 hours
one 5.4 MB video (`LOOP_HQ_00001_.mp4`, 2112x1408 / 8 s) produced **4,216 files
and 9.8 GB** of `storyboard_LOOP_HQ_*.png` + `poster_LOOP_HQ_*.png`. Deleting
them all did not stop it — 6 fresh files appeared within 45 s.

Two behaviours that are each deliberate combine into an unbounded disk write:

1. `run-completion.js` bounds completion replays at `maxUnackedDeliveries = 3`,
   but `releaseUnackedDelivery()` DECREMENTS that budget on a transport refusal,
   documented as intentional: *"a run whose frames are ALWAYS refused is retried
   without limit (#370 / #1739 untouched: nothing ever landed)"*. With the bridge
   down every frame is refused, so the budget never accumulates and the 20 s
   sweep re-arms indefinitely. That part is working as designed.

2. Each of those retries re-enters `produce()` in `run-completion-frame.js:767`:

   ```js
   const storyboardIdentity = v?.storyboardIdentity || createStoryboardIdentity();
   ```

   `createStoryboardIdentity()` mints a FRESH per-attempt identity by design
   (#1718 / #1834 cache-busting), and `storyboardUploadName()` /
   `storyboardPosterUploadName()` therefore produce a NEW filename every attempt,
   so line 825 and line 888 upload two more PNGs into `temp/` each sweep.

The `v?.storyboardIdentity` fallback shows the record is meant to be able to
carry an identity across attempts, but nothing writes it back after a successful
compose, so the reuse path is never taken. The retry is bounded in the dimension
that was considered (agent turns) and unbounded in the one that was not (bytes
written), and the artifact is byte-identical every time — the source video has
not changed.

The user-visible symptom is also that the same storyboard keeps reappearing in
the panel chat, which is how this was noticed.

### Steps to reproduce

1. Queue a video render via `panel_run` and let it finish.
2. Have the completion frame's transport refuse — here the panel MCP server
   disconnected from the agent session, so no receipt could ever arrive.
3. Leave the ComfyUI tab open.
[...]

Issue #2233, opened 2026-09-04
Long panel_set_widget text writes time out after applying


artokun:
### What happened / root cause
Two panel_set_widget calls for ordinary CLIPTextEncode text values were applied on the live canvas, but the panel did not return graph_set_widget acknowledgements within 90000 ms. A following short filename_prefix write succeeded. Immediate panel_query_graph readback showed both timed-out text mutations had applied correctly. This points to the live panel text-widget mutation path losing or delaying its acknowledgement, especially for longer multiline values or a backgrounded tab.

### Steps to reproduce
1. Open a bound ComfyUI workflow with CLIPTextEncode nodes.
2. Call panel_set_widget on node 3 text with a multiline prompt around 800 characters.
3. Call panel_set_widget on node 4 text with a negative prompt around 650 characters.
4. Both calls may time out after 90000 ms.
5. Call panel_query_graph for nodes 3 and 4; the new values are present.
6. A short panel_set_widget filename_prefix mutation succeeds.

### Exact error (scrubbed)
Error: Panel tab ~/workflows/??????_Chibi+IconDesigner.json did not reply to "graph_set_widget" within 90000 ms - the ComfyUI tab may be backgrounded or frozen. This command MUTATES and was already delivered to the tab, so it may have been applied despite the missing reply.

Retry tokens from this occurrence: aad465cb-aa91-401d-ac45-c1a594033f7f and 33e85f69-44f4-461c-87bb-985550f48a8c.

### Fix
applied locally: no
upstream-only: yes

Precise change needed: ensure the graph_set_widget frontend handler sends its acknowledgement immediately after the widget value is committed, including for multiline text values and backgrounded tabs. If the acknowledgement can be delayed by browser throttling, the broker should reconcile the mutation by graph identity/node/widget readback before returning outcome-unknown.

Temporary workaround: query the exact node immediately after timeout and continue only when the intended value is confirmed.

### Environment
Windows 11
ComfyUI 0.34.0
ComfyUI frontend 1.49.6
Python 3.12.11
PyTorch 2.14.0+cu130
GPU NVIDIA GeForce RTX 4070 SUPER 12 GB
comfyui-mcp 0.52.180
panel 0.15.159

artokun:
[...]

PR #2237, opened 2026-09-04
chore: release v0.15.169

artokun:
Release v0.15.169 covering:

- #2233 / #2236 — graph_set_widget acknowledges a long text write without waiting on rAF
- #2234 / #2235 — refused completion retries reuse the storyboard identity and skip re-upload


artokun:
Checked this one for release integrity: the `[0.15.169]` section promotes **exactly** the two entries sitting in main's `## [Unreleased]` right now — `graph_set_widget` long-text ack (#2233/#2236) and the refused-completion storyboard identity. Nothing merged-but-unlisted, nothing listed-but-unmerged.

**One sequencing note, since it is cheap to act on and expensive not to.** Merging this re-conflicts every open panel PR on `CHANGELOG.md` — a release promotes the whole Unreleased block, so all eight branches that touch it go `CONFLICTING` at once. I swept them ~20 minutes ago after #2236 landed and they are all clean and building right now:

```
#2215 #2214 #2210 #2204 #2202 #2200 #2196 #2190
```

So anything in that list you intend to merge is cheaper to merge **before** this, not after — otherwise each needs a repair commit that restarts its four required checks, and with eight of them the queue rarely holds still long enough for any to be green simultaneously. That queue-thrash is the cost measured in comfyui-mcp#2830, which proposes per-PR changelog fragments as the root fix; the same argument applies to this repo.

If you would rather merge this first, say so and I will re-sweep the eight afterwards — it is mechanical, and I have the procedure. Just flagging that the order is worth a moment's thought rather than being incidental.

PR #2238, opened 2026-09-04
fix(#1215): refuse leftover previous-tab graph after a failed open

### LINKED ISSUE #1215: After panel_open_workflow tab switch, graph reads (panel_graph_outline/query) silently return the PREVIOUS tab's graph while the fence/active both name the new workflow

artokun:
## Summary
After `panel_open_workflow` switched tabs, graph reads could still serve the previous canvas while the fence, `panel_list_workflows`, and `panel_set_workflow_target({mode:"current"})` all named the new workflow.

#1951/#2038 already close the capture/repaint poison. The remaining hole is the open that **moves the pointer** then fails the safe-repaint rebind because the target has no comparable `activeState.nodes`. Existing predicates fail-open there (`graphReadDesynced` only fires on an empty live root; shape comparison is inconclusive when `nodes` is missing; an untagged leftover canvas has no uuid mismatch). Outline/query then returned the prior graph.

Fail closed: a nonempty live root with no readable current state is `[root-state-unreadable]`. `panel_set_workflow_target` is not a remedy. Already-current untagged capture with matching state (#2038) still admits.

Closes #1215

## Tests
```
node --test browser_tests/unit/graph-binding.test.mjs browser_tests/unit/open-active-settle.test.mjs
```
200 pass, 0 fail (with changelog-integrity).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,14 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+- After panel_open_workflow switches tabs but cannot repaint, panel_graph_outline / panel_query_graph refuse instead of serving the previous tab's graph under the new workflow's fence — panel_set_workflow_target is not a remedy (#1215)
+
 ## [0.15.169] - 2026-09-04
 
 ### Fixed
 - graph_set_widget acknowledges a long CLIPTextEncode / multiline text write as soon as the live editor holds the value, instead of waiting on a backgrounded-tab rAF flush until the 90s relay times out (#2233, #2236)
 - Refused completion retries of a finished video reuse the composed storyboard identity and skip re-upload, so a down bridge cannot fill ComfyUI/temp with unique `storyboard_*.png` / `poster_*.png` copies every sweep (#2234, #2235)
-
 
 ## [0.15.168] - 2026-09-04
 
--- a/browser_tests/unit/graph-binding.test.mjs
+++ b/browser_tests/unit/graph-binding.test.mjs
@@ -21,6 +21,7 @@ import {
   activeWorkflowProvenEmpty,
   graphEmptyBindingUnproven,
   graphReadDesynced,
+  graphRootUnprovenAgainstActiveState,
   missingNodeStateReportsNodes,
   graphRootMidPopulation,
   graphRootMismatchesActiveWorkflow,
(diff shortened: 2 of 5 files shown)


artokun:
Adversarial pass (codex is out of tokens, so these are mine). I tried three ways to break it and could not; recording what I checked so the review is falsifiable rather than a thumbs-up.

**1. False positive on programmatic edits — does not apply.** My first guess was that this refuses a perfectly good canvas: I measured on a live rig today that ComfyUI's tracker does not capture programmatic graph writes, so `activeState` lags the canvas after `panel_add_node`. But lagging is not *absent* — `activeState.nodes` stays a well-formed array, so `activeWorkflowCurrentState` returns non-null and this guard stays silent. The comment already says that case belongs to the shape guard, and the ordering in `resolveGraphBindingVerdict` puts `root-shape-mismatch` first, so it genuinely gets there.

**2. Fail-open coverage is real, not asserted.** The directions that must NOT refuse are each pinned: empty live canvas, absent workflow service, subgraph scope, and no arguments at all. That is the direction a fail-closed guard gets wrong silently, and it is the direction most reviews skip.

**3. Cross-PR interaction — none found.** I checked whether the in-flight API/prompt-graph load path could leave a workflow without comparable state and trip this. It cannot from here: that work is comfyui-mcp#2598, on the MCP side, and the tracker serializes the live LiteGraph in UI form regardless of the source document's format, so `nodes` stays an array.

**One residual I could not settle without a rig.** The guard is deliberately not gated on `includeBaselineReadGuard`, so it also governs reads. That leaves the window between a canvas being loaded and the tracker publishing `activeState` — a read landing inside it would refuse with `root-state-unreadable` on a canvas that is actually correct. `settleOpenedWorkflowReadable` / `settleOpenedWorkflowActive` appear to exist for exactly that window on the open path; what I cannot tell statically is whether every entry point that reaches `resolveGraphBindingVerdict` has passed through a settle first, or only the open path has.

If a reader hits `root-state-unreadable` immediately after something *other* than a tab switch, that is the shape to suspect. Worth a sentence in the message or the comment naming which paths are settled, so the next person does not have to derive it.

artokun:
Settled the residual from my last comment — it was answerable by tracing, not only on a rig.

**The asymmetry is 16 to 1.**

```
assertGraphBoundToActiveWorkflow(…)   16 call sites   (one definition, one comment excluded)
settleOpenedWorkflowReadable(…)        1 call site    (inside the open path)
```

`resolveGraphBindingVerdict` has exactly one caller — `assertGraphBoundToActiveWorkflow` — and that is invoked from sixteen places across the read and mutation surface (`graph_outline` at 10066, and fifteen more). The settle runs at one, on open. So the new `root-state-unreadable` refusal is reachable from fifteen entry points that have not passed through a settle.

**That is not automatically a bug, and the distinction matters.** For the case this PR is built on it is *correct*: a user switching tabs in ComfyUI's own UI never goes through `panel_open_workflow`, so no settle runs, the canvas is still the previous graph, and refusing is exactly right. The guard firing without a settle is the feature.

The exposure is the narrower one: if `activeState` can be transiently absent while the canvas is *already correct* — a completed switch, a reconnect, a reload — then a read landing in that window refuses spuriously, and the message tells the user to re-open a workflow that is already the right one. Whether that window exists on the non-open paths is the part I still cannot answer statically; it needs the rig.

**Cheap way to find out without one:** the refusal is distinctive, so `root-state-unreadable` appearing in a session where the user did *not* just switch tabs is the tell. If you would rather not wait for that signal, gating the new branch on `includeBaselineReadGuard` for the read paths only would keep the mutation protection (which is where a wrong graph does damage) while leaving reads to the existing guards until the window is characterised.

Not asking for a change — the fail-closed direction is defensible for a guard about serving the wrong graph. Recording the count so the next person does not have to derive it.

artokun:
This went red after my review comments — here is the cause, which is mechanical rather than a problem with the design.

```
# tests 8315   pass 8304   fail 7
Actual message: "switchRepaintUnproven is not defined"
```

Not the read-refusal I speculated about. It is a **ReferenceError inside the `new Function` harnesses**.

`switchRepaintUnproven` is module-scope state (`let switchRepaintUnproven = false;`, line 1303) and is read at line 10066 inside `assertGraphBoundToActiveWorkflow`. Three test files rebuild that function from source and inject its dependencies **by name** — `stale-tag-read-executor.test.mjs` lists 21 of them explicitly. A module-scope `let` is not among them, so the rebuilt copy throws the moment the line runs.

**The branch updated one harness of three:**

| harness | injects `switchRepaintUnproven` |
|---|---|
| `graph-binding.test.mjs` | yes (8 references) |
| `stale-root-fence-after-tab-switch.test.mjs` | **missing** |
| `stale-tag-read-executor.test.mjs` | **missing** |

All 7 failures are in those two. The fix is to add `"switchRepaintUnproven"` to each one's parameter-name list and pass a value — `false` for the ordinary case, `true` for any test that wants the unproven-repaint path.

**Worth flagging as a recurring shape rather than a one-off.** This is the third instance today of the same trap: `link-id-headroom.js` documents it for `graph_connect`; I hit it myself on #2196 a couple of hours ago with `ReferenceError: warningField is not defined` after factoring a helper to module scope; and now this. Any new module-scope binding consumed inside a harness-rebuilt function fails this way, and it always fails *only* in the harnesses, never in the browser — so it looks like a test bug and is actually a coupling between production scope and a hand-maintained injection list.

If it is worth preventing rather than re-hitting, the cheap gate is a test asserting that every harness rebuilding `assertGraphBoundToActiveWorkflow` injects the same name set — the list is discoverable from the source region, and the three files already agree on 21 of them.

PR #2240, opened 2026-09-04
test(panel): pin the phrase comfyui-mcp keys its models fallback on

artokun:
Found reviewing comfyui-mcp#2793. This repo owns a sentence another repo greps, and nothing on either side would notice it changing.

## The coupling

`comfyui-mcp`'s `list_local_models` recovery decides the panel cannot serve `models/<category>` by matching this module's rejection text:

```ts
return /operation must be one of/i.test(parts.join(NEWLINE));
```

and then falls back to reading the model list out of `object_info` (comfyui-mcp#2511).

The only structured signal on the wire is `invalid_input`, which this module *also* uses for "operation is required" — too coarse to tell the two apart — so the prose is what carries the distinction.

## Why a test rather than a comment alone

Reword that sentence and the recovery stops firing **silently**. `list_local_models` goes back to returning nothing useful, with no error raised in either repo. Nobody rewording a message in this file has any reason to suspect a tool in another repository depends on its phrasing.

The mcp-side test cannot catch it either: it mocks the rejection with its own hardcoded copy of the sentence, so it keeps passing after this repo diverges.

This makes the reword fail **here**, in the repo doing the rewording.

## Deliberately narrow

Pinned on the substring mcp actually matches, not the whole sentence. Verified both directions:

```
reworded 'must be one of' -> 'must be'   test FAILS   (the guard bites)
reordered the operation list             test PASSES  (no false alarm)
```

So the allowlist can grow, shrink or be reordered freely — only the matched phrase is frozen.

## Not an endorsement of prose coupling

The better fix is a distinct error code (`unsupported_operation`) that mcp matches instead of the sentence, which needs a change on both sides and is your call. This is a guard until that lands, and the comment on the constant says so and names the code as the intended replacement.

Full panel suite green: 8,315 pass, 0 fail.


artokun:
Adversarial pass (mine — codex is out through Sept 6). **The pin is the right shape and I verified the contract end to end. One residual it cannot cover, which is worth one extra line.**

## Verified across the repo boundary

Both sides use the identical predicate, which is the thing most likely to have drifted:

```
comfyui-mcp (#2793)   return /operation must be one of/i.test(parts.join("\n"));
panel  (this PR)      assert.match(text, /operation must be one of/i, …)
```

and the panel message really does carry it (`ALLOWED_OPERATION_NOTE`, thrown via `invalidInput`). The test is **behavioural** — it calls the resolver, catches the throw, and matches the thrown message — not a source grep, which is the distinction that makes it worth having.

I also checked the thing that silently defeats every prose predicate in this repo: `ALLOWED_OPERATION_NOTE` is a bare English constant, and `fetch-comfyui-read.js` contains **zero `tr(` calls**. So the phrase is not translated, and mcp's grep is locale-safe today.

## The residual

The pin catches a **rewording**. It cannot catch a **translation**, because it runs under one locale.

If a future i18n pass wraps this string — `tr("...", "operation must be one of …")` — this test still passes: `tr` resolves the `en` catalog, which `i18n:build` generates from the very same source string. Meanwhile a user on any of the other 11 catalogs gets a translated message, `/operation must be one of/i` misses, and comfyui-mcp silently stops falling back to `object_info` for `models/<category>`. The failure is invisible from the English side, which is the side every test and every developer runs on.

The comment at the top of the test is a good defence against a human doing this deliberately. It is no defence at all against a sweep that wraps user-facing strings in `tr()` without reading them — which is exactly how this repo's i18n coverage grew.

**Cheapest closure is a negative source assertion**, which is what source assertions are actually good for:

```js
assert.doesNotMatch(source, /tr\(\s*["'][^"']*operation must be one of/i,
  "this message must stay UNtranslated — comfyui-mcp greps it across the repo boundary");
```

That fails the moment someone makes it translatable, in English, before any locale is built. Nothing blocking.

artokun:
Closed the residual in `3e337285`.

A negative source assertion — `fetch-comfyui-read.js` must contain no `tr(` call — so the contract breaks **in English, at the moment the string becomes translatable**, rather than silently for the 11 non-English catalogs at runtime.

**Mutation-verified with a realistic sweep.** The naive mutation (wrap in `tr()`) only proves the module fails to load, since `tr` is not imported there — the file dies and every test "fails" for the wrong reason. Wrapping it *and* defining `tr`, which is what an actual i18n pass would do, kills exactly this test and nothing else.

## The part worth reporting against myself

**My first version of this assertion was vacuous, and the mutation is the only reason I know.** I wrote the regex through a heredoc, which turned `\b` into a literal **backspace byte (0x08)** — the committed line read:

```
/<0x08>tr\s*\(/
```

It could never match anything. It passed on clean code, passed on the mutation, and looked completely normal in a diff; `cat -A` rendering it as `^H` is what exposed it. Had I trusted the green run, this PR would have shipped a guard that asserts nothing — which is worse than no guard, because the next person reads it as coverage.

Same trap the panel has hit before with `\b` in generated matchers. The durable lesson is narrower than "escaping is hard": **a regex containing `\b`, `\s` or `\(` should not be written through a shell heredoc at all** — use an editor, and check `cat -A` on the result before believing a passing run.

**Verification:** `npm run test:unit` **8,342 pass / 0 fail**.

artokun:
**Correcting one word in my earlier comment.** I wrote that if this phrase were translated, comfyui-mcp's grep would miss and the `models/<category>` fallback would stop firing **"invisibly"**. That is wrong, and I only found out by reading the consuming side (mcp#2793) afterwards.

The consumer does not swallow an unmatched error:

```ts
} catch (panelErr) {
  if (!isUnsupportedModelsPanelRead(panelErr)) throw panelErr;
  memo.modelsOpUnsupported = true;
}
```

An error the predicate does not recognise is **rethrown**, and its caller (`fetchModelsRoute`) does not catch it either. So a translated or reworded panel message does not produce a silent empty inventory — it surfaces the panel's actual rejection to the caller. Loud, not silent.

I did not trace it all the way to the tool boundary, so I am not claiming exactly what the user sees; I am claiming it is not swallowed at the two levels I read.

**The guard is still worth having, for a smaller reason than I gave.** An error where a working recovery should have happened is a real regression — `list_local_models` on a remote instance stops returning inventory it could have recovered from `object_info`. But that is a visible failure someone would report, not the invisible rot I described. I overstated it.

Worth noting what makes the fail-safe direction correct here: the predicate is used to *narrow* an error into a known-recoverable case, and everything it does not recognise is rethrown. That is the opposite of the shape I flagged on #2200 earlier today, where a catch swallowed by default — same construct, opposite polarity, and the polarity is what decides whether an unknown failure is loud or lost.

artokun:
**Retracting the closing paragraph of my previous comment.** I wrote that panel#2200 and mcp#2793 are "the same construct, opposite polarity — one swallows by default, one rethrows by default, and the polarity is what decides whether an unknown failure is loud or lost."

That is not true. Side by side:

```js
// panel#2200, run-prompt-snapshot.js:112
if (!isDynamicWidgetMissingError(error)) throw error;

// mcp#2793, model-resolver.ts:563
if (!isUnsupportedModelsPanelRead(panelErr)) throw panelErr;
```

Same shape. **Both rethrow the unknown error; neither swallows by default.** There is no polarity difference. I had read the two hours apart, kept each as a summary rather than as text, and generalised a tidy distinction across them that the code does not have.

The only real difference is what happens to the *known* case — #2200 swallows and continues, #2793 routes to an alternate recovery — which is a much smaller point and not a review heuristic.

**What stands, unaffected:**

- The correction itself: mcp **rethrows** an unrecognised error, so a translated phrase surfaces a real failure rather than silently dropping the fallback. My original "invisibly" was wrong, and that remains wrong.
- The guard added in `3e337285`, for the reason given: an error where a working recovery should have happened is a genuine regression, and it fails in English before any locale is built.
- panel#2200's separate finding, which was about test coverage — mutating that catch to a blanket swallow left all 8,344 panel tests green — and is unrelated to this.

Flagging it here rather than quietly, since I stated the claim in this thread.

artokun:
**Right, and verified against comfyui-mcp's actual source rather than from memory.**

The comment quoted the matcher as `parts.join(NEWLINE)`. `isUnsupportedModelsPanelRead` in `comfyui-mcp/src/services/model-resolver.ts` has:

```js
return /operation must be one of/i.test(parts.join("\n"));
```

`NEWLINE` is not a symbol that exists there.

This comment's whole job is to document a fragile cross-repository contract — the phrase `operation must be one of` is the only signal distinguishing "this panel cannot serve `models/<folder>`" from "operation is required", because both surface as `invalid_input`. A misquoted contract is **worse than an unquoted one**: someone verifying it greps for a symbol that cannot be found and concludes the contract has moved, which is exactly the wrong conclusion, since the matched sentence is unchanged.

Now quoted verbatim, with the file it came from named so the next reader can check it in one step.

`npm run test:unit` 8354 pass / 0 fail, `[check-panel-scope] OK`.

The other four panel findings in this batch I have read but not taken — the A2UI height cap (`calc(100% - 2rem)` winning when the log is shorter than `70vh`), the startup dismissal race against `historyRestoreReady`, and two on #2200's dynamic-widget reconcile. All look plausible; none is a one-line correction like this one, and each wants its own pass.

PR #2242, opened 2026-09-04
fix(2139): the unknown-staleness arm must not tell a dirty tab to load over its edits

### LINKED ISSUE #2139: panel_unexpose_subgraph_* leaves ChangeTracker behind + root-shape-mismatch; save and further edits blocked

artokun:
Found by auditing the population rather than reviewing a PR: sibling gaps were the finding in four of five consecutive reviews, so I enumerated every panel message recommending a tool that REPLACES live canvas state and checked which carried an unsaved caveat. 19 sites, 12 without one — 11 of those turned out fine. This is the one that is not.

## The defect

`workflow_open`'s staleness reply has a ternary chain. Every arm consults `dirtyNow || wasDirty` before recommending a disk load — the CONFLICT arm says so outright, and the arm after it is the proven-clean `else` — except one:

```js
: staleInfo.stale === "unknown"
  ? {
      stale: "unknown",
      reloaded: false,
      stale_hint:
        "Could not verify whether the on-disk file still matches this tab (disk read " +
        "unavailable or slower than its deadline). Treat the canvas as possibly stale; " +
        "call panel_load_workflow to be sure you have the on-disk version.",
    }
```

`panel_load_workflow` loads the on-disk copy **over the canvas**. On a tab with unsaved edits, following that instruction discards them.

Two things make it worse than a missing caveat:

- **It is the arm that admits it verified nothing.** The message opens by saying the disk read failed, then issues an unconditional instruction to overwrite.
- **The trigger is transient and common** — a disk read "unavailable or slower than its deadline" — not a rare state.

This is issue #2139's item (3) one tool over: *the tool recommends the destructive exit*.

## The fix

The dirty arm names the loss and reorders the advice; the clean arm keeps the original, useful recommendation.

Positive evidence only, per #2204: `isModified === true` is evidence of drift, `false`/`undefined` is evidence of nothing. I measured that on an isolated rig — a programmatic `LiteGraph.createNode` + `graph.add` and a direct widget write both leave the flag **false** — so the clean arm says "save first if you are unsure" rather than calling the tab clean.

**Inline, introducing no new identifiers.** Several `new Function` harnesses rebuild `workflow_open` and inject dependencies by name, so a module-scope helper would throw `ReferenceError` there. `dirtyNow` and `wasDirty` are already local to this handler, so nothing needed injecting and no harness changed.

## Verification

- **Mutation-verified**: reverting to the original unconditional hint kills the new test and nothing else.
- The test is scoped to the branch region and **comment-stripped**, because the fix ships with a comment naming these very identifiers — an unscoped `assert.match` over the handler body would pass on the comment alone.
- `npm run test:unit`: **8,341 pass / 0 fail**. `check-panel-scope` OK. `check-changelog` OK.

Does not close #2139 — that issue tracks several items and this is one site within item (3). The other sibling I found, `workflowListReadinessRefusalError` recommending `panel_open_workflow(path)` after a reconnect, is written up on #2204 and not touched here.


artokun:
Self-review found a defect in my own fix, corrected in `02f57fe0`.

The original change added `conflict: true` to the dirty branch of the `stale: "unknown"` arm. That is an overclaim. A conflict is **"the file changed on disk AND this tab has unsaved edits"** — and on that arm the disk read failed, so only the second half is known. Asserting a conflict claims precisely the thing the sentence beside it admits it could not verify.

It matters because of who reads the key. Nothing branches on `conflict` in code — the **agent** reads it, so an unverified `true` converts *"I could not check whether the file changed"* into *"the file changed"*. That is the same failure this PR exists to fix, pointed the other way: the original bug told a dirty tab to overwrite itself, and my fix told the caller a conflict existed that may not.

Now the dirty branch warns about what is at risk without asserting the half that was never established. The CONFLICT arm above keeps `conflict: true` — there staleness *is* established, so both halves hold.

**Mutation-verified:** restoring `conflict: true` kills the new test, which is scoped to the unknown arm's region so it cannot pass on the legitimate claim 35 lines above it.

Worth naming how this was found. I had reviewed a dozen other PRs today against the positive-evidence rule — `isModified === true` only (#2204), a failed read is not an empty inventory (#2793), unreadable is UNKNOWN not zero (#2771) — and did not apply it to the thing I wrote an hour earlier. The fix was mutation-verified and fully green when I posted it; correctness of the *guard* is not the same property as truthfulness of the *report*, and only the first one had a test.

**Verification:** `npm run test:unit` **8,350 pass / 0 fail**, `check-panel-scope` OK, `check-changelog` OK.

artokun:
**Both Copilot findings on this PR were real. The first is the same shape as panel#2044: the remedy could cause the loss it warns about.**

Both arms said *"save first"*. But:

```js
// web/js/lib/workflow-save.js
export async function classifyInPlaceOverwrite(wf, currentPath, readDiskBytes) {
  if (!wf || wf.isPersisted === true) return "skip";   // already overwrite:true — untouched
```

An ordinary save of an already-persisted tab **overwrites in place without comparing to disk**. And this branch exists precisely because the on-disk contents could not be established — so the advice could clobber the very external change the sentence above it says might exist.

Both arms now point at a NEW path (Save As / a different filename), which keeps both versions, and say why in one clause.

**Second finding, also real, and it would have hidden a swap.** The arm test asserted only that both strings were PRESENT:

```js
assert.match(region, /dirtyNow \|\| wasDirty/);
```

`!(dirtyNow || wasDirty)` still contains that substring, and the slice below still reads the FIRST message object — so the arms could be inverted and a dirty tab would receive the clean-tab advice with all 52 tests green. It now pins the mapping: true branch carries the has-unsaved-edits text, false branch the reports-none text, and the condition may not be negated.

**The part worth reporting:** with only the mapping test in place, restoring the old `"Save first, then re-read"` string **passed all 52 tests**. The remedy change had no guard of its own. There is now an explicit assertion that neither arm sends the caller to an in-place overwrite, and re-running that mutation fails it.

Both changes mutation-verified. `npm run test:unit` **8354 pass / 0 fail**, `[check-panel-scope] OK`.

This is the second defect found in my own work on this PR today — the first was `conflict: true` on an arm whose staleness was explicitly unverified. Both were green and mutation-verified when I pushed them; neither mutation asked whether the thing being asserted was a safe thing to tell a user.

artokun:
**Pulled both emitted `stale_hint` strings out of the branch and read what they actually tell a caller to do.**

**Dirty arm:**

> Could not verify whether the on-disk file still matches this tab (disk read unavailable or slower than its deadline), **AND this tab has unsaved edits.** Treat the canvas as possibly stale, but **do NOT reach for `panel_load_workflow` first**: it loads the on-disk version over the canvas and those unsaved edits are discarded. **Save to a NEW path** (Save As…)…

**Clean arm:**

> Could not verify whether the on-disk file still matches this tab… Treat the canvas as possibly stale; call `panel_load_workflow` to be sure you have the on-disk version. **This tab reports no unsaved edits — though an edit written programmatically can leave that flag unset**, so if you are unsure…

Both halves of the original defect are closed:

- The **unconditional** `panel_load_workflow` recommendation is gone from the dirty arm, and replaced with an explicit warning against it plus a remedy that keeps both versions.
- The **"save first"** hazard is gone: the surviving save advice is *Save As to a new path*, not an in-place save. That matters because `classifyInPlaceOverwrite` returns `"skip"` for an already-`isPersisted` tab, so an ordinary save would have overwritten without reading disk — on the one branch that exists because the disk contents could not be established.

The clean arm's caveat is the positive-evidence rule applied correctly: `isModified === false` is not evidence of a clean tab, because a programmatic node add or a direct widget write both leave the flag unset. It recommends the load but declines to call the tab clean.

And `conflict: true` is correctly **not** set on either arm — a conflict asserts the file changed *and* the tab is dirty, while staleness here is explicitly unverified. Setting it would claim the very thing the sentence admits it could not check.

**One methodological note.** My first pass ran a crude predicate over the strings and it reported "tells them to save first: **true**" on the dirty arm — because the text contains "Save to a NEW path". The predicate was wrong, not the code. I mention it because that is exactly the failure this PR is about: a check that pattern-matches a phrase without reading what it actually instructs.

PR #2247, opened 2026-09-04
fix(984): panel_get_errors must not report clean for missing models on a promoted subgraph rail

artokun:
## Diagnosis

Recurrence of #984 on current main (0.15.172): turn-start still named four absent `MiniMaxH3/...` assets promoted through native subgraph host 1512, but `panel_get_errors` returned `audit_complete:true`, `errored_count:0`, empty `missing_models`. Same false-clean at root and inside the subgraph.

This is **not** the original load-time-store-empty hole. PR #991 / v0.11.83 already shipped the live `/object_info` scan. The new hole is the live **cross-check** that **subtracts** from Pinia `missingModel`:

1. The store blames the inner locator **or** host id 1512. After promotion the inner widget often holds an on-disk fallback while the **host rail** still names the missing file. Treating “inner no longer holds the filename” as a user fix dropped the candidate.
2. `get_errors` then sets `trustCombo` after refresh. `refreshComboOptionsFromDefs` keys off `node.type`, so a virtual GraphNode host is never rebuilt. A stale (or empty) host-rail combo that still **lists** the MiniMaxH3 path was treated as proof the file exists, so `missing_models` went empty.
3. The live scan skipped the virtual host entirely, and skipped inner loaders whose widgets are link-driven from the subgraph input, so `unavailable_widget_values` did not put the four files back. Both halves empty → `clean`.

## What this does **not** re-ship

- PR #991 / v0.11.83 live-scan introduction
- The original “stores empty at load” path

## Fix

- `assetCandidateStillReferenced` also looks at the promoted host rail (and the reverse).
- `assetCandidateResolvesLive` combo-clears only from **inner** nodes that currently hold the file — never from a host-rail combo list.
- `graph_get_errors` scans nested subgraphs from the bound root, and maps promoted host rails onto the inner loader class so `/object_info` can judge the **host** value.

Fail-closed: a genuine miss stays reported. An empty/stale combo on a host rail or inner node is not proof the file exists.

## Tests

- `#984 recurrence: a promoted HOST rail that still names the missing file keeps the inner locator candidate`
- `#984 recurrence: a stale HOST-rail combo must not combo-clear a still-missing promoted file`
- `#984 recurrence: an empty HOST-rail combo is not proof the file exists`
- `#984 recurrence: locator + trustCombo still reports when only the HOST rail names the miss`
- `#984 recurrence: the two halves naming the same promoted miss count once and are not clean`
- `#984 recurrence: a virtual subgraph HOST still reports a promoted missing model`
- Existing #984 suite in `browser_tests/unit/asset-staleness.test.mjs` still passes

## Notes

Do not merge. Do not cut a release. Parent overlay-merges after CI CLEAN.

Closes nothing yet — tracking #984.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+- panel_get_errors no longer reports a clean graph when missing models are still named on a promoted native-subgraph host rail: the load-time store blames the inner locator, and the inner widget may hold an on-disk fallback while the host still names the absent file. The live combo scan now also walks nested subgraphs from the bound root, so inner loaders are judged even when the user is looking at the root canvas (#984)
+
 ## [0.15.172] - 2026-09-04
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project are documented here. This project adheres to
 ## [Unreleased]
 
 ### Fixed
-- panel_get_errors no longer reports a clean graph when missing models are still named on a promoted native-subgraph host rail: the load-time store blames the inner locator, and the inner widget may hold an on-disk fallback while the host still names the absent file. The live combo scan now also walks nested subgraphs from the bound root, so inner loaders are judged even when the user is looking at the root canvas (#984)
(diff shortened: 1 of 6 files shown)


artokun:
Closing as duplicate of already-merged #2245, shipped in v0.15.173 (#2246). #984 is closed. Do not overlay-merge this CONFLICTING leftover.

Issue #1325, opened 2026-08-18
Graph reads work but mutations remain stuck in backend-reconnecting after Wan render


artokun:
### What happened / root cause
After a successful Wan render, panel graph reads continued to work from the open canvas, but every graph mutation was refused indefinitely with [backend-reconnecting]. panel_set_workflow_target(mode: current) reported graph_binding: bound/already_current, yet panel_disconnect and panel_set_widget still failed. The headless health endpoint simultaneously returned ECONNREFUSED for 127.0.0.1:8188, suggesting the backend went down while the panel retained readable cached graph state and never cleared its reconnect mutation guard.

### Steps to reproduce
1. Run a 49-frame Wan 2.2 native dual-pass workflow through panel_run.
2. Wait for successful SaveAnimatedWEBP completion.
3. Call panel_graph_outline: succeeds and returns 25-node graph.
4. Call panel_disconnect or panel_set_widget: repeatedly returns [backend-reconnecting].
5. Call panel_set_workflow_target(mode: current): reports bound/already_current.
6. Wait and retry: mutations remain blocked.

### Exact error (scrubbed)
Error: [backend-reconnecting] ComfyUI's backend connection is down right now (a restart or reconnect is in progress), so graph_set_widget was NOT applied - nothing changed.

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: distinguish cached-canvas readability from backend readiness in binding status; when the backend reconnects, reliably clear the mutation guard. If the backend is actually down, graph_binding should report a reconnecting/down state rather than bound/already_current, and expose an actionable recovery path/status.

### Environment
Windows 11; ComfyUI 0.33.1; frontend version unavailable because backend health endpoint was unreachable; AMD Radeon RX 9070 XT 16 GB; torch 2.12.0; Python 3.12; comfyui-mcp 0.52.1; panel 0.15.1.

artokun:
Claimed by grok-swarm 2026-08-19T07:02:09Z — driving this in worktree `/c/Users/Artokun/code/comfyui-mcp-panel/.worktrees/wt-1325` (branch `fix/1325-mutations-stuck-reconnecting`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
[...]

PR #2248, opened 2026-09-04
fix(2854): a stale socket flag refused every panel_run while every other call worked

artokun:
## Diagnosis

The emitting code IS in the panel — the report could not find it because the string is composed, not literal. `refuseUnstableRunDispatch` joins `comparison.changed`, so `"backend socket down"` is an array element and the sentence reads "backend socket down changed while…".

It is **not** a transition check. `compareRunDispatchIdentity` pushes that label whenever *either* observation is `"down"`:

```js
if (left.backendSocketState === "down" || right.backendSocketState === "down") changed.push("backend socket down");
```

Both observations were `"down"` because the provider read the sticky flag directly:

```js
backendSocketState = comfyBackendSocketDown === true ? "down" : transportState;   // web/js/comfyui-mcp-panel.js
```

Every other `panel_*` path asks `comfyBackendIsDown()` instead, and that helper carries the **#1325** rule:

```js
export function backendSocketIsDown({ flaggedDown, socketReadyState }) {
  if (flaggedDown !== true) return false;
  if (socketReadyState === WS_OPEN) return false;   // flagged-down + OPEN is STALE, not down
  return true;
}
```

#1325's own comment describes this session exactly: ComfyUI dispatches `status` with a null payload from `_pollQueue` whenever `GET /prompt` fails; a long GPU-bound render blocks the backend event loop, the poll fails, the flag arms — and `reconnected` never fires because the websocket **never left OPEN**, so the flag never clears.

That accounts for every observation in the report without needing a second cause:

| observed | why |
|---|---|
| `panel_graph_outline` says `backend_socket: "up"` | it uses the combined predicate; OPEN wins over the stale flag |
| reads, `set_widget`, `connect`, `set_node_mode` all work | same — every mutation path uses `comfyBackendIsDown()` |
| Queue button in the ComfyUI UI works | never touches the panel's dispatch fence |
| survives MCP disconnect/reconnect | the flag is cleared by `noteComfyBackendAlive()`, which the run path never reaches |
| survives `panel_set_workflow_target` rebinds | binding is a different fence |
| `panel_run` refuses **forever** | it alone read the raw flag |

The reconnect gate three lines above the identity read already used `comfyBackendIsDown()` and passed — so the same function disagreed with itself about whether the socket was down, a few lines apart.

## Fix

One line: ask the same question every other caller asks.

```js
backendSocketState = comfyBackendIsDown() ? "down" : transportState;
```

Strictness is unchanged where it matters, and the tests pin all four directions:

| flag | readyState | before | after |
|---|---|---|---|
| armed | OPEN | `down` → refuse | **`available` → dispatch** |
| armed | CLOSED (3) | `down` | `down` |
| armed | unreadable | `down` | `down` (#646 fail-closed) |
| clear | unreadable | `unknown` | `unknown` |

## Tests

`browser_tests/unit/run-dispatch-identity.test.mjs` — 18 pass.

The existing source pin asserted the **buggy** line verbatim, so the defect was pinned in place; it is updated, plus a `doesNotMatch` on the raw-flag form. The three new tests drive the decision through the real `backendSocketIsDown` / `backendSocketTransportState` helpers, and one deliberately runs the **pre-fix** form to pin it as the thing that produced the refusal.

Mutation-verified: restoring the raw-flag read fails the provider pin (17 pass / 1 fail).

`npm run test:unit`: **8355 pass, 0 fail**, including `[check-panel-scope] OK — every name resolves`, which is the gate that matters for calling `comfyBackendIsDown()` from inside `buildPanel`.

## Not verified

I could not reproduce the reporter's arming condition on this machine — it needs a render long enough to block ComfyUI's event loop until `_pollQueue` fails. The mechanism is read from the code and from #1325's own account of the same flag, and the fix is exercised at the decision boundary rather than through a live stuck socket.

**Fixes `artokun/comfyui-mcp#2854`** — filed against the orchestrator repo because the
reporter could not find the emitting string and concluded it lived there. The code is in
the panel, so the fix is here.

**This PR will NOT auto-close that issue.** GitHub honours closing keywords only within a
repository, and `#2854` in this repo resolves to nothing (panel issue numbers are still in
the 2200s). Whoever merges this needs to close `artokun/comfyui-mcp#2854` by hand.


artokun:
Audited the sibling call sites before this merges, since a one-line fix to a shared flag usually leaves relatives. **It does not — the scope here is right, and widening it would be a regression.**

`comfyBackendSocketDown` is read raw in eight places on `main`. Three are assignments, three pass it as a named `flaggedDown:` / `socketDown:` field (which is what the helper and the retain-callbacks want), and three are conditions:

```
web/js/comfyui-mcp-panel.js:2533   !comfyBackendSocketDown && runStartedAtEpoch === backendReconnectEpoch && …
web/js/comfyui-mcp-panel.js:2690   !comfyBackendSocketDown && …
web/js/comfyui-mcp-panel.js:2710   !comfyBackendSocketDown && …
```

Those three look like this PR's line but are the opposite case. They compute `runIsCurrent` — *is this refresh run still trustworthy* — and a stale-true flag makes the verdict **false**, so a possibly-pre-restart refresh does not re-open combo trust. That fails **safe**, and the file already says so:

> a missing model whose loader combo was never rebuilt would then have been SUPPRESSED as stale. That is the one direction this flag must never fail in.

Swapping them to `comfyBackendIsDown()` would make `runIsCurrent` **true** during exactly the flagged-but-OPEN interval that rule exists to distrust — re-opening combo trust in the window the comment forbids. So a later "consistency" pass replacing every raw read with the helper would reintroduce the bug this file was written to avoid.

This PR's line is the genuinely different one: `backendSocketState` gates *dispatch*, so a stale-true flag refuses runs that would have succeeded — failing closed in the harmful direction rather than the conservative one.

Net: one call site was correct to change, and the three that resemble it are correct to leave. No follow-up needed.

artokun:
**Pre-merge reachability check: the fix reaches the reported path end to end.** Tracing it because the recurring failure in this queue has been a change that is correct in the function it touches and never reaches the path the issue is about.

```
comfyui-mcp-panel.js:41657   backendSocketState = comfyBackendIsDown() ? "down" : transportState;   ← the fix
comfyui-mcp-panel.js:41666   returned as `backendSocketState` on the dispatch identity object
comfyui-mcp-panel.js:41680   runDispatchIdentityRef = panelRunDispatchIdentity
comfyui-mcp-panel.js:20080   const dispatchIdentityReader = runDispatchIdentityRef
lib/run-dispatch-identity.js:81
      if (left.backendSocketState === "down" || right.backendSocketState === "down") {
        changed.push("backend socket down");
```

That pushed string is verbatim the one in the issue title — *"panel_run always fails with `backend socket down changed while…`"*. So the value the fix changes is the value that produces the refusal, with no branch in between that could bypass it.

Two things this also confirms:

- **`backendSocketState` has exactly one producer and one consumer** (41657 → line 81), so there is no sibling call site reading the raw flag on this path. I checked the other raw `comfyBackendSocketDown` reads separately — the three `runIsCurrent` freshness verdicts must keep the raw flag, and my earlier comment on this PR records why changing them would regress.
- **The refusal is `||` across both sides**, so a stale `"down"` on either the before- or after-identity refuses. That is why the reporter saw *every* dispatch refused rather than an intermittent one.

Unrelated but worth stating on the same thread: the "no longer reproduces" follow-up on comfyui-mcp#2854 does not indicate this is fixed. The reporter restarted ComfyUI, and the flag's initializer (`let comfyBackendSocketDown = false`) clears it on any version — the raw read is still on panel `main` at v0.15.174, newer than the version they upgraded to. That issue needs closing by hand when this merges, and not before.

Issue #2027, opened 2026-08-29
panel_refresh_nodes clears schema when a stale browser bundle cannot refresh large object_info


artokun:
### What happened / root cause
The live browser tab was running panel 0.15.123 while the installed panel bundle on disk was 0.15.124. On a large custom-node install, panel_refresh_nodes returned object_info_fetch_failed and cleared the last-known schema. A subsequent panel_set_widget correctly observed that /models/checkpoints already listed the newly downloaded checkpoint, but the stale /object_info combo still exposed only four old options. The current installed 0.15.124 source contains the larger-refresh-budget fix, but an already-open tab cannot pick it up without a cache-bypassing reload.

### Steps to reproduce
1. Open ComfyUI with panel 0.15.123 loaded in the browser.
2. Update/install panel 0.15.124 on disk without hard-refreshing the tab.
3. On an install with a very large /object_info, download a new checkpoint.
4. Call panel_refresh_nodes.
5. Call panel_set_widget on CheckpointLoaderSimple.ckpt_name with the newly downloaded checkpoint.

### Exact error (scrubbed)
panel_refresh_nodes:
{ "ok": true, "refreshed": false, "reason": "object_info_fetch_failed", "detail": "(Failed to fetch)" }
Remedy said every transport failed and nothing was refreshed; the command also cleared the last-known schema.

panel_set_widget:
Value "pony-diffusion-v6.safetensors" is not a valid option. Valid options: four old checkpoints.
Model inventory disclosure in the same result: /models/checkpoints includes "pony-diffusion-v6.safetensors", but refreshed /object_info does not.

### Fix
applied locally: no
upstream-only: yes
The on-disk 0.15.124 bundle already contains the large-/object_info refresh-budget fix, so patching that file would not change the running browser module. Suggested product fix: before panel_refresh_nodes clears the last-known schema, compare the running PANEL_VERSION to the installed version marker. If stale, return a dedicated stale_bundle verdict with a Ctrl+Shift+R remedy instead of attempting the refresh; preserve the existing schema. This makes the true cause explicit and avoids worsening the next widget edit.

### Environment
Windows 11
ComfyUI 0.34.2
ComfyUI frontend 1.49.6
NVIDIA GeForce RTX 5090, 32 GB VRAM
comfyui-mcp 0.52.146
running panel 0.15.123; installed panel 0.15.124

---
[...]

PR #2253, opened 2026-09-05
fix(2252): refuse panel_run on a stale live bundle instead of silent drop

artokun:
## Problem

On a live tab running panel 0.15.173 while 0.15.174 is installed, `panel_run` accepted the request and then silently dropped dispatch. Scoped and full runs produced no `prompt_id`; ComfyUI never logged `got prompt`; the queue stayed idle. One scoped attempt created a pending item that the guard correctly removed after no dispatch was observed.

This is not #2248 / comfyui-mcp#2854: those refuse loudly with `backend socket down`. This path returned `queued_unknown` with no reason and no hard-refresh requirement.

## Fix

Fail-closed: compare the running `PANEL_VERSION` to the installed pack (`/comfyui_mcp_panel/version`) **before** `beginPanelRun` / `graphToPrompt` / `queuePrompt` / `dispatchScopedRun`. A proven mismatch returns `queued: false`, `reason: "stale_bundle"`, and a Ctrl+Shift+R remedy. An unreadable probe still fails open (same rule as #2027).

Unverified scoped pending-item removal is unchanged (`cancelPendingScopedQueueItem` on the dispatch timeout path).

## Tests

`node --test browser_tests/unit/*.test.mjs` — 8369 pass, 1 todo.

New `browser_tests/unit/run-stale-bundle.test.mjs` fails without the gate (graph_run proceeds into dispatch) and passes with it. Also pins fail-open on a missing probe and that tagged pending-item removal still splices only this run's item.

## Changelog

```
## [Unreleased]

### Fixed

- panel_run refuses a stale live browser bundle with a Ctrl+Shift+R hard-refresh requirement instead of silently dropping dispatch; unverified scoped pending items are still removed fail-closed (#2252)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to this project are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- panel_run refuses a stale live browser bundle with a Ctrl+Shift+R hard-refresh requirement instead of silently dropping dispatch; unverified scoped pending items are still removed fail-closed (#2252)
+
+
 ## [0.15.174] - 2026-09-04
 
 ### Fixed
--- a/browser_tests/unit/module-graph-link.test.mjs
+++ b/browser_tests/unit/module-graph-link.test.mjs
@@ -36,7 +36,7 @@ import { commandIsCanvasIndependent } from "../../web/js/lib/workflow-chat-ident
 import { sealProvenRootBinding } from "../../web/js/lib/graph-binding.js";
 import { composeShowMediaReply } from "../../web/js/lib/media-preview.js";
 import { composeRunCompletionFrame } from "../../web/js/lib/run-completion-frame.js";
-import { describeNodeDefRefresh, describeStaleBundleRefresh } from "../../web/js/lib/node-def-refresh.js";
+import { describeNodeDefRefresh, describeStaleBundleRefresh, describeStaleBundleRun } from "../../web/js/lib/node-def-refresh.js";
 import { confirmCanvasNavigation } from "../../web/js/lib/canvas-navigation.js";
 import {
   watchPostReconnectSettle,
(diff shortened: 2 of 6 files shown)


artokun:
**Independent read — this addresses the reported symptom precisely, and the placement is the part that makes it work.**

I looked at comfyui-mcp-panel#2252 before this PR existed and flagged one thing to preserve: the guard's fail-closed removal of an unverified scoped pending item. **This makes that moot, which is better than preserving it** — the refusal happens before any pending item exists:

```js
const staleBundle = await refuseStaleBundleRun();
if (staleBundle) return staleBundle;
// ...then beginPanelRun(), graph binding, queuePrompt, scoped dispatch
```

Ahead of `beginPanelRun()`, prompt construction, `queuePrompt`, and scoped dispatch. Nothing is created, so nothing needs unwinding.

**The message answers the actual complaint.** The reported failure was `queued_unknown` with no `prompt_id` and an idle queue — the caller could not tell whether a run had been accepted. Now:

> This tab is running panel `{running}` while the installed pack is `{installed}`. Hard-refresh this tab (Ctrl+Shift+R) to load panel `{installed}` before `panel_run`. **Nothing was queued.**

with `queued: false` on the returned object. Concrete versions, an executable remedy, and an unambiguous statement about what did *not* happen — which is the half a silent drop leaves the caller to guess.

**Fail-open is the right direction and is implemented that way.** `refuseStaleBundleRefresh()` returning falsy yields `null`, so an unreadable version probe lets the run proceed rather than refusing everything. A probe that cannot read its own version should not become a global block.

**And it reuses the refresh-path helper** rather than restating the rule, so run and refresh cannot drift apart — the "fixed one of two paths" failure this codebase keeps meeting.

One note for triage: this is a **different bug** from comfyui-mcp#2854, and comfyui-mcp-panel#2248 does not overlap it. That one is an explicit refusal from a stale socket flag; this is a silent drop from a bundle mismatch. Both are "panel_run does not reach ComfyUI", which is how they would get merged in triage.

